PR #12, opened 2017-11-13
Add user stats to home view

mFelgate:


src/services/session.js
```diff
@@ -27,7 +27,38 @@ const getAll = () => http.get('sessions').then(sessions => sessions.reverse());
  */
 const getCurrent = () => http.get('sessions/current');
 
+/**
+ * Get days left in the term
+ * @return {Promise.<Session>}
```
AdamVig: This function does not return a Promise resolved by a `Session`. Can we create a new `@typedef` like the one above, and have it define what the object returned by this function will look like?

src/services/user.js
```diff
@@ -0,0 +1,39 @@
+
```
AdamVig: Can we get rid of this blank line? The opening comment should be at the very top of the file.

src/services/user.js
```diff
@@ -0,0 +1,39 @@
...
+ * User
+ *
+ * @module User
+ */
+import http from './http';
+import session from './session';
+import storage from './storage';
+
+/**
+ * Get all activities for a session, sorted alphabetically by description
```
AdamVig: Can we update this JSDoc block so it documents the function below it?

src/services/user.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {String} sessionCode Identifier for a session
+ * @return {Promise.<User[]>}
+ */
+const getAttendedEvents = (username, termCode) =>
+  http.get(`events/chapel/${username}/${termCode}`);
+
+const getChapelCredits = async () => {
```
AdamVig: This function needs a JSDoc block. Can we add one? We will want a new `@typdef` that defines what the object returned by this function looks like, too.

src/views/ActivitiesAll/index.js
```diff
@@ -12,6 +12,7 @@ import session from '../../services/session';
 import GordonActivityGrid from './components/ActivityGrid';
 import GordonLoader from '../../components/Loader';
 
+
```
AdamVig: Can we get rid of this extra blank line?

src/views/Home/components/DaysLeft/index.js
```diff
@@ -0,0 +1,56 @@
+import React, { Component } from 'react';
+import Card, { CardContent } from 'material-ui/Card';
+import { Doughnut } from 'react-chartjs-2';
+
+import { gordonColors } from '../../../../theme';
+import session from '../../../../services/session';
+import GordonLoader from '../../../../components/Loader';
+
+
```
AdamVig: Can we get rid of this blank line?

src/views/Home/components/DaysLeft/index.js
```diff
@@ -0,0 +1,56 @@
...
+    this.setState({ loading: false });
+  }
+  render() {
+    let content;
+    const daysleft = this.state.daysLeft[0];
+    const pastDays = this.state.daysLeft[1] - daysleft;
+    const Data = {
+      datasets: [{ data: [pastDays, daysleft], backgroundColor: [gordonColors.primary.blue] }],
+    };
+    if (this.state.loading === true) {
```
AdamVig: With this approach (in this component and the other one), the homepage just looks like two random loading spinners until the data has arrived. A better approach would be to swap out the `<Doughnut ... />` element for a loader, then replace the loader with the chart once the data has loaded. This way, there would be context for what the user is waiting for ("oh, my chapel credits are loading"), and the card would be around the loader, so it would look better.

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
AdamVig: @mFelgate It looks like this didn't change in your latest commit. Can you make the name change I described above?

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,56 @@
...
+    let content;
+    // console.log(chapelEvents);
+    const numEvents = (this.state.chapelCredits.current);
+    // console.log(this.state.chapelCredits);
+    const remaining = this.state.chapelCredits.required - numEvents;
+    const Data = {
+      datasets: [{ data: [numEvents, remaining], backgroundColor: [gordonColors.primary.blue] }],
+    };
+    if (this.state.loading === true) {
+      content = <GordonLoader />;
```
AdamVig: Same feedback as above - it would make more sense to swap just the chart out for a loader, instead of the whole card.

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,56 @@
...
+    this.loadChapel();
+  }
+  async loadChapel() {
+    const chapelCredits = await user.getChapelCredits();
+    this.setState({ loading: false });
+    this.setState({ chapelCredits });
+  }
+  render() {
+    let content;
+    // console.log(chapelEvents);
```
AdamVig: Can we remove these commented out lines?

src/views/Home/events-chapel-user-session.json
```diff
@@ -4,7 +4,7 @@
     "CHDate": "2017-09-27T10:25:00",
```
AdamVig: Now that we are using live data, we can remove this file and `session-days-left.json`. Can you take care of that?

src/views/Home/components/DaysLeft/index.js
```diff
@@ -0,0 +1,56 @@
...
+      content = <GordonLoader />;
+    } else {
+      content = (
+        <Card>
+          <CardContent>
+            <figure>
```
AdamVig: The `<figure>` tag doesn't make much sense in this context. You could either get rid of it and just have an `h3` and the chart, or move the chart to inside the `<figure>`. This goes for the other component, too.

src/views/Home/components/DaysLeft/index.js
```diff
@@ -0,0 +1,56 @@
...
+  async getDaysLeft() {
+    const daysLeft = await session.getDaysLeft();
+    this.setState({ daysLeft });
+    this.setState({ loading: false });
+  }
+  render() {
+    let content;
+    const daysleft = this.state.daysLeft[0];
+    const pastDays = this.state.daysLeft[1] - daysleft;
+    const Data = {
```
AdamVig: Can we change this name from `Data` to `data`? Having a capital first letter isn't correct camelcase.

AdamVig:
Matt, a few thoughts:
1. Can you either disable the hover on the charts, or add labels so that the hover user experience is good? Currently, it just says "undefined" when you hover.
2. Can you add text to the cards that says "__ days left" and "__ out of __ CL&W credits"?
3. It looks like on the chapel progress graph, the chapel credits the user has completed show up as gray, and the ones left to complete are blue. Shouldn't it be reversed, and be fully blue when the user has completed all of his or her chapel credits?

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
+    "chart.js": "2.7.1",
```
AdamVig: You should also have `react-chartjs-2` listed as a dependency. Looks like it got lost when you merged master into this branch. Run `npm install --save react-chartjs-2` to fix this problem.

src/routes.js
```diff
@@ -4,6 +4,7 @@ import ActivityEdit from './views/ActivityEdit';
 import Home from './views/Home';
 import Help from './views/Help';
 
+
```
AdamVig: Can we remove this extra blank line?

src/services/session.js
```diff
@@ -29,7 +29,7 @@ const getCurrent = () => http.get('sessions/current');
 
 /**
  * Get days left in the term
- * @return {Promise.<Session>}
+ * @return {Promise.<DaysLeft[]>}
```
AdamVig: You cannot use a JSDoc type like `DaysLeft` without first defining it. See, for example, the type definition for `Auth`:

```JavaScript
/**
 * @global
 * @typedef Auth
 * @property {String} token Bearer token
 * @property {Date} expires Date/time when token expires
 */
```

You can read more about JSDoc typedefs [here](http://usejsdoc.org/tags-typedef.html). We make all of our typedefs `@global` so they can be used in any file.

src/services/user.js
```diff
@@ -9,13 +8,17 @@ import session from './session';
...
- * Get all activities for a session, sorted alphabetically by description
+ * Get events attended by the user
  * @param {String} sessionCode Identifier for a session
- * @return {Promise.<User[]>}
+ * @return {Promise.<attendedEvents[]>}
  */
 const getAttendedEvents = (username, termCode) =>
   http.get(`events/chapel/${username}/${termCode}`);
 
+  /**
```
AdamVig: This line's indentation is off. The starting `/**` of a documentation block should be indented to match the thing it is documenting. In this case, `const getChapelCredits = ...` is indented zero spaces, so this line should be indented zero spaces as well.

src/services/user.js
```diff
@@ -9,13 +8,17 @@ import session from './session';
 import storage from './storage';
 
 /**
- * Get all activities for a session, sorted alphabetically by description
+ * Get events attended by the user
  * @param {String} sessionCode Identifier for a session
- * @return {Promise.<User[]>}
+ * @return {Promise.<attendedEvents[]>}
```
AdamVig: Same as my comment above about `DaysLeft`, you have to define this type before you can use it in a documentation block. Note that the `[]` syntax means "an array of this type", so you would want the name of the type you use to be singular, like `AttendedEvent`.

src/views/Home/components/DaysLeft/index.js
```diff
@@ -0,0 +1,68 @@
...
+      loading: true,
+    };
+  }
+  componentWillMount() {
+    this.getDaysLeft();
+  }
+  async getDaysLeft() {
+    const daysLeft = await session.getDaysLeft();
+    this.setState({ daysLeft });
+    this.setState({ loading: false });
```
AdamVig: You can combine this `setState` with the previous line:
`this.setState({ daysLeft, loading: false });`

src/views/Home/components/DaysLeft/index.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+  render() {
+    defaults.global.legend.display = false;
+    let content;
+    const daysleft = this.state.daysLeft[0];
+    const pastDays = this.state.daysLeft[1] - daysleft;
+    const Data = {
```
AdamVig: This variable name should be lowercase.

src/views/Home/components/DaysLeft/index.js
```diff
@@ -0,0 +1,68 @@
...
+      datasets: [{ data: [pastDays, daysleft], backgroundColor: [gordonColors.primary.blue] }],
+      labels: [
+        'Days Finished',
+        'Days Remaining',
+      ],
+    };
+    const Options = {
```
AdamVig: This variable name should be lowercase too.

src/views/Home/components/chapelProgress/index.js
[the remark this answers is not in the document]
AdamVig: @mFelgate It looks like the folder name still starts with a lowercase letter. I think you are running into a problem with Git Bash, where it sometimes does not catch changes in file/folder name case. See [this StackOverflow answer](https://stackoverflow.com/questions/14578680/changing-case-of-folder-via-git-bash-on-windows) for help fixing the problem.

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,62 @@
+import React, { Component } from 'react';
+import { Doughnut } from 'react-chartjs-2';
+import Card, { CardContent } from 'material-ui/Card';
+
+import { gordonColors } from '../../../../theme';
+import user from '../../../../services/user';
+import GordonLoader from '../../../../components/Loader';
+
+export default class ChapelPsrogress extends Component {
```
AdamVig: The name of the component has a typo in it.

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,62 @@
...
+      chapelCredits: {},
+    };
+  }
+  componentWillMount() {
+    this.loadChapel();
+  }
+  async loadChapel() {
+    const chapelCredits = await user.getChapelCredits();
+    this.setState({ loading: false });
+    this.setState({ chapelCredits });
```
AdamVig: This line can be combined with the previous line:
`this.setState({ chapelCredits, loading: false });`

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,62 @@
...
+      content = (
+        <Card>
+          <CardContent>
+            <figure>
+              <figcaption>
+                <h3>Chapel Progress</h3>
+                <h4>{current}/{required} Attended Chapels Events</h4>
+              </figcaption>
+            </figure>
+            <Doughnut data={Data} />
```
AdamVig: Can we put this `<Doughnut>` inside the `<figure>` like you did in `DaysLeft`?

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,62 @@
...
+        data: [current, remaining],
+        backgroundColor: [gordonColors.primary.blue],
+      }],
+      labels: [
+        'Chapel Events Attended',
+        'Events Remaining',
+      ],
+    };
+    if (this.state.loading === true) {
+      content = <GordonLoader />;
```
AdamVig: With this approach (in this component and the other one), the homepage just looks like two random loading spinners until the data has arrived.

A better approach would be to swap out the `<Doughnut ... />` element for a loader, then replace the loader with the chart once the data has loaded. This way, there would be context for what the user is waiting for ("oh, my chapel credits are loading"), and the card would be around the loader, so it would look better.

As a result, this would look like:
```JavaScript
    if (this.state.loading === true) {
      content = <GordonLoader />;
    } else {
      content = (
        <figure>
          <figcaption>
            <h4>{current}/{required} Attended Chapels Events</h4>
          </figcaption>
          <Doughnut data={Data} />
        </figure>
      );
    }
    return (
      <Card>
        <CardContent>
          <h3>Chapel Progress</h3>
          { content }
        </CardContent >
      </Card>
    );
```

src/views/Home/components/chapelProgress/index.js
```diff
@@ -0,0 +1,62 @@
...
+    } else {
+      content = (
+        <Card>
+          <CardContent>
+            <figure>
+              <figcaption>
+                <h3>Chapel Progress</h3>
```
AdamVig: We should use `<Typography>` elements for all text so that it remains consistent across the app. Can you change these two headers (and the corresponding headers in `DaysLeft`) to use `<Typography>`?

src/views/Home/components/ChapelProgress/index.js
```diff
@@ -35,25 +36,28 @@ export default class ChapelPsrogress extends Component {
...
-        'Events Remaining',
+        'CL&W Credits',
+        'CL&W Credits Remaining',
       ],
     };
     if (this.state.loading === true) {
       content = <GordonLoader />;
     } else {
       content = <Doughnut data={data} />;
+      if (current === 1) {
```
AdamVig: It is not grammatically correct to pluralize "CL&W" (it would be "Christian Life and Worships").

Can we change the subheader from `${current} CL&W` and `${current} CL&W's` to `${current} CL&W Credit` and `${current} CL&W Credits`?

src/services/user.js
```diff
@@ -7,10 +7,18 @@ import http from './http';
 import session from './session';
 import storage from './storage';
 
+/**
+ * @global
+ * @typedef User
```
AdamVig: Since we already have a global typedef called `User`, we should rename this one to something like `CLWCredits` or `ChapelCredits`.

src/services/user.js
```diff
@@ -7,10 +7,18 @@ import http from './http';
...
+ * @property {int} current user's current cl&w credits
```
AdamVig: JavaScript doesn't actually have an `int` type, so this and the line below it should be `Number` types.

src/services/user.js
```diff
@@ -7,10 +7,18 @@ import http from './http';
...
+ * @property {int} required user's required cl&w credits
+ */
+
 /**
  * Get events attended by the user
- * @param {String} sessionCode Identifier for a session
- * @return {Promise.<attendedEvents[]>}
+ * @param {String} termCode code for the semester
+ * @param {String} username username of the user
```
AdamVig: Swap this line with the previous line. `@param` definitions should be in the order of the actual functions parameters.

src/services/user.js
```diff
@@ -7,10 +7,18 @@ import http from './http';
...
+ * @return {Promise.<attendedEvents>}
```
AdamVig: What does `attendedEvents` reference? It seems like you forgot to create a typedef for it. Here's an example that might help:

Replace this line with: `@return {Promise.<AttendedEvent[]>}`. This means "this function returns a promise that is resolved by an array of attended event objects.
Create the following typedef at the top of the file:
```
/**
 * @global
 * @typedef AttendedEvent
 * @property {proptype} PropName description of property
 */
```

That last line is an example. You need to add an `@property` line for each property in the `AttendedEvent` objects that are returned from the API.

package.json
```diff
@@ -39,6 +39,7 @@
     "build": "npm run build-css && react-scripts build",
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject",
+<<<<<<< HEAD
```
AdamVig: You need to resolve this merge conflict.

The app build script reads the proxy configuration from this file, and the file is currently impossible to parse, so this is probably the reason your API requests are failing.

src/services/user.js
```diff
@@ -9,28 +9,44 @@ import storage from './storage';
...
   /**
  * return the number of cl&w credits aquired, and number of credits required.
- * @return {Promise.<User[]>}
+ * @return {Promise.<CLWCredits[]>}
  */
 const getChapelCredits = async () => {
   const { username } = storage.get('credentials');
   const termCode = session.getTermCode();
   const attendedEvents = await getAttendedEvents(username, termCode);
+  console.log(attendedEvents);
```
AdamVig: Can we remove this?

src/services/user.js
```diff
@@ -9,28 +9,44 @@ import storage from './storage';
...
+ * @param {String} termCode code for the semester
  * @return {Promise.<attendedEvents>}
  */
 const getAttendedEvents = (username, termCode) =>
   http.get(`events/chapel/${username}/${termCode}`);
 
   /**
  * return the number of cl&w credits aquired, and number of credits required.
- * @return {Promise.<User[]>}
+ * @return {Promise.<CLWCredits[]>}
```
AdamVig: Since this function actually returns a single instance of `CLWCredits`, not an array, you can drop the brackets. The result will look like this:
```JavaScript
@return {Promise.<CLWCredits>}
```

src/services/user.js
```diff
@@ -9,28 +9,44 @@ import storage from './storage';
...
+ * @property {Object} Occurrences An array with the locations, start, and end times of future events
+ * @property {String} Organization Organization hosting the event
+ * @property {Number} Required Required chapel for the student
+
 /**
  * Get events attended by the user
- * @param {String} termCode code for the semester
  * @param {String} username username of the user
+ * @param {String} termCode code for the semester
  * @return {Promise.<attendedEvents>}
```
AdamVig: I see that you added a `@typedef` for the type that this function returns, but you did not update the documentation for the function. Since you're returning an array of attended events, this line should look like:
```JavaScript
@return {Promise.<AttendedEvent[]>}
```

src/services/user.js
mFelgate: well its returning on abject containing two numbers which are why the brackets are there. I added the description but it didn't show up in the commit.

AdamVig: Brackets indicate an *array* of whatever type is defined in the `@typedef`. As you said yourself, this function returns one *object*. Therefore, there should be no brackets.

.env
```diff
@@ -1 +1,2 @@
-REACT_APP_API_URL=/
+REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
```
AdamVig: This should be just `REACT_APP_API_URL=/` because we are now using the proxy configuration.

src/services/user.js
```diff
@@ -0,0 +1,66 @@
...
+ * @property {Object} Occurrences An array with the locations, start, and end times of future events
+ * @property {String} Organization Organization hosting the event
+ * @property {Number} Required Required chapel for the student
+*/
+
+/**
+ * Get events attended by the user
+ * @param {String} username username of the user
+ * @param {String} termCode code for the semester
+ * @return {Promise.<attendedEvents[]>} An object of all CL&W events attended by the user
```
AdamVig: This needs to be changed from `attendedEvents[]` to `AttendedEvent[]`. As it is now, it will not compile correctly because the name used here does not match the `@typedef`'s name.

src/views/Home/components/ChapelProgress/index.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+  async loadChapel() {
+    this.setState({ loading: true });
+    const chapelCredits = await user.getChapelCredits();
+    this.setState({ loading: false, chapelCredits });
+  }
+  render() {
+    let content;
+    let subheader;
+    const { current, required } = (this.state.chapelCredits);
```
AdamVig: We can remove the parentheses around `this.state.chapelCredits`.

src/views/Home/components/ChapelProgress/index.js
```diff
@@ -0,0 +1,65 @@
...
+        'CL&W Credits',
+        'CL&W Credits Remaining',
+      ],
+    };
+    if (this.state.loading === true) {
+      content = <GordonLoader />;
+    } else {
+      content = <Doughnut data={data} />;
+      if (current === 1) {
+        subheader = `${current} CL&W`;
```
AdamVig: This currently reads, for example, "5 CL&W" which means "5 Chapel Life and Worship." You need to add "Credit" to the end for the phrase to make any sense: "5 CL&W Credit", which translates to:
```JavaScript
subheader = `${current} CL&W Credit`;
```

src/views/Home/components/ChapelProgress/index.js
```diff
@@ -0,0 +1,65 @@
...
+      } else {
+        subheader = `${current} CL&W's`;
```
AdamVig: Same here, this currently reads, for example, "5 CL&W's" which means "5 Chapel Life and Worship's." You need to add "Credits" to the end for the phrase to make any sense: "5 CL&W Credits", which translates to:
```JavaScript
subheader = `${current} CL&W Credits`;
```

src/views/Home/components/ChapelProgress/index.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+    return (
+      <Card>
+        <CardContent>
+          <CardHeader
+            title="Chapel Progress"
```
AdamVig: I think this title should be "CL&W Credits," not "Chapel Progress."

My rationale is that the other chart, days left, is entitled "Days Left In Semester." If they were to be made consistent, we could either make the change I suggest above, or change the days left chart's title to "Semester Progress."

src/services/user.js
```diff
@@ -34,7 +34,7 @@ import storage from './storage';
  * Get events attended by the user
  * @param {String} username username of the user
  * @param {String} termCode code for the semester
- * @return {Promise.<attendedEvents[]>} An object of all CL&W events attended by the user
+ * @return {Promise.<AttendedEvents[]>} An object of all CL&W events attended by the user
```
AdamVig: This should be `Promise.<AttendedEvent[]>` (drop the "s" on `AttendedEvents`) because that's the name of the `@typedef`.

src/views/Home/components/ChapelProgress/index.js
```diff
@@ -45,16 +45,16 @@ export default class ChapelProgress extends Component {
     } else {
       content = <Doughnut data={data} />;
       if (current === 1) {
-        subheader = `${current} CL&W`;
+        subheader = `${current} CL&W credits`;
```
AdamVig: This should be singular ("credit" not "credits") because it is in the clause `current === 1`, meaning that there is only one credit.

src/views/Home/components/ChapelProgress/index.js
```diff
@@ -45,16 +45,16 @@ export default class ChapelProgress extends Component {
...
       } else {
-        subheader = `${current} CL&W's`;
+        subheader = `${current} CL&W credit's`;
```
AdamVig: There should be no apostrophe ("credits" not "credit's") because it is not possessive.

PR #13, opened 2017-11-15
Add Help page

RFStauffer:


RFStauffer:
image

src/views/Help/help.scss
```diff
@@ -0,0 +1,19 @@
+@import '../../vars';
+
+ul{
```
AdamVig: Can we get rid of these styles, now that we are relying on `Typography` instead?

src/views/Help/index.js
```diff
@@ -0,0 +1,174 @@
+import Typography from 'material-ui/Typography';
+import React, { Component } from 'react';
+
+import './help.css';
+
+export default class Help extends Component {
+  render() {
+    return (
+      <div>
```
AdamVig: Do you think we can remove all of the `<div>` elements in the page?

My guess is that you might be able to remove all of them except for the outermost one (JSX requires a single container element to be used). It's worth a shot, because the fewer elements we use, the better (both for performance and maintainability).

However, if you start removing them and it breaks the layout, then it's probably not worth spending much time trying to figure out.

RFStauffer:
image

AdamVig:
Fantastic work! Thanks for taking the time to fix typos and inconsistencies between the two pages. That will have a lasting impact!

PR #15, opened 2017-11-17
Create static HTML About page, moved link from nav drawer to global menu

ajabbot:
localhost-3000-about
localhost-3000-about nexus 6p


src/components/Header/components/GlobalMenu/index.js
```diff
@@ -43,6 +43,7 @@ export default class GordonGlobalMenu extends Component {
         >
           <MenuItem onClick={this.onClose}><Link to="/help">Help</Link></MenuItem>
           <MenuItem onClick={this.onClose}><Link to="">Logout</Link></MenuItem>
+          <MenuItem onClick={this.onClose}><Link to="/about">About</Link></MenuItem>
```
AdamVig: By convention I've seen on other sites, I think Logout should be last on this list, and probably also separated by a divider. You can use the [`Divider`](https://material-ui-next.com/demos/dividers/) component for that.

src/views/About/about.scss
```diff
@@ -0,0 +1,59 @@
+@import '../../vars';
+
+p{
```
AdamVig: We don't have a Sass linter set up (maybe we should!), but it is conventional to put a space between a selector (`p` in this situation) and the opening curly brace (`{`).

src/views/About/about.scss
```diff
@@ -0,0 +1,59 @@
...
+  margin-left: 15px;
+}
+a{
+  color: #0b91fd;
```
AdamVig: I think you might not need to set the link color explicitly (Material-UI should handle that), but in case you do need to override it here, can you use a variable from `./src/_vars.scss` instead of the hex code?

My reasoning for that is, it is difficult to look at a hex code and know "Oh, that's the primary cyan from the Gordon color palette," which makes it difficult to maintain styles that use hex codes frequently. Using variables makes styles much more maintainable and easier to read.

src/views/About/about.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/*.highlight {
```
AdamVig: Can we get rid of these unused styles?

src/views/About/index.js
```diff
@@ -0,0 +1,151 @@
+import React, { Component } from 'react';
+
+import Typography from 'material-ui/Typography';
+import './about.css';
+
+export default class About extends Component {
+  render() {
+    return (
+      <div className="container page-title about-container about">
```
AdamVig: Nice job converting all the `class` attributes to `className`.

Since we aren't using any styles in this file anymore, you should be able to remove the `className` attributes on all elements!

src/views/About/index.js
```diff
@@ -0,0 +1,151 @@
...
+                <li>Institutional reporting</li>
+              </Typography>
+            </ul>
+          </div>
+        </div>
+        <div className="about-content">
+          <h3 className="titletext">
+            <Typography type="title" gutterBottom>
+                Origins
+            </Typography></h3>
```
AdamVig: It looks like this `</h3>` should be on its own line. ESLint picked it up when I opened the file in my editor. Can you look through the file and correct any other linting errors?

To double check, run `npm run lint` in your terminal to see if there are any linting errors you missed.

src/views/About/index.js
```diff
@@ -0,0 +1,151 @@
...
+                <br /> And to our mentors and coaches in development and deployment, thank you for
+               your guidance and encouragement to try new things.
+              </Typography>
+            </p>
+          </div>
+          <div>
+            <h5>
+              <Typography type="caption" gutterBottom>
+              Found a bug? Report to
+                <a href="mailto:cts@gordon.edu?Subject=Gordon 360 Bug" target="_top"> CTS</a>
```
AdamVig: Instead of just "CTS" being a link, I think "Report to CTS" should be a link. This will make it more obvious to users that the text is clickable.

Also, I don't think we need `target="_top"` here - it doesn't have any effect, as far as I can tell.

src/views/About/index.js
```diff
@@ -0,0 +1,151 @@
+import React, { Component } from 'react';
+
+import Typography from 'material-ui/Typography';
```
AdamVig: As a convention, I have been grouping external imports at the very top of the file, then a blank line, then local imports. I also loosely alphabetize imports by the name of the imported file (`material-ui` vs. `react`). By this convention, your first three lines would look like this:
```
import Typography from 'material-ui/Typography';

import './about.css';
```

Like I said, this is a convention, not a rule - so feel free to develop a better convention and document it somewhere, but for now this is the way that most of the other files are written, so let's stick with it. I should document it somewhere!

src/views/About/index.js
```diff
@@ -0,0 +1,151 @@
...
+          </div>
+          <div>
+            <h4>
+              <Typography type="subheading" gutterBottom>
+              Software Development Team, 2016/2017
+              </Typography>
+            </h4>
+            <p>
+              <Typography type="body2" gutterBottom>
+              Ross and Chris Qiao &apos;17
```
AdamVig: It looks like this line got cut off. On the current site, it reads:
> Bradley Boutcher '18, Matthew Felgate '18, Jenny Kim '19, Sam Nguyen '19, Joseph Ross and Chris Qiao '17

src/views/About/index.js
```diff
@@ -0,0 +1,151 @@
...
+                <li>Subscriber email feeds</li>
+                <li>Membership affinity connections</li>
+                <li>Student leader and Advisor Activity self-management</li>
+                <li>Rosters and administrative records</li>
+                <li>Student Membership and Leadership Experience Transcript</li>
+              </Typography>
+            </ul>
+          </div>
+          <div>
+            <h4 className="titletext">
```
AdamVig: The `Typography` component actually renders its own header element (`h2` in this case), so you can remove the surrounding `h4` tags.

src/views/About/index.js
```diff
@@ -0,0 +1,115 @@
...
+          </Typography>
+        </p>
+        <Typography type="subheading" gutterBottom>
+              Software Development Team, 2016/2017
+        </Typography>
+        <p>
+          <Typography type="body2" gutterBottom>
+              Bradley Boutcher &apos;18, Matthew Felgate &apos;18, Jenny Kim &apos;19, Sam Nguyen
+              &apos;19, Joseph Ross
+              Ross and Chris Qiao &apos;17
```
AdamVig: Looks like "Ross" is repeated twice here.

PR #16, opened 2017-11-19
Update roadmap

AdamVig:
Updated the roadmap to show what is done and what is in progress.

I also made some changes to the roadmap itself: merging pages, adding things we missed, and removing checkboxes for things we are not going to build.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -41,7 +41,7 @@ To check off items on the list, open `ROADMAP.md` in your editor and place an `x
     - `StatDaysLeftInSemester`
   - unknowns
     - what is the best charting library to use with React?
-- [ ] carousel
+- [ ] carousel (in progress @ajabbot)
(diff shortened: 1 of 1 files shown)


AdamVig:
@ajabbot Sounds good! Go ahead and mark it as in progress. (Pull, make that change, commit,and push.)

ROADMAP.md
```diff
@@ -15,35 +15,38 @@ To check off items on the list, open `ROADMAP.md` in your editor and place an `x
 
 ## Application
 
-- [ ] toolbar
+- [x] toolbar
   - name of application, global menu, and people search
```
RFStauffer: Do we have a people search yet?

AdamVig: Not yet, but there is a separate item for it at the end of the roadmap. This item is just a description of what is in the toolbar on the current 360 site.

PR #17, opened 2017-11-27
Page title

AdamVig:
This PR adds dynamic page and document titles, fixes the header scrolling with content, and fixes some syntax in the Help and About pages.


src/app.js
```diff
@@ -41,11 +37,14 @@ export default class App extends Component {
...
         <GordonNav onDrawerToggle={this.onDrawerToggle} drawerOpen={this.state.drawerOpen} />
         <main className="app-main">
-          <Route exact path="/" component={Home} />
-          <Route exact path="/about" component={About} />
-          <Route path="/activities" component={ActivitiesAll} />
-          <Route path="/activity/:activityId/edit" component={ActivityEdit} />
-          <Route path="/help" component={Help} />
+          {routes.map(route => (
+            <Route
+              key={route.path}
```
ajabbot: Are key and path supposed to be the same thing?

AdamVig: Good question! The `key` is a required attribute and must be unique for each route. I would use the array index of the routes, but our ESLint rules say to not do that (I didn't look into it, but I'm sure they have a good reason for it). So, since paths should pretty much always be unique (technically you can have two routes with the same path, but we probably will not use that feature), they seem like the best candidate for a unique key in this situation (short of generating a unique key somewhere else).

ajabbot: Ok, that makes sense

src/components/Header/index.js
```diff
@@ -4,11 +4,30 @@ import Typography from 'material-ui/Typography';
 import IconButton from 'material-ui/IconButton';
 import MenuIcon from 'material-ui-icons/Menu';
 import PropTypes from 'prop-types';
-
 import React, { Component } from 'react';
+import DocumentTitle from 'react-document-title';
```
ajabbot: When I try to build this branch, it fails to compile and says 'module not found: Can't resolve 'react-document-title'. Do I need to npm install react-document-title?

AdamVig: Just `npm install`. Since I saved the dependency to `package.json`, running that one command will ensure that all the listed modules are installed.

ajabbot: Ok, thank you. Changes look good!

AdamVig:
This pull request does not address dynamic page titles, which will come into play when we add the activity profile and user profile views. For those views, we will want to use the user's name or the activity's name as the title, which will be a bit more complex than this current implementation.

Making a note of this so that we don't forget to build that feature.

PR #19, opened 2017-12-01
Add analytics and error handling

AdamVig:
- Add Google Analytics dependency
- Change router configuration to support firing an event every time the route changes
This enables support for tracking pageviews based on route changes.
- Add `Error` component for displaying errors
- Add analytics service that abstracts analytics functions (in case underlying API for analytics ever changes)
- Add environment variable for Google Analytics ID

The error component looks like this:
screen shot 2017-11-30 at 23 46 00


src/app.js
```diff
@@ -17,10 +17,19 @@ export default class App extends Component {
   constructor(props) {
     super(props);
 
+    // Only use analytics in production
```
ajabbot: What's the reasoning for only enabling analytics during production?

AdamVig: Analytics are useful for analyzing how much usage the app is getting. Since we can provide the `user_role` dimension to the analytics platform, we will be able to see usage by user type ("student", "faculty", "police", etc.).

During development, our usage will not be realistic (I would not actually be using the app at all if I wasn't working on it, for example), so logging that usage would throw off our overall analytics numbers. Furthermore, I often log in using a test account that has the "god" role, so if we were logging usage during development, it would look like there were very active users with the "god" role, but in reality it would just be me working on the app.

ajabbot: Ok, that makes sense. Thank you.

PR #22, opened 2017-12-03
User profile preview

AdamVig:
Adds the current user's profile picture, name, and email address to the top of the navigation drawer.

This pull request also adds a user service (which will need to be merged with the user service in #12) that has methods to get user images, get local info stored in the auth token, and get user profiles.

screen shot 2017-12-02 at 22 29 25


src/components/Nav/components/NavAvatar/index.js
```diff
@@ -19,36 +20,64 @@ class GordonNavAvatar extends Component {
...
+    this.setState({ name, username });
+    const [
+      { Email: email },
+      { def: defaultImage, pref: preferredImage },
+    ] = await Promise.all([
+      await user.getProfileInfo(),
+      await user.getImage(),
+    ]);
+
+    const image = preferredImage || defaultImage;
```
mFelgate: So this is the first time I'm seeing this notation. I get the idea of it but how does it decide which one to do, or is it logically have to be one or the other?

AdamVig: Good question! `||` means "or", and JavaScript uses [falsy/truthy checking](https://www.sitepoint.com/javascript-truthy-falsy/), so this line essentially means:
"the variable `image` should equal `preferredImage` if it is truthy or `defaultImage` otherwise".

When `preferredImage` is defined, it will be truthy, so it will be used as the value of `image`, and when it is undefined, it will be falsy, so `defaultImage` will be used as the value of `image` instead.

PR #23, opened 2017-12-04
Fix session and user service JSDoc

AdamVig:
There were a few errors in the JSDoc that were causing the precommit hook to fail.


AdamVig:
Going to override the review requirement and merge now so I can move on to other things without this getting in the way.

PR #24, opened 2017-12-04
Replace Markdownlint with ReMark

AdamVig:
`markdownlint-cli` does not support globs on Windows, making it a poor choice. ReMark does not have as many rules available as Markdownlint, but will be sufficient to enforce best practices and consistency in Markdown.

This change will fix the pre-commit script problems on Windows. After this pull request is merged, pull master and run `npm install` and then Windows developers should be able to commit again.


README.md
```diff
@@ -23,7 +23,6 @@ The following extensions are recommended for any code editor used to develop thi
 
 - [EditorConfig](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) ensures that indentation style, line endings, and file endings are consistent across editors and operating systems
 - [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) for JavaScript (`.js` files)
-- [Markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) for Markdown (`.md` files)
```
RFStauffer: No mention of using ReMark instead of `Markdownlint`, but 1) would you need to mention it? If so, 2) would here would make sense?

AdamVig: Good question. I could not find a ReMark extension that provided live feedback like the Markdownlint extension did, so I simply took out the Markdown linting extension recommendation. ReMark is also a lot less strict, so it should not require live feedback as much.

PR #25, opened 2017-12-05
Remove token refresh

AdamVig:
- Fix CL&W credits chart bugs
    - rename component to match title: `ChapelProgress` to `CLWCredits`
    - change chart column size on medium screen to fill the space better
    - display CL&W credits chart even when there are zero credits This was done by setting a default `required` value of `30`, which will be correct for the majority of students, and will not bother users who do not have a CL&W credit requirement.
- Improve component error handling
    Now, many of the components that use data from the backend will throw an error if any of their data requests fail. At the moment, these errors are handled in two ways: 1. if the error is an authentication error, the user is redirected to the login view 2. otherwise, the error is displayed using `GordonError` This error handling is provided by the `componentDidCatch` method of `App`. In the future, this could be made more granular by using React error boundary components, Material-UI snackbars, and more specific error types.
- Add debug configuration for VS Code
    This allows for debugging with Chrome.
- Remove token refresh from auth service
    Since token expiration is now disabled on the backend, the buggy token refresh support can be removed from the auth service.
    - remove token refresh methods
    - store token in localStorage key `token` instead of as a property in key `auth`
    - add error service to create errors based on HTTP errors from the backend
    - add `AuthError` This allows conditionals like `if (error instanceof AuthError) { ... }`, which will make it easy to redirect the user to the login view. when there is an authentication error.
    - Add error handling to `parseResponse` method in http service
    - Use `parseResponse` from http service in auth service `getAuth` method
    - Refactor HTTP functions in http service to have less code duplication
    - add `NotFoundError` This allows for more effective handling of missing resources and incorrect URLs.

### DIFF
--- a/src/services/user.js
+++ b/src/services/user.js
@@ -6,6 +6,7 @@
 
 import jwtDecode from 'jwt-decode';
 
+import { AuthError } from './error';
 import http from './http';
 import session from './session';
 import storage from './storage';
@@ -174,27 +175,6 @@ const getAttendedEvents = (username, termCode) =>
   http.get(`events/chapel/${username}/${termCode}`);
 
 /**
- * return the number of cl&w credits aquired, and number of credits required.
- * @return {Promise.<CLWCredits>} An Object of their current and requiered number of CL&W events,
- */
-const getChapelCredits = async () => {
-  const { username } = storage.get('credentials');
-  const termCode = session.getTermCode();
-  const attendedEvents = await getAttendedEvents(username, termCode);
-
-  // Get required number of CL&W credits for the student, defaulting to zero
-  let required = 0;
-  if (attendedEvents.length > 0) {
-    ([{ Required: required }] = attendedEvents);
-  }
-
-  return {
-    current: attendedEvents.length,
-    required,
-  };
-};
-
-/**
  * Get image for a given user or the current user if `username` is not provided
  * @param {String} [username] Username in firstname.lastname format
  * @return {Promise.<String>} Image as a Base64-encoded string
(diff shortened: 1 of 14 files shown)


AdamVig:
That commit description was actually intentional. I often condense several changes into a commit, but itemize all of the small changes that make up the commit so it's clear what happened in the commit.

Matt originally had the column width set to 6, but we decided the charts looked silly at that size. A better solution would probably be to shorten the title to "Days Left" so that it would not wrap on smaller screen sizes. What do you think?

RFStauffer:
Oh ok.

Are we going to have anything else, like an image, on the home screen? That might determine how the cards are placed on the page. I was looking at the sm and xs and noticing how the cards go to the edge and fill the page, so I thought it could be similar in md and lg. As it is, the sm at 959px makes the charts quite big, I think bigger than the charts in md, then resizes to smaller once the screen is md size. I guess I just thought the spacing on both sides was odd when resizing in md and lg. We could potentially rename the card header, but I just remember from your mobile app the "Days Left in Semester" made sense, otherwise, it would be "Days Left," in what?

AdamVig:
All good points! Addison is working on the carousel, which will be the only other thing on the home page in the near future. The carousel will be on its own row below the charts.

As for removing "...In Semester," you are right: it would make the chart ambiguous. I think we could change it so that the title would be "Days Left" and the subheading would be "16 Days Left In Semester." That way, a user could easily clarify the meaning of the chart. Does that sound like a good solution?

RFStauffer:
Good solution. Having the carousel below the charts will stress the importance of having the cards even on the bottom.

src/services/user.js
```diff
@@ -215,18 +195,39 @@ const getLocalInfo = () => {
...
+ * Get the number of cl&w credits aquired, and number of credits required.
+ * @return {Promise.<CLWCredits>} An Object of their current and requiered number of CL&W events,
+ */
+const getChapelCredits = async () => {
+  const { user_name: username } = getLocalInfo();
+  const termCode = session.getTermCode();
+  const attendedEvents = await getAttendedEvents(username, termCode);
+
+  // Get required number of CL&W credits for the user, defaulting to thirty
+  let required = 30;
```
mFelgate: this is okay but the required field is in the account view. At some point we have to move over to that api for required credits cuz 30 will still be inaccurate for a large portion of the student body

PR #26, opened 2017-12-05
Add docs script and move guides to separate file

AdamVig:


AdamVig:
Thanks Rebekah! That file provides configuration for JSDoc, which generates the documentation.

PR #29, opened 2017-12-15
Activity profile read only page

RFStauffer:
Multiple contacts: Computer Science club activity_201709_comps
Full Info: AJ Gordon Scholars Programactivity_201709_ajg
Minimum info: AJ Gordon Missions Fellows activity_201709_ajmiss


src/views/ActivityProfile/activity-profile.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../vars';
+
+.gordon-activity-profile {
+  .activity-image {
```
AdamVig: With larger activity images, the image goes beyond the width of the card.
With a longer list with names of different lengths, the email icon 
screen shot 2017-12-16 at 12 58 20
We can fix that by limiting the width of the image so it fits in its container.
```CSS
.activity-image {
  margin-top: 25px;
  margin-bottom: 25px;

  img {
    max-width: 100%;
  }  
}
```

src/views/ActivityProfile/index.js
```diff
@@ -0,0 +1,174 @@
...
+    } else {
+      const { ActivityDescription: activityDescription } = this.state.activityInfo;
+      const { ActivityImagePath: activityImagePath } = this.state.activityInfo;
+      const followers = this.state.activityFollowers;
+      const members = this.state.activityMembers;
+      const { SessionDescription: sessionDescription } = this.state.sessionInfo;
+      let admins;
+      let groupContacts;
+      if (this.state.activityGroupAdmins.length > 0) {
+        groupContacts = (
```
AdamVig: There are a few things that we could improve about this part of the component's code. First, the whole section could be combined into a single variable, which would remove the confusion of using both `groupContacts` and `admins` (it took me a while to figure out that they were two variables for the same section). Second, the `.map()` function is actually generating a whole independent `<List>` for each item, when it should instead be generating `<ListItem>`s for each item of the list, contained by a single `<List>` component. Third, the parameter passed to the `.map()` function's callback should be singular rather than plural, because it represents a single `activityGroupAdmin`, not multiple. Lastly, the `<List>` component has a `className` of `gordon-activity-profile`, which is redundant in this case because this class is already present on the `<CardContent>` component.

Here is how I would refactor it:

```JavaScript
      let groupContacts;
      if (this.state.activityGroupAdmins.length > 0) {
        groupContacts = (
          <Typography type="body1">
            <strong>Group Contacts:</strong>
            <List dense disablePadding>
              {this.state.activityGroupAdmins
                .map(activityGroupAdmin => (
                  <ListItem key={activityGroupAdmin.Email}>
                    &emsp;{activityGroupAdmin.FirstName} {activityGroupAdmin.LastName}
                    <IconButton
                      classes={({ root: 'email-button' })}
                      color="primary"
                      href={`mailto:${activityGroupAdmin.Email}`}
                      padding={0}
                    >
                      <Email
                        color="primary"
                        style={{ width: 16, height: 16 }}
                      />
                    </IconButton>
                  </ListItem>
                ))}
            </List>
          </Typography>
        );
      }
```

Can we change it to something like the above so the code communicates a bit more clearly?

RFStauffer:
localhost_3000_activity_201709_comps-12 21
localhost_3000_activity_201709_ajg-12 21
Subscribe and Join buttons not functioning, but this is a read-only form of the page.

src/services/activity.js
```diff
@@ -32,6 +48,42 @@ const getAll = sessionCode =>
...
+ * @returns {Number} Number of followers
+ */
+const getFollowersNum = (activityCode, sessionCode) =>
+  http.get(`memberships/activity/${activityCode}/followers/${sessionCode}`);
+
+/**
+ * Get group administrators for an activity
+ * @param {String} activityCode Identifier for an activity
+ * @param {String} sessionCode Identifier for a session
+ * @returns {Person[]} List of group administrators
```
AdamVig: What does `Person` refer to? In order to use a JSDoc type, you must first define that type.

Look at the docs for [`@typedef`](http://usejsdoc.org/tags-typedef.html) to learn more. Use the type definitions in the other services for examples.

RFStauffer:
Ok, thanks. I think I would like to merge it (can you accept?), and then if I have time, or Matt has time, we can work on breaking it down before the semester begins.

PR #34, opened 2018-01-03
Add Event calendar

mFelgate:


RFStauffer:
Same with Addison about the filters

image
There's something odd happening with the drawers. The above one seems docked, and then when I click on something, another one appears over it and isn't docked.
image
`Develop` branch doesn't have this issue.

RFStauffer:
`Description` not `Discription` spelling error for additional info for events

src/views/Events/components/EventItem/index.js
```diff
@@ -0,0 +1,87 @@
...
+              {subheader}
+            </Typography>
+            <Typography type="caption" >
+              {event.location}
+            </Typography>
+          </CardContent>
+          <Collapse in={this.state.open} timeout="auto" unmountOnExit>
+            <CardContent>
+              <Typography>
+                Discription
```
RFStauffer: Description not Discription spelling error for additional info for events

src/views/Events/index.js
```diff
@@ -0,0 +1,202 @@
...
+                 }
+                  label="Athletics"
+                />
+              </FormGroup>
+              <Divider light />
+              <FormGroup row>
+                <FormControlLabel
+                  control={
+                    <Switch
+                      checked={this.state.chapelCredits}
```
RFStauffer: Why is this label different from the rest? It doesn't seem to work as the others do.

src/views/Events/index.js
```diff
@@ -0,0 +1,202 @@
...
+                    />
+                 }
+                  label="Fair or Expos"
+                />
+                <FormControlLabel
+                  control={
+                    <Checkbox
+                      checked={this.state.sports}
+                      onChange={this.filterEvents('Sports')}
+                    />
```
RFStauffer: The `Athletics` filter doesn't seem to be working. The checkbox isn't changing and the events aren't being filtered.
Also, it is somewhat confusing to have `sports` when the actual filter label is called `Athletics`.

mFelgate:
hey guys! this is way too late but yeah, I was actually showing someone the spacing and forgot to change it back, the vents show up now that there are some listed. and i also changes some colors in the app as a whole

src/views/Events/index.js
mFelgate: that was a database thing it is good now

mFelgate:
and fixed spacing!

PR #41, opened 2018-02-11
Pull Request for CPS 492- Using variables instead of 

nicholas-miller:
Inviting @gordon-cs/cps492-2018 for the pull request assignment.

package-lock.json can be ignored. Primary changes are with regard to the way data is read by the Orientation index.js, replacing numbers and direct references with variables instead, allowing for easier modification. Current variable values are hooked up to random number generators in order to test proper operation- such as the avatar turning red when incomplete and green when complete.

Would especially appreciate feedback on specific location of variables within the code, and suggestions related to the current naming conventions.

The statements 'green-avatar', 'yellow-avatar' and 'red-avatar' refer to specific classes in the CSS file that define the specific color.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha1-+PLIh60Qv2f2NPAFtph/7TF5qsg=",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
       "dev": true
     },
     "accepts": {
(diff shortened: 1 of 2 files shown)


DonutPirate:
LGTM. Since you're the one able to run tests I would make sure to mess around with possible problems with floats when you're calculating task completion percentages, but it looked fine. Also since "Completed!" ends in a punctuation mark you might want to consider adding a period to "Not received", so they match better stylistically. That could go either way though; just a suggestion.

Natoman95:
I think it could do with some commenting. I like to split up the HTML into logical sections with comments above them, especially with a long chunk of HTML like this. It'd make it much easier to tell what elements are meant to display what. A class comment describing the overall purpose of the class would be useful as well. I can't really tell what the page is supposed to be doing on the whole.

cfrandsen09:
I agree with what Nathan said about comments, both for the class and the code. Variable names are good and useful, I'm able to figure out what they are meant to do. I would suggest using a boolean for the if statement at line 27 to indicate whether the ID photo has been sent in or not (I realize it's currently set up for testing).

BradleyBoutcher:
LGTM. Commenting, as others have said, would be useful for code reviews and for the future of the code. It's important to state what exactly you're trying to do, so we can determine if you're doing it properly. I'd also recommend using more className conventions, to make it easier to track down specific items.

mFelgate:
make sure you use typography for the text, you can see examples on the material website and on the development branch.

PR #44, opened 2018-04-16
Orientation Roadmap Updates

nicholas-miller:
Using variables instead of hard data, setting all the variables outside of the HTML, added comments, and added official descriptions courtesy of Chris Carlson.


sam-nguyen7:
lgtm. variables are always better than hard code

PR #47, opened 2018-05-13
Activity profile static page (with master)

RFStauffer:
Some examples.
activity-edit member
localhost_3000_activity_201609_ajg


RFStauffer:
Hmm. This seems unrelated to my branch since I didn't touch CLW credits. I can still build properly, but I have gotten some home page errors before like this, so it might be finicky. Can you try to skip this by typing in the URL "http://localhost:3000/activities" to get to the list of activities?

PR #65, opened 2018-06-14
Transcript Page

ajabbot:
The Transcript page has been formatted with activities grouped by session, and the css has been cleaned up. the activities filter background color has also been updated for consistency.

Desktop:
image

Mobile:
image

Download PDF:
image


NathanielRudenberg:
The name and email address on the navbar are pushed over to the right, which causes clipping and leaves unused space to the left:
button_go_to_profile

Any chance that could be fixed before merging?

src/services/user.js
```diff
@@ -315,6 +315,46 @@ const getMemberships = async id => {
   memberships = await http.get(`memberships/student/${id}`);
   return memberships;
 };
+
+const getTranscriptInfo = async id => {
+  let transcriptInfo = await getMemberships(id);
+  console.log(transcriptInfo);
+  transcriptInfo.activities = [0];
+  let k = 0;
```
russtuck: There are a lot of loop variables here (4).  Could you explain what they are for?  

"k" is puzzling me the most.

src/services/user.js
```diff
@@ -315,6 +315,46 @@ const getMemberships = async id => {
   memberships = await http.get(`memberships/student/${id}`);
   return memberships;
 };
+
+const getTranscriptInfo = async id => {
```
russtuck: A comment describing the result (structure and content) would be helpful.

src/services/user.js
```diff
@@ -315,6 +315,46 @@ const getMemberships = async id => {
...
+      } else {
+        let obj = [];
+        while (k !== first) {
+          obj.push(transcriptInfo[k].ActivityDescription);
+          k++;
+        }
+        transcriptInfo.activities.push(obj);
+      }
+    }
+    first++;
```
russtuck: It looks like "first" will be incremented twice per iteration, moving ahead twice as fast as "second".  Is that what's intended?

ajabbot: This was written by Matt, I'll try to make sense of the logic and document it accordingly

src/views/Transcript/Components/ActivityList/index.js
```diff
@@ -0,0 +1,75 @@
+import Grid from 'material-ui/Grid';
+import Divider from 'material-ui/Divider/Divider';
+import React, { Component } from 'react';
+import List from 'material-ui/List';
+import Typography from 'material-ui/Typography';
+import './activityList.css';
+
+export default class Activities extends Component {
```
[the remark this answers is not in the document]
ajabbot: Components that will be used by several other components (such as loader and error) should be created in under the src/components/ folder. Components that are only implemented for specific views (such as Transcript, Carousel, etc), should be created under a folder for their respective view

src/views/Transcript/Components/ActivityList/index.js
```diff
@@ -0,0 +1,75 @@
...
+        </Grid>
+      </Grid>
+    );
+
+    return content;
+  };
+
+  render() {
+    return (
+      <div>
```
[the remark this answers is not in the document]
ajabbot: Only one heading and content is generated per activity, but the Transcript Component displays a list of this ActivityList component, so more than one are shown on the Transcript. I have updated the naming of this component to better reflect this.

src/views/Transcript/Components/ActivityList/index.js
```diff
@@ -0,0 +1,75 @@
...
+            </List>
+          </Grid>
+        </Grid>
+        <div className="divider">
+          <Divider />
+        </div>
+        <Grid />
+      </div>
+    );
+    if (!this.props.isUnique) {
```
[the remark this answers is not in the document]
ajabbot: This is a prop passed down to it from its parent component. i have added a comment to clarify this

src/views/Transcript/Components/TranscriptActivity/index.js
```diff
@@ -3,9 +3,14 @@ import Divider from 'material-ui/Divider/Divider';
 import React, { Component } from 'react';
 import List from 'material-ui/List';
 import Typography from 'material-ui/Typography';
-import './activityList.css';
+import './transcriptActivity.css';
 
-export default class Activities extends Component {
+//This components is a child of the Transcript component. Seperates Headings from content in order
```
russtuck: Much improved, thank you.

"This component" (no "s"). 
"Separates" (not "Seperates")

src/views/Transcript/index.js
```diff
@@ -19,6 +22,8 @@ export default class Transcript extends Component {
       loading: true,
       profile: {},
     };
+    //Variable used by SessionComparitor to determine when activies in the list are part of a new session
+    //Necessary in order to remember session info about previous item in list while looping though
```
russtuck: Thanks.  Would it make sense to name the variable "previousSessionVal" (or "prevSessionVal")?

src/services/user.js
```diff
@@ -316,42 +316,24 @@ const getMemberships = async id => {
   return memberships;
 };
```
russtuck: Do we have a convention for comment format?  In particular, do we prefer 
  // Comment
or
  //Comment
?
(Personally, I prefer a space after //.  But being consistent with existing code is more important than personal preference.)

src/views/Transcript/index.js
```diff
@@ -44,15 +49,20 @@ export default class Transcript extends Component {
     }
   }
 
-  //Compares each activity from TranscriptInfo to determine how to group by session
+  //Compares activity from activityList to previous activity' session to to determine how to group
```
russtuck: Please end this line with a period, so it doesn't read like "... how to group isUnique ..."
(Next line also needs a period.)

src/views/Transcript/index.js
```diff
@@ -44,15 +49,20 @@ export default class Transcript extends Component {
...
+  //isUnique value is passed as a prop, along with Activity object, to TranscriptActivity Component
+  //Returns TranscriptActivity component which is passed into activityList
   sessionComparitor = activity => {
```
russtuck: Given what all it does, is there a better name?
Maybe activityInSession? 
(That's not great.  You can probably do better.)

src/views/Transcript/index.js
```diff
@@ -44,15 +49,20 @@ export default class Transcript extends Component {
     }
   }
 
-  //Compares each activity from TranscriptInfo to determine how to group by session
+  //Compares activity from activityList to previous activity' session to to determine how to group
+  //isUnique value is passed as a prop, along with Activity object, to TranscriptActivity Component
+  //Returns TranscriptActivity component which is passed into activityList
```
russtuck: Very helpful comment.  Thank you.

Maybe "...which should be passed into..." would be more accurate, since that's done outside this function.

src/views/Transcript/transcript.scss
```diff
@@ -18,14 +20,6 @@
     visibility: visible;
     overflow: visible;
   }
-  #section-to-print {
```
russtuck: Nice job eliminating what's not needed, as well as documenting the file!

src/services/user.js
ajabbot: We don't have a convention (Adam probably did, but I don't think we've been following one). Should we decide on one? Spaces are good with me, I've been commenting with no whitespace, but I'm not particularly attached to that style

PR #70, opened 2018-06-15
Profile page

NathanielRudenberg:
The UI of the profile page has been cleaned up some:
profile_page_cleaned_up_desktop


src/views/Profile/index.js
```diff
@@ -94,7 +94,7 @@ export default class Profile extends Component {
 
     const style = {
       img: {
-        width: '75%',
+        maxWidth: '100%',
```
russtuck: Since you've tried various values, would it be helpful to put a comment here about why this is right, and what the possible values mean?

PR #93, opened 2018-06-21
Social media links

EmJoBi:
Social media links implemented. Please help me improve the code if you see ways to do so!

-views/MyProfile
-views/MyProfile/Components/LinksDialog
-src/user


src/views/Profile/Components/LinksDialog/index.js
```diff
@@ -0,0 +1,226 @@
+import React from 'react';
```
ajabbot: I think this is an extra file left over from when you switched to myProfile. Can this be removed?

src/socialMedia.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
ajabbot: Does this file need to be in the /src directory, or could it be moved to the /src/MyProfile/Components/LinkDialog directory of something? It works fine (and I really like your implimentation of this), I would just prefer to not add files to the /src directory unless necessary

EmJoBi:
See what you think about borders around the icons. It makes them look like buttons, but I'm super on the fence about them.

PR #94, opened 2018-06-21
Toolbar changes

NathanielRudenberg:
- The NavBar is no longer visible on desktop. Instead, the main webpage content fills the entire browser width and the NavLinks are now located in the toolbar :sparkles: 
desktop toolbar

- The toolbar NavLinks become hidden on mobile and the NavBar becomes visible as a NavDrawer.
mobile home view

screenshot from 2018-06-21 16-13-57

Users can now access "My Profile" by clicking on the Nav Avatar in the toolbar and selecting the option from the dropdown menu.
global menu

Pair programmed.


src/components/Nav/index.js
```diff
@@ -18,6 +30,16 @@ export default class GordonNav extends Component {
       </div>
     );
 
+    // let navCSS;
```
ajabbot: Can this be removed?

PR #98, opened 2018-06-22
Upload/Crop Tweaks & Hide/Show Profile Picture

scmac97:
screenshot from 2018-06-22 14-02-05

****Added Features:****
- Hide/Show Profile Picture
- Explanatory tooltips
- Better cropper-sizing "heuristic"

Any suggestions?


scmac97:
I haven't been able to reproduce the Unhandled Rejection error, no matter how many times I toggle the hide/show button. Maybe this is a problem with http.js? That's really strange; I have no idea why that happened.

scmac97:
The avatar component displayed on the toolbar will have to be modified to include the same methods Max added to the NavAvatar component on the navbar.

scmac97:
I've updated the colors and order of the buttons. [This page](https://material.io/design/components/dialogs.html#actions) on the Material Design site shows the cancel button on the left, like Dr. Tuck said. It also says to have buttons be disabled until they are active, so I've kept that (hope you don't mind). :)

They also have all additional buttons left justified (like you mentioned), but I haven't been able to do that. When I've tried using a grid, things started acting weird, but I might not be doing it right. Feel free to make that change if you want to.

image
image

The Google colors looked so cheerful, though... :)

scmac97:
We've determined that (2) is a fluke caused by EXIF image rotation metadata. And (3) is an error already extant in the develop branch; I'm not sure that it is related solely to the image uploader.

PR #101, opened 2018-06-22
Avatar Improvements

scmac97:
Added animation to drop shadow. Yay!


scmac97:
Made further changes: now avatar images that aren't square will properly fill the avatar circle; looks a bit nicer! :)

PR #105, opened 2018-06-25
Add Gordon seal as toolbar background

NathanielRudenberg:


scmac97:
BTW, this looks so much better than the old one!

PR #106, opened 2018-06-25
Fix my profile

TheGreatMaximus:
Added some UI Changes to make profile information align correctly and have faculty information formatted correctly. Also added a slider instead of a button.


.env.production
```diff
@@ -1 +1 @@
-REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+REACT_APP_API_URL=https://360Api.gordon.edu/
```
ajabbot: I think this should remain as train until we merge into master

src/views/MyProfile/index.js
```diff
@@ -54,8 +55,7 @@ export default class MyProfile extends Component {
     };
   }
 
-  handleExpandClick() {
-    this.changePrivacy();
+  handleChangePrivacy() {
```
ajabbot: It looks like you renamed handleExpandClick(), but left it the same in line 35...
image

src/views/MyProfile/index.js
```diff
@@ -64,7 +64,7 @@ export default class MyProfile extends Component {
   };
 
   handleCloseSubmit = () => {
-    if (this.state.preview) {
+    if (this.state.preview != null) {
```
ajabbot: I'm not sure this is necessary, did you just add it to improve readability?

TheGreatMaximus: I think I did it because there was no reason to compare something to null when state preview will evaluate to true or false based on if it exists.

PR #109, opened 2018-06-25
Update to stable material ui

ajabbot:
I've updated the imports, and some function names and naming conventions to support the transition over to v1.2.3 of Material-UI. Everything seems to be mostly working, but some more rigorous testing is probably necessary. Some minor issues I've noticed:

- Grid spacing now defaults to 0, so we will have to add spacing between our cards and stuff
- The search text field has a weird white card over it, autocomplete still works fine though
- The Events loading circle is kinda weird, not sure why, it's using the same GordonLoader
- Buttons now default to not being text buttons, we can update them to outlined buttons (which look dope) and contained buttons (which is what they were before) as we see fit


scmac97:
Just FYI—the default grid spacing used to be 16. So we'll be typing lots of 16s tomorrow. :)

PR #117, opened 2018-06-27
New Search Bar!

scmac97:
image

It animates (width and drop shadow) on focus.


softwary:
It looks nice so far on desktop view.

Issues on mobile view:
The avatar runs off the page and there is a white column that appears underneath it now.

I like this new look though, nice job.

scmac97:
Added more fixes; works better now. The issue with mobile view occurred with the old search bar, too. ATM, it only occurs when the title is long (looks fine on "Home", which is short). On phone-sized screens we'll have to replace the search bar with a search button which toggles a slide-down search bar.

PR #126, opened 2018-07-02
 People search only lets you reach your first search, subsequent searches do not work (7)

NathanielRudenberg:
People search is now functional. You can search for new people indefinitely without needing to refresh the page to get new data to show up.


softwary:
Yay! It works. Nice fix.

NathanielRudenberg:
Woah, that is strange. I do remember I pulled changes from develop into this branch, then un-staged and dismissed those changes as it did not accomplish what I had intended. I'll look into this.

NathanielRudenberg:
Closing this, since I would rather recreate the fix and not reintroduce the problem instead of trying to undo the new problem ¯\_(ツ)_/¯

PR #129, opened 2018-07-02
Include past events

EmJoBi:
Added button to Events page to toggle inclusion of past events. The database seems to record past events back to the start of the term. It is currently showing events of the summer term.


EmJoBi:
It looks a bit better now. See if you think it's good enough for the time being. The grid padding is being a bit tricky.

PR #178, opened 2018-07-06
Set mobile phone privacy on MyProfile now works (3)

softwary:


softwary:
**This still needs to have a snackbar (or other type of message) -confirmation in order to be merged!**

PR #191, opened 2018-07-09
Past events

EmJoBi:
If someone would like to look at this with me, I believe it's reached a point where it can be merged. 

In order to test it, I'm running it with a change to my local instance of the back end, which is why I'm saying look with me. 

For you guys, the page should just display "No Events to Show".


scmac97:
Just glanced at the changed files briefly. Did you mean to include package.json and package-lock.json in your commit?

PR #226, opened 2018-07-12
added autocomplete tag and username id to login

TheGreatMaximus:
Not sure if this is normal with all chrome autofill but when the page loads the information is automatically entered rather than having to specify which username and password pair you would like to use for the form


TheGreatMaximus:
screenshot from 2018-07-12 11-08-23

PR #228, opened 2018-07-12
Private Involvements on MyProfile (3)

softwary:
Fixes issue #229 
When you are in a secret involvement, there will no longer be a toggle. Instead it will be a lock icon, signifying to the user that they can't change whether this is displayed or not (it just will _not_ be).

Mobile view:
image

Desktop:
image


scmac97:
Would it be helpful to add a tool-tip that pops up when you hover over the lock icon which explains that the club itself is hidden, so they can't choose to show it?

PR #234, opened 2018-07-13
Added report bug to help page

ajabbot:
Based on feedback from our user testing.
Users went to the help page first to find the report bug feature, and for some reason it was only on the about page. I made the link an uncontained button instead of just a linked text, to be consistent with the rest of the website.

image


TheGreatMaximus:
Oh I see you just reordered the code a bit, it looks good

PR #236, opened 2018-07-13
Remove duplicate Navatar on mobile (2)

softwary:
Fixes issue #233 

Right Corner Navatar is gone, those menu options are now in the NavDrawer.
image

**Search bar:**
(351 just under xs)
image

(399)
image

(951 just under md)
image

(999)
image

(every size above 1000)
image


TheGreatMaximus:
Looks good thanks for the pictures

PR #245, opened 2018-07-16
Made student id grayed out

ajabbot:
Fixes issue#243

image


TheGreatMaximus:
Can you add grey text to the home street and home as well?

PR #247, opened 2018-07-16
Remove snackbar from mobile phone toggle, add check mark (1)

NathanielRudenberg:
snackbar screenshot


TheGreatMaximus:
Can you provide pictures?

NathanielRudenberg:
...Of what? I don't know how to take a picture of an element when it no longer appears after a certain action, and I've already added a screenshot showing the other change I made which was adding the checkmark.

TheGreatMaximus:
It wasn't clear where this snackbar was appearing that is why I asked for a picture of the page and not just the snackbar. Also if you are using snapshot there is a delay feature for taking a snapshot its rather useful.

PR #249, opened 2018-07-16
Feedback update

ajabbot:
Got rid of the padding around the feedback form.

Desktop:
image

Mobile:
image


TheGreatMaximus:
It looks like will update for components is legacy now

TheGreatMaximus:
This looks good to me, but you may want to remove Component will update because it is considered legacy.

PR #257, opened 2018-07-17
Round login box corners and add background image

scmac97:
image


scmac97:
- Different resolution background image loads based on browser window size (e.g. to save data on mobile devices)
- Added drop shadow and slight transparency

PR #273, opened 2018-07-18
Time since sent

EmJoBi:
Added time since sent to requests. Any advice for improvement?

screenshot from 2018-07-18 11-52-48


softwary:
This looks so nice!!!!!!!!!!!!!!

PR #304, opened 2018-07-30
GoStalk :}

softwary:
**Students' View:**
image

**Only non-students have the option to click "Include Alumni"**
How it looks when you land:
image

What it looks like when you expand everything:
image

The drop downs work :)
image

Once you search, it collapses into this smaller view:
image

Each row looks like this:
image

On Mobile:
image

If a profile doesn't exist:
image


softwary:
One issue we are having:
The checkbox takes a hot second to load on the page.

PR #306, opened 2018-08-01
Requests

EmJoBi:
I added notification badges to show number of requests for each group you are admin of. This way each panel can be default to closed and not fill up the page.

Doing this messed up the requests of activities profile page since we were sharing the component. I took it out of the component to fix this and moved it back into RequestDetail.


src/views/Home/components/Requests/components/RequestsReceived/index.js
```diff
@@ -60,11 +65,29 @@ export default class RequestReceived extends Component {
...
+        <Badge color="error" badgeContent={requests.length} style={badge} className="badge">
+          <Typography variant="title">{this.props.involvement.ActivityDescription}</Typography>
+        </Badge>
+      );
+    } else {
+      title = <Typography variant="title">{this.props.involvement.ActivityDescription}</Typography>;
+    }
+
+    //Requests and buttons
+    let displayedRequets;
```
ajabbot: *displayedRequests

EmJoBi: I moved it a little bit, but for some reason I can't move it to the left with leftPadding, only outward diagonally with padding.

PR #330, opened 2018-08-22
Involvements Bug

softwary:
Fixes the issue where:
"When I click on a past involvement that I wasn't in, then I get the "loading circle" forever (or at least for longer than my patience)."

The issue was that any time anyone wanted to access a specific Involvement, the front end was trying to get a list of all the members' emails (when it should only give it to the admins).


TheGreatMaximus:
This looks good and makes sense to me

PR #349, opened 2018-10-02
Hopefully fixes student/admin problem of viewing dining balance

EmJoBi:
I am using both college_role and personType now because college_role won't work and I'm not sure how well personType would work for student/faculty/staff combinations. So I'm using both. Please correct me if this approach is unnecessary.


jsenning:
This update has been superseded by other work.  College role is no longer a factor in determining meal plan or dining balances.

PR #350, opened 2018-10-03
Tweak CL&W Chart

scmac97:
Replace redundant "CL&W Credits" subtitle at top of card with "CL&W Credits Left" (to match "Days Left").

image


softwary:
So it's not actually "CL&W Credits Left," though. This is your total amount of CL&W credits – thus this doesn't actually make sense. 

(I saw where you were going though. If the number was inverted, it'd make sense, but it's counting up, not down.)

scmac97:
Are you sure? This is what it looks like for me now:
image

Currently the grey text at the top right is redundant, since it mirrors the big cyan number at the bottom right. I thought it would be nice to make it match the grey text on the top left for the number of days left. So in this case, I currently have 11 credits, and I still need to get 9 more. (I'm a commuter, so I only need to get a total of 20 CL&W credits.)

Does that make sense?

PR #382, opened 2018-11-13
tests now compiles. testing error page

sam-nguyen7:
Here's an example of a test that fails. Now that the tests actually compile I'll start with the error page tests.
screen shot 2018-11-13 at 6 51 35 am


sam-nguyen7:
just merging this into my testing branch. No need for review!

PR #401, opened 2019-03-05
Super Admin Page

ajabbot:
image

Includes functions for super admins, including
 - linking to open and closed involvements
 - list of super admins, with ability to remove them
 - add super admin based on given email (@gordon.edu domain inclusion not required)

Does not include (future functionality)
 - stats (number of members, leaders, subscribers) next to each open involvement


softwary:
Changes:
- When removing a super admin, there should be a check/confirmation dialog.
- Chris C. requests that we be able to add admins via their username, not just email. This is possible, just not obvious.
- Expand the 'add super admin' dialog box

Just for clarification:
There are group admins that are admins for each specific group (in the membership table): they have the power to 'confirm final roster'.

Super Admins (in the admin table), on the other hand, can add/remove admins, and have the privileges of an admin in every single group.

softwary:
Centered the SuperAdmin List just a little bit.

image

sam-nguyen7:
Checked it out with Addison and looks good to me.

PR #425, opened 2019-06-07
Links

limyer:
Added Links feature below My Profile tab. Currently supporting both Web and Mobile.

Links are right now hard-coded on src/components/QuickLinksDialog/components/LinksList.


src/components/QuickLinksDialog/index.js
```diff
@@ -0,0 +1,77 @@
...
+import MuiDialogContent from '@material-ui/core/DialogContent';
+import { withStyles } from '@material-ui/core/styles';
+import Typography from '@material-ui/core/Typography';
+
+import PropTypes from 'prop-types';
+
+import GordonLinksList from './components/LinksList';
+import './quicklinksdialog.css';
+
+const styles = theme => ({
```
joshuaRogers24: I would see if you can create a .scss file and make "root" and "button" entries in the style list. This would follow a consistent format with the rest of the code.

joshuaRogers24: Oh, my bad. You have a .scss already! Can root and button be entries on the style sheet or are they something different?

limyer: Root and button are the simple ones that can stay on the original theme and I just added scss for future style reformation.

src/components/QuickLinksDialog/components/LinksList/index.js
```diff
@@ -18,19 +18,19 @@ export default class GordonLinksList extends Component {
...
+                className="gordon"
                 src="https://www.gordon.edu/favicon.ico"
-                alt="my gordon"
+                alt="gordon"
                 width="16"
                 height="16"
               />
             </ListItemIcon>
-            <ListItemText primary="Gordon Main" />
+            <ListItemText primary="Gordon College Official" />
```
russtuck: I'd suggest just "Gordon College" since all of these are official websites of the college.

limyer: I will make that change in UserViewsEnhancement

PR #427, opened 2019-06-07
Profile tidying

joshuaRogers24:


joshuaRogers24:
A few not so urgent TODOs:
-make sure the LinkedIn button appears when it should
-Make sure the font matches production (pull master into develop and preserve master's font)
-put all grid attributes into style sheets
-remove unnecessary sizing attributes

joshuaRogers24:
I reached a point where I became limited by what is known as CSS Flexbox, which is what the code seems to use for designing the ui's framework. I found that using CSS Grid Layout will allow for the fine tuning I was seeking, but it would require a lot if refactoring. I will probably move on to a more important project for a time and leave my changes un-merged.

joshuaRogers24:
**NOT READY FOR REVIEW**

PR #458, opened 2019-06-25
Cct format dates

### LINKED ISSUE #447: (2 pts) Format activity dates on CCT

joshuaRogers24:
resolves #447 , combining consecutive sessions into a hyphenated range to be used for an activity's duration on the CCT.

### DIFF
--- a/src/views/CoCurricularTranscript/index.js
+++ b/src/views/CoCurricularTranscript/index.js
@@ -186,7 +186,7 @@ export default class Transcript extends Component {
 
     // sort activities by ActivityCode
     while (activities.length > 0) {
-      let curAct = activities[0];
+      let curAct = activities.shift();
 
       // keep track of the activity code which will be used to identify all activities of the same
       // code so they can be grouped into one activity component
@@ -194,8 +194,7 @@ export default class Transcript extends Component {
 
       // Pop first session code from array and split into months and years, which are saved as
       // the initial start and end dates
-      // (shift is like pop, removes and returns the first item in the array)
-      let sess = activities.shift().SessionCode;
+      let sess = curAct.SessionCode;
       let startMon = this.sliceStart(sess);
       let endMon = this.sliceEnd(sess);
       let startYear = this.sliceYear(sess),
--- a/src/views/CoCurricularTranscript/index.js
+++ b/src/views/CoCurricularTranscript/index.js
@@ -131,11 +131,26 @@ export default class Transcript extends Component {
       case '01':
         return 'May';
       case '05':
-        return 'Aug';
+        return 'Sep';
       default:
         console.log('An unrecognized semester code was provided');
         return '';
     }
+  };
+
+  /* Param: expects an array of [later month (format: Mon), later year (YYYY), earlier month,
+            earlier year]
+     Returns: true if given dates are consecutive, false otherwise */
+  checkConsecutiveness = dates => {
+    console.log(dates);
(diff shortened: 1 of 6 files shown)


src/views/CoCurricularTranscript/index.js
```diff
@@ -170,36 +186,33 @@ export default class Transcript extends Component {
...
+      // keep track of the activity code which will be used to identify all activities of the same
+      // code so they can be grouped into one activity component
+      let curActCode = curAct.ActivityCode;
 
-      // Pop first session code from array and split into months and years
-      let sess = sessions.shift();
+      // Pop first session code from array and split into months and years, which are saved as
+      // the initial start and end dates
+      // (shift is like pop, removes and returns the first item in the array)
+      let sess = activities.shift().SessionCode;
```
EjPlatzer: Why shift activities here instead of shifting it for curAct above and then taking sessioncCode from curAct?

src/views/CoCurricularTranscript/index.js
```diff
@@ -170,36 +186,33 @@ export default class Transcript extends Component {
...
+      // For each other activity matching curActCode, if it is consecutive to the current end date,
+      // save its end date as the new end date, otherwise, add the current start and end dates to
+      // the string 'duration' (because the streak is broken) and prepare to start a new streak.
+      // Loop assumes activities will be sorted by session and Activity Code.
       let duration = '';
-      let prevSess = sess;
-      while (sessions > 0) {
-        sess = sessions.shift();
-        if (false && prevSess) {
+      while (activities.length > 0 && activities[0].ActivityCode === curActCode) {
```
EjPlatzer: The first half of this while is redundant with the above while. Maybe the above can be an if instead?

src/views/CoCurricularTranscript/index.js
EjPlatzer: Why shift activities here instead of at curAct above and then just take sessionCode from curAct here?

src/views/CoCurricularTranscript/index.js
EjPlatzer: The first half of this while loop is redundant with the while loop it is within. Maybe the above can be changed to an if, and these can be un-nested?

src/views/CoCurricularTranscript/index.js
```diff
@@ -170,36 +186,33 @@ export default class Transcript extends Component {
...
-      // Translate the sessions array into a duration:
+      // keep track of the activity code which will be used to identify all activities of the same
+      // code so they can be grouped into one activity component
+      let curActCode = curAct.ActivityCode;
 
-      // Pop first session code from array and split into months and years
-      let sess = sessions.shift();
+      // Pop first session code from array and split into months and years, which are saved as
+      // the initial start and end dates
+      // (shift is like pop, removes and returns the first item in the array)
```
EjPlatzer: This comment isn't necessary.

src/views/CoCurricularTranscript/index.js
joshuaRogers24: good point :)

src/views/CoCurricularTranscript/index.js
joshuaRogers24: Do you mean line 197 or 195-197?

EjPlatzer: Just 197

PR #479, opened 2019-06-28
Shows API version at the bottom of About page

shyung18:
Added version.js that gets from api/version 
Returns API version (hashcode and date) at the bottom of About page


src/views/About/index.js
```diff
@@ -4,9 +4,30 @@ import Grid from '@material-ui/core/Grid';
...
+  }
+
+  componentWillMount() {
+    this.loadData();
+  }
+
+  async loadData() {
+    const versionPromise = Version.getVersion();
+    const version = await versionPromise;
+    console.log('Version:', version);
```
EjPlatzer: We should remove this console log as well.

PR #503, opened 2019-07-10
Default Session in Involvements Page

shyung18:
If there are no activities in the current session, then change default session to the most recent session (which is fall semester during the summer)

### DIFF
--- a/src/views/ActivitiesAll/index.js
+++ b/src/views/ActivitiesAll/index.js
@@ -59,14 +59,13 @@ export default class GordonActivitiesAll extends Component {
         activities,
         allActivities: activities,
         myInvolvements: myInvolvements,
-        loading: true,
         sessions,
         types,
       });
 
       if (activities.length === 0) {
         var recentSession;
-        recentSession = this.state.sessions[0].SessionCode;
+        recentSession = sessions[0].SessionCode;
         this.setState({ session: recentSession, currentSession: sessionCode });
         const [activities, types, sessions] = await Promise.all([
           activity.getAll(recentSession),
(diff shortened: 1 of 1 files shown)


src/views/ActivitiesAll/index.js
```diff
@@ -58,10 +59,35 @@ export default class GordonActivitiesAll extends Component {
...
         myInvolvements: myInvolvements,
-        loading: false,
+        loading: true,
         sessions,
         types,
       });
+
+      if (activities.length === 0) {
+        var recentSession;
+        recentSession = this.state.sessions[0].SessionCode;
```
EjPlatzer: Do we need sessions from the state? Can't we use the variable instantiated within this method?

src/views/ActivitiesAll/index.js
```diff
@@ -58,10 +59,35 @@ export default class GordonActivitiesAll extends Component {
         activities,
         allActivities: activities,
         myInvolvements: myInvolvements,
-        loading: false,
+        loading: true,
```
EjPlatzer: We probably don't need to update loading to true here since we already set that above. It only needs to be changed when we set it to false at the end of the method, right?

src/views/ActivitiesAll/index.js
```diff
@@ -58,10 +59,35 @@ export default class GordonActivitiesAll extends Component {
...
+          activities,
+          allActivities: activities,
+          myInvolvements: myInvolvements,
+          loading: false,
+          sessions,
+          types,
+        });
+      }
+      this.setState({
+        loading: false,
```
EjPlatzer: If we set loading:false in the setState directly above this one, why do we need to do it again here?

PR #504, opened 2019-07-10
Error Logging for ID Uploader

isaacbleecker:
Added generic error logging service to UI. Includes method to get user's browser and OS. Currently only used by ID Uploader, to record successes and failures, as well as the browser / os.
It uses a new logging route in the API, to record logs to the database.

Hopefully, this logging service will be helpful for collecting errors from actual website usage. 

Currently the size of the error message seems to be limited -- might need to look into sending multipart file content, or something that can handle larger messages. We were unable to send an entire 6 mb blob, but for typical strings it is currently fine.

### DIFF
--- a/src/services/errorLog.js
+++ b/src/services/errorLog.js
@@ -1,0 +1,85 @@
+/**
+ * Error Log
+ *
+ * @module errorLog
+ */
+
+import http from './http.js';
+
+const postErrorLog = message => {
+  let currentTime = new Date();
+  let data = {
+    LOG_MESSAGE: message,
+    LOG_TIME: currentTime,
+  };
+  return http.post('log/add', data);
+};
+
+const postErrorMessage = message => {
+  return http.post('log', message);
+};
+
+//Code modified from https://medium.com/creative-technology-concepts-code/detect-device-browser-and-version-using-javascript-8b511906745
+function matchItem(string, data) {
+  var i = 0,
+    regex,
+    match;
+
+  for (i = 0; i < data.length; i += 1) {
+    regex = new RegExp(data[i].value, 'i');
+    match = regex.test(string);
+    if (match) {
+      return {
+        name: data[i].name,
+      };
+    }
+  }
+  return { name: 'unknown' };
+}
+
+const parseNavigator = navigator => {
+  //Code modified from https://medium.com/creative-technology-concepts-code/detect-device-browser-and-version-using-javascript-8b511906745
+  var oses = [
+    { name: 'Windows Phone', value: 'Windows Phone' },
+    { name: 'Windows', value: 'Win' },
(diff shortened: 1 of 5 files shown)


src/services/errorLog.js
```diff
@@ -0,0 +1,85 @@
...
+  };
+  return http.post('log/add', data);
+};
+
+const postErrorMessage = message => {
+  return http.post('log', message);
+};
+
+//Code modified from https://medium.com/creative-technology-concepts-code/detect-device-browser-and-version-using-javascript-8b511906745
+function matchItem(string, data) {
```
joshuaRogers24: can you use the React function declaration syntax here, as you did on lines 9 and 18?

src/services/errorLog.js
```diff
@@ -0,0 +1,85 @@
...
+      };
+    }
+  }
+  return { name: 'unknown' };
+}
+
+const parseNavigator = navigator => {
+  //Code modified from https://medium.com/creative-technology-concepts-code/detect-device-browser-and-version-using-javascript-8b511906745
+  var oses = [
+    { name: 'Windows Phone', value: 'Windows Phone' },
```
joshuaRogers24: is oses plural for os? Is there a clearer name for this array?

src/services/membership.js
```diff
@@ -53,7 +53,7 @@ import http from './http';
  * @return {Promise<any>} Response
  */
 function addMembership(data) {
-  return http.post(`memberships`, data).catch(reason => {
+  return http.post('memberships', data).catch(reason => {
     console.log(reason);
```
joshuaRogers24: was this just a "bunny-trail" change, irrelevant to the error log?

src/views/IDUploader/index.js
```diff
@@ -57,17 +57,20 @@ export default class IDUploader extends Component {
   };
 
   async postCroppedImage(croppedImage, n) {
-    console.log('ID photo resubmission #' + n);
+    let logMessage = 'ID photo submission #' + n + ' from ' + errorLog.parseNavigator(navigator);
```
joshuaRogers24: is there a more descriptive name for the 'n' parameter?

joshuaRogers24: Is 'n' clear enough for this parameter's name?

src/services/membership.js
isaacbleecker: Yes, this one was just a change for consistency, as the special back tick characters only need to be used when there is a multi-line string or a variable/expression inside the string.
It's not relevant to error logging.

src/services/errorLog.js
isaacbleecker: That was the intent. Perhaps just naming it "os" would be clearer? Or I could use "OperatingSystems" ?

PR #508, opened 2019-07-11
Cct leadership redesign

### LINKED ISSUE #500: (3 pts) Display cct memberships split into overall involvement with duration and leader involvement with duration
E.G. 
"Computer Science Club ----- Fall 2016 - Spring 2020
        Leader ----- Fall 2018 - Fall 2019

EjPlatzer:
After talking to Chris Carlson, it was decided that the Co-Curricular Transcript should display memberships in order from most recently ended to least recently ended, and that each membership should only display in one category, showing the full duration of the user's membership, and-if needed-the duration of the user's leadership.

This pull request fixes #500 and #499

### DIFF
--- a/src/views/CoCurricularTranscript/Components/CoCurricularTranscriptActivity/coCurricularTranscriptActivity.scss
+++ b/src/views/CoCurricularTranscript/Components/CoCurricularTranscriptActivity/coCurricularTranscriptActivity.scss
@@ -25,7 +25,7 @@
   grid-template-columns: 5% 70% 25%;
   grid-template-rows: auto auto;
   justify-items: stretch;
-  grid-row-gap: 20px;
+  grid-row-gap: 0px;
 }
 
 .organization-role {
@@ -38,6 +38,15 @@
   grid-column: 3;
   grid-row: 1;
   text-align: left;
+}
+
+.leadership-line {
+  grid-column: 2;
+  display: grid;
+  grid-template-columns: 5% 70% 25%;
+  grid-template-rows: auto auto;
+  justify-items: stretch;
+  grid-row-gap: 20px;
 }
 
 @media print {
--- a/src/views/CoCurricularTranscript/Components/CoCurricularTranscriptActivity/coCurricularTranscriptActivity.scss
+++ b/src/views/CoCurricularTranscript/Components/CoCurricularTranscriptActivity/coCurricularTranscriptActivity.scss
@@ -25,7 +25,6 @@
   grid-template-columns: 5% 70% 25%;
   grid-template-rows: auto auto;
   justify-items: stretch;
-  grid-row-gap: 0px;
 }
 
 .organization-role {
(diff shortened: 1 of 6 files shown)


src/views/CoCurricularTranscript/Components/CoCurricularTranscriptActivity/index.js
```diff
@@ -1,80 +1,132 @@
...
+  };
+
+  // Prepares a list of sessions to be displayed as one coherent string representing the timepsan
+  // of the membership.
+
+  // Param: sessionsList - a list of sessionCodes
+  // Returns: A string representing the duration of the user's membership based on the sessionsList
+  formatDuration = sessionsList => {
+    let duration = '';
+    sessionsList.sort(function(sessA, sessB) {
```
joshuaRogers24: Would it be better to be consistent and use react function syntax?

EjPlatzer: An arrow function wouldn't work here, since the function is being defined as the parameter of the sort function.

src/views/CoCurricularTranscript/index.js
```diff
@@ -86,132 +86,71 @@ export default class Transcript extends Component {
...
 
-      // Flush the remaining start and end info to duration.
-      // Again, don't show the year twice if the months are of the same year
-      if (startYear === endYear) {
-        duration += startMon;
-      } else {
-        duration += startMon + ' ' + startYear;
-      }
-      duration += '-' + endMon + ' ' + endYear;
+      let sessionsOrdered = sessions.sort(function(a, b) {
```
joshuaRogers24: Would it be helpful to add a simple comment that says "sorts sessions from least to most recently"?

EjPlatzer: I don't think so, but if others do, then we could certainly add it. This is one of the default behaviors of sort() available in most JavaScript tutorials, so I don't think it requires clarification.

src/views/CoCurricularTranscript/Components/CoCurricularTranscriptActivity/coCurricularTranscriptActivity.scss
```diff
@@ -25,7 +25,7 @@
   grid-template-columns: 5% 70% 25%;
   grid-template-rows: auto auto;
   justify-items: stretch;
-  grid-row-gap: 20px;
+  grid-row-gap: 0px;
```
joshuaRogers24: Does a 0 pixel gap look different than just not having a gap? (Who knows, weird things go on with these grids)

EjPlatzer: no, it does not. I'll remove that line until we clean up formatting.

PR #515, opened 2019-07-12
Show facstaff dining balance in green by default, or in either

jsenning:
OPTIONAL: This seemed like a good idea to me.

This pull-request changes the color of the faculty staff dining balance to yellow if the balance is below $20 and to orange (really quite red) when the balance is below $10.


src/views/Home/components/DiningBalance/index.js
```diff
@@ -58,6 +58,15 @@ export default class DiningBalance extends Component {
       content = <GordonLoader />;
     } else {
       if (this.facStaffBalance) {
+        //balance color will be green or yellow (low balance) or orange (really low balance)
```
joshuaRogers24: So, does this 'if' block only run if facStaffBalance is non-zero?

joshuaRogers24: Oh, your comment is slightly unclear. It sounds like a color of green or yellow represents low balance. Replacing the first 'or' with a comma would probably sound right.

joshuaRogers24: Oh, I realized a better question to ask than my first one is "does this 'if' block only run if facstaffBalance is non-null?"

jsenning: The "if" statement was pre-existing.  An string will evaluate to False in a Boolean expression if it is empty, otherwise it will be True (https://stackoverflow.com/questions/8692982/in-javascript-is-an-empty-string-always-false-as-a-boolean). 

Comments have been rewritten!

PR #528, opened 2019-07-15
Update deps

### LINKED ISSUE #486: (3 pts) Investigate (and try out) how to update important dependencies, especially React and Material-ui. Figure out what is needed to reform the code structure according to the newest method.

joshuaRogers24:
### Background
For a couple years, dependency versions had been frozen, kept from updating to major releases. This put a limit on what new dependencies we could add. When we discovered that this also prevented us from setting up a Jasmine-Karma testing environment without numerous vulnerabilities, updating these dependencies became a must. 

### Process
The process began with updating only the dependencies whose outdated-ness caused vulnerabilities according to `npm audit`: react-scripts, stylelint, and node-sass-chokidar. This led to updating other dependencies which relied on the first three. Finally, the rest of the dependencies were told to accept the latest version (prefixing their package.json entries with `>=`. All this meant the following code changes were necessary so that the site would not crash: 

- adjustment of the `test` entry under the `scripts` section of package.json, as directed [here](https://github.com/facebook/create-react-app/blob/master/CHANGELOG-2.x.md#the-default-jest-environment-was-changed-to-jsdom)
- migration of the proxy object's data in package.json to a new file (setupProxy.js) and rewriting it in the format described [here](https://github.com/facebook/create-react-app/blob/master/CHANGELOG-2.x.md#object-proxy-configuration-is-superseded-by-srcsetupproxyjs)
- addition of `browserlist` config to package.json based on example [here](https://github.com/facebook/create-react-app/blob/b0cbf2caa18ee8267855b14578ebc3dee826f552/packages/react-scripts/package.json#L83-L94)
- removal of `--recursive` flag from `package.json>scripts>watch-css` since, as noted [here](https://github.com/michaelwayman/node-sass-chokidar/blob/master/CHANGELOG.md#100), support was removed
- correction of icon names to match how Font Awesome now names them, as seen [here](https://react-icons.netlify.com/#/icons/fa)
- rewrite of the use of Dropzone and Downshift components since they now only support the "render prop" as a child and not as an attribute
- removal of base64toblob from package.json since it was unused

Some changes were just to avoid warnings:
- addition of `rel="noopener norferrer"` to any anchor tags which used target="_blank" 
- switch from edit social media links button to an IconButton, since anchors cannot go without an href attribute

The rest of the changes were my own tidying, removing useless code, and automatic style linting.

### Note: 
Documentation will have to be edited to reflect changes.

Resolves #486


src/views/MyProfile/index.js
```diff
@@ -244,6 +244,7 @@ export default class Profile extends Component {
       const [{ def: defaultImage, pref: preferredImage }] = await Promise.all([
         await user.getImage(),
       ]);
+      console.log(preferredImage);
```
EjPlatzer: Do we still need this console.log?

joshuaRogers24: I don't think so. Good catch.

PR #532, opened 2019-07-15
Back button for people search gives back the past results #511

shyung18:
URL stores the search parameters
Back button restores the searched results from the past


src/views/PeopleSearch/index.js
```diff
@@ -116,6 +116,161 @@ class PeopleSearch extends Component {
...
+            </div>
+          )
+        }
+      </Media>
+    );
+    return content;
+  }
+
+  componentDidUpdate() {
+    window.onpopstate = e => {
```
isaacbleecker: Do you need the "e" here? Or could it be empty parentheses "()" ?

PR #538, opened 2019-07-17
Progressive Web App

jahnueldorelus:


public/index.html
```diff
@@ -1,16 +1,46 @@
...
     <!-- Gordon blue -->
-    <meta name="theme-color" content="#014983">
+    <meta name="theme-color" content="#014983" />
     <!--
       manifest.json provides metadata used when your web app is added to the
       homescreen on Android. See https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/
     -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
-    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+    <link rel="manifest" href="./manifest.json" />
```
EjPlatzer: I'm not sure if you purposefully ignored the above comment or not, but it seems like either we should obey it or we should change it to reflect what we're doing.

public/index.html
```diff
@@ -1,16 +1,46 @@
...
+    <link rel="shortcut icon" href="./favicon.ico" />
```
EjPlatzer: See above

public/sw.js
```diff
@@ -0,0 +1,345 @@
...
+  '/involvements',
+  '/feedback',
+  '/people',
+  '/about',
+  '/help',
+  '/admin',
+  '/myprofile',
+  // 'https://cloud.typography.com/7763712/7294392/css/fonts.css', // Doesn't work in Developlent
+  '/static/js/bundle.js',
+  //'/static/js/bundle.js',
```
EjPlatzer: This looks like a copy of the above line.

PR #559, opened 2019-07-18
Add Guest view to 360Train

EjPlatzer:
This pull request will add the ability for guests (unauthenticated users) to access a limited version of the 360Train site. 

The major changes are:

- Unauthenticated users are no longer limited to the login page. That page has been migrated within the Home view. Instead, when a user is unauthenticated, each component is responsible for rendering itself appropriately.
- When a user is not logged in, the RightNavatar dropdown menu no longer has a link to MyProfile, and the sign out button becomes a sign in button that redirects to the Home (login) page.
- The same as above is true for the mobile-only Nav Drawer.
- Home will render the login page when the user isn't logged in, but will render the regular view (carousel, CL&W credits, etc.) when they are authenticated.
- The involvements view will not render a section for "My involvements" when the user isn't logged in, and the page for an individual involvement will not display any membership information.
- The events view will only display Public events when the user is not logged in.
- The people search Tab/View is disbaled when the user is not logged in.
- The same as above is true of the people search bar.

All pages that are accessible via the unauthenticated GUI have been edited to display the appropriate information. However, many pages that aren't accessible via the GUI (but are still sometimes accessible via URL) have not been edited to render a proper support for guests. They will fail to fetch because the user is not authenticated and the page will crash.


src/views/Login/index.js
```diff
@@ -70,17 +71,32 @@ export default class Login extends Component {
...
+          message: 'cache-static-dynamic-files',
+          token: storage.get('token'),
+          termCode: session.getTermCode(),
+        });
+        // Stores the current term in Local Storage for later use when updating the cache
+        storage.store('currentTerm', session.getTermCode());
+        // Saves the network state as online in local storage
+        localStorage.setItem('network-status', JSON.stringify('online'));
+      } else {
+        console.log('SERVICE WORKER IS NOT AVAILABLE');
```
isaacbleecker: It might be nice if this error was more descriptive of the problem, potentially for frustrated users that know how to view the console.

PR #564, opened 2019-07-19
Bring develop into PWA

russtuck:
We don't want to do this exactly, because it undoes some of the PWA work.  We're trying to figure out how to fix that.

DO NOT MERGE.


russtuck:
Closing because the problem was fixed another way: revert the revert, then do a merge.

PR #589, opened 2019-07-24
Fixed the CL&W Credits Days Left Component 

EjPlatzer:
When guest views were added, a small bug on the home page was created, in which the DaysLeft doughnut would not update properly. If the user logs in as a non student, logs out, and then logs in as a student, the component would still render the non student version, without CL&W Credits Tracking. Now it will rerender with the new personType of the user whenever the user logs back in.


philleegordon:
Looks good to me

PR #603, opened 2019-07-26
Alignment of buttons

philleegordon:
Buttons are at least mostly aligned now


src/views/Home/components/CLWCreditsDaysLeft/index.js
```diff
@@ -192,17 +191,19 @@ export default class CLWCreditsDaysLeft extends Component {
...
-                <Button variant="contained" style={style.button}>
-                  ATTENDED EVENTS
-                </Button>
-              </Link>
-            </Grid>
-          </Grid>
+                  <Button
+                    variant="contained"
+                    style={style.button}
+                    onClick={() => (window.location.href = 'https://360.gordon.edu/attended/')}
```
EjPlatzer: We should change this to "window.location.pathname = '/attended'", that way it stays within the site that it is clicked from.

PR #609, opened 2019-07-26
Leftover UI fixes Part 1

joshuaRogers24:
### Summary

There were some UI fixes that we lost in the merge into develop. This fixes the hover behavior of links (so that they are only underlined or change color when we want them to be) and removes the blue border around involvements tiles.


src/views/MyProfile/index.js
```diff
@@ -774,7 +776,7 @@ class MyProfile extends Component {
                                 <CardHeader title="Involvements" />
                               </Grid>
                               <Grid item xs={5} align="right">
-                                <Link className="gc360-link-color" to="/transcript">
+                                <Link className="gc360-link" to="/transcript">
                                   <Button variant="contained" style={style.button}>
                                     Co-Curricular Transcript
```
EjPlatzer: Can we switch this to say Experience Transcript instead?

PR #623, opened 2019-08-15
Update auth.js

jason-whitehouse:
Altered auth.js and re-wrote getAuth() function with no-cors header and better readability.


src/services/auth.js
```diff
@@ -34,13 +34,18 @@ const getAuth = (username, password) => {
   if (username.includes('@gordon.edu')) username = username.replace('@gordon.edu', '');
   else if (username.includes('Gordon.edu')) username = username.replace('@Gordon.edu', '');
 
-  const headers = new Headers({ 'Content-Type': 'application/x-www-form-urlencoded' });
-  const body = new URLSearchParams({
+  //const headers = new Headers({ 'Content-Type': 'application/x-www-form-urlencoded' });
```
russtuck:
```suggestion
```

PR #649, opened 2019-11-18
Add mailbox location

### LINKED ISSUE #620: Add Student/Staff Mail Box Number to the student's address shown Public

babbett:
Added mailbox number and dorm to the profile view.

profile

Partially resolves #620


babbett:
The most recent version of Natou and Josh's API wasn't working for me, so I have not been able to test this with actual data yet. Everything looks pretty good, but we are very low on real estate for the desktop People Search results list. I combined the email and mailbox number into a single 2 lined column, and fixed some centering issues that caused. If anyone can think of a better way to fit them all together, let me know.

PR #661, opened 2019-12-12
added button to link attended page to events page

lavenwinkle:
Screen Shot 2019-12-12 at 16 57 27
Added a button that links to the Events page on both the view if user has chapel credits and if they do not.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5678,12 +5678,20 @@
       }
     },
     "eslint-utils": {
-      "version": "1.4.0",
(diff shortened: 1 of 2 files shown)


lavenwinkle:
I looked back through and I had considered putting the button under the events, but since the events are listed with the most recent ones at the top, I felt that the natural flow would be to return to the top of the page when reviewing events attended or staying within the top few if a student was double-checking their credits, so I put the button at the top.

PR #668, opened 2020-01-23
Add mailbox location to activity profile

babbett:
Added ability to see mailbox numbers of activity members on the member list.


babbett:
The admin view:
adminview

The member view:
memberview

babbett:
The change is currently live in the develop backend. Should be all set!

PR #671, opened 2020-01-30
Banner page

ajabbot:
image
The carousel is now working (using a new component), and the margin issues are fixed.

Couple things to note:
     Keyboard functionality is gone in this component :(
     The on-hover pointer is removed. This can be added back in the future if we think it important.

image
A new page has been created for sourcing banner images. We need to create a slide to advertise this (and link from it to this page). This way clubs can advertise their upcoming events relatively simply.

There were a bunch of dependencies I updated (including React and material ui). There are breaking changes going from react-router-dom 5.0.1 to 5.1.2. This should have been done in a separate commit, sorry!


src/routes.js
```diff
@@ -93,4 +94,9 @@ export default [
     path: '/admin',
     component: Admin,
   },
+  {
+    name: 'Banner',
```
NathanielRudenberg: I wonder whether something like "submit_banner" would be a slightly more appropriate name? It's not important though, the current one works just fine

PR #678, opened 2020-02-20
Fix empty mailbox

babbett:
On the involvements page, if a user had no mailbox it would still show "Box #". This change removes that issue, now "Box #" is only shown if the user actually has a mailbox.


babbett:
Annotation 2020-02-23 232755

Adds a simple check to remove these empty "Box #" fields.

PR #679, opened 2020-02-20
Add 'no description' text for events without a null description

babbett:
Solves #680 
On the events page, some events have a "No description available" description, while some just have an empty field. Now both cases just display "No description available".


babbett:
events360

This is what it looks like currently, the fix just adds "No description available" under the career fair's description as well.

PR #702, opened 2020-04-05
Devtomaster

ajabbot:
Banner image fix, bug fixes, ui improvements


src/components/EventList/components/EventItem/index.js
```diff
@@ -1,63 +1,64 @@
...
-    Event_ID: PropTypes.string.isRequired,
-    Event_Name: PropTypes.string.isRequired,
-    Event_Title: PropTypes.string.isRequired,
-    Event_Type_Name: PropTypes.string,
-    Category_ID: PropTypes.number,
-    Description: PropTypes.string,
-    Occurrences: PropTypes.array,
-  }).isRequired,
-};
+import PropTypes from 'prop-types';
```
NathanielRudenberg, line 1: Any idea why this diff spans the whole file even though it looks unchanged?

ajabbot: Oh odd, good catch. I have no idea

src/views/BannerSubmission/index.js
```diff
@@ -0,0 +1,85 @@
...
+      version: null,
+    };
+  }
+
+  componentWillMount() {
+    this.loadData();
+  }
+
+  async loadData() {
+    const versionPromise = Version.getVersion();
```
NathanielRudenberg, line 25: I wonder whether it would be cleaner to do it this way:
```suggestion
    const version = await Version.getVersion();
```

Not mission critical, but just a suggestion

src/views/EventsAttended/index.js
```diff
@@ -63,12 +63,13 @@ export default class EventsAttended extends Component {
         );
       } else {
         content = (
-          <Grid item>
+          <Grid item align="center">
```
NathanielRudenberg, line 66: Oh good, finally 😆

src/views/Login/login.scss
```diff
@@ -7,7 +7,7 @@
   opacity: 0.92;
   box-shadow: 0 5px 22px rgba(0, 0, 0, 0.5);
 }
-img {
+.login-img {
```
NathanielRudenberg, line 10: Do you think we might still need the `img {}` selector?

ajabbot: This addressed some formatting issue (I think on the login page) and made this css only apply to the login image. There's still img styling somewhere

PR #710, opened 2020-04-10
Dev to master april10

NathanielRudenberg:
Contains timesheets


NathanielRudenberg:
Travis keeps failing and I have no idea why

README.md
```diff
@@ -22,12 +22,13 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
...
 
-Make sure Node.js is set up on your machine (https://nodejs.org/en/)
-
-To run the app for the first time, run the following commands:
-
-- `npm install`
-- `npm start`
+Make sure Node.js is set up on your machine. The project uses version `9.11.1` as later versions break it.
+- Download the latest release of NVM (Node Version Manager) from [here](https://github.com/coreybutler/nvm-windows/releases). `Select nvm-setup.zip`.
+    - Extract the file and run it, making sure to install it to the account you are logged into (example: if you're logged in as anthony.aardvark, make sure it installs to anthony.aardvark and not cct.service). If it says Node is already installed, proceed anyway.
```
joshuaRogers24, line 27: Would the note about cct.service be confusing to someone developing on their own device (not on the virtual machine)?

src/components/Header/index.js
```diff
@@ -203,6 +205,14 @@ export default class GordonHeader extends Component {
       }
     }
 
+    //Add to return statement when re-enabling work link
```
joshuaRogers24, line 208: When this is incorporated, will it need different versions for offline or guest 360 as the People Tab did?

src/components/Nav/components/NavLinks/index.js
```diff
@@ -360,6 +376,32 @@ export default class GordonNavLinks extends Component {
         );
       }
 
+      //Re-enable when adding the link to timesheets page
```
joshuaRogers24, line 379: Should timesheets be available to guests?

src/routes.js
```diff
@@ -94,9 +95,14 @@ export default [
     path: '/admin',
     component: Admin,
   },
+  {
+    name: 'Timesheets',
+    path: '/t1m35433t5',
```
joshuaRogers24, line 100: The path attribute will be changed when timesheets goes fully public, correct?

src/services/jobs.js
```diff
@@ -0,0 +1,81 @@
...
+
+import http from './http';
+
+/**
+ * Get active jobs for current user
+ * @param {String} details The user's details
+ * @return {Promise.<String>} User's active jobs
+ */
+const getActiveJobsForUser = details => {
+  return http.post(`jobs/getJobs`, details);
```
joshuaRogers24, line 14: How does http.post() work for retrieving data? I would expect a http.get().

src/services/jobs.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+/**
+ * Get active jobs for current user
+ * @param {Number} eml we don't know what this means yet
+ * @param {DateTime} shiftStart The start time of the shift
+ * @param {DateTime} shiftEnd The end time of the shift
+ * @param {Number} hoursWorked The number of hours worked
+ * @param {String} shiftNotes Shift notes
+ * @param {String} lastChangedBy The person who last updated the shift
```
joshuaRogers24, line 32: Should the lastChangedBy parameter be listed here? It is not actually in the declaration.

src/views/Timesheets/components/SavedShiftsList/SavedShiftsList.scss
```diff
@@ -0,0 +1,19 @@
+@import '../../../../vars';
+
+// Make rows alternate white and grey
+/* stylelint-disable */
+.shift-list > .shift-item:nth-child(even) {
+  background: #ebeaea;
+  &:hover {
+    background-color: $primary-cyan;
+    //   cursor: pointer;
```
joshuaRogers24, line 9: Is there a plan for the cursor: pointer lines here and on 17?

src/views/Timesheets/components/SavedShiftsList/index.js
```diff
@@ -0,0 +1,306 @@
...
+            <Typography>
+              {this.state.reportingSupervisor !== null && this.state.reportingSupervisor.name}
+            </Typography>
+          </MenuItem>
+        </Select>
+      </FormControl>
+    );
+
+    let content = <></>;
+    if (this.props.shifts.length === null) {
```
joshuaRogers24, line 210: Will shifts.length be zero if the user has no shift history, or will the Loader show for that case also?

src/views/Timesheets/components/SavedShiftsList/index.js
```diff
@@ -0,0 +1,306 @@
...
+              alignItems="center"
+              alignContent="center"
+              style={styles.boxShadow}
+            >
+              {header}
+              {shiftsList}
+            </Grid>
+          </CardContent>
+          {cardTitle === "Submitted Shifts" &&
+              <CardContent>
```
joshuaRogers24, line 239: Should lines 239-248 and 251-260 be exactly the same?

src/views/Timesheets/components/SavedShiftsList/index.js
```diff
@@ -0,0 +1,306 @@
...
+                </Button>
+                </Grid>
+              </Grid>
+            </Grid>}
+          </CardActions>
+        </Card>
+        </>
+      );
+    }
+    return <>{content}</>;
```
joshuaRogers24, line 285: Would it work to just return `content`? (with no braces around it)

src/views/Timesheets/components/ShiftItem/index.js
```diff
@@ -0,0 +1,139 @@
...
+                  </Button>
+                </Grid>
+              </Grid>
+            </DialogContent>
+          </Dialog>
+        </Grid>
+      </Grid>
+    );
+
+    let shiftItemIcon = (STATUS !== 'Submitted' && STATUS !== 'Approved') ? (
```
joshuaRogers24, line 89: Would it be more readable to ask whether the status is ‘Saved’ or ‘Rejected’ instead of asking what it is not?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+  const [shiftTooShort, setShiftTooShort] = useState(false);
+  const [timeOutIsBeforeTimeIn, setTimeOutIsBeforeTimeIn] = useState(false);
+  const [enteredFutureTime, setEnteredFutureTime] = useState(false);
+  const [hoursWorkedInDecimal, setHoursWorkedInDecimal] = useState(0.0);
+  const [userShiftNotes, setUserShiftNotes] = useState('');
+  const [isOverlappingShift, setIsOverlappingShift] = useState(false);
+  const [shiftDisplayComponent, setShiftDisplayComponent] = useState(null);
+  const [snackbarOpen, setSnackbarOpen] = useState(false);
+
+  const handleTimeOutIsBeforeTimeIn = (timeIn, timeOut) => {
```
joshuaRogers24, line 40: Could this (`handleTimeOutIsBefore...`) have a name that better matches its function- or could you split it into multiple functions?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+      setHoursWorkedInDecimal(roundedHourDifference);
+      let hoursWorked = Math.floor(calculatedTimeDiff);
+      let minutesWorked = Math.round((calculatedTimeDiff - hoursWorked) * 60);
+
+      if (minutesWorked >= 60) {
+        hoursWorked++;
+        minutesWorked = 0;
+      }
+
+      if (timeDiff < 0) {
```
joshuaRogers24, line 54: Can 54-70 be rewritten to a total of 3 lines using the format `setFunction(lessOrGreaterStatement)`?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+      jobs.getActiveJobsForUser(details).then(result => {
+        setUserJobs(result);
+      });
+    };
+
+    const getSavedShiftsForUser = () => {
+      return jobs.getSavedShiftsForUser();
+    };
+
+    const handleDateChange1 = date => {
```
joshuaRogers24, line 98: Would a more fitting name for these (and the one on 103) be handleTimeChangeIn and Out?

joshuaRogers24: "this" not "these"

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+        let roundedHourDifference2 = (Math.round(calculatedTimeDiff2 * 4) / 4).toFixed(2);
+
+        saveShift(
+          selectedJob.EMLID,
+          timeIn2.toLocaleString(),
+          timeOut2.toLocaleString(),
+          roundedHourDifference2,
+          userShiftNotes,
+        ).then(result => {
+          setSnackbarOpen(true);
```
joshuaRogers24, line 137: Do you need to do the same cleanup here as you did at 152? (setting things to null and 0, etc.)

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+        <></>
+      );
+
+    const isLeapYear = date => {
+      if (date.getFullYear() % 4 === 0) {
+        if (date.getFullYear() % 100 === 0) {
+          if (date.getFullYear() % 400 !== 0) {
+            return 'false';
+          }
+          if (date.getFullYear() % 400 === 0) {
```
joshuaRogers24, line 197: Could 197, 201, and 205 just be `else`?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+        if (date.getDate() === 30) {
+          nextDate = 1;
+          monthToReturn = (date.getMonth() + 1) % 12;
+          yearToReturn = date.getFullYear();
+        } else {
+          nextDate = date.getDate() + 1;
+          monthToReturn = date.getMonth();
+          yearToReturn = date.getFullYear();
+        }
+      } else if (!is30DayMonth) {
```
joshuaRogers24, line 263: Could 263 just be `else`?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+          yearToReturn = date.getFullYear() + 1;
+        } else {
+          nextDate = date.getDate() + 1;
+          monthToReturn = date.getMonth();
+          yearToReturn = date.getFullYear();
+        }
+      } else if (is30DayMonth) {
+        if (date.getDate() === 30) {
+          nextDate = 1;
+          monthToReturn = (date.getMonth() + 1) % 12;
```
joshuaRogers24, line 256: Is the modulo 12 necessary if you handle the December case separately?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+          The entered shift conflicts with a previous shift.
+      </Typography>
+      );
+    }
+    else {
+      errorText = <></>;
+    }
+
+    const onDatetimeSelectorClose = () => {
+      setIsOverlappingShift(false);
```
joshuaRogers24, line 364: Why is overlapping shift set false here (as opposed to somewhere else)?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+      if (selectedDateIn !== null && selectedDateOut !== null) {
+        getActiveJobsForUser();
+        checkForFutureDate();
+      }
+    }
+
+    const handleShiftNotesChanged = event => {
```
joshuaRogers24, line 371: Is there any chance of the shift notes input being malicious?

src/views/Timesheets/index.js
```diff
@@ -0,0 +1,539 @@
...
+                    </Grid>
+                    <Grid item xs={12} sm={6} md={3}>
+                      <TimePicker
+                        variant="inline"
+                        margin="normal"
+                        id="time-picker-in"
+                        label="Time In"
+                        value={selectedDateIn}
+                        onChange={date => {
+                          let dateToChange = date;
```
joshuaRogers24, line 416: Can this just be `onChange={handleDateChange2}`, as was done on 433?

README.md
NathanielRudenberg: It might possibly be. We could potentially change it to "zelda.zebra" or "rico.raccoon" or something like that

src/components/Header/index.js
NathanielRudenberg: Good question, right now there is a guest view so that guests cannot access the timesheets page. It's a good idea for us to add the offline view too.

src/components/Nav/components/NavLinks/index.js
NathanielRudenberg: It's already unavailable to guests, we might just need to disable the link on the guest view

src/routes.js
adamprinciotta: Yes, we will change it to /timesheets instead of the one that is there now. We made it much less guessable while we are in the testing phase

src/views/Timesheets/components/SavedShiftsList/SavedShiftsList.scss
adamprinciotta: Thanks for picking that up, we will remove those

src/services/jobs.js
NathanielRudenberg: With `http.get()`, any data that we send would need to be in a URL parameter. We needed to send a DateTime object, and there wasn't a good way to make that into a URL parameter because it would contain a lot of characters that need to be sanitized if they're part of a url. `http.post()` allows you to send data as an object without sanitization, which is more ideal for this scenario.

src/services/jobs.js
NathanielRudenberg: I thought we got rid of that parameter, maybe git merge weirdness put it back. We'll look into it

src/views/Timesheets/components/SavedShiftsList/index.js
adamprinciotta: If a saved shifts list is empty, it will not show the card

src/views/Timesheets/components/SavedShiftsList/index.js
adamprinciotta: Good catch, we will Or the cardTitles to check if it is either Submitted Shifts or Approved Shifts and display the same content if that is true

src/views/Timesheets/components/ShiftItem/index.js
adamprinciotta: Yeah, I had mentioned it while we were working on it and decided since it was working as is, might as well keep it. But changing it would make it more readable

src/views/Timesheets/components/SavedShiftsList/index.js
NathanielRudenberg: Ah, the weirdly useful quirk of JSX. Because `content` is a variable, we need the braces here to access what's stored in it. In this case, it's just more JSX, but it's displayed differently based on truth conditions, which is why it's in a variable in the first place. If we didn't have the braces, instead of the actual content, the page would display the literal word "content" in it's place.

src/views/Timesheets/components/SavedShiftsList/index.js
NathanielRudenberg: so basically something like 
```jsx
{(cardTitle === "Approved Shifts" || cardtitle === "Submitted Shifts") &&
    <CardContent>
      <Grid container>
        <Grid item xs={12} sm={6}>
          <Typography variant="h6">Total hours worked: {totalHoursWorked}</Typography>
        </Grid>
        <Grid item xs={12} sm={6}>
          <Typography variant="h6">Estimated gross pay: ${totalEstimatedPay}</Typography>
        </Grid>
      </Grid>
    </CardContent>
}
```

src/views/Timesheets/index.js
NathanielRudenberg: You're absolutely right, I can't believe we didn't do that in the first place 😆 
Yes, we will make these lines look like `setTimeOutIsBeforeTimeIn(timeDiff < 0)`

Good catch!

src/views/Timesheets/index.js
NathanielRudenberg: Yes, their current names are a relic of the early stages of development. We changed the names of the objects they deal with, but forgot to change the names of the method... whoops

src/views/Timesheets/index.js
adamprinciotta: We could change it to something like handleTimeErrors

src/views/Timesheets/index.js
adamprinciotta: It will reach that code that sets them null every time this method is run

src/views/Timesheets/index.js
NathanielRudenberg: Probably. Because it's so finicky I'm kind of afraid to touch it, but we will try it anyway

src/views/Timesheets/index.js
NathanielRudenberg: Most likely, we'll look into it

src/views/Timesheets/index.js
adamprinciotta: You are probably right, we will test that

src/views/Timesheets/index.js
NathanielRudenberg: My thinking was that if a user ever sees the associated error, after they make a change to their timein/timeout the error should go away. Now that you mention it, that sounds like an operation that should go in one of the `onChange` functions instead of this one.

src/views/Timesheets/index.js
NathanielRudenberg: I'm fairly certain there's no chance, but it's worth doing a bit of research on it. Here's what happens in the back end:

```c#
IEnumerable<StudentTimesheetsViewModel> result = null;

var id_num = new SqlParameter("@ID_NUM", studentID);
var eml = new SqlParameter("@eml", jobID);
var shiftStartDateTime = new SqlParameter("@shift_start_datetime", shiftStart);
var shiftEndDateTime = new SqlParameter("@shift_end_datetime", shiftEnd);
var hours_worked = new SqlParameter("@hours_worked", hoursWorked);
var notes = new SqlParameter("@shift_notes", shiftNotes);
var changedBy = new SqlParameter("@last_changed_by", lastChangedBy);

try
           {
               result = RawSqlQuery<StudentTimesheetsViewModel>.StudentTimesheetQuery("student_timesheets_insert_shift @ID_NUM, @eml, @shift_start_datetime, @shift_end_datetime, @hours_worked, @shift_notes, @last_changed_by", id_num, eml, shiftStartDateTime, shiftEndDateTime, hours_worked, notes, changedBy);
           }
```

My understanding is that because shiftNotes is passed in as a parameter instead of explicitly being part of the query, it is safe from malicious input. But again, we really should research this first. Thanks for bringing it up!

src/views/Timesheets/index.js
NathanielRudenberg: It can't be `handleDateChange2` because that is specific to date out, while this one is specific to date in. This touches on the need to rename those two functions again. 

One thing I'm not sure about: why did we do the whole `let dateToChange = date;                           handleDateChange1(dateToChange);` instead of `handleDateChange1(date);`? Beats me. We will look into this.

src/components/Nav/components/NavLinks/index.js
NathanielRudenberg: I thought of something else, it might make more sense for us to put the link in the Links dialog instead of in the navbar.
image

src/services/jobs.js
NathanielRudenberg: Update: evidently this one just slipped through our fingers and Travis passed it anyway. I'll get rid of it

src/views/Timesheets/index.js
NathanielRudenberg: After looking this up, I found out that `SqlParameter` does, in fact, prevent SQL injection. It doesn't prevent xss, which is a vulnerability that I'm not sure we prevent anywhere on the site. That's important, and it should probably be an item on the list for the summer practicum. if it happens this summer.

NathanielRudenberg:
Thanks Josh, these are super helpful. We'll get on this in the next day or so!

###### Dear Travis: ???????? Your failing command worked _just fine_ on other branches. Please, why this? why now?

NathanielRudenberg:
@joshuaRogers24 I have made some changes according to your comments. I'll open a pull request into develop with those changes so we can get them approved there before merging into master.

src/views/Timesheets/index.js
joshuaRogers24: Yeah, maybe in the same place where you set it true, you also set it false- under the right conditions, of course.

NathanielRudenberg: The issue is that it's set to true in response to an http fetch. When it's set, we can't run another fetch and try to set it to false because to do that we'd not to save again, but that's disabled when it's set to true

src/views/Timesheets/index.js
joshuaRogers24: Good!

src/views/Timesheets/index.js
joshuaRogers24: Oh, I see. That's difficult. . .

src/views/Timesheets/index.js
joshuaRogers24: Oh, okay!

src/views/Timesheets/index.js
joshuaRogers24: Thanks, glad to help!

src/views/Timesheets/index.js
joshuaRogers24: Oh, you noticed that! I was wondering the same. Did you have a reason?

joshuaRogers24: And just as I post this I see the change where you replaced it- sorry you had to hear me harping on it after you fixed it!

src/services/jobs.js
joshuaRogers24: okay, and you get the active jobs back in response?

NathanielRudenberg:
Why won't Travis build it 😭

NathanielRudenberg:
...It worked!

NathanielRudenberg:
@joshuaRogers24 This PR now has the changes you requested! Thank you for all your help these last few days.

NathanielRudenberg:
yikes

PR #711, opened 2020-04-12
Resolve review comments

NathanielRudenberg:
This pull request contains changes that were requested by @joshuaRogers24. The idea is to merge this into develop, then we can update pull request #710 to reflect these changes.

### DIFF
--- a/src/views/Timesheets/index.js
+++ b/src/views/Timesheets/index.js
@@ -196,15 +196,15 @@ const Timesheets = (props) => {
       if (date.getFullYear() % 4 === 0) {
         if (date.getFullYear() % 100 === 0) {
           if (date.getFullYear() % 400 !== 0) {
-            return 'false';
+            return false;
           } else {
-            return 'true';
+            return true;
           }
         } else {
-          return 'true';
+          return true;
         }
       } else {
-        return 'false';
+        return false;
       }
     };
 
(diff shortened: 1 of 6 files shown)


joshuaRogers24:
Looks good so far. The one thing I still am not sure about is the README- which isn't really part of timesheets, so I wouldn't hold the pull request back for it. Anyway, I don't think I understand the context for the NVM instructions. I was thinking someone might be downloading nvm on their own computer- which would not have a cct.service account to confuse with their own. That is why I think the note about cct.service is confusing. Am I missing something?

src/views/Timesheets/index.js
```diff
@@ -418,7 +418,7 @@ const Timesheets = (props) => {
                         value={selectedDateIn}
                         onChange={date => {
                           let dateToChange = date;
-                          handleDateChange1(dateToChange);
+                          handleDateChangeIn(dateToChange);
```
joshuaRogers24: Whoops! I said handle...2 instead of handle...1 in my earlier comment. The suggestion I meant to make was that 
`date = > { 
let dateToChange = data; 
handleDataChangeIn(dateToChange); }`

just be replaced with `handleDateChangeIn`. 
But am I missing something?

joshuaRogers24: Nevermind the above comment. I just saw the change 👍

NathanielRudenberg:
The NVM instructions are for anyone who does not yet have Node installed, whether on their own computer or on the VM. The instructions should work for either case. I had the note about making sure it installs to the right account because I had to manually change the installation path myself. I changed the cct.service bit to another fake account to hopefully help with confusion.

PR #715, opened 2020-04-17
Shift notes tooltips

NathanielRudenberg:
Add ability to see shift notes and rejection comments
Consolidated the datetime pickers
Restored keyboard support
image

image

### DIFF
--- a/src/views/Timesheets/components/ShiftItem/index.js
+++ b/src/views/Timesheets/components/ShiftItem/index.js
@@ -391,7 +391,7 @@ export default class ShiftItem extends Component {
                       editing: !this.state.editing,
                       newDateTimeIn: new Date(SHIFT_START_DATETIME),
                       newDateTimeOut: new Date(SHIFT_END_DATETIME),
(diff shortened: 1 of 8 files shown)


src/views/Timesheets/components/ShiftItem/index.js
```diff
@@ -292,7 +307,7 @@ export default class ShiftItem extends Component {
     } else {
       timeInDisp = <Typography variant="body2">{dateTimeIn}</Typography>
       timeOutDisp = <Typography variant="body2">{dateTimeOut}</Typography>
-      hoursWorkedDisp = <Typography variant="body2">{HOURS_WORKED}</Typography>
+      hoursWorkedDisp = <Typography variant="body2">{HOURS_WORKED.toFixed(2)}</Typography>
```
adamprinciotta, line 310: Did we set up the .toFixed(2) in every spot we should?

NathanielRudenberg: oops

PR #722, opened 2020-04-23
Add snackbar to involvements

babbett:
Two cases when adding someone to an involvement:
1) User is not in the involvement

Screenshot (93)

2) User is already in the involvement

Screenshot (92)

This also adds a snackbar to the removal of a member from an involvement, but since that removing someone from an involvement doesn't really work right now, that snackbar is commented out.

On almost any action the screen is refreshed. It seems unnecessary and I may find a way to stop this from happening. I will then add snackbars to actions that previously would have refreshed the page. Examples of this are editing involvements, confirming the final roster, editing members, and changing admin status.


babbett:
I am a fool and forgot the third case. Trying to add someone who's username does not exist results in this snackbar:

Screenshot (94)

src/views/ActivityProfile/components/Membership/components/MemberList/index.js
```diff
@@ -131,10 +137,25 @@ export default class MemberList extends Component {
   // Called when user confirms removal
   async confirmLeave() {
     await membership.remove(this.props.member.MembershipID);
+    // Once we fix the bug preventing this, remove comment to use snackbar
```
ajabbot, line 140: Wasn't this fixed in the backend?

babbett:
I uncommented the portion that now works. In order to validate that a user has been removed from the activity, I added a membership.search() after issuing the membership.remove(). membership.search() returns a 3 element array, where the first element is a boolean which is false if the user is in the involvement and true if the user is. I use this to conditionally display the snackbar after leaving an involvement. 

While this works, I'm not sure if it is the most efficient way to check if a user was removed. It didn't seem like the membership.remove() returned something that I could use to test if the call succeeded or failed. If I am wrong please let me know!

PR #738, opened 2020-06-03
(S20) Resolving Good First Issues

### LINKED ISSUE #639: URL in office hours loses description ":" when clicked
My office hours description includes the URL.  It is displayed correctly.  However, when someone clicks on it, the "https://" becomes "htts//".  The result is the user gets a DNS error.

It might be an error in URL unescaping.

(Chrome's message is 
  This site can’t be reached
  https’s server IP address could not be found.
  DNS_PROBE_FINISHED_NXDOMAIN
)

### LINKED ISSUE #640: Email logo covers email address in public profile
But only for longish email addresses.  For example, it's a problem for jonathan.senning but not for russ.tuck.

cpabbot:
Minor bug fixes:
- Fixed bug with overlapping long emails covering the email icon (resolves #640),
- Fixed issue with layout sizing on people search. The person "type" was given a larger buffer than the correct size given by the headers. This resolved the email address / mailbox section from overflowing side of card.
- Centered people search results on mobile. Potential new issue to note: fixed widths to 250px (to maintain image alignment).
- Corrected the width and location of the "schedule" card on profiles
- Resolved url issue on "schedule" card where links would be https://http// etc. The issue was a logic && mistake. (resolves #639)

### DIFF
--- a/src/views/PeopleSearch/components/MobilePeopleSearchResult/index.js
+++ b/src/views/PeopleSearch/components/MobilePeopleSearchResult/index.js
@@ -109,6 +109,7 @@ export default class PeopleSearchResult extends Component {
           <Grid
             container
             alignItems="center"
+            justify="center"
             spacing={2}
             style={{
               padding: '1rem',
@@ -123,7 +124,12 @@ export default class PeopleSearchResult extends Component {
                 placeholderColor="#FFF"
               />
             </Grid>
-            <Grid item>
+            <Grid
+              item
+              style={{
+                width: '250px',
+              }}
+            >
               <Typography variant="h5">{fullName}</Typography>
               <Typography variant="body2">{nickname}</Typography>
               <Typography variant="body2">{personClassJobTitle}</Typography>
--- a/src/views/PeopleSearch/components/MobilePeopleSearchResult/index.js
+++ b/src/views/PeopleSearch/components/MobilePeopleSearchResult/index.js
@@ -127,7 +127,9 @@ export default class PeopleSearchResult extends Component {
             <Grid
               item
               style={{
-                width: '250px',
+                // a set width is necessary to keep profile images in line 
+                // while maintaining center alignment
+                width: '260px',
               }}
             >
               <Typography variant="h5">{fullName}</Typography>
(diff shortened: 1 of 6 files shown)


src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -132,7 +132,7 @@ export default class PeopleSearchResult extends Component {
             <Grid item xs={2}>
               <Typography>{Person.LastName}</Typography>
             </Grid>
-            <Grid item xs={2}>
+            <Grid item xs={1}>
```
cpabbot, line 135: just correcting the layout so the header widths match the table cell widths essentially

src/views/Profile/index.js
```diff
@@ -356,8 +356,9 @@ export default class Profile extends Component {
                   </Card>
                 </Grid>
 
-                <Grid item xs={12} lg={12} align="center">
-                  <Grid container xs={12} lg={10} spacing={2} justify="center">
+                <Grid item xs={12} lg={10} align="center">
```
cpabbot, line 359: Explanation of numerical size changes: previously the schedule was in a small grid within a full width container. Now it is in a full width grid the size of a small container. Makes little sense without seeing it, but fixes the centering issue

src/views/Profile/profile.scss
```diff
@@ -18,16 +18,11 @@
     }
   }
   &_email {
-    display: grid;
```
cpabbot, line 21: Why oh why did they use CSS grid here instead of flexbox? The world may never know.
CSS grid should be used for 2D layouts and is a relatively newly adopted standard.

src/views/PeopleSearch/components/MobilePeopleSearchResult/index.js
```diff
@@ -123,7 +124,14 @@ export default class PeopleSearchResult extends Component {
...
               />
             </Grid>
-            <Grid item>
+            <Grid
+              item
+              style={{
+                // a set width is necessary to keep profile images in line 
+                // while maintaining center alignment
+                width: '260px',
+              }}
```
cpabbot, line 133: added a comment here. Not generally good to have fixed pixel widths like this here, but it seemed like the simplest solution for fixing mobile alignment.

PR #740, opened 2020-06-04
S20 profilepage dropdownbar alignment

aridospassos:
Took out the minor changes made on the victory promise component and simply kept the drop down alignment in both the profile and my profile pages.

### DIFF
--- a/src/views/MyProfile/index.js
+++ b/src/views/MyProfile/index.js
@@ -497,7 +497,7 @@ class MyProfile extends Component {
                     justify="flex-start"
                     spacing={2}
                   >
-                    <Grid item xs={12} md={profileCardSize} lg={profileCardSize}>
+                    <Grid item xs={12} md={profileCardSize} lg={8}>
                       <Card>
                         <CardContent>
                           <Grid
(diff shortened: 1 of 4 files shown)


src/views/MyProfile/index.js
```diff
@@ -497,7 +497,7 @@ class MyProfile extends Component {
                     justify="flex-start"
                     spacing={2}
                   >
-                    <Grid item xs={12} md={profileCardSize} lg={profileCardSize}>
+                    <Grid item xs={12} md={profileCardSize} lg={8}>
```
russtuck: Is there a symbol that makes sense here instead of 8?  I'm sorry to see a variable replaced with a "magic number".

src/views/MyProfile/index.js
```diff
@@ -497,7 +497,7 @@ class MyProfile extends Component {
                     justify="flex-start"
                     spacing={2}
                   >
-                    <Grid item xs={12} md={profileCardSize} lg={8}>
+                    <Grid item xs={12} md={profileCardSize} lg={profileCardSize}>
```
russtuck: So this undoes the opposite change?

russtuck:
I'm closing this, since it is replaced by https://github.com/gordon-cs/gordon-360-ui/pull/748

PR #743, opened 2020-06-05
S20 People Search 'No results'

changman0:
src/components/Header/components/PeopleSearch/index.js: Added 'No results' dropdown when there are no people search suggestions & fixed bug where backspacing the search and entering when no suggestions dropdown is shown redirects the user to the first previously available top suggestion.

src/components/Header/components/PeopleSearch/people-search.scss: Added class for 'No results' for gray text and italics.

Edit: Added renderNoResult() function to avoid DRY code. Changed MIN_QUERY_LENGTH from 3 to 2 to include people searches for individuals with 2 letter first or last names.

### DIFF
--- a/src/components/Header/components/PeopleSearch/index.js
+++ b/src/components/Header/components/PeopleSearch/index.js
@@ -15,7 +15,7 @@ import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import './people-search.css';
 import peopleSearch from '../../../../services/people-search';
-const MIN_QUERY_LENGTH = 3;
+const MIN_QUERY_LENGTH = 2;
 
 //  TextBox Input Field
 const renderInput = inputProps => {
@@ -49,6 +49,7 @@ export default class GordonPeopleSearch extends Component {
   constructor(props) {
     super(props);
     this.getSuggestions = this.getSuggestions.bind(this);
+    this.renderNoResult = this.renderNoResult.bind(this);
     this.renderSuggestion = this.renderSuggestion.bind(this);
     this.reset = this.reset.bind(this);
     this.handleKeys = this.handleKeys.bind(this);
@@ -147,13 +148,23 @@ export default class GordonPeopleSearch extends Component {
     );
   }
 
+  renderNoResult() {
+    return(
+    <MenuItem className="people-search-suggestion" style=
+    {{paddingBottom: '5px'}}>
+      <Typography className="no-results" variant="body2">
+        No results
+      </Typography>
+    </MenuItem>)
+  }
+
   renderSuggestion(params) {
     const { suggestion, itemProps } = params;
     let suggestionIndex = this.state.suggestionIndex;
     let suggestionList = this.state.suggestions;
     // Bail if any required properties are missing
     if (!suggestion.UserName || !suggestion.FirstName || !suggestion.LastName) {
-      return;
+      return null;
     }
     return (
        <MenuItem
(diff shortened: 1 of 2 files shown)


src/components/Header/components/PeopleSearch/index.js
```diff
@@ -15,7 +15,7 @@ import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import './people-search.css';
 import peopleSearch from '../../../../services/people-search';
-const MIN_QUERY_LENGTH = 3;
+const MIN_QUERY_LENGTH = 2;
```
Matt-Ramos, line 18: Yes, thank you for listening to my suggestion! This will make it possible for those with two letter first/last names to be effectively searched

src/components/Header/components/PeopleSearch/index.js
```diff
@@ -110,6 +111,10 @@ export default class GordonPeopleSearch extends Component {
       if (suggestionIndex === -1) suggestionIndex = suggestionList.length - 1;
       this.setState({ suggestionIndex });
     }
+    if (key === 'Backspace') { // Bug fixed where suggestions are remembered after backspacing
+      const suggestions = [];
```
russtuck, line 115: Should this distinguish whether the resulting search string is still long enough?  Maybe I just don't understand the code well enough, but this looks like it will get rid of suggestions if I type 5 letters and then 1 backspace (leaving 4 letters, which should be enough to get suggestions).  Does clearing suggestions cause them to be re-fetched?

src/components/Header/components/PeopleSearch/index.js
```diff
@@ -152,69 +167,69 @@ export default class GordonPeopleSearch extends Component {
...
-          {/* If the first name matches either part (first or last name) of the query, don't
-              highlight occurrences of the query in the first name part of the username.
-
-              If the username contains a period, add it back in.
-
-              If the username contains a period,
-              If the last name matches either part (first of last name) of the query, don't
-              highlight occurrences of the query in the last name part of the username. */}
-          {!suggestion.FirstName.match(
+       <MenuItem
```
russtuck, line 170: It looks like this is now indented 3 spaces, instead of 2 as elsewhere.  Is that intentional?  Why?

src/components/Header/components/PeopleSearch/index.js
```diff
@@ -349,7 +370,13 @@ export default class GordonPeopleSearch extends Component {
                       }),
                     )}
                   </Paper>
-                ) : null}
+                ) : isOpen && this.state.suggestions.length === 0 &&
```
russtuck, line 373: Is there a way to avoid having 2 copies of this code? (Here and line 333)

src/components/Header/components/PeopleSearch/index.js
changman0: So, I did some manual testing and it still re-fetches suggestions. I think it just has to do with clearing the suggestions on the key event and then the query repopulates the array, updates suggestions again, and shows them.

src/components/Header/components/PeopleSearch/index.js
changman0: Unintentional. Will revert change with next pull.

src/components/Header/components/PeopleSearch/index.js
changman0: I'm not sure there's a simple way to do this. because I need lines 377 and 379 to hold line 378 and when I put that into a function it doesn't work as intended. If you're talking about the if else statements, I'm also not sure if it would be more efficient to turn it into a function because I think I would need to pass isOpen to it.

changman0:
I checked to make sure suggestions would still show up when needed after using backspace and it works as intended.
I copied and pasted the original renderSuggestion() function from github to get the original spacing back.
I left a comment about the last request. I'm not sure if I misunderstood it, because I don't think it would be much of a benefit to changing the if else.

src/components/Header/components/PeopleSearch/index.js
russtuck: I was thinking of the whole if/else, with almost-identical <Downshift> inside each.   I'm concerned with DRY code (don't repeeat yourself) for correctness.  I don't think efficiency is different enough to worry about.  Since it's a larger problem than just your change, let's submit this.  

If you have time afterwards, it would be nice to refactor (if possible) so there's one copy with the "online" parameter used internally.

PR #768, opened 2020-06-11
S20 wellness animation

changman0:
Fixed the compatibility across different operating systems. Need to be checked on Mac OS and Windows.


changman0:
bruh

PR #786, opened 2020-06-19
Refined the no results dropdown 

changman0:
Added CSS animations and a loading dropdown to smooth out the choppy no results to people suggestions list dropdown.

I tried to use GordonLoader, but the way I implemented the 'No results' dropdown didn't seem to work well with that solution.


src/components/Header/components/PeopleSearch/index.js
```diff
@@ -60,6 +60,8 @@ export default class GordonPeopleSearch extends Component {
       highlightQuery: String,
       loginDialog: false,
       network: 'online',
+      time: new Date().toLocaleString([], { hour: '2-digit', minute: '2-digit' }),
```
russtuck, line 63: Not needed?

src/components/Header/components/PeopleSearch/people-search.scss
```diff
@@ -40,8 +40,29 @@
   }
 
   .no-results {
```
russtuck, line 42: Maybe add a comment explaining that this shows loading for 350ms, then no results, and a word or two about how (opacity).

changman0:
The develop branch updated off another merge. So, this needs to be approved again.

PR #788, opened 2020-06-22
(S20) Create 404 Page

### LINKED ISSUE #675: More helpful/entertaining 404 page
Perhaps partner with the Scotty Mascot for a more fun landing page

cpabbot:
Created a 404 landing page, which appears when a user types in a url under `360.Gordon.edu/any-incorrect-location` that does not correspond with an existing route. (Closes #675)

Currently, it displays a simple message and our mascot, with an easter egg that appears after 5-10 seconds. Feedback is welcome!! But I wanted to put this PR in so we at least have a base 404 to work off of.

404page


src/views/Page404/components/ScottieDog/scottieDog.scss
```diff
@@ -0,0 +1,114 @@
...
+    border-right: solid 0.5em transparent;
+    border-bottom: solid 4.5em $dog-color;
+    transform-origin: bottom;
+    animation: wag-tail 1s infinite;
+}
+@keyframes wag-tail {
+    50% {
+        transform: translateX($dog-body-width - 2em + 0.8em) translateY(-4.5em + 1em) rotate(10deg);
+    }
+}
```
cpabbot, line 114: I didn't know how to make the scss files compile into css files the way it is done elsewhere, but it appears that React supports importing sass files anyways. Assuming this should be changed, how does that work?

Also, when I edit that I'll add a new line to the end of the file. I didn't realize that was an important practice.

src/views/Page404/components/ScottieDog/scottieDog.scss
```diff
@@ -0,0 +1,114 @@
...
+
+.dog {
+    position: relative;
+    /*** font size of root element controls size of dog ***/
+    // font-size: 30px;
+
+    /*** Used for walking across screen animation (optional) ***/
+    animation: dog-walk linear 30s;
+    animation-delay: 5s;
+    left: 150%;
```
russtuck, line 13: It appears that this starts the dog off to the right of the screen.  That makes it take about 10 seconds, instead of 5 (animation-delay), for the dog to appear.  Is that intentional?  (Mentioning that in comment above would save people wondering.)

Dog is cute!

src/views/Page404/mascot.svg
```diff
@@ -0,0 +1,93 @@
+<?xml version="1.0" standalone="no"?>
```
russtuck, line 1: Where did this file come from?
I haven't seen Gordon use a lion that looks like this before.  Did you consider using the "rampant lion"?
I'm not sure where to find it, but I did find this: 
https://i.turner.ncaa.com/sites/default/files/images/logos/schools/bgd/gordon.svg
It's white-on-white, so easier to see here: 
https://www.ncaa.com/schools/gordon
We could ask Stephen Dagley (Gordon's webmaster) for a good image.

src/views/Page404/components/ScottieDog/scottieDog.scss
russtuck: .scss to .css should be automatic in the build process, and it seems to be working here.

Finishing the file with a newline is nice because it's logical and consistent, and because some tools (more often on Linux than Windows) care.  I don't think the absence breaks anything in our typical work flow.

src/views/Page404/components/ScottieDog/scottieDog.scss
cpabbot: Yeah good point, I can add a comment about that. The extra distance was important for the 'end' of the walk to make sure the dog is off the screen, but this part I should be able to just make it close to 100% and/or comment it.

src/views/Page404/mascot.svg
cpabbot: This lion is the new Gordon mascot, which changed toward the middle-end of this past academic year. I just converted it to an svg for quality and then changed the color to match the Gordon blue. The other one (Gordon logo) could work as well but I just thought the mascot worked well.

src/views/Page404/components/ScottieDog/scottieDog.scss
cpabbot: Okay got it. I was confused though because other places the scss file actually goes directly into a css file (in VSCode) and the index.js files import the css file (not the scss file like I did). I wonder which way is the best practice and how the other way even works

PR #791, opened 2020-06-22
S20 wellness check logged out

jz1210:
Put the wellness tab and button for logged-out view


jahnueldorelus:
There are some files I recreated when I re-made the Desktop view menu. Due to this, if I approve the Pull Request, it's guaranteed to break my code. Therefore, I will close this request and do a "git merge" instead which would allow me to place this incoming code into the correct places. 

Apart from that, I reviewed the code and it looks well-maintained.

PR #799, opened 2020-06-29
S20 chapel link

Matt-Ramos:


Matt-Ramos:
Clicking "need more chapel credits?" button from attended events page now links to events with chapel credits only. The URL also changes its params based on the filter that's selected. Works with back button and direct link as well.

src/views/Events/index.js
```diff
@@ -50,22 +50,108 @@ export default class Events extends Component {
...
+  componentDidMount() {
     this.loadEvents();
+    window.addEventListener('resize', this.resize);
   }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.resize);
+  }
+
+  async loadPrevious() {
```
aridospassos, line 73: a comment here explaining whats going on on this particular part would be really helpful.

src/views/Events/index.js
```diff
@@ -50,22 +50,108 @@ export default class Events extends Component {
...
+      this.setState({ filteredEvents: events, loading: false });
+
+      // If the include past filter is on, we get the events from the past
+      if (this.state.includePast) {
+        /* When togglePastEvents() runs, it changes the state of includePast to its opposite.
+         * So, when includePast is true, we set it to false so that togglePastEvents() will set it
+         * back to true
+         */
+        this.setState({ includePast: false });
+        this.togglePastEvents();
```
aridospassos, line 113: Might be good to put the state change after the function call. I could see the re-rendering causing some problems because it technically re-renders before getting to the function call.

src/views/Events/index.js
```diff
@@ -50,22 +50,108 @@ export default class Events extends Component {
...
+      const events = await gordonEvent.getFilteredEvents(this.state);
+      this.setState({ filteredEvents: events, loading: false });
+
+      // If the include past filter is on, we get the events from the past
+      if (this.state.includePast) {
+        /* When togglePastEvents() runs, it changes the state of includePast to its opposite.
+         * So, when includePast is true, we set it to false so that togglePastEvents() will set it
+         * back to true
+         */
+        this.setState({ includePast: false });
```
russtuck, line 112: I don't think there is any guarantee that setState() here will take effect before the next line.  See:
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous
https://reactjs.org/docs/react-component.html#other-apis-1

src/views/Events/index.js
```diff
@@ -101,6 +188,7 @@ export default class Events extends Component {
       const events = gordonEvent.getFutureEvents(allEvents); //Filter out past events initially
       this.setState({ allEvents, events, loading: false, filteredEvents: events });
     }
+    this.loadPrevious();
```
russtuck, line 191: Why is this added?  A comment explaining the reasoning would help.

src/views/Events/index.js
```diff
@@ -30,7 +30,7 @@ export default class Events extends Component {
       search: '',
       chapelCredits: false,
       art: false,
-      sports: false,
+      athletics: false,
```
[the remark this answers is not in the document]
Matt-Ramos: good catch, fixed. reverted back to sports as it wasn't working

src/views/Events/index.js
Matt-Ramos: When togglePastEvents() runs, it changes the state of includePast to its opposite. So, when includePast is true, we set it to false so that togglePastEvents() will set it back to true.

If I put it after the function call then if i had the include past filter selected and refreshed the page, it would go away.

src/views/Events/index.js
Matt-Ramos: added comment

src/views/Events/index.js
```diff
@@ -50,22 +50,108 @@ export default class Events extends Component {
...
+        athletics,
+        calendar,
+        cec,
+        fair,
+        chapelOffice,
+        studentLife,
+      });
+
+      this.setState({ loading: true });
+      const events = await gordonEvent.getFilteredEvents(this.state);
```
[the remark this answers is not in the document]
Matt-Ramos: if I don't pass in the states, which are the filters in this case, it wont correctly filter

src/views/Events/index.js
Matt-Ramos: If I remove the setState or move it after the toggle, it doesn't work appropriately

src/views/Events/index.js
Matt-Ramos: added comment explaining that this was added to load previously set filters when loading the events either as an authenticated user or as a guest

src/views/Events/index.js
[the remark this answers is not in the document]
russtuck: OK.  The state is not much more than the filters (just hasFilters and loading), so it's reasonable to pass the whole state.

src/views/Events/index.js
russtuck: I understand it won't work without it.  My concern is that in might sometimes not work even with it.

russtuck: It's not save to call togglePastEvents() here.  I think you'll need to call just part of it.  See my note there about how to refactor.

src/views/Events/index.js
```diff
@@ -79,6 +227,7 @@ export default class Events extends Component {
       const futureEvents = gordonEvent.getFutureEvents(this.state.allEvents);
       await this.setState({ events: futureEvents });
     }
+    this.createURLParameters();
```
russtuck, line 230: I'd suggest breaking togglePastEvents() into 2 functions.  One is the toggle part, and the other is the rest, starting here.  Then you can call both, or just one.

src/views/Events/index.js
```diff
@@ -123,8 +123,9 @@ export default class Events extends Component {
          * So, when includePast is true, we set it to false so that togglePastEvents() will set it
          * back to true
          */
-        this.setState({ includePast: false });
-        this.togglePastEvents();
+        this.setState({ includePast: false }, () => {
+          this.togglePastEvents();
```
russtuck, line 127: This code is OK, but it's creating a new unnamed function that just calls a named function.  It would be slightly simpler and cleaner to say
  this.setState({ includePast: false }, this.togglePastEvents);

PR #810, opened 2020-06-30
S20 timesheets clock in out

aridospassos:
added basic front end working clock in and out button to the time sheets page. Styling and other peripherals  can be added as a separate part later on.

### DIFF
--- a/src/services/jobs.js
+++ b/src/services/jobs.js
@@ -92,7 +92,7 @@ const clockOut = () => {
 };
 
 const deleteClockIn = () => {
-  return http.get(`jobs/deleteClockIn`);
+  return http.put(`jobs/deleteClockIn`);
 };
 
 export default {
--- a/src/views/Timesheets/index.js
+++ b/src/views/Timesheets/index.js
@@ -66,8 +66,7 @@ const Timesheets = (props) => {
   
   // disabled lint in some lines in order to remove warning about race condition that does not apply
   // in our current case.
-  // eslint-disable-next-line
-  useEffect(async ()=>{
+  useEffect(()=>{
     // updates ui with the current status of the users clocked in feature
     // either clocked in and ready to clock out or the apposite.
     async function getClockInOutStatus(){
@@ -76,14 +75,13 @@ const Timesheets = (props) => {
             
             if(status[0].currentState){
               setClockInOut("Clock Out");
-              //
               handleDateChangeInClock(new Date(status[0].timestamp));
     
             } else {
               setClockInOut("Clock In");
               }
          } catch(error) {
-            //did nothing
+            //do nothing
            }
     }
 
--- a/src/views/Timesheets/index.js
+++ b/src/views/Timesheets/index.js
@@ -69,6 +69,7 @@ const Timesheets = (props) => {
   useEffect(()=>{
     // updates ui with the current status of the users clocked in feature
     // either clocked in and ready to clock out or the apposite.
+    // status is notted by either true or false. true being clocked in.
     async function getClockInOutStatus(){
       try{
             let status = await jobs.clockOut();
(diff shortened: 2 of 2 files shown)


src/services/jobs.js
```diff
@@ -83,6 +83,18 @@ const submitShiftsForUser = (shiftsToSubmit, submittedTo) => {
...
+  return http.post(`jobs/clockIn`, data);
+};
+
+const clockOut = () => {
+  return http.get(`jobs/clockOut`);
+};
+
+const deleteClockIn = () => {
+  return http.get(`jobs/deleteClockIn`);
+};
```
cpabbot, line 96: You may want to change the type of HTTP requests here, but I left comments on the API PR on that

src/views/Timesheets/index.js
```diff
@@ -62,7 +62,48 @@ const Timesheets = (props) => {
...
+  useEffect(async ()=>{
+    // updates ui with the current status of the users clocked in feature
+    // either clocked in and ready to clock out or the apposite.
+    async function getClockInOutStatus(){
+      try{
+            let status = await jobs.clockOut();
+            
+            if(status[0].currentState){
+              setClockInOut("Clock Out");
+              //
```
cpabbot, line 79: unfinished comment? Also perhaps it would be helpful to explain what line 77 is checking. I have no idea what status[0] is nor what the currentState is (guessing its the clock in out state, but perhaps that could be made more clear with a comment or two?)

src/views/Timesheets/index.js
```diff
@@ -62,7 +62,48 @@ const Timesheets = (props) => {
   const [saving, setSaving] = useState(false);
   const [snackbarText, setSnackbarText] = useState('');
   const [snackbarSeverity, setSnackbarSeverity] = useState('');
+  const [clockInOut, setClockInOut] = useState('Clock In');
```
cpabbot, line 65: Figuring this all out helped me get a basic understanding of hooks, good work figuring out how to do that effectively!

src/views/Timesheets/index.js
```diff
@@ -356,6 +397,31 @@ const Timesheets = (props) => {
...
+          await jobs.clockIn(false);
+          let clockOutDate = new Date();
+          handleDateChangeOut(clockOutDate);
+          await jobs.deleteClockIn();
+      } 
+      if(clockInOut === "Reset") {
+          setClockInOut("Clock In");
+          setSelectedDateIn(null);
+          setSelectedDateOut(null);
+
```
cpabbot, line 420: I was initially hesitant of this, because the variable is a string and it seemed potentially easy to get confused, but now that I understand hooks a bit better this makes sense. It seems to be pretty efficient code. Someone else should perhaps double check but I think it works well (maybe add a comment before the clockin method to clarify what it is doing, and I also wonder if there is a better name, since the clockin function is also handling clock out and reset. Perhaps "handleClockEvent" or something might be more accurate?)

src/views/Timesheets/index.js
aridospassos: I return data using the <IEnumerable> type which is a list of lists. However I only really have 1 list in that list. The benefit from using it come in the back end when we are parsing data that comes back from the database. it lets us specify exactly what we a want to do with every piece of data coming from the data. The downside to it is that I have to reference what I get back in the front end with a [0] so that it knows i'm talking about the first item in the list. Even though there's only 1.

aridospassos: added a comment about status and what it holds

src/views/Timesheets/index.js
jahnueldorelus: Everything looks good to me. There are just two things. The first one is the same as what Cameron pointed out. I was confused seeing that function was called clockIn even though it handled clocking in, out, and resetting. The second thing are the if statements. Did you mean to use "else if" for line 408 and 515? Without it, each if statement will process. It's not really a big deal at all since it very minuscule code haha.

PR #811, opened 2020-06-30
S20 faculty wellness check fix

aridospassos:
fixed bug causing wellness check to crash on faculty page due to lack of phone number variable.

### DIFF
--- a/src/views/Home/components/Question/questionData.js
+++ b/src/views/Home/components/Question/questionData.js
@@ -34,14 +34,19 @@ export async function getQuestions() {
   
   await getUserData();
   let backendQuestions = await wellness.getQuestion();
-
-  let phoneNumber = `(${user.MobilePhone.substring(0, 3)}) ${user.MobilePhone.substring(
-    3,
-    6,
-  )}-${user.MobilePhone.substring(6)}`;
-
-  //comment below disables the eslint warning so we can execute dynamic string parsing without warnings
   
+  let phoneNumber;
+  
+  if(user.MobilePhone === undefined){
+     phoneNumber = "your phone number";
+  } else{
+    phoneNumber = `(${user.MobilePhone.substring(0, 3)}) ${user.MobilePhone.substring(
+        3,
+        6,
+       )}-${user.MobilePhone.substring(6)}`;
+  }
+  
+
   /* eslint-disable no-template-curly-in-string */
   let wellnessQuestion = backendQuestions[0].question.replace("${user.FirstName}", `${user.FirstName}`);
   wellnessQuestion = wellnessQuestion.replace("${user.LastName}", `${user.LastName}`);
--- a/src/views/Home/components/Question/questionData.js
+++ b/src/views/Home/components/Question/questionData.js
@@ -35,15 +35,19 @@ export async function getQuestions() {
   await getUserData();
   let backendQuestions = await wellness.getQuestion();
   
-  let phoneNumber;
+  let contactInfo;
   
-  if(user.MobilePhone === undefined){
-     phoneNumber = "your phone number";
-  } else{
-    phoneNumber = `(${user.MobilePhone.substring(0, 3)}) ${user.MobilePhone.substring(
-        3,
-        6,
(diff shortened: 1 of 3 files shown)


src/views/Home/components/Question/questionData.js
```diff
@@ -34,14 +34,19 @@ export async function getQuestions() {
   
   await getUserData();
   let backendQuestions = await wellness.getQuestion();
-
-  let phoneNumber = `(${user.MobilePhone.substring(0, 3)}) ${user.MobilePhone.substring(
-    3,
-    6,
-  )}-${user.MobilePhone.substring(6)}`;
-
-  //comment below disables the eslint warning so we can execute dynamic string parsing without warnings
```
cpabbot, line 43: Should you keep this comment?

src/views/Home/components/Question/questionData.js
```diff
@@ -34,14 +34,19 @@ export async function getQuestions() {
...
   
+  let phoneNumber;
+  
+  if(user.MobilePhone === undefined){
+     phoneNumber = "your phone number";
```
cpabbot, line 41: Wondering if this should be "no phone number found" or something, but probably doesn't matter that much. Good bug catch

src/views/Home/components/Question/questionData.js
aridospassos: Dr. tuck wants us to explain why we used an eslint disable comment if we have to use one so that the next person working on it know why.

src/views/Home/components/Question/questionData.js
aridospassos: so this phoneNumber varible is eaither going to display a persons phone number or " your phone number". This is like this for the purpose of clarity when the people in the database are writing the questions. They are only going to specify the variable and then the rest is on us. so if they right "we will contact you at {phone number}" it will show up as "we will contact you at 781 526 5849" or " we will contact you at your phone number"

src/views/Home/components/Question/questionData.js
cpabbot: Yeah that's why I'm wondering why you have it deleted? Seems like a helpful comment

src/views/Home/components/Question/questionData.js
cpabbot: Ah okay thanks for the clarification, this makes perfect sense in context of what the rest of the sentence would be

src/views/Home/components/Question/questionData.js
```diff
@@ -34,14 +34,19 @@ export async function getQuestions() {
...
+  } else{
+    phoneNumber = `(${user.MobilePhone.substring(0, 3)}) ${user.MobilePhone.substring(
+        3,
+        6,
+       )}-${user.MobilePhone.substring(6)}`;
+  }
+  
+
   /* eslint-disable no-template-curly-in-string */
```
cpabbot, line 50:
```suggestion
  // comment below disables the eslint warning so we can execute dynamic string parsing without warnings
  /* eslint-disable no-template-curly-in-string */
```

src/views/Home/components/Question/questionData.js
aridospassos: oh I didn't think I deleted it. thanks, I committed the suggested changes.

src/views/Home/components/Question/questionData.js
```diff
@@ -34,14 +34,20 @@ export async function getQuestions() {
...
-    3,
-    6,
-  )}-${user.MobilePhone.substring(6)}`;
-
-  //comment below disables the eslint warning so we can execute dynamic string parsing without warnings
   
+  let phoneNumber;
+  
+  if(user.MobilePhone === undefined){
+     phoneNumber = "your phone number";
```
Matt-Ramos, line 41: I think that if there's no mobile phone, then we should try home phone, and if theres no home phone then maybe this fix or email works.

aridospassos: fixed it!

Matt-Ramos:
The changes LGTM

PR #823, opened 2020-07-07
Add link to author profile for news items 

Marilyn-S:


src/views/News/components/NewsItem/index.js
```diff
@@ -24,6 +25,7 @@ export default class NewsItem extends Component {
     const { posting } = this.props;
     const { size } = this.props;
     const postingDescription = posting.Body;
+    const authorLink = posting.author.replace(" ", ".");
```
cpabbot, line 28: This works, but you can use posting.ADUN to give you the username. I added the posting.author one in the format method of the news.js service

src/views/News/components/NewsItem/index.js
```diff
@@ -32,7 +34,11 @@ export default class NewsItem extends Component {
           <Grid container onClick={this.handleExpandClick} className="news-item">
             <Grid item xs={12}>
               <Typography variant="h6" className="news-heading" style={{fontWeight: "bold"}}> {posting.Subject} </Typography>
-              <Typography className="news-content" style={{textTransform: "capitalize"}}> {posting.author} ({posting.datePosted}) </Typography>
+              <Link className="text-link" to={`/profile/${authorLink}`}>
```
cpabbot, line 37:
```suggestion
              <Link className="text-link" to={`/profile/${posting.ADUN}`}>
```

src/views/News/components/NewsItem/index.js
cpabbot, line 28:
```suggestion

```

src/views/News/components/NewsItem/index.js
```diff
@@ -56,7 +62,11 @@ export default class NewsItem extends Component {
               <Typography className="news-column" style={{fontWeight: "bold"}}>{posting.Subject}</Typography>
             </Grid>
             <Grid item xs={3}>
-              <Typography className="news-column" style={{textTransform: "capitalize"}}>{posting.author}</Typography>
+            <Link className="text-link" to={`/profile/${authorLink}`}>
```
cpabbot, line 65:
```suggestion
            <Link className="text-link" to={`/profile/${posting.ADUN}`}>
```

src/views/News/components/NewsItem/newsItem.scss
```diff
@@ -13,6 +13,10 @@
     transition: background-color 150ms;
 }
 
+.text-link:hover {
+    color: white !important;
+}
```
cpabbot, line 18: !importants always make me a bit queasy haha, but if that's the only way to override the MUI css then so be it

src/views/News/components/NewsItem/newsItem.scss
```diff
@@ -13,6 +13,10 @@
     transition: background-color 150ms;
 }
 
+.text-link:hover {
```
cpabbot, line 16: Optional but wondering if this class would be more accurately named something a little more specific like 'newsAuthorProfileLink' (or with dashes as apparently is the convention that I don't follow lol) but totally up to you!!

src/views/News/components/NewsItem/index.js
Marilyn-S: Thanks! I knew there had to be a better way but I couldn't think of anything. I'll get on this right away.

src/views/News/components/NewsItem/newsItem.scss
Marilyn-S: Good idea - always wise to lean on the obvious side when naming things. I'll probably use the dashes for consistency since other styles referenced in the NewsItem file uses the same convention.

Marilyn-S: Or maybe a hybrid of both for naming conventions - like news-authorProfileLink?

src/views/News/components/NewsItem/newsItem.scss
Marilyn-S: Same. But yeah I haven't been able to override the MUI otherwise so far.

PR #829, opened 2020-07-12
(S20) News Page - "Delete News" UI

### LINKED ISSUE #805: (3) Deleting personal news in UI

cpabbot:
#### Initial UI For Delete News on the news page -> DEVELOP
News is still an unfinished feature. This pull request adds the ability to delete news displayed on the News Page UI from the database.

Currently Working:
- Delete news: a button that shows *only* for news items that the currently authenticated user has authored
- Snackbar displays success/failure message
Closes #805

Needs Work:
- Improve snackbar feedback (icon change? like with membership requests)
- Maybe see if there is any way to make snackbars not be terrible (have to pass through every child to send to each component, probably no workarounds).
- Delete button does not contrast well with approved news items (navy on cyan, perhaps change all news items to gray like unapproved news items)
- (Most things from News 2.0)
   - More consistent item heights
  - Perhaps allow re-rendering of data without refreshing page (move data retrieval from mount method to render method?)
  - Edit personal news items
  - Limit number of news entries get pulled into home card (or manage somehow)
  - Home card listings are sorted by category but not in the way originally intended (which is probably better anyway)
  - Search bar
  - Filter news page

image

### DIFF
--- a/src/views/News/components/NewsList/index.js
+++ b/src/views/News/components/NewsList/index.js
@@ -49,7 +49,7 @@ export default class NewsList extends Component {
     window.removeEventListener('resize', this.resize);
   }
 
-  render() {;
+  render() {
     const { news } = this.props;
     const { personalUnapprovedNews } = this.props;
     let postings;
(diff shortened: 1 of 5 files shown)


src/views/News/components/NewsList/index.js
```diff
@@ -49,7 +49,7 @@ export default class NewsList extends Component {
     window.removeEventListener('resize', this.resize);
   }
 
-  render() {
+  render() {;
```
Marilyn-S, line 52: Was there supposed to be a semicolon here?

cpabbot: Wow good catch! I don't know how it was still working but fixed!

PR #831, opened 2020-07-13
Updated Service Worker

jahnueldorelus:
Recreated Service Worker into different parts for organization and readability.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -92,18 +92,18 @@
     <!-- Different font key for development and production -->
     <!-- Font will not work in development and will cause a 403, because the only domains allowed are *.gordon.edu -->
     <!-- Prod font key -->
-    <link
+    <!-- <link
       rel="stylesheet"
       type="text/css"
       href="https://cloud.typography.com/7763712/7294392/css/fonts.css"
-    />
+    /> -->
 
     <!-- Dev font key -->
-    <!--<link
+    <link
       rel="stylesheet"
       type="text/css"
       href="https://cloud.typography.com/7763712/6754392/css/fonts.css"
-    />-->
+    />
   </head>
   <body>
     <script src="./pwa.js"></script>
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,3 +1,10 @@
+/**
+ * ES-Lint warnings are disabled because there are functions/variables that are defined here but
+ * referenced in another script and this script itself is referencing variables from another script.
+ * Also, the global variable "self" is used instead of "window.self" because "window" is not
+ * defined within this scope. ES-Lint dislikes "self" but accepts "window.self"
+ */
+
 /* eslint-disable no-undef */
 /* eslint-disable no-restricted-globals */
 /* eslint-disable no-unused-vars */
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,6 +1,7 @@
-/* eslint-disable no-loop-func */
 /* eslint-disable no-undef */
 /* eslint-disable no-restricted-globals */
+/* eslint-disable no-unused-vars */
+
 /**
  * Caches files and responds to network requests when offline.
  *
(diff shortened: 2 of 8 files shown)


public/index.html
```diff
@@ -92,18 +92,18 @@
...
+    /> -->
 
     <!-- Dev font key -->
-    <!--<link
+    <link
       rel="stylesheet"
       type="text/css"
       href="https://cloud.typography.com/7763712/6754392/css/fonts.css"
-    />-->
+    />
```
cpabbot, line 106: I had no idea this was a thing haha, we'll just have to be sure to change it back when we push to master

public/sw.js
```diff
@@ -1,6 +1,7 @@
-/* eslint-disable no-loop-func */
 /* eslint-disable no-undef */
 /* eslint-disable no-restricted-globals */
+/* eslint-disable no-unused-vars */
```
cpabbot, line 3: wondering if this should be here? I know the unused variables warnings are a pain but it's helpful to make sure we don't keep any lying around

public/sw.js
```diff
@@ -1,6 +1,7 @@
-/* eslint-disable no-loop-func */
```
cpabbot, line 1: I don't know what this does but as long as you do then that's good

public/index.html
jahnueldorelus: Lol yeah, I never actually knew why we had it since it never worked in Develop haha

public/sw.js
jahnueldorelus: Very true and I have a good reason for doing so. I added those comments in the end after I coded everything and checked to make sure that there were no unused variables. The service worker only imports scripts. Importing scripts and importing files like you regularly would in a component are not the same. When you import a script, you're importing both the script's functions and global variables. In the service worker, I imported 3 scripts which makes them connect all together as one. Due to this, I could define a global variable in one script and call it in another even if it's not defined in that other script. Thanks to Dr. Tuck who helped me find that out, I took advantage of that to reduce the number of variables I would have to create. The extension ES-Lint doesn't really like that at all and screams at me for doing so 😂 So I had to put those comments there for each script so that the code would pass in Travis-CI

public/sw.js
jahnueldorelus: ES-Lint prevents functions from being created within a loop. In example, ES-Lint wouldn't allow this:

for (var i = 0; i < 10; i++) {
    funcs[i] = function() {
        return i;
    };
}

To be honest, I'm not sure why it was there in the first place. I removed that code and ES-Lint showed no warnings haha.

public/sw.js
cpabbot: Ah okay gotcha, that makes sense! I would almost add a quick comment explaining that at some point, but optional (or you can add it in your next branch). Might be useful for future people

public/sw.js
cpabbot: Oh interesting! That's really good you were able to remove it though - probably the less hacks and disables the better.

src/services/auth.js
```diff
@@ -37,17 +37,17 @@ const getAuth = (username, password) => {
...
-    grant_type: 'password'
+    grant_type: 'password',
   });
-  
+
   const request = new Request(`${base}token`, {
-    method: 'post', 
+    method: 'post',
     headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
     //mode: 'no-cors',
```
cpabbot, line 46: This here I remember seeing. The CORS error appears in our api whenever an incorrect route is attempted (I need to add that to the documentation) for some reason, so I'm guessing (unless you know what it's for) it was a past attempt to fix this (that's one of the solutions shown if you Google the error). So this could maybe be removed unless you think otherwise, optional change.

jahnueldorelus: I would remove it but unfortunately, CORS is really important. It helps to make sure that the backend serves requests only from Gordon 360 and not other possibly malicious sites. In general, you could remove CORS if the request is a GET and it's for non-sensitive data. For everything else, you'd probably want it active lol

jahnueldorelus:
> Approving but check the couple minor comments I added. Optional and you can always change them in a future PR if that works better. LGTM.

I've looked

PR #842, opened 2020-07-14
Updated My Profile View in Offline Mode

jahnueldorelus:
This pull request modifies offline mode for the My Profile URL. Originally when offline, clicking on the My Profile link would bring the user to their public profile page. This has been modified so that the user goes to their My Profile page instead. While offline on the My Profile page, all services that would require the user's browser to communicate with Gordon 360's API will be hidden. In other words, the user will only see their information and no buttons or switches that would allow them to make any changes.

The service worker has updated its removal of the old cache. Console logging has been disabled too

### DIFF
--- a/public/sw_user_cache.js
+++ b/public/sw_user_cache.js
@@ -1,3 +1,10 @@
+/**
+ * ES-Lint warnings are disabled because there are functions/variables that are defined here but
+ * referenced in another script and this script itself is referencing variables from another script.
+ * Also, the global variable "location" is used instead of "window.location: because "window" is not
+ * defined within this scope. ES-Lint dislikes "location" but accepts "window.location"
+ */
+
 /* eslint-disable no-undef */
 /* eslint-disable no-restricted-globals */
 /* eslint-disable no-unused-vars */
@@ -205,7 +212,6 @@ async function cacheUserFiles() {
       canceledUserFetches = [];
       failedUserFetches = [];
       successfulUserFetches = [];
-      userRemoteLinks = [];
       // Console logs the result of attempting to cache all of the user's files
       cachedAllSuccessfully
         ? console.log(`%c${successfulEmoji} Cached All User Files Successfully`, successfulLog)
--- a/public/sw_user_cache.js
+++ b/public/sw_user_cache.js
@@ -61,11 +61,13 @@ async function createRemoteUserLinks() {
       `${apiSource}/news/not-expired`,
       `${apiSource}/version`,
       `${apiSource}/events/chapel/${termCode}`,
+      `${apiSource}/vpscore`,
+      `${apiSource}/schedule`,
     ];
 
-    // Fetches the user's profile info to create the remote links. Their profile info is also cached
     try {
-      await fetch(
+      // Fetches the user's profile info to create the remote links. Their profile info is also cached
+      let userProfile = await fetch(
         (request = new Request(`${apiSource}/profiles`, { method: 'GET', headers })),
       ).then(async response => {
         // Checks to make sure the response of the fetch is okay before adding links to the list
@@ -78,7 +80,6 @@ async function createRemoteUserLinks() {
           // Creates the user's remote links
           const profile = await response.json();
           userRemoteLinks.push(
-            `${apiSource}/memberships/student/${profile.ID}`,
             `${apiSource}/memberships/student/username/${profile.AD_Username}/`,
             `${apiSource}/profiles/${profile.AD_Username}/`,
             `${apiSource}/profiles/Image/${profile.AD_Username}/`,
@@ -88,6 +89,7 @@ async function createRemoteUserLinks() {
             `/profile/${profile.AD_Username}`,
           );
           saveSuccessfulUserLink(userRequiredSource);
+          return profile;
         } else {
           // Response is not okay so creating the links failed
           areLinksCreated = false;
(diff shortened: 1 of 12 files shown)


public/sw_user_cache.js
```diff
@@ -285,12 +284,6 @@ async function removeUserCache() {
           cache.delete(item);
           userRemoteLinks = userRemoteLinks.filter(link => link !== item.url);
         }
-        // Removes '/profile/firstName.lastName' from the list of user's remote links and cache
-        // since that file is associated with the user.
-        else if (item.url.match(location.origin) && item.url.includes('/profile/')) {
-          cache.delete(item);
-          userRemoteLinks = userRemoteLinks.filter(link => link !== item.url);
-        }
       });
```
jahnueldorelus, line 287: The code above has been deleted because we no longer cache the user's public profile. Therefore this check is no longer required

src/components/MyProfileActivityList/index.js
```diff
@@ -92,17 +92,28 @@ class MyProfileActivityList extends Component {
...
+                  ) : (
+                    <div style={membershipItemStyle}>
+                      <Typography>
+                        <b>{Membership.ActivityDescription}</b>
+                      </Typography>
+                      <Typography>{Membership.SessionDescription}</Typography>
+                      <Typography>{Membership.ParticipationDescription}</Typography>
+                    </div>
+                  )}
                 </ListItem>
```
jahnueldorelus, line 117: This here is to prevent the Involvement Activity from being a link since you cannot view activity profiles offline

src/components/MyProfileActivityList/index.js
```diff
@@ -114,30 +125,33 @@ class MyProfileActivityList extends Component {
...
+                  </Grid>
+
+                  <Grid item xs={12} align="center">
+                    <Typography style={toggleTextStyle}>
+                      {Membership.Privacy ? 'Private' : 'Public'}
+                    </Typography>
+                  </Grid>
                 </Grid>
-              </Grid>
+              )}
```
jahnueldorelus, line 154: There's a check that happens to see if the user is offline. If so, the switch that would allow them to change the privacy of an activity is removed. Visually, it's actually hidden because the space that switches occupied remains visible while the switch itself is removed. This is to prevent the spacing of the card from changing when offline.

src/components/MyProfileActivityList/index.js
```diff
@@ -148,13 +162,20 @@ class MyProfileActivityList extends Component {
...
+                  to={`/activity/${Membership.SessionCode}/${Membership.ActivityCode}`}
+                  style={membershipItemStyle}
+                >
+                  <img src={Membership.ActivityImagePath} alt="" style={imgStyle} />
+                </Link>
+              ) : (
+                <div style={membershipItemStyle}>
+                  <img src={Membership.ActivityImagePath} alt="" style={imgStyle} />
+                </div>
+              )}
```
jahnueldorelus, line 178: Same case as src/components/MyProfileActivityList/index.js above!

src/components/MyProfileActivityList/index.js
```diff
@@ -175,20 +196,32 @@ class MyProfileActivityList extends Component {
...
+                    </Link>
+                  ) : (
+                    <div style={membershipItemStyle}>
+                      <Typography>
+                        <b>{Membership.ActivityDescription}</b>
+                      </Typography>
+                      <Typography>{Membership.SessionDescription}</Typography>
+                      <Typography>{Membership.ParticipationDescription}</Typography>
+                    </div>
+                  )}
```
jahnueldorelus, line 220: Same case as src/components/MyProfileActivityList/index.js above!

src/components/MyProfileActivityList/index.js
```diff
@@ -221,13 +257,20 @@ class MyProfileActivityList extends Component {
...
+                  to={`/activity/${Membership.SessionCode}/${Membership.ActivityCode}`}
+                  style={membershipItemStyle}
+                >
+                  <img src={Membership.ActivityImagePath} alt="" style={imgStyle} />
+                </Link>
+              ) : (
+                <div style={membershipItemStyle}>
+                  <img src={Membership.ActivityImagePath} alt="" style={imgStyle} />
+                </div>
+              )}
```
jahnueldorelus, line 273: Same case as src/components/MyProfileActivityList/index.js above!

src/components/Nav/components/NavAvatar/index.js
```diff
@@ -76,33 +76,6 @@ class GordonNavAvatar extends Component {
...
-      ) {
-        this.setState({ network: 'offline' });
-      }
-    });
-
-    /* Gets status of current network connection for online/offline rendering
-     *  Defaults to online in case of PWA not being possible
-     */
-    const networkStatus = JSON.parse(localStorage.getItem('network-status')) || 'online';
-
```
jahnueldorelus, line 105: This event listener is no longer needed because it was used only to determine if the avatar should be a link to the My Profile page or the user's public profile page

src/components/OfflineBanner/index.js
```diff
@@ -59,15 +59,17 @@ export default class OfflineBanner extends Component {
...
+    ) {
       return 'rounded-corners MuiGrid-grid-lg-10 MuiGrid-grid-md-12';
     }
     // Classes for People Search, Timesheets, Feedback, and My Profile view
     else if (
       this.props.currentPath === '/people' ||
       this.props.currentPath === '/student-timesheets' ||
-      this.props.currentPath === '/feedback' ||
-      this.props.currentPath === '/myprofile'
+      this.props.currentPath === '/feedback'
```
jahnueldorelus, line 72: Since the user's My Profile page is shown instead of their public profile, I had to change the size of the offline banner to fit the user's My Profile page

src/views/MyProfile/Components/VictoryPromiseDisplay/index.js
```diff
@@ -371,19 +371,21 @@ export default class VictoryPromiseDisplay extends React.Component {
...
-                </Typography>
-              </Grid>
+                  <Typography
+                    variant="body1"
+                    className="victory-promise-container-card-container-link-text"
+                    onClick={() => this.openVPLink()}
+                  >
+                    Click here for more information!
+                  </Typography>
+                </Grid>
```
jahnueldorelus, line 387: If the user is offline on their My Profile page, the link that's displayed in the victory promise is removed since the user doesn't have an internet connection

src/views/MyProfile/index.js
```diff
@@ -93,109 +93,55 @@ const MyProfile = props => {
...
+                <Grid container spacing={2}>
+                  {profileInfo}
+                  {officeInfo}
                 </Grid>
               </Grid>
-            </div>
-          )}
-        </div>
-      );
-    }
```
jahnueldorelus, line 148: The code below is removed because the My Profile page is now available offline

PR #854, opened 2020-07-23
(S20) Handshake Link on Profile

cpabbot:
## Handshake Social Link (UI)

_Note: This is a Completed PR Started During Summer Practicum_

Simply adds a handshake link to the profile social media links! Accepts either an 'app.handshake...' (hard coded) or 'gordon.handshake...' url prefix to validate link.


EjPlatzer (approved):
Looks great! Glad we caught the issue with the second prefix getting spliced incorrectly.

PR #863, opened 2020-07-24
added route name changes to match back end

aridospassos:
Route name changes to match the back end changes.

### DIFF
--- a/src/services/jobs.js
+++ b/src/services/jobs.js
@@ -67,7 +67,7 @@ const saveShiftForUser = async (
       HOURS_TYPE: hoursType,
       SHIFT_NOTES: shiftNotes,
     };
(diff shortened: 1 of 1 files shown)


src/services/jobs.js
```diff
@@ -33,7 +33,7 @@ const getActiveJobsForUser = (canUseStaff, details) => {
  */
 const getSavedShiftsForUser = (canUseStaff) => {
   if (canUseStaff) {
-    return http.get(`jobs/getSavedShiftsStaff/`);
+    return http.get(`jobs/savedShiftsStaff/`);
```
Matt-Ramos, line 36: in the API its return http.get(`jobs/savedShiftsForStaff/

PR #864, opened 2020-07-24
S20 staff timesheets

changman0:
Timesheets is ready to push. Staff timesheets will never allow staff to pick a job until the if statement is removed and staff timesheets is fully functional


bennettforkner:
Staff Timesheets (along with student timesheets unfortunately) will become obsolete with the new HR & Payroll system that is being implemented this summer. Criterion will allow us to have a much nicer online interface for managing time, so we plan to use that for staff and eventually for students as well. Because of this and since this pull request has not been deployed, **I officially recommend that we close this pull request** and abort the changes. @EjPlatzer @russtuck @ChrisImagineer

EjPlatzer:
I agree with @bennettforkner that this is probably not worth pursuing, since it will become obsolete once our new all-in-one HR and Payroll solution is in place. If all goes according to plan, the official Criterion time reporting system will replace timesheets on 360 (and GO Gordon) completely. In the meantime, I think we should keep this around in case we have need to come back to it at a later date. That would mean - in my mind - closing this pull request and the related issues, but keeping the branch. That way all of the code and previous discussion will still exist, and the issues and this pull request could be re-opened if needed.

bennettforkner:
@EjPlatzer agreed. Closing this PR but keeping the branch open for future reference...

PR #866, opened 2020-07-24
S20 show advisor

jz1210:
This is the UI part of the advisor. Backend will return first name, last name, and ADusername for advisors. Advisor info (first name and Lastname) will put into the public student profile list. If a student does not have an advisor, it will show (not assigned) in the advisor info list.


src/components/MajorList/index.js
```diff
@@ -36,13 +36,13 @@ export default class Majors extends Component {
     }
 
     // Creates the list item's title
-    if (this.props.majors.length === 1) {
+    if (this.props.majors.length < 2) {
       majorPrefix = (
```
jahnueldorelus, line 40: This has been updated so that if a person doesn't have any majors, the Major list item will still appear.

PR #873, opened 2020-08-09
S20 timesheets staff disable

jahnueldorelus:
- Timesheets is unavailable to non-students

### DIFF
--- a/src/views/Timesheets/index.js
+++ b/src/views/Timesheets/index.js
@@ -28,14 +28,15 @@ import './timesheets.css';
 import GordonLoader from '../../components/Loader';
 import { makeStyles } from '@material-ui/core/styles';
 import SimpleSnackbar from '../../components/Snackbar';
+import user from '../../services/user';
 
-const useStyles = makeStyles((theme) => ({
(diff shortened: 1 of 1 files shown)


src/views/Timesheets/index.js
```diff
@@ -582,209 +599,233 @@ const Timesheets = (props) => {
...
-                          classes={{ tooltip: classes.customWidth }}
-                          interactive
-                          disableFocusListener
-                          disableTouchListener
-                          title={
-                            canUseStaff
-                              ? 'Staff Timesheets Info' // need to update for staff
-                              : // eslint-disable-next-line no-multi-str
-                                'Student employees are not permitted to work more than 20 total hours\
+    if (networkStatus === 'online' && isUserStudent && props.Authentication) {
```
jahnueldorelus, line 602: I've changed the condition statement here. In the new condition statement, we check to make sure the user is logged in, a student, and online before displaying the Timesheets page. If the user is a staff, they will see a MUI card saying that Timesheets is unavailable. For both students and non-students, if they're offline, they will see a MUI card saying that they're offline.

PR #878, opened 2020-08-14
Change shift rounding from 15 minutes to 5 minutes (quarter to 12th of an hour)

chris-hansen:
Recording time more precisely will help make it affordable to hire Class Assistants to do cleaning between classes (lots of people, working frequent short shifts between their classes).

### DIFF
--- a/src/views/Timesheets/components/ShiftItem/index.js
+++ b/src/views/Timesheets/components/ShiftItem/index.js
@@ -222,8 +222,8 @@ export default class ShiftItem extends Component {
       zeroLengthShift = timeDiff === 0;
       shiftTooLong = calculatedTimeDiff > 20;
       let roundedHourDifference = (Math.round(calculatedTimeDiff * 12) / 12).toFixed(2);
-      if (roundedHourDifference == 0.00) {
-        roundedHourDifference = 0.08; //minimum 1/12th hour for working a shift (5 minutes)
+      if (roundedHourDifference == "0.00") {
+        roundedHourDifference = "0.08"; //minimum 1/12th hour for working a shift (5 minutes)
       } 
 
       this.setState({
--- a/src/views/Timesheets/components/ShiftItem/index.js
+++ b/src/views/Timesheets/components/ShiftItem/index.js
@@ -222,10 +222,10 @@ export default class ShiftItem extends Component {
       zeroLengthShift = timeDiff === 0;
       shiftTooLong = calculatedTimeDiff > 20;
       let roundedHourDifference = 0;
-      if (calculatedTimeDiff > 0 && calculatedTimeDiff < 0.25) {
-        roundedHourDifference = 0.25;
-      } else if (calculatedTimeDiff >= 0.25) {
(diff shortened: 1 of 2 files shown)


src/views/Timesheets/index.js
```diff
@@ -138,13 +138,13 @@ const Timesheets = props => {
     if (timeIn !== null && timeOut !== null) {
       checkForFutureDate(timeIn, timeOut);
       let timeDiff = timeOut.getTime() - timeIn.getTime();
-      let calculatedTimeDiff = timeDiff / 1000 / 60 / 60;
-      let roundedHourDifference = 0;
-      if (calculatedTimeDiff > 0 && calculatedTimeDiff < 0.1) {
-        roundedHourDifference = 0.1;
-      } else if (calculatedTimeDiff >= 0.1) {
-        roundedHourDifference = (Math.round(calculatedTimeDiff * 10) / 10).toFixed(2);
+      let calculatedTimeDiff = timeDiff / 3600000; //3,600,000 milliseconds in an hour.
```
russtuck, line 141: It's just a matter of preference, but the original code is clearer to me.

src/views/Timesheets/index.js
```diff
@@ -138,13 +138,13 @@ const Timesheets = props => {
...
+      let roundedHourDifference = (Math.round(calculatedTimeDiff * 12) / 12).toFixed(2);
```
russtuck, line 142: This reorganization of the code is a nice improvement.

However, is the "toFixed(2)" needed?  I don't think so, and the code would be a little clearer without it, so a further small improvement might be possible.

jason-whitehouse: I think the .tofixed(2) is necessary to display a result number value of such as 1 as a text value of 1.00 on the screen.  Although I'm not 100% sure.  In terms of rounding and database insertion it's definitely not.

jason-whitehouse:
OK! Do I want to Squash and Commit now??  There are about 6 of the 10 commits that are just syntax errors with curly braces and using == when === is apparently required in the UI javascript.

src/views/Timesheets/index.js
russtuck: Ah, that sounds reasonable.  Thanks.

src/views/Timesheets/components/ShiftItem/index.js
```diff
@@ -212,21 +212,19 @@ export default class ShiftItem extends Component {
     let shiftTooLong = false;
     let isOverlappingShift = false;
     let timeDiff;
-    let calculatedTimeDiff = timeDiff / 1000 / 60 / 60;
+    let calculatedTimeDiff = timeDiff / 3600000; //3,600,000 milliseconds in an hour.
```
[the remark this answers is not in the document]
jason-whitehouse: I think all files are using "3600000" without the need to update either of them any more.  Do you see one I missed?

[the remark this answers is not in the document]
russtuck: I thought I did, but I don't now, so I must have been confused.  Sorry about that.

PR #888, opened 2020-10-12
(F20) Student News Bug Fix - shows unapproved news even if no other news

cpabbot:
Adds a simple logical condition so that unapproved news shows if it exists rather than just "No news to show"


russtuck (approved):
Good fix.

PR #890, opened 2020-10-15
Update privacy

EjPlatzer:
Fix privacy for students marked as semi-private - their on/off campus status and dorm should be hidden from other students.

Because this is an issue of privacy concerns, it can't wait for the features in develop to become stable, so this is being merged directly into master. It has been tested thoroughly locally.

This branch updates the front end to properly display the updated (obscured) info received from the backend. However, all obscuring of info is done in the backend, so the info will still display until the backend update is merged.

Semi-private students now have an on/off campus value of 'P' for private, and the front end will display 'Private as requested' when this is received. 

Furthermore, both private and semi-private students will show on/off campus status and dorm to FacStaff as sensitive info.

Finally, when a request is sent for which the profile returned is null (i.e. because the profile is private to the logged in user), the user will be redirected to the 404page at the url /profilenotfound, rather than hanging on the loading indicator.

### DIFF
--- a/src/components/ProfileList/index.js
+++ b/src/components/ProfileList/index.js
@@ -154,7 +154,8 @@ class ProfileList extends Component {
     if (!this.props.myProf) {
       this.setState({
         campusLocationDisclaimer:
-          this.props.profile.KeepPrivate === 'S' || this.props.profile.KeepPrivate === 'P',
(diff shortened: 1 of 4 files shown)


jason-whitehouse (approved):
Looks good!

PR #892, opened 2020-10-15
Fix casing on variable for disclaimer

EjPlatzer:
The casing of the variable 'OnOffCampus' was incorrect in the check for the campusLocationDisclaimer, which was causing the disclaimer to always be true, making the campus location red for all users.


bennettforkner (approved):
Spectacular work here.

Issue #894, opened 2020-10-16
Timesheets Month-Day reversed with different Browser locale


EjPlatzer:
If the user has set their locale to somewhere outside the United States where date convention is MM/dd instead of dd/MM. the timesheets date entry will silently fail by handing off the incorrect date (e.g. February 10 instead of October 2nd) to the backend. 

Often but not always, this will result in the user not being able to see their jobs at all, since they are in the wrong term or an invalid date range.

Sometimes, though, the user will jobs from different terms or will just end up submitting a shift for the wrong dates (e.g. January 2 instead of February 1).

PR #896, opened 2020-10-19
Send date to backend as ISO string, not locale

EjPlatzer:
This commit fixes a timesheets bug (#894) where users set to a different locale (e.g. had browser language set to german) would not be able to retrieve their jobs. This was because when fetching jobs, the front end would send shift_start_datetime and shift_end_datetime as locale strings, and many international locales flip day and month in date strings (i.e. dd/MM/yy instead of MM/dd/yy). This would give the wrong date to the backend. Since ISO date strings are standardized and unaffected by locale, the backend should now always receive the correct date.


bennettforkner (approved):
Approved!

PR #899, opened 2020-10-22
Release: Bug fixes for timesheets and student news

EjPlatzer:
This pull request incorporates two small bug fixes:

1. Student news bug: unapproved news was not showing up for users when there was no approved news to show.

2. Timesheets bug: Users in specific browser locales (e.g. German) where the standard date format is `dd/MM/yy` instead of `MM/dd/yy` were having errors with the dates on their shifts. The frontend was sending dates to the backend as a localized string, so the backend would switch month and day, which would either be an impossible date (e.g. some day of the 20th month), or a very wrong date (February 10th instead of October 2nd).


src/views/News/index.js
```diff
@@ -297,10 +297,10 @@ export default class StudentNews extends Component {
     const networkStatus = JSON.parse(localStorage.getItem('network-status')) || 'online';
 
     if(this.props.Authentication) {
-
+      
       if (this.state.loading === true) {
         content = <GordonLoader />;
-      } else if (this.state.news.length > 0) {
+      } else if (this.state.news.length > 0 || this.state.personalUnapprovedNews.length > 0) {
```
joshuaRogers24, line 303: Shouldn't we check the length of this.state.filterednews instead of this,state,news? It seems that currently if filterednews was empty but there was some unexpired news, the news Card would just be empty and not say "No News to Show". Also, if we do decide to not check the length of 'news', there would be no need for this.state.news in the component and it could be removed.

joshuaRogers24: The addition looks good, though!

src/views/Timesheets/index.js
```diff
@@ -168,10 +168,10 @@ const Timesheets = props => {
   if (props.Authentication) {
     const getActiveJobsForUser = (dateIn, dateOut) => {
       let details = {
-        shift_start_datetime: dateIn.toLocaleString(),
-        shift_end_datetime: dateOut.toLocaleString(),
+        shift_start_datetime: dateIn.toISOString(),
+        shift_end_datetime: dateOut.toISOString(),
```
joshuaRogers24, line 172: Good! This was tested with a computer set to a different locale, right?

joshuaRogers24 (commented):
Both were good catches and clean solutions! Note the comments I made. I know one is not necessarily about this issue. We can fix it in another pull request, but I thought this was a good place to mention it.

src/views/News/index.js
EjPlatzer: @cpabbot can probably speak better to this then I can, but it looks like currently, filtering news is disabled, so news and filtered news will have the same value. However, when search is implemented, we will need to think of what to display in the case that there is no news matching the result.

Probably good to proactively update this if to check filterednews as you suggested Josh, good catch!

src/views/Timesheets/index.js
EjPlatzer: Yes, it was tested with computers in a US English locale and locales that were previously causing issues (such as Germany).

joshuaRogers24 (approved):
Looks good! I made an issue out of the question I had about code outside the scope of these changes.

PR #901, opened 2020-10-24
Fix attnd events sort

joshuaRogers24:
# Background

Because the `sortByTime` function was never exported from event.js, its use in user.js to sort attended chapel events never had any effect. Attended chapel events were not sorted. 
To make matters more complicated, a recent change in where chapel event data originates has left most chapel events without `Occurrences` values, which is what `sortByTime` uses to sort events.  
The goal of this fix was to make sure attended events are sorted for uses.

#  Solution

A new sorter function was created in user.js, `sortAtndEventsByTime`, that sorts based off of `Occurrences` only if it is non-null and by `CHDate` otherwise. `CHDate` works well as a default for sorting attended chapel events because it is the time when a student is awarded credit for the event and thus should never be null. (At least, based on how this time is always close to but rarely the same as the end time of the event, I assumed that is what `CHDate` is.) By calling `console.log()` on `events.sort(sortAtndEventsByTime)`, which is the same thing that the 'EventsAttended' component displays, it is clear that this new sorter works: (The datetimes with arrows pointing to them and the one crossed out are where `Occurrences` was used to sort, the remaining ones (bracketed) were where `CHDate` was used to sort.)
Screenshot 1
By making some dummy-data, a single event that is just 1 minute later than a real event, you can also see that the time zone on the end `Occurences` datetime does not cause problems when comparing to `CHDate`, which is missing the time zone in its format:
Screenshot 2

### DIFF
--- a/src/services/user.js
+++ b/src/services/user.js
@@ -396,7 +396,7 @@ const getLocalInfo = () => {
  * is non-null, since users would rather know when the event was
  * than when they got credit.
  * 
- * @param {JSON} event 
+ * @param {JSON} event : an event the user attended
  * @returns {DateTime} event.CHDate or event.Occurrences[0][0]
  * (since Occurences is a list of lists of start and end times
  * for each re-occurence of an event)
@@ -417,8 +417,9 @@ function getAtndEventTime(event) {
  * can sometimes be weeks after an event due to slow
  * processing.
  * 
- * @param {JSON} a 
- * @param {JSON} b 
+ * @param {JSON} a : an event
+ * @param {JSON} b : another event
+ * @returns {int} -1 if a's time is less than b's, 1 if it's more, 0 if they're equal
  */
 function sortAtndEventsByTime(a, b) {
 
--- a/src/services/user.js
+++ b/src/services/user.js
@@ -390,17 +390,26 @@ const getLocalInfo = () => {
   }
 };
 
+function sortAtndEventsByTime(a, b) {
+  if (a.CHDate < b.CHDate) {
+    return -1;
+  }
+  if (a.CHDate > b.CHDate) {
+    return 1;
+  }
+  return 0;
+}
+
 //Call function to retrieve events from database then format them
 const getAttendedChapelEventsFormatted = async () => {
   const termCode = session.getTermCode();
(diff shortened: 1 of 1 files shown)


src/services/user.js
```diff
@@ -390,17 +390,36 @@ const getLocalInfo = () => {
   }
 };
 
+function sortAtndEventsByTime(a, b) {
+
+  let timeA = a.CHDate;
```
EjPlatzer, line 395: Good job finding a different property to sort on. It might be helpful to leave a comment here about the meaning and relative accuracy of CHDate versus Occurrences[0][0], so that some future maintainer is not mystified by this code.

src/services/user.js
```diff
@@ -390,17 +390,36 @@ const getLocalInfo = () => {
...
+  if (a.Occurrences[0]) {
+    timeA = a.Occurrences[0][0];
+  }
+  let timeB = b.CHDate;
```
EjPlatzer, line 399: Since these two code blocks are totally repetitive, we could possibly extract them into a different method (e.g. getAtndEventTime). That way you just call `timeA = getAtndEventTime(a)` and likewise for b. This method would be more clear and singular in it's intent, and the complexity of getting the most accurate time available is abstracted to it's own method.

EjPlatzer (commented):
This seems like  a good temporary fix, but ultimately I think we ought to focus on getting complete event data, and sorting on the more accurate Occurrences datetime. If CHDate really is the time that credit was awarded, it's possible for that time to be a day or more delayed (e.g. if students are awarded credit for a non-chapel event like Departmental Convocation where they use the old sheets instead of a scanner, and there is a delay in processing those sheets). This could result in chapel events being displayed in a different order than they actually occurred - not a huge deal, but not ideal.

src/services/user.js
joshuaRogers24: Good point.

src/services/user.js
joshuaRogers24: Good idea!

EjPlatzer (dismissed):
LGTM!

EjPlatzer (approved):
Good to have the extra documentation. Thanks Travis!

PR #904, opened 2020-10-30
Apartapp search styling

NerdyGriffin:
Experimental feature: Added an optional prop to the GordonPeopleSearch to allow for triggering callback instead of creating a link, thus increasing the modularity of the GordonPeopleSearch component and eliminating a lot of redundant code from the ApartApps implementation.

I also made several structuring and styling changes to the apartment application to make better use of Material-UI components to simplify things further.

Lastly, I rearranged some of the file/folder locations of the ApartmentApp components so that they better fit the style guide.


NerdyGriffin:
I should have split these features/changes up into 2 or 3 separate pull requests, because some of them are not related to each other conceptually, even though strongly depend on the changes made in previous commits. In the future, I will try to split things up better to make the code review easier for the reviewer(s).

src/components/ApartmentApplicantList/index.js
```diff
@@ -2,23 +2,28 @@ import React, { Component } from 'react';
...
     super(props);
-    this.renderApplicant = this.renderApplicant.bind(this);
+    this.handleSelection = this.handleSelection.bind(this);
     this.handleRemove = this.handleRemove.bind(this);
   }
 
+  handleSelection = theChosenOne => {
+    // Make sure the chosen username was not null
+    if (theChosenOne && theChosenOne !== null) {
+      // Send the selected username to the parent component
```
joshuaRogers24, line 22: Can you just say `if (theChosenOne) {...` to check whether it is null or not? 
(Same question about `profile` on line 29)

src/components/ApartmentApplicantList/index.js
```diff
@@ -27,66 +32,27 @@ export default class ApplicantList extends Component {
...
+              onSearchSubmit={this.handleSelection}
+              Authentication={this.props.Authentication}
+            />
+          </ListItem>
+          {this.props.applicants.map(profile => (
+            <ApplicantListItem
+              key={profile.AD_Username}
+              profile={profile}
+              isPrimaryApplicant={profile === this.props.userProfile}
+              onApplicantRemove={this.handleRemove.bind(this, profile)}
```
joshuaRogers24, line 53: I have never seen bind be used on anything else than `this`. How does that work?

src/components/Header/components/PeopleSearch/index.js
```diff
@@ -138,7 +149,7 @@ export default class GordonPeopleSearch extends Component {
     var hasMatched = false;
     return (
       <span>
-        {parts.map((part) =>
+        {parts.map(part =>
```
joshuaRogers24, line 152: Do the majority of these `<someParam> => ...` style actions have parentheses around the `<someParam>`? I think I just saw a pull request the other day from one of the 360 team that added the parentheses to each instance of such syntax.

src/components/Header/components/PeopleSearch/index.js
```diff
@@ -334,17 +349,33 @@ export default class GordonPeopleSearch extends Component {
...
-                  // only bottom padding changed and 'no-results' class used
-                    <Paper square className="people-search-dropdown">
-                      {this.renderNoResult()}
-                    </Paper>) : null}
-              </span>
-            )}
-          </Downshift>
-        );
+                ) : (
+                  <Paper square className="people-search-dropdown">
```
joshuaRogers24, line 353: How does it work when a ternary operator has 5 operands? In other words, in which cases do each of these colon-separated operands take effect?

src/views/ApartmentApp/index.js
```diff
@@ -65,11 +76,11 @@ export default class ApartApp extends Component {
...
     if (profileToRemove && profileToRemove !== null) {
-      let applicantList = this.state.applicants; // make a separate copy of the array
-      let index = applicantList.indexOf(profileToRemove);
+      let applicants = this.state.applicants; // make a separate copy of the array
+      let index = applicants.indexOf(profileToRemove);
       if (index !== -1) {
-        applicantList.splice(index, 1);
-        this.setState({ applicants: applicantList });
+        applicants.splice(index, 1);
+        this.setState({ applicants });
```
joshuaRogers24, line 83: Should we have an error here in the case of index being -1 to warn that they are trying to remove someone who is not there?

src/views/ApartmentApp/index.js
```diff
@@ -154,66 +165,81 @@ export default class ApartApp extends Component {
...
-              </div>
-            )}
-          </div>
+        return this.state.loading ? (
+          <GordonLoader />
+        ) : (
+          <Grid container direction="row-reverse" justify="center" spacing={2}>
+            <Grid item xs={12} md={4}>
+              <Card>
+                <CardContent
```
joshuaRogers24, line 174: Could this styling be done in a .scss file? It could then save the repetition on 194

src/views/ApartmentApp/index.js
```diff
@@ -154,66 +165,81 @@ export default class ApartApp extends Component {
...
+                          readOnly: true,
+                        }}
+                      />
+                    </Grid>
+                  </Grid>
+                </CardContent>
+              </Card>
+              <br />
+              <Card>
+                <CardContent
```
joshuaRogers24, line 227: Previous question

joshuaRogers24 (changes requested):
Great work, Christian! I love seeing a little cleanup done along the way. Please see my comments and respond.

src/components/ApartmentApplicantList/index.js
NerdyGriffin: Yes... I think I had it originally check for something other than null when developing the method inside the subclass, but then that got changed and I guess I forgot that I could just remove that second part

src/components/ApartmentApplicantList/index.js
NerdyGriffin: It is used whenever setting a callback on a component to trigger a method of the parent class. If you do not use the `bind` with the callback function, then it will actually call the method every time the component is created/rendered.

src/components/Header/components/PeopleSearch/index.js
NerdyGriffin: This was done by the Prettier formatter, it likes to do that a lot, but also sometimes Prettier will specifically remove them too. As a guess: maybe it adds them when it is in a JSX expression inside the `{ ... }` but maybe it removed them if it is a method body?
TLDR: Idk why, but Prettier adds parentheses in some cases and removes them in other cases

joshuaRogers24:
After looking at it in action, just a few more thoughts:

- Should the icon for removing match what we see in other parts of 360? (see timesheets (a trash bin) and membership requests (an  X)
- Can the text color for name and class standing be changed to black instead of dark blue?
- Can the text in the peopleSearch bar be changed to "Add applicants" or something so it is more clear how a user would do that?
-The error messages are not the same style as others on the site-  I think there is usually red text explaining the error. Would that be easy to add or should we wait to change it later?

src/components/ApartmentApplicantList/index.js
joshuaRogers24: Ah, whoops!

src/components/Header/components/PeopleSearch/index.js
joshuaRogers24: Huh, okay. It's fine then.

src/components/ApartmentApplicantList/index.js
joshuaRogers24: Okay, so if you did not bind to profile, the application page would re-render every time we removed a profile?

src/components/Header/components/PeopleSearch/index.js
NerdyGriffin: It is just like two nested `if` statements, it works like
```
if (condition1) {              // condition1 ? (
  if (condition2) {            //   condition2 ? (
    return (<FirstThing />);   //     <FirstThing />
  } else {                     //   ) : (
    return (<SecondThing />);  //     <SecondThing />
  }
} else {                       // ) : (
  return (<ThirdThing />);     //   <ThirdThing />
}                              // )
```
I tried it out a bunch in CodePen before putting in the real code, because I saw this same thing done in other parts of 360 and I was also confused about that at first

NerdyGriffin: You could easily nest the second condition inside the outer `else` case too, just by putting that part after the first colon. It is far more versatile than I thought at first

src/views/ApartmentApp/index.js
NerdyGriffin: Yes, it is a check to ensure that `splice` does not get called with an invalid `index`, because otherwise it will "crash" the whole page. It was just the standard way that I saw this done when comparing multiple online guides, I guess technically you could use a `try ... catch` but that is somewhat less elegant.

src/views/ApartmentApp/index.js
NerdyGriffin: It could be done that way, but it would actually add significantly more lines of code because the `Grid` components are essentially just referencing a set a modular css definitions from `material-ui`.
While they may be confusing to work with at first, I have found that the `Grid` components make the styling a lot more consistent and reliable across various browsers, because it allows you to define the styling with those shorthand properties that would equivalently take maybe a dozen line in the .scss file (especially the rules for handling the breakpoints)

NerdyGriffin:
> 
> 
> After looking at it in action, just a few more thoughts:
> 
>     * Should the icon for removing match what we see in other parts of 360? (see timesheets (a trash bin) and membership requests (an  X)
> 
>     * Can the text color for name and class standing be changed to black instead of dark blue?
> 
>     * Can the text in the peopleSearch bar be changed to "Add applicants" or something so it is more clear how a user would do that?
>       -The error messages are not the same style as others on the site-  I think there is usually red text explaining the error. Would that be easy to add or should we wait to change it later?

I will respond to these in the same order:
- I did not think to check the icons used in other parts. I originally tried it with the trashcan icon but it seemed a bit confusing at the time, but I can easily try that again. I will make a note to look at other 360 pages as well, because that is an easily solution
- The text color is actually a bug, because the list items are actually links to the persons profile page. I had intended to fix that coloring, but I guess I forgot, so I will definitely do that
- I was actually very much wondering what text should be for that. It is a very easy change for me to make, I actually had custom text in it before but I changed it back because I wasn't sure how to phrase the text and put it on the todo list for later.
- The error message using `alert` are only a placeholder, I planned to implement them like in the IDUploader or one of the other pages, but I did not go down that rabbit-hole yet because I wanted to keep this specific feature change focused on the search and applicants list. There are actually some other pages on 360 that I saw using `alert` instead of the stylized error dialogs, but I agree that it should be change over to the stylized versions. 
  - I originally had it displaying with red text underneath the search bar, but that feature stopped working when I tried to make it an optional feature of the peopleSeach like I did with `disableLink`, so I removed that part for now and put on my list to explore further later.

src/components/Header/components/PeopleSearch/index.js
joshuaRogers24: Oh, okay! I did not notice the other `?`s

src/views/ApartmentApp/index.js
joshuaRogers24: Okay, but should we have an error message for the user in the case of index being -1?

src/views/ApartmentApp/index.js
joshuaRogers24: Good point. What about the style for CardContent, though?

src/views/ApartmentApp/index.js
NerdyGriffin: Since the remove button only exists if the person is being displayed in the list, then there isn't really a need to put an error message (no need to worry about a user clicking a button that does not exist)

src/views/ApartmentApp/index.js
NerdyGriffin: That styling is actually kind-of redundant and unnecessary in this page, it was leftover from when we originally copied from timesheets. I have since removed that and also split up the cards to match the visual style things like the Involvements page or MyProfile (I just discovered the material-ui `CardHeader` components which the MyProfile page uses, and those also greatly simplify the code

src/components/ApartmentApplicantList/index.js
NerdyGriffin: No, it would actually to call the `handleRemove` method every time we render the remove button (Which would essentially remove every person as soon as they are added). 

The `bind(this, profile)` part is not binding it to `profile`, it is just binding to `this` and passing `profile` to the `handleRemove` as a parameter.

NerdyGriffin:
I believe I have completed the requested changes, let me know if I missed anything

src/components/ApartmentApplicantList/index.js
joshuaRogers24: Oh, okay!

src/views/ApartmentApp/index.js
joshuaRogers24: That's fair.

src/views/ApartmentApp/index.js
```diff
@@ -39,34 +54,24 @@ export default class ApartApp extends Component {
...
-      alert('Something went wrong while trying to get the profile of ' + username);
+    let applicants = this.state.applicants; // make a separate copy of the array
+    // Get the profile of the selected user
+    let applicantProfile = await user.getProfileInfo(username);
+    // Check if the selected user is a student
+    if (!String(applicantProfile.PersonType).includes('stu')) {
+      // Display an error if the selected user is not a student
+      let newErrorText = applicantProfile.fullName + ' is not a student';
+      this.setState({ errorDialogOpen: true, errorDialogText: newErrorText });
+      console.log('Error: ' + newErrorText);
```
joshuaRogers24, line 65: Is this console log leftover from testing?

src/views/ApartmentApp/index.js
```diff
@@ -39,34 +54,24 @@ export default class ApartApp extends Component {
...
+    } else if (applicants.some(applicant => applicant.AD_Username === username)) {
+      // Display an error if the selected user is already in the list
+      let newErrorText = applicantProfile.fullName + ' is already in the list';
+      this.setState({ errorDialogOpen: true, errorDialogText: newErrorText });
+      console.log('Error: ' + newErrorText);
```
joshuaRogers24, line 70: and this one?

src/views/ApartmentApp/index.js
NerdyGriffin: Yes, I guess there is no need for it anymore, now that the stylized dialog has been added

joshuaRogers24 (approved):
Looks good! (Nick says “Gorgeous”). Just get rid of unnecessary console.logs.

PR #905, opened 2020-10-30
Close addMember dialog and refresh members on successful add

### LINKED ISSUE #870: Add Member confirmation incomplete
Add member confirmation does not clear, and the membership dialog box remains with that member's information on the screen. Should the X closing the confirmation (or a timeout) also close the dialog box or wipe the dialog box? Or is there intention to leaving the dialog box with the previously-entered data?

EjPlatzer:
closes #870 - The addMember dialog will now close itself and refresh members when a member is added.

Because the addMember dialog is in the Membership child component of ActivityProfile, whereas members are retrieved and managed by the ActivityProfile component itself, the dialog cannot cleanly refetch members. The two ways to manage this would be to drill a function down through props or to refresh the page. Props drilling is a more efficient solution, but it increases complexity and reduces maintainability really quickly. Since involvement profiles aren't displayed offline, refreshing should never cause an issue. It's not ideal because it means more fetching than is strictly necessary, but it is the cleaner solution.


cpabbot (approved):
LGTM

PR #909, opened 2020-11-04
Adding feature to save the state of the application

NerdyGriffin:
Part of first/second user story for MVP


src/components/ApartmentApplicantList/index.js
```diff
@@ -1,25 +1,47 @@
...
 import GroupAddIcon from '@material-ui/icons/GroupAdd';
+import ErrorIcon from '@material-ui/icons/Error';
+import CheckCircleIcon from '@material-ui/icons/CheckCircle';
+import { gordonColors } from '../../theme';
 import GordonPeopleSearch from '../Header/components/PeopleSearch';
+import GordonLoader from '../../components/Loader';
 import ApplicantListItem from './components/ApplicantListItem';
 import '../../views/ApartmentApp/apartmentApp.css';
 import '../../views/PeopleSearch/components/PeopleSearchResult/peopleSearchResult.css';
+const MIN_NUM_APPLICANTS = 3;
```
joshuaRogers24, line 13: I think there can actually be as few as 2 applicants in the case of the Village

src/services/housing.js
```diff
@@ -22,6 +22,21 @@ const getHousingInfo = async () => {
...
+ * @param {String} primaryUsername the student username of the person filling out the application
+ * @param {StudentProfileInfo} applicants Array of StudentProfileInfo objects
+ * @return {Promise.<String>} User's active jobs
+ */
+const saveApartmentApplication = async (primaryUsername, applicants) => {
+  let applicationDetails = {
+    PRIMARY_USERNAME: primaryUsername,
+    APPLICANT_ARRAY: applicants.map(profile => profile.AD_Username),
+  };
+  return await http.post(`housing/submit/`, applicationDetails);
```
joshuaRogers24, line 36: I think we should call the route 'save', not 'submit', but you probably were just going off of what Nick had...

src/views/ApartmentApp/index.js
```diff
@@ -18,54 +19,123 @@ export default class ApartApp extends Component {
...
   }
 
   /**
    * Callback for apartment people search submission
    * @param {String} searchSelection Username for student
    */
-  onSearchSubmit = (searchSelection) => {
+  onSearchSubmit = searchSelection => {
+    this.setState({ updating: true });
     if (searchSelection && searchSelection !== null) {
```
joshuaRogers24, line 87: Line 87, it  looks like you missed one of the redundant null checks.

src/services/housing.js
NerdyGriffin: I completely agree, especially since there will actually be a 'submit' for when they are done with the application

src/components/ApartmentApplicantList/index.js
NerdyGriffin: I have commented this out in the next version up, but I realized there should not be a minimum check for save, since we allow saving a partially completed state, but we will need to have this check with the submit button

src/views/ApartmentApp/index.js
NerdyGriffin: Oh, how did that one sneak passed 😆

src/views/ApartmentApp/index.js
```diff
@@ -74,60 +144,64 @@ export default class ApartApp extends Component {
...
-  checkPersonType(profile) {
-    let personType = String(profile.PersonType);
-    this.setState({ isStu: personType.includes('stu') });
-    this.setState({ isFac: personType.includes('fac') });
-    this.setState({ isAlu: personType.includes('alu') });
-  }
+  handleSaveButtonClick = () => {
+    // The method is separated from callback because the housing API service must be handled inside an async method
+    let debugMessage = 'DEBUG: Save button was clicked';
+    console.log(debugMessage);
```
joshuaRogers24, line 156: Do you need this concole log?

joshuaRogers24: *console

joshuaRogers24: oh, the debug message itself is just for debugging, isn't it?

src/views/ApartmentApp/index.js
```diff
@@ -74,60 +144,64 @@ export default class ApartApp extends Component {
...
+    this.saveButtonAlertTimeout = null;
+    let saveSuccessful = null;
     try {
-      // TODO - Once saving application has been implemented in the backend, this will be replaced with a call to the load the application info. The getHousingInfo was made obsolete after the Hello World
-      let housingInfo = await housing.getHousingInfo();
-      let onOffCampus = String(housingInfo[0].OnOffCampus);
-      let onCampusRoom = String(housingInfo[0].OnCampusRoom);
-      this.setState({ onOffCampus, onCampusRoom, loading: false });
+      const saved = await housing.saveApartmentApplication(primaryUsername, applicants);
+      console.log(saved.primaryUsername); // `saved` is a placeholder name
```
joshuaRogers24, line 172: Is this console log necessary?

src/views/ApartmentApp/index.js
```diff
@@ -74,60 +144,64 @@ export default class ApartApp extends Component {
...
+      saveSuccessful = saved;
     } catch (error) {
       // Do Nothing
     }
+    if (saveSuccessful) {
+      this.setState({ saving: 'success' });
```
joshuaRogers24, line 178: Do you want to use the state 'savingSuccess' and 'true' here? (and set it to false on 183?)

src/views/ApartmentApp/index.js
```diff
@@ -182,49 +256,54 @@ export default class ApartApp extends Component {
...
+                        applicants={this.state.applicants}
+                        userProfile={this.state.userProfile}
+                        saving={this.state.saving}
+                        savingSuccess={this.state.savingSuccess}
+                        onSearchSubmit={this.onSearchSubmit}
+                        onApplicantRemove={this.onApplicantRemove}
+                        onSaveButtonClick={this.handleSaveButtonClick}
+                        Authentication={this.props.Authentication}
+                      />
+                      {/* <GordonDialogBox
```
joshuaRogers24, line 271: make sure you get rid of this and its import once you are sure you do not need it.

src/components/ApartmentApplicantList/index.js
```diff
@@ -71,6 +99,33 @@ export default class ApplicantList extends Component {
...
+            <Grid item xs={12} sm={6}>
+              {this.props.applicants.length < MIN_NUM_APPLICANTS && this.state.saveButtonClicked ? (
+                <Typography variant="overline" color="error">
+                  Must have at least {MIN_NUM_APPLICANTS} applicants
+                </Typography>
+              ) : null}
+            </Grid>
+            <Grid item xs={12} sm={3}>
+              {this.props.saving ? (
+                this.props.saving === 'success' ? (
```
joshuaRogers24, line 111: Should you use 'savingSuccess' here and on 113 instead of saving? You would have to rearrange the if-else pattern here, if so (by making the GordonLoader case the 'else' of the outer-if and not the second inner-if).

joshuaRogers24: Hmm, on second thought, then the loader would not show in the case of "it is saving but it is undecided about failure or success"..

joshuaRogers24 (changes requested):
Great work! See my comments.

src/components/ApartmentApplicantList/index.js
NerdyGriffin: I thought I had completely removed "savingSuccess" and was using "saving" as both an indicator of saving in progress and indicator of saving complete, making the "savingSuccess" variable uneccessary. (Saving = false means not in progress, so show the save button. True means saving, so show progress circle. "Success" means saving it done and successful, so it will register like saving=true to hide the button, but the checking the value will tell it to give the checkmark instead of progress wheel. And "failed" is same idea but with error icon)
Since that savingSuccess variable is still there, then it either got messed up when I did a merge or maybe I forgot to commit/push a change from one of my computers

NerdyGriffin: I seems I made that change 2 branches up from this branch, so that is my fault for not thinking to include those things with the "save-application" branch.

src/views/ApartmentApp/index.js
NerdyGriffin: It is for debug to test whether the save was successful, and will eventually be removed or replaced with an `if` and other operations once the save feature has been implemented in the backend

src/views/ApartmentApp/index.js
NerdyGriffin: I added some comments to clarify which things are for debugging and/or placeholder.
They will be highlighted if you are using the Better Comments extension in VSCode

src/views/ApartmentApp/index.js
NerdyGriffin: `this.state.savingSuccess` was deprecated and I forgot to remove it, which I have now done

src/views/ApartmentApp/index.js
```diff
@@ -2,8 +2,9 @@
 import React, { Component } from 'react';
 import 'date-fns';
 import { Grid, Card, CardHeader, CardContent, Button, Typography } from '@material-ui/core/';
-import GordonDialogBox from '../../components/GordonDialogBox';
+// import GordonDialogBox from '../../components/GordonDialogBox';
```
joshuaRogers24, line 5: As long as this unused import was removed in another branch, I won't worry about it here

joshuaRogers24: Ah! I see you removed it in #910. Great!

glgnkim (approved):
Pair-reviewed with Josh via zoom. Great work!

joshuaRogers24 (approved):
Knowing some changes will be fixed in later branches: LGTM

PR #910, opened 2020-11-04
ApartApps Design Fix: Added option to change the primary applicant

NerdyGriffin:
This is a feature that is crucial to the usefulness of the MVP, because without this option there would be no way for a primary applicant to leave an application group. 
On the database, we will need to add a column to the table of applications for storing a reference to the primary applicant so that it may be stored, access, and modified when necessary.


src/views/ApartmentApp/index.js
```diff
@@ -133,11 +144,38 @@ export default class ApartApp extends Component {
...
+      }
+    }
+  };
+
+  handleChangePrimaryAccepted = () => {
+    if (this.state.newPrimaryApplicant) {
+      try {
+        this.saveApplication(this.state.newPrimaryApplicant.ID, this.state.applicants);
+      } catch (error) {
+        this.snackbarText = 'This feature is not yet implemented.';
```
joshuaRogers24, line 165: This is changed in a later branch, right? (One that was made when save is working?)

src/views/ApartmentApp/index.js
```diff
@@ -186,20 +224,26 @@ export default class ApartApp extends Component {
...
+  handleCloseDialog = (event, reason) => {
     if (reason === 'clickaway') {
       return;
     }
+    this.handleCloseOkay();
+  };
 
-    this.setState({ snackbarOpen: false });
+  handleCloseOkay = () => {
+    this.setState({ submitDialogOpen: false, editDialogOpen: false });
```
joshuaRogers24, line 246: I see you removed the code that set the dialog text to null. Is this because it will always be set to something else before the dialog box opens? We do not need to worry about opening the dialog box without the text being changed?

src/views/ApartmentApp/index.js
```diff
@@ -186,20 +224,26 @@ export default class ApartApp extends Component {
...
     }
   }
 
-  handleCloseOkay = () => {
-    this.setState({ submitDialogOpen: false, errorDialogOpen: false, errorDialogText: null });
+  handleCloseSnackbar = (event, reason) => {
+    if (reason === 'clickaway') {
+      return;
+    }
+    this.setState({ snackbarOpen: false });
```
joshuaRogers24, line 235: Do you not set the snackbarOpen state to false if they clickaway? Are you preventing them from clicking out of the dialog box or do you just not need to set the state to false in the case of clickaway- it closes on its own?

src/views/ApartmentApp/index.js
```diff
@@ -226,6 +270,15 @@ export default class ApartApp extends Component {
         }
       });
 
+      const primaryApplicantAlertTest = (
```
joshuaRogers24, line 273: Did you mean to call this AlertText as opposed to AlertTest?

src/views/ApartmentApp/index.js
```diff
@@ -260,24 +313,27 @@ export default class ApartApp extends Component {
...
-                        text={this.state.errorDialogText}
-                        buttonClicked={this.handleCloseOkay}
-                        buttonName={'Okay'}
-                      /> */}
+                      <AlertDialogBox
+                        open={this.state.editDialogOpen}
+                        onClose={this.handleCloseDialog}
+                        severity={'warning'}
+                        title={'Change primary applicant?'}
+                        text={primaryApplicantAlertTest}
```
joshuaRogers24, line 331: Might need to change to "text"

src/views/ApartmentApp/index.js
```diff
@@ -2,13 +2,15 @@
...
-// import GordonDialogBox from '../../components/GordonDialogBox';
 import GordonLoader from '../../components/Loader';
+import AlertDialogBox from '../../components/AlertDialogBox';
 import SimpleSnackbar from '../../components/Snackbar';
 import ApplicantList from '../../components/ApartmentApplicantList';
 import user from '../../services/user';
 import housing from '../../services/housing';
 import './apartmentApp.css';
+const MAX_NUM_APPLICANTS = 8;
+// const MIN_NUM_APPLICANTS = 2;
```
joshuaRogers24, line 13: Should you just delete the MIN?

NerdyGriffin: Probably. If so, I will do so in apartapp-landing-page #911  because most of this student-related code was moved from `.../ApartmentApp/index.js` to `.../ApartmentApp/components/StudentApplication/index.js`

src/views/ApartmentApp/index.js
NerdyGriffin: Nick caught this in his review of #911 and the change has been made 👍

src/views/ApartmentApp/index.js
NerdyGriffin: This makes it so that the snackbar will not close if the user clicks somewhere else on the page. In other word, it restrict it to only close if the user clicks the close button deliberately or the snackbar automatically times-out after 6 seconds

src/views/ApartmentApp/index.js
joshuaRogers24: Okay, sounds good

src/components/ApartmentApplicantList/index.js
```diff
@@ -94,19 +95,24 @@ export default class ApplicantList extends Component {
...
                   />
                 )}
               </List>
             </Grid>
-            <Grid item xs={12} sm={6}>
-              {this.props.applicants.length < MIN_NUM_APPLICANTS && this.state.saveButtonClicked ? (
+            <Grid item xs={9}>
+              {this.props.saving === 'failed' ? (
+                <Typography variant="overline" color="error">
+                  Something when wrong while trying to save the application
```
joshuaRogers24, line 107: Did you mean "Something went wrong"

src/views/ApartmentApp/index.js
NerdyGriffin: I am not sure which line you are referring to from the old code, 
In the latest version, the dialog box uses a constant text (it gets the text from the 'primaryApplicantAlertText' variable, for example), so the text if never changed.
I originally thought I would use one dialog for all alerts and edit the text, but the flaw with that design is the each alert situation needs to map the 'confirm' and/or 'close' button to a unique callback function for that situation. 
If dialog boxes are needed for other messages or confirmations in the future, it is best to create them as separate DialogBox components, which I why I created the new `AlertDialogBox` component

NerdyGriffin: I tried an approach of handleing the callback's behavior dynamically when it was setup to change the text, but that was bugging due to weird race conditions. I found that attempting to use a single dialog box dynamically led to a lot of unnecessary computational complexity compared to using multiple statically-defined dialog boxes. (which leads to better page response/performance in the web browser)

src/views/ApartmentApp/index.js
NerdyGriffin: Yes, I believe the branch for #911 uses the new `change-modifier` endpoint to change the primary applicant

src/components/ApartmentApplicantList/index.js
NerdyGriffin: Yes, thank you,

NerdyGriffin: Change has been pushed

src/views/ApartmentApp/index.js
NerdyGriffin: In #911 I added a comment earlier this week that explains the purpose of that line, because it caused the same confusing when Nick reviewed the code

src/components/ApartmentApplicantList/components/ApplicantListItem/index.js
```diff
@@ -132,19 +156,41 @@ export default class ApplicantListItem extends Component {
...
-              <ClearIcon color="disabled" />
-            </IconButton>
-          ) : (
-            <IconButton
-              edge="end"
-              aria-label="remove"
+          <Button
+            aria-controls="applicant-menu"
+            aria-haspopup="true"
+            disabled={this.props.isPrimaryApplicant}
```
glgnkim, line 162: Josh's comment: This seems to only allow the user to edit a list item for someone who is not the Primary Applicant. Am I reading that right, and, if so, why keep PrimaryApplicants from editing their own info? Maybe I'm confused because I don't know what sort of things are being edited...

src/components/ApartmentApplicantList/components/ApplicantListItem/index.js
```diff
@@ -8,9 +8,15 @@ import {
...
-  IconButton,
+  Button,
+  Menu,
+  MenuItem,
+  ListItemIcon,
 } from '@material-ui/core';
 import PersonIcon from '@material-ui/icons/Person';
+import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
+import StarBorderIcon from '@material-ui/icons/StarBorder';
+// import ContactMailIcon from '@material-ui/icons/ContactMail';
```
glgnkim, line 19: Are you planning to use this import later on? Or can we just delete it?

src/components/ApartmentApplicantList/components/ApplicantListItem/index.js
NerdyGriffin: That prop is in the list item, and each list item represents on of the applicants in the list. I believe that prop indicates whether the person represented by that list item is the primary applicant, not about the user.

I will check the surrounding code in a few minutes to confirm or deny

NerdyGriffin: This is indeed the case. 
The reason it is disabled for the primary applicant is that the only options in that edit menu are "Make primary applicant" and "Remove". 
There is no need to enable the "Make primary applicant" button if the person is already the primary applicant, and the user is not allowed to remove the primary applicant because they must assign a new primary applicant before they can be removed from the group. Since those are the only options in that edit menu, and both actions are designed to be impossible to use _on_ the primary applicant, it makes sense to disable the menu containing those options.

src/components/ApartmentApplicantList/components/ApplicantListItem/index.js
NerdyGriffin: I do not remember what I used that icon for at the time, but I know it is no longer needed in that file. So we can delete that import

src/components/ApartmentApplicantList/components/ApplicantListItem/index.js
joshuaRogers24: Okay, that makes sense.

src/views/ApartmentApp/index.js
joshuaRogers24: Ah yes, I see it. You moved it to components/StudentApplication/index.js. Looks good.

src/views/ApartmentApp/index.js
joshuaRogers24: Nice, that makes sense, and that comment was a good idea

src/views/ApartmentApp/index.js
joshuaRogers24: Nice! The Alert Dialog Box component looks like a great solution!

joshuaRogers24 (approved):
Looks good! Sorry for the wait. I should have reviewed this before #421.

PR #912, opened 2020-11-05
Wellness Refactoring and Updates

### LINKED ISSUE #886: Wellness check always present upon login, even if it has been completed
It goes away if you navigate to another tab and back, at least

Noted on 10/9/2020

EjPlatzer:
This is the first in a series of changes scheduled for the wellness/health check in system between now and the start of the Spring semester. These changes have been requested by the Senior COVID 19 team. Closes #886 

This pr consists of front-end only changes. Specifically, it includes:

1. Changing the color for symptoms (from red to yellow) in anticipation of a three-tier wellness system.

2. Refining the wellness question presentation so that answer disclaimers are color-coded and a symptoms-positive answer requires a confirmation dialog before submitting - with the intent to reduce the number of users who mistakenly submit a symptoms-positive response (usually because they think "yes" means "I'm good" or similar).

3. Adding a "Report Symptoms" button to the Wellness page, which is only displayed when the user has a status of NO_SYMPTOMS. This will allow people to update their status midday if they start experiencing symptoms after previously reporting no symptoms. Note that it is only possible to escalate to symptoms positive. Users are not able to de-escalate themselves to no symptoms.

4. Extensive refactoring of the Wellness, HealthStatus, Home, and WellnessQuestion components and the wellness service. This was done primarily to simplify and streamline these components to prevent adding bugs while updating them. It had the added benefit of fixing some pre-existing bugs (see issue 886 above, and many other bugs in the conditional rendering of the home page), as well as vastly increasing the clarity and maintainability of this part of the codebase.

One element of the refactoring (changing the case on the authentication prop passed to the home page) had repercussions for ~25 unrelated files, since this prop was being passed by the App root via the router to every page and the headers. This has the unfortunate side-effect of bloating this pr to make it seem much broader than it actually is. However, the change does make the code more consistent and understandable. The only changes made to these files was the new casing of props.authentication and miscellaneous linting. They have been tested to ensure that they still function properly.


bennettforkner (approved):
Wellness updates look adequate and the new coloring system make it easy to understand the gravity of your answer to the wellness check-in.

PR #914, opened 2020-11-11
Adjust event categorization

Matt-Ramos:


EjPlatzer (approved):
Looks good! We'll want to clean this all up once we know more about the events coming through now.

PR #916, opened 2020-11-12
Wellness check revamp

EjPlatzer:
This pull request includes three updates to the Wellness UI:

1. Update the wellness service and it's dependent components to make use of the refactored API. This includes using the new routes and making use of the simple objects being returned instead of the single-element array returned by the old routes.

2. Adding the WellnessQuestion to the Wellness page, in addition to the Home page. This will prevent the scenario where a user goes to view there status without having a valid status (new since 5AM).

2. Further refactoring, especially splitting out reused sections (WellnessQuestion and SymptomsDialog) into their own component.

### DIFF
--- a/src/components/SymptomsDialog/index.js
+++ b/src/components/SymptomsDialog/index.js
@@ -1,0 +1,47 @@
+import React from 'react';
+import {
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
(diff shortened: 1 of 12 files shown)


bennettforkner (approved):
Looks well done. Responsible changes in refactoring a lot of css/js so that wellness status is not a binary option, but can be used for a number of different values.

PR #922, opened 2020-11-18
Suggested Bug fix for compilation error

NerdyGriffin:
This is a suggested fix for two errors that appear during compilation: One is the missing onSignOut for the NavAvatar, the other is an upcoming deprecation warning regarding the syntax used for `createBrowserHistory` in `app.js`


EjPlatzer (approved):
Excellent bug fixes. Looks great!

PR #923, opened 2020-11-19
Fixed error caused by missing key in child of Downshift

NerdyGriffin:
Whenever the PeopleSearch bar in the header displays the list of suggested results, the browser console would always show an error about the missing `key` in the child elements of the `Downshift`. I found the corresponding line that was missing a unique key and added one 
This issue kept coming up and annoying me while I was trying to work on things for ApartApps, so I decided to try to fix it myself


EjPlatzer (dismissed):
Great fixes, thanks Christian. Looks good to me!

EjPlatzer (approved):
Good catch with the inputDisabled prop.

PR #924, opened 2020-11-20
Display FacStaff Mailstop on profile and search

EjPlatzer:
When FacStaff have a mailstop (Mail_Location), show it on their profile and in people search results, similar to students' mailbox number.


bennettforkner (approved):
Changes are clear and responsible. Allows for easy understanding of information as well as differentiation between mail stops and mail boxes.

PR #925, opened 2020-11-23
Fix Timesheets bug with Timezones

EjPlatzer:
A fix to a previous timesheets bug (see issue #894 and pr #896), introduced a new bug wherein the backend was receiving ISO dates and parsing them correctly, but passing them to the database in UTC time, where the database was treating them as local time. 

The solution is to pass dates from the front end in the correct timezone of the user (which the API can then parse into local time), but formatted according to our rules. This has been accomplished by moving all date string formatting into the jobs service (rather than being handled inconsistently by the component calling service methods). 

In the process, it became obvious that a jobs refactor was necessary. Therefore, the fixed UI makes use of new, simplified routes which are consistent with the conventions of ASP.NET in general and the rest of the 360 API in particular.

### DIFF
--- a/src/services/jobs.js
+++ b/src/services/jobs.js
@@ -5,6 +5,11 @@
 
 import http from './http';
 
+const dateFormat = new Intl.DateTimeFormat('en', {
+  timeStyle: 'medium',
+  dateStyle: 'short',
+});
+
 /**
  * Get Whether or not the user can use staff timesheets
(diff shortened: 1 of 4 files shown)


EjPlatzer:
This has been made obsolete by #1007 

The refactoring of the Timesheets API is more work than is worthwhile for a system that may be replaced in the near future.

PR #926, opened 2020-11-24
Apartapp develop

NerdyGriffin:
The completion of the first few user stories for the ApartApps senior project.
DO NOT delete this branch after the merge; it will act as our teams "develop" branch from the remainder of our senior project


src/components/Header/components/PeopleSearch/index.js
```diff
@@ -87,6 +87,13 @@ export default class GordonPeopleSearch extends Component {
     this.setState({ suggestions });
   }
 
+  handleClick = theChosenOne => {
+    if (theChosenOne && this.props.disableLink) {
```
Nick-Noormand, line 91: What is the **this.props.disableLink** for?

Nick-Noormand (commented):
Overall it looks good, but there are some parts that are still unclear. It is really great that you are compartmentalizing components to make things easier. Some confusing parts but looks all fine and good.

src/components/ApartmentApplicantList/components/ApplicantListItem/index.js
```diff
@@ -0,0 +1,152 @@
...
+    let personType, personClassJobTitle, nickname;
+
+    // set nicknames up
+    if (
+      profile.NickName !== null &&
+      profile.NickName !== '' &&
+      profile.FirstName !== profile.NickName
+    ) {
+      nickname = '(' + profile.NickName + ')';
+    }
```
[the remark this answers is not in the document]
NerdyGriffin: this line 80 takes the user's nickname (if they have one) and encapsulates it in parentheses, so that it may be displayed as "firstname (nickname) lastname ", like on the student profile page
For example, it would display you as "Nicholas (Nick) Noormand" instead of "Nicholas Nick Noormand "

src/components/ApartmentApplicantList/index.js
```diff
@@ -0,0 +1,134 @@
...
+
+// Create a list of applicants, displayed by name, username, and class standing.
+export default class ApplicantList extends Component {
+  constructor(props) {
+    super(props);
+    this.handleSelection = this.handleSelection.bind(this);
+    this.handleRemove = this.handleRemove.bind(this);
+    this.handleSaveButtonClick = this.handleSaveButtonClick.bind(this);
+    this.state = {
+      saveButtonClicked: false,
```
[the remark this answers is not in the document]
NerdyGriffin: That was in the very outdated version of the code, and it was part of one of my earlier attempts to change the look or behavior of the save button during or after saving. That whole feature is handled in a completely different way now, so that state variable is long-gone

src/components/Header/components/PeopleSearch/index.js
NerdyGriffin: By default, the menu items of the people search bar act as a link to that user's profile. The `this.props.disableLink` allows you to optionally disable the link feature, so that you may instead assign that "click" or "select" action to some custom callback function.
In this particular line that you commented on, it is being used to only allow the `handleClick` to perform its stuff if `disableLink` is true, so that the `handleClick` will behave exactly the same as before whenever `disableLink` is false (or not set at all)
The goal is to leave the default behavior of the people search exactly the same as it always has been, so it continues to work in the places that it is already being used.

Nick-Noormand (approved):
The changes that have been made look good. We should verify that it is alright to merge our code into _develop_ when the code isn't fully complete.

NerdyGriffin:
Feel free to review design changes of the ApartApp UI in addition to typical code review. I am happy to explore any suggestions and improvements to the code I wrote

EjPlatzer (changes requested):
I don't have a lot of time to do a very in-depth review right now, but I would definitely say this needs much more work before it's stable enough to merge into develop. I'm leaving comments for the few surface-level things I noticed with some quick testing. Hopefully I'll be able to perform a deeper code review soon.

The biggest issue I ran into was that if I open the ApartApp page, click on a link to someone's profile, and refresh the new tab that the profile was opened in, the entire ApartApp page freezes and the only way to recover it is to close the tab.

I also think you need to work on the responsiveness of the layout a little bit more. For example, if you open Chrome DevTools and enable the mobile view and choose any iPhone size, you'll notice that the Applicant List has fields clipping over each other. The Hall preferences can do the same thing on smaller sizes like iPhone 5.

Finally, something that's relevant for the consistency of the 360 codebase: when components are only relevant to one page, we usually keep them within the components subfolder of that view. The only things that go in the src/components folder are usually general purpose and reused components. For example, AlertdialogBox, if it has the potential to be reused in many contexts, definitely belongs in src/components. But if ApartmentHallSelection is only going to be used in the ApartmentApp view, it should be in src/views/ApartmentApp/components, not src/components.

EjPlatzer:
> I don't have a lot of time to do a very in-depth review right now, but I would definitely say this needs much more work before it's stable enough to merge into develop. I'm leaving comments for the few surface-level things I noticed with some quick testing. Hopefully I'll be able to perform a deeper code review soon.
> 
> The biggest issue I ran into was that if I open the ApartApp page, click on a link to someone's profile, and refresh the new tab that the profile was opened in, the entire ApartApp page freezes and the only way to recover it is to close the tab.
> 
> I also think you need to work on the responsiveness of the layout a little bit more. For example, if you open Chrome DevTools and enable the mobile view and choose any iPhone size, you'll notice that the Applicant List has fields clipping over each other. The Hall preferences can do the same thing on smaller sizes like iPhone 5.
> 
> Finally, something that's relevant for the consistency of the 360 codebase: when components are only relevant to one page, we usually keep them within the components subfolder of that view. The only things that go in the src/components folder are usually general purpose and reused components. For example, AlertdialogBox, if it has the potential to be reused in many contexts, definitely belongs in src/components. But if ApartmentHallSelection is only going to be used in the ApartmentApp view, it should be in src/views/ApartmentApp/components, not src/components.

This sounds very negative but I should point out that overall I think you're developing a good, usable UI that fits well into 360. Keep up the good work!

NerdyGriffin:
Thank you for the quick feedback. I am very intrigued by some of those issues because they sound familiar. I though I had fixed the clipping UI elements, so I will work on that today, and the page freezing is new to me so I will be investigating that as well. I have plenty of time to work on it today so I will probably start by restructuring to move those components into views/ApartmentApp/components

src/components/ApartmentApplicantList/index.js
```diff
@@ -0,0 +1,118 @@
...
+  handleRemove = (profile) => {
+    // Make sure the chosen profile was not null
+    if (profile) {
+      // Send the selected profile to the parent component
+      this.props.onApplicantRemove(profile);
+    }
+  };
+
+  handleSaveButtonClick = () => {
+    this.props.onSaveButtonClick();
```
EjPlatzer, line 55: You could probably just pass this function directly as the onClick prop, rather than creating a handle method that simply calls the function.

src/components/ApartmentApplicantList/index.js
```diff
@@ -0,0 +1,118 @@
+import React, { Component } from 'react';
+import 'date-fns';
```
EjPlatzer, line 2: importing but not using `date-fns` again.

src/components/ApartmentHallSelection/components/HallListItem/index.js
```diff
@@ -0,0 +1,135 @@
+import React, { Component } from 'react';
+import 'date-fns';
```
EjPlatzer, line 2: There it is again.

src/components/ApartmentHallSelection/index.js
```diff
@@ -0,0 +1,132 @@
+import React, { Component } from 'react';
+import 'date-fns';
```
EjPlatzer, line 2: Hello `date-fns` my old friend.

src/components/ApartmentHallSelection/index.js
```diff
@@ -0,0 +1,132 @@
...
+            </Grid>
+            <Grid item xs={3}>
+              <SaveButton saving={this.props.saving} onClick={this.handleSaveButtonClick} />
+            </Grid>
+          </Grid>
+        </CardContent>
+      </Card>
+    );
+  }
+}
```
EjPlatzer, line 132: Here's a more substantial example of refactoring into a functional component, where the useState and useEffect hooks are needed.
```suggestion
import React, { useState, useEffect } from 'react';
import { Grid, Card, CardHeader, CardContent, List, Typography, Button } from '@material-ui/core';
import AddIcon from '@material-ui/icons/Add';
import HallListItem from './components/HallListItem';
import SaveButton from '../ApartAppSaveButton';
import goStalk from '../../services/goStalk';
import housing from '../../services/housing';
import '../../views/ApartmentApp/apartmentApp.css';
import '../../views/PeopleSearch/components/PeopleSearchResult/peopleSearchResult.css';

// Create a list of selection boxes to choosing preferred halls
const HallSelection = ({
  authentication,
  saving,
  primaryUsername,
  preferredHalls,
  onHallInputChange,
  onHallRemove,
  onHallAdd,
  onSaveButtonClick,
}) => {
  const [halls, setHalls] = useState([]);

  useEffect(() => {
    const loadHalls = async () => {
      let unfilteredHalls;
      try {
        // Get the halls available for apartments, filtered by the gender of the primary applicant
        unfilteredHalls = await housing.getApartmentHalls(primaryUsername);
      } catch {
        //! DEBUG: Fills in halls dropdown when the housing api endpoint is not yet implemented
        // This list of halls is references from the 'Hall' dropdown on the PeopleSearch page
        unfilteredHalls = await goStalk.getHalls();
      }
      //Remove spaces from strings
      let availableHalls = unfilteredHalls.map((hall) => hall.trim());
      setHalls(availableHalls);
    };

    if (authentication) {
      loadHalls();
    }
  }, [authentication, primaryUsername]);

  const handleInputChange = (hallSelectionValue, hallRankValue, index) => {
    console.log('Called "handleInputChange" in ApartmentHallSelection component');
    console.log('hallName: ' + hallSelectionValue); //! DEBUG
    console.log('hallRank: ' + hallRankValue); //! DEBUG
    console.log('index: ' + index); //! DEBUG
    onHallInputChange(hallSelectionValue, hallRankValue, index);
  };

  const handleRemove = (index) => {
    // Make sure the chosen index was not null
    if (index !== null) {
      // Send the selected index to the parent component
      onHallRemove(index);
    }
  };

  return (
    <Card>
      <CardHeader title="Preferred Halls" className="card-header" />
      <CardContent>
        <Grid container justify="space-between" spacing={2}>
          <Grid item xs={11}></Grid>
          <Grid item xs={12}>
            <List className="hall-list" aria-label="apartment preferred halls">
              {preferredHalls ? (
                preferredHalls.map((preferredHall, index) => (
                  <HallListItem
                    key={preferredHall.hallName + index}
                    index={index}
                    availableHalls={halls}
                    primaryUsername={primaryUsername}
                    preferredHalls={preferredHalls}
                    onHallInputChange={handleInputChange}
                    onHallRemove={handleRemove}
                    authentication={authentication}
                  />
                ))
              ) : (
                <HallListItem
                  key={''}
                  index={0}
                  availableHalls={halls}
                  primaryUsername={primaryUsername}
                  preferredHalls={preferredHalls}
                  onHallInputChange={handleInputChange}
                  authentication={authentication}
                />
              )}
            </List>
          </Grid>
          <Grid item xs={12}>
            <Button variant="contained" color="default" startIcon={<AddIcon />} onClick={onHallAdd}>
              Add a Hall
            </Button>
          </Grid>
          <Grid item xs={9}>
            {saving === 'failed' ? (
              <Typography variant="overline" color="error">
                Something when wrong while trying to save the application
              </Typography>
            ) : preferredHalls.length >= halls.length ? (
              <Typography variant="overline" color="error">
                You have reached the maximum number of halls ({halls.length})
              </Typography>
            ) : null}
          </Grid>
          <Grid item xs={3}>
            <SaveButton saving={saving} onClick={onSaveButtonClick} />
          </Grid>
        </Grid>
      </CardContent>
    </Card>
  );
};

export default HallSelection;
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,681 @@
...
+  /**
+   * Loads the user's profile info only once (at start)
+   */
+  async loadProfile() {
+    this.setState({ loading: true });
+    try {
+      let applicants = this.state.applicants;
+      applicants.push(this.props.userProfile);
+      this.setState({ applicants });
+      this.setState({ loading: false });
```
EjPlatzer, line 79: These could be one single setState call.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,681 @@
...
+      preferredHalls: [{ hallName: '', hallRank: 1 }],
+    };
+    // Off-campus program info is stored as a Map, where the Key is a student's username and the corresponding Value is the department of that student's off-campus program
+    this.offCampusProgramInfo = new Map();
+    this.state.applicants.forEach((profile) =>
+      this.offCampusProgramInfo.set(profile.AD_Username, ''),
+    );
+    this.editDialogText = '';
+    this.snackbarText = '';
+    this.snackbarSeverity = '';
```
EjPlatzer, line 61: Why aren't these state variables?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,681 @@
...
+        </span>
+      );
+
+      /**
+       * Gets status of current network connection for online/offline rendering
+       * Defaults to online in case of PWA not being possible
+       */
+      const networkStatus = JSON.parse(localStorage.getItem('network-status')) || 'online';
+
+      if (networkStatus === 'online' && this.props.authentication) {
```
EjPlatzer, line 455: You already checked authentication above, no need to check it again.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,681 @@
...
+        return (
+          <div>
+            {this.state.loading ? (
```
EjPlatzer, line 458: Should probably make this a separate if else, rather than burying it in the return of this if block. I think the flow would be more readable that way.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,681 @@
...
+
+      /**
+       * Gets status of current network connection for online/offline rendering
+       * Defaults to online in case of PWA not being possible
+       */
+      const networkStatus = JSON.parse(localStorage.getItem('network-status')) || 'online';
+
+      if (networkStatus === 'online' && this.props.authentication) {
+        return (
+          <div>
```
EjPlatzer, line 457: Is this top-level empty div necessary?

EjPlatzer (changes requested):
I made some minor comments, including a repeated pattern of importing `date-fns` when it wasn't used anywhere.

I also think this code could benefit from a larger refactor. Class components have been replaced within React by the cleaner and more concise functional components using hooks. We should not be creating any new Class components in our codebase. Prefer functional components and hooks instead. Based on the code you've written, I'm confident that functional components could reduce the size of your code by almost 30% and make it much clearer and easier to understand and extend.

src/components/AlertDialogBox/index.js
```diff
@@ -0,0 +1,74 @@
...
+            variant="contained"
+            onClick={this.props.cancelButtonClicked}
+            color="primary"
+            autofocus
+          >
+            {this.props.cancelButtonName}
+          </Button>
+          <Button variant="contained" onClick={this.props.confirmButtonClicked} color="primary">
+            {this.props.confirmButtonName}
+          </Button>
```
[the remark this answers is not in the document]
NerdyGriffin: In that case, it may be better to instead make that an optional feature in GordonDialogBox, since this component was made to workaround the limitations of the existing GordonDialogBox component

NerdyGriffin:
We will be using it, 

> I made some minor comments, including a repeated pattern of importing `date-fns` when it wasn't used anywhere.
> 
> I also think this code could benefit from a larger refactor. Class components have been replaced within React by the cleaner and more concise functional components using hooks. We should not be creating any new Class components in our codebase. Prefer functional components and hooks instead. Based on the code you've written, I'm confident that functional components could reduce the size of your code by almost 30% and make it much clearer and easier to understand and extend.

Some parts of this code were originally functional components because we referred to Timesheets as our "starter code", so it should be a conceptually simple task to convert the components to that form. I have already begun making this change and it should hopefully be merged in the next few day

src/components/ApartAppSaveButton/index.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from 'react';
+import 'date-fns';
+import { Button } from '@material-ui/core';
+import ErrorIcon from '@material-ui/icons/Error';
+import CheckCircleIcon from '@material-ui/icons/CheckCircle';
+import { gordonColors } from '../../theme';
+import GordonLoader from '../Loader';
+import '../../views/ApartmentApp/apartmentApp.css';
+const styles = {
```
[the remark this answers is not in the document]
NerdyGriffin: I am not sure how to make that work with the material-ui icons, but I will look into it right now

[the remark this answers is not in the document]
NerdyGriffin: The change was easier than I thought. It has been committed to #954

src/services/housing.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {Number} applicationID the application ID number
+ * @param {String} newPrimaryUsername the student username of the person who will be allowed to edit this application
+ * @return {Promise.<Boolean>} Status of whether or not the operation was successful
+ */
+const changeApplicationModifier = async (applicationID, newPrimaryUsername) => {
+  let newModifierDetails = {
+    AprtAppID: applicationID,
+    Username: newPrimaryUsername,
+  };
+  return await http.post(`housing/change-modifier/`, newModifierDetails);
```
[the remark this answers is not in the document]
NerdyGriffin: This change has been made in #953 or #963 and will be merged after testing and review

src/components/ApartmentHallSelection/index.js
NerdyGriffin: Thank you, this is very helpful

src/components/ApartmentApplicantList/components/ApplicantListItem/index.js
```diff
@@ -0,0 +1,195 @@
...
+      this.handleMenuClose();
+    }
+  };
+
+  handleRemove = (profile) => {
+    // Make sure the chosen profile was not null
+    if (profile) {
+      // Send the selected profile to the parent component
+      this.props.onApplicantRemove(profile);
+      this.handleMenuClose();
```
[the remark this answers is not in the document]
NerdyGriffin: I think that the menu would still be closed by other actions such as clicking outside the menu, but that menu has since been removed so it is no longer needed

src/components/ApartmentApplicantList/index.js
NerdyGriffin: I either did not know that you could do that, or I couldn't get it to work with the class components. I looks like I tried that after the conversion to functional components but I could not get it to work, so I will look into that tonight most likely

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: removed and improved in #956

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: fixed in #956

src/components/ApartmentApplicantList/index.js
NerdyGriffin: Nvm it does work. This change will be merged in soon

src/components/AlertDialogBox/index.js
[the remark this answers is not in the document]
NerdyGriffin: See #1013 for implementation of this change

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,884 @@
...
+      if (newApplicationID === null || newApplicationID === -1) {
+        // Intentionally trigger the 'catch'
+        throw new Error("Invalid value of 'newApplicationID' = " + newApplicationID);
+      }
+      setApplicationID(newApplicationID);
+      let applicationDetails = await housing.getApartmentApplication(newApplicationID);
+      if (applicationDetails) {
+        if (applicationDetails.DateSubmitted) {
+          setDateSubmitted(applicationDetails.DateSubmitted);
+        }
```
EjPlatzer, line 174: I believe you can use JavaScript's nullish coalescing to drastically simplify this entire `if` block. For example:
```suggestion
  setDateSubmitted(applicationDetails?.DateSubmitted ?? null);
```

This should work for all of the `set...`'s below.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,884 @@
...
+          });
+        }
+      }
+    } catch {
+      // No existing application was found in the database,
+      // or an error occurred while attempting to load the application
+      setApplicationID(-1);
+      if (!editorUsername) {
+        if (
+          applicants.every((applicant) => applicant.Profile.AD_Username !== userProfile.AD_Username)
```
EjPlatzer, line 199: This is really minor but I think we could use `!applicants.some(...)` instead of `applicants.every(...)`. This should have a minor benefit to performance since it will finish earlier in the case that a matching applicant is found, and I think it's also (arguably) clearer becuase it literally means 'if there isn't some applicant that matches this username'

EjPlatzer (commented):
Just a couple of thoughts and suggestions. Looks like it's coming along nicely!

EjPlatzer (commented):
Just a couple of thoughts and suggestions. Looks like it's coming along nicely!

src/views/ApartmentApp/components/StudentApplication/index.js
EjPlatzer: I realize now that since you're doing a negative comparison, the shortcutting behavior is already achieved, so then it comes down to a syntax/semantics opinion, and that isn't as important. What you have already works well.

src/views/ApartmentApp/index.js
```diff
@@ -0,0 +1,204 @@
...
+  }, [authentication]);
+
+  const loadPage = async () => {
+    setLoading(true);
+    const [profileInfo, isHousingStaff] = await Promise.all([
+      user.getProfileInfo(),
+      housing.checkHousingStaff(),
+    ]);
+    setUserProfile(profileInfo);
+    profileInfo.PersonType.includes('stu') ? setIsUserStudent(true) : setIsUserStudent(false);
```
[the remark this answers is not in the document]
NerdyGriffin: Oh that's a really good point

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,884 @@
...
+
+/**
+ * @typedef { import('../../../../services/housing').ApartmentChoice } ApartmentChoice
+ */
+
+/**
+ * Renders a card displaying the apartment application instructions
+ * @returns {JSX.Element} JSX Element for the instructions card
+ */
+const InstructionsCard = () => (
```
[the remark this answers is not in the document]
NerdyGriffin: All of those sub-components have now been moved to their own folders, that branch should get merged in here shortly

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Typically this part only gets run the first time the page is loaded (or at least that is the intent), so the applicants array is expected to be empty or one element...

This makes me realized I could actually just compare `applicants[0].Profile.AD_Username` since the application editor is always the first element of the array by design

NerdyGriffin: It technically should not need to check this at all, since it is meant to happen only with the editor has not been set at the page load, 
but regardless of whether I put the `setEditorUsername` before or after that `setApplicants` line, there was always a race condition of it somehow triggering nearly simultaneously and causing it to enter the current user into the array twice on page load. 
So the applicants array check is to prevent that race condition

EjPlatzer: The race condition you're describing I believe is the normal behavior of useState. Updates to state are batched and run altogether according to some complex React-internal logic. This prevents excessive rerenders, because otherwise every set[WhateverStateVariableName]() call would cause an independent rerender.

Usually, if you have state that is derived from other state, you want a useEffect callback that will which the state that is derived from and update the derived state appropriately. 

I'm not sure I fully understand this usecase, so I can't say for sure whether that's what you need here, but hopefully that's helpful.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,884 @@
...
+    } catch {
+      // No existing application was found in the database,
+      // or an error occurred while attempting to load the application
+      setApplicationID(-1);
+      if (!editorUsername) {
+        if (
+          applicants.every((applicant) => applicant.Profile.AD_Username !== userProfile.AD_Username)
+        ) {
+          setApplicants((prevApplicants) =>
+            prevApplicants.concat({ Profile: userProfile, OffCampusProgram: '' }),
```
EjPlatzer, line 202: I'm not sure `Array.concat()` is the best method to use here. That method is intended to join two arrays together. If you're simply appending an element to the array, you could use `Array.push()`, or better yet just the spread operator, e.g.:
```suggestion
            [...prevApplicants, {Profile: userProfile, OffCampusProgram: '' }]
```

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: I believe I have fixed this in a more recent push, but the best solution is just check if the array is empty,

It is basically saying (through the combination of all the nested if/else) that if there is not an existing application, then this must be a new application, so make sure the current user gets added into the array. 

Two answer your other comment,
The other problem (which has been a big problem in other parts of the Apartapp stuff) is that the useEffect will only do a shallow comparison if the "derived state" is an array of custom objects, so it actually often does not handle the updates correctly (either triggering at the wrong time or failing to trigger when an object within the array should be effected, etc.)

I think in this use case, the simplest solutions is the say
```
if (!applicants.length) {
  // insert the current user 
}
```
Since the goal is simply to insert the current user as an applicant iff the array hasn't been filled in yet.

NerdyGriffin: Now that I think about it, I guess this could be split up in various way, such as doing it outside the `if (!editorUsername)` or even before the call of the load endpoint, since the entire applicant array is replaced by whatever is loaded from the backend (if an application is found), otherwise it should just initialize to only the current user.

Or, it may be possible to initialize that via the `= useState(......)` that initializes the state variables

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Interesting, I like it.
I have only recently started to learn more about how and when you can use the spread operator

src/services/housing.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+  } catch (e) {
+    // handle thrown 404 errors
+    if (e.status !== 404) throw e;
+    err = e;
+    applicationID = false;
+    console.log('A 404 code indicates that an application was not found for this applicant');
+  }
+  // handle "unhandled" requests, `this.status = 404`s, and 404 errors
+  var status = (this.status = (err && err.status) || this.status || 404);
```
EjPlatzer, line 115: I'm not sure this code is correct, and if it is, it could probably be cleaned up. Firstly, in an arrow function like this, the `this` keyword is lexically bound which in this case means it will refer to the `window` object. I don't think you mean to be reading (and setting !) the `status` property on the window object (here's some detail on [what `window.status` is](https://developer.mozilla.org/en-US/docs/Web/API/Window/status)). It seems like you're trying to set the variable defined by `var status` to the value of `err.status`, or `404` if there is no `err.status`. I think `const status = err?.status ?? 404` would suffice.

Also, please avoid using the `var` keyword to define variables. It has unexpected scoping behaviors which can result in hard-to-notice bugs. Use `const` for a variable like this that doesn't need to be mutated after it's defined, or `let` for a mutable value.

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/applicationRow.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../../../../../../../vars';
+
+.apartment-sub-table {
+  padding-top: 0 !important;
+  padding-bottom: 0 !important;
```
EjPlatzer, line 5: The `!important` css property is a major code smell. Anywhere it is used, you should be able to achieve the same effect by simply adding more specificity to the selector (e.g. by adding more classnames, prepending the html element the class is applied to, etc.).

Creating css rules with !important means that those rules can only be over-written by other rules marked !important with a higher specificity. In this case, it isn't likely to matter, but as a rule it's bad practice.

NerdyGriffin: I'll look into that. The Material-UI things tend to be very particular and stubborn unless you use inline styles like they do in the Material-UI documentation. 
They have a prop for `class` that can be used to easily override the default component styling, but that feature does not work with styles defined in css/scss, only for the inline objects with useStyle and such.

NerdyGriffin: At the time that I made that style change, I had no idea how else to get it to work, but I have since learned how to override the Material UI things when I needed the make the striped/alternative colors for the table rows

EjPlatzer: The main reason that Material-UI styles tend to override ours is that they're loaded last, and thus take precedence over our bespoke css files that are loaded before the MUI ones. 

The two easiest ways to override them are to either: 1) include the html element in the selector (e.g. `div.apartment-sub-table`) or 2) include one of the MUI classes on the element you're selecting (e.g. `.MuiGrid-root.apartment-sub-table`). Both of those selectors will have higher specificity than the MUI ones (which in my experience only include one class), and so will take precedence.

src/services/housing.js
NerdyGriffin: Yeah, you can tell I didn't check it very hard that night just by the fact that there is a `var` in it. As someone whose favorite language is C++, I despise the behavior of `var` 😆

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/applicationRow.scss
NerdyGriffin: Ah, it seem the `span.apartment-sub-table` style would be the easier solution, since it is much easier to check which HTML type the thing is than getting the MUI root, since there are many MUI sub-types in the styles and I always end up just trying them and seeing which one makes it work the way that I wanted

Additionally, using the `div.custom-class` , `span.custom-class` form makes the style file much easier to read

NerdyGriffin:
I believe the main features should now be working correctly, aside from the off-campus program card which is still in development.
Feel free to test it out and try to break the page, incase there are still bugs that I have either missed or forgotten about

src/services/housing.js
```diff
@@ -0,0 +1,202 @@
...
+  // Calculate the total and average points for each application
+  applicationDetailsArray.forEach((applicationDetails) => {
+    let totalPoints = 0;
+    applicationDetails.Applicants.forEach((applicant) => {
+      totalPoints += applicant.Points;
+    });
+    let avgPoints = totalPoints / applicationDetails.Applicants.length;
+    applicationDetails.TotalPoints = totalPoints;
+    applicationDetails.AvgPoints = avgPoints;
+  });
```
EjPlatzer, line 188: Could this not be done in the API? It seems odd to fetch data from the API and then immediately format that data again. Usually, an API endpoint should do all the general puprose data transformation ahead of time, so the  front end requests the data in it's final form and is only responsible for displaying it.

NerdyGriffin: Thank you for pointing this out, I forgot about this.

This was discussed back and forth quite a lot, and yes it could easily be done in the API. I think before we implemented it, there were reasons that we thought it would be better to put it here, but the process of implementing that endpoint revealed various things about how best to handle that stuff, and so I believe the actual implementation of the endpoint now would make that calculation super easy to put in the backend. 
I think the reasoning before had to do with the points being stored with the applicant data in the AA_Applicants table in the database, and thus the applicants, apartment choices, and such would have been retrieved somewhat separately, but we instead figured out that it is more efficient for both UI and API to handle it all in the current nested JSON structure, so the points are now generated/calculated at the time that the data is loaded or retrieved for the staff menu. 
TLDR our original design _we thought_ that calculating the point in the frontend made more sense, but in the new/current design it is obvious that the points total and average should be in the API, and I will make this change tonight

NerdyGriffin: This is now ready in #1066

src/views/ApartmentApp/components/StudentApplication/components/ApplicationDataTable/index.js
```diff
@@ -0,0 +1,73 @@
...
+  function createData(label, value) {
+    return { label, value };
+  }
+
+  let rows = [
+    createData(
+      'Last Submitted: ',
+      DateTime.fromJSDate(dateSubmitted).toISODate() ?? 'Not yet submitted',
+    ),
+    createData('Last Modified: ', DateTime.fromJSDate(dateModified).toISODate() ?? 'Not yet saved'),
```
EjPlatzer, line 34: `dateSubmitted` and `dateModified` are not JavaScript dates, they are strings, so this is always failing and returning `Not yet submitted`/`Not yet saved`. 

Instead, you can use Luxon to convert from a string directly:
```suggestion
      DateTime.fromISO(dateSubmitted).toISODate() ?? 'Not yet submitted',
    ),
    createData('Last Modified: ', DateTime.fromISO(dateModified).toISODate() ?? 'Not yet saved'),
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+  const [newEditorProfile, setNewEditorProfile] = useState(null); // Stores the StudentProfileInfo of the new editor before the user confirms the change
+  const [changeEditorDialogOpen, setChangeEditorDialogOpen] = useState(false);
+  const [submitDialogOpen, setSubmitDialogOpen] = useState(false); // Use this for submitting app (later feature)
+  const [snackbarOpen, setSnackbarOpen] = useState(false);
+  const [snackbarText, setSnackbarText] = useState('');
+  const [snackbarSeverity, setSnackbarSeverity] = useState('info');
+  const [saveButtonAlertTimeout, setSaveButtonAlertTimeout] = useState(null);
+
+  /**
+   * Loads the user's saved apartment application, if one exists
```
EjPlatzer, line 77: This only needs to run when the component first loads (and if the logged in user changes), correct?

The way it's currently running is causing all sorts of issues with your local state when you try to load a saved application.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+      // or an error occurred while attempting to load the application
+      setApplicationID(null);
+      setEditorUsername((prevEditorUsername) => prevEditorUsername ?? userProfile.AD_Username);
+      setApplicants((prevApplicants) =>
+        prevApplicants.length ? prevApplicants : [{ Profile: userProfile, OffCampusProgram: '' }],
+      );
+      setUnsavedChanges(true);
+    };
+
+    if (!applicationID || !applicationDetails) {
```
EjPlatzer, line 93: This check is not necessary, since it will always be true when we first load.
```suggestion
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+              });
+          }
+        })
+        .catch((error) => {
+          handleError(error);
+        });
+      setLoading(false);
+    }
+
+    return () => (isSubscribed = false);
```
EjPlatzer, line 145: I believe this is intended to prevent state changes to this component after it unmounts. However, I don't think you realized that it gets run on every re-render, because the previous call to `useEffect` is tossed out and the new one is enabled. That's preventing a lot of your state updates from working.

You should just get rid of this. Updating unmounted components is not actively harmful, and it would only happen if someone leaves the page while the initial load is still taking place, so it should be a rare circumstance.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+  }, [userProfile, applicationDetails, applicationID, editorUsername, applicants]);
```
EjPlatzer, line 146: You only want to re-run this useEffect when userProfile changes. The rest of these should not be dependencies of this effect.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+      setLoading(true);
+      // Check if the current user is on an application. Returns the application ID number if found
+      housing
+        .getCurrentApplicationID()
+        .then((newApplicationID) => {
+          console.log('Retrieved Application ID: ' + newApplicationID); //! DEBUG
+          if (newApplicationID === null || newApplicationID === -1) {
+            // Intentionally trigger the 'catch'
+            throw new Error("Invalid value of 'newApplicationID' = " + newApplicationID);
+          } else if (isSubscribed) {
```
EjPlatzer, line 103:
```suggestion
          } else {
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+            setApplicationID(newApplicationID);
+            setUnsavedChanges(false);
+            housing
+              .getApartmentApplication(newApplicationID)
+              .then((newApplicationDetails) => {
+                if (isSubscribed && newApplicationDetails) {
```
EjPlatzer, line 109:
```suggestion
                if (newApplicationDetails) {
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+                          const newApplicantProfile = await user.getProfileInfo(
+                            newApplicantInfo.Username,
+                          );
+                          return { Profile: newApplicantProfile, ...newApplicantInfo };
+                        }
+                      });
+                    } else {
+                      return prevApplicants;
+                    }
+                  });
```
EjPlatzer, line 129: This does not work, because your arrow function supplied to map is asynchronous, you're actually returning a Promise. Also, since this is only running when the page first loads (or userProfile changes), you know that `applicants` will always be an empty array, so you don't need to check for the case that it isn't.
```suggestion
                  Promise.all(
                    newApplicationDetails.Applicants.map(async (newApplicantInfo) => {
                      if (newApplicantInfo.Profile !== null) {
                        return newApplicantInfo;
                      } else {
                        const newApplicantProfile = await user.getProfileInfo(
                          newApplicantInfo.Username,
                        );
                        return { Profile: newApplicantProfile, ...newApplicantInfo };
                      }
                    }),
                  ).then((newApplicants) => setApplicants(newApplicants));
```

EjPlatzer: Also, is it possible for an `Applicant` returned from the database to have no `Profile` attached? If not, than all this is unnecessary, and if so than maybe you should fix that instead?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -0,0 +1,794 @@
...
+  });
+  /** @type {[Number, React.Dispatch<React.SetStateAction<Number>>]} */
+  const [applicationID, setApplicationID] = useState(null); // Default value of -1 indicate to backend that the application ID number is not yet known
+  const [dateSubmitted, setDateSubmitted] = useState(null); // The date the application was submitted, or null if not yet submitted
+  const [dateModified, setDateModified] = useState(null); // The date the application was last modified, or null if not yet saved/modified
+  const [editorUsername, setEditorUsername] = useState(null); // The username of the application editor
+  /** @type {[ApartmentApplicant[], React.Dispatch<React.SetStateAction<ApartmentApplicant[]>>]} Array of applicant info */
+  const [applicants, setApplicants] = useState([]);
+  /** @type {[ApartmentChoice[], React.Dispatch<React.SetStateAction<ApartmentChoice[]>>]} Array of apartment choice info */
+  const [preferredHalls, setPreferredHalls] = useState([]); // Properties 'HallName' and 'HallRank' must be capitalized to match the backend
```
EjPlatzer, line 63: I'm very curious why you have separate state variables here for things that are part of the `applicationDetails` state. Keeping two copies of the same data is the easiest way of getting broken state that behaves in unpredictable ways. Why not just use the `applicationDetails` state itself?

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -0,0 +1,154 @@
...
+      );
+    }
+  };
+
+  /**
+   * Saves the nickname of the given user if available
+   * @param {StudentProfileInfo} profile The StudentProfileInfo object for the student represented by this list item
+   */
+  const createNickname = (profile) => {
+    let Name = String(profile.fullName);
```
EjPlatzer, line 70: The `fullName` property is added to a user's profile by the `formatName` function in `user.js`, called when getting a user's Profile. That means that when you use a Profile straight from the backend (i.e. when it is loaded as part of an Applicant), this property does not exist, so you end up with `undefined (nickname)` as their name - e.g.:
image
The easiest fix is to probably find all references to `fullName` in your code and switch them over to use 
```
`${Profile.FirstName} ${Profile.LastName}`
```

EjPlatzer (commented):
Some initial comments. I'm still working through a lot of things, but these stuck out first because they prevented me from testing anything else. I'll continue to work through the changes and add more comments as I have them.

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -0,0 +1,154 @@
...
+    let FirstName = Name.split(' ')[0];
+    setHasNickname(FirstName !== profile.NickName && profile.NickName !== '');
```
EjPlatzer, line 72: `profile.FirstName` is what you want here I think. `profile.fullName` is a frontend-added property that is simply composed of Profile.FirstName + ' ' + Profile.LastName

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -0,0 +1,154 @@
...
+    // Make sure the chosen profile was not null
+    if (profile) {
+      // Send the selected profile to the parent component
+      onApplicantRemove(profile);
+    }
+  };
+
+  const displayName = hasNickName
+    ? profile.fullName + ' (' + profile.NickName + ')'
+    : profile.fullName;
```
EjPlatzer, line 101: Since `fullName` is added by `user.js` in the frontend, you can't depend on having it. If any of the listed applicants were loaded from the backend rather than being added by the frontend, they will not have this property. Your two options are two add the property yourself in the backend, or to sinply use First and Last name normally. I would recommend the latter.
```suggestion
    ? `${profile.FirstName} ${profile.LastName} (${profile.NickName})`
    : `${profile.FirstName} ${profile.LastName}`;
```

src/views/ApartmentApp/components/StudentApplication/components/ApplicationDataTable/index.js
NerdyGriffin: That makes sense, I was having a hard time figuring out exactly how the date information is sent from the backend to the frontend. **You may commit this suggestion for me**, since I do not actually have write access to this branch directly due to the protected-branch settings

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
EjPlatzer: I think once you have the simplified version, this helper function is no longer necessary, since it's only one line, and is only used in the useEffect above.
```suggestion
    setHasNickname(profile.FirstName !== profile.NickName && profile.NickName !== '');
```

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Yes, that is the intended behavior (run at first load). I have been struggling to actually get that you work, and I have been having problems with the `setLoading(false)` being run before the things actually finish loading, giving the user one of the error views from `src/views/ApartmentApp/index.js` rather than the GordonLoader while the page is loading.

I think this problem should first be addressed in `src/views/ApartmentApp/index.js` because some of the problems starting there are causing the weirdness in StudentApplication, but I agree both need to be fixed

Any help/suggestions would be much appreciated, and also I suggest looking at #1063 in case I tried anything in there that relates to this problem

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: That is how it _should_ be, but in reality, it often attempts to run the useEffect at least twice at first load, causing many of the other problems. This check was the only way I could figure out how to force it to not attempt the loading process multiple times, and as you commented above it is clearly not working correctly

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: I only added that because there were errors with the thing attempting to make state changes after unmount, and that would cause the page to freeze or crash instead of updating the state when doing something like adding an applicant.

For some reason, the StudentApplication component would unmount every time the user adds an applicant forcing the component to update the array of applicants. It is as if it is unmounting and then mounting a "new" instance or something, but still carrying over the state somehow? I don't fully understand the detailed of this part, all I know was that the page crashed before I added the "isSubscribed" thing and the errors reported that is had attempted to update the state of an unmounted component, and after I added that check it worked correctly during my testing.

I know this probably is a symptom of a larger problem with the setup of this useEffect, we can talk more later

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: I actually plan to completely replace the separate state variables with that `applicationDetails` state like you described because I realized it is beneficial to the reliability of other new features, and IMO it makes the code logic more intuitive to read. I just haven't gotten around to fully implementing that change yet

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
NerdyGriffin: I think I wrote that (or, actually, copy and paste from somewhere else within 360) that code last semester, and I didn't know that profile.FirstName existed at the time

So yes, I agree this whole nickname handling section needs some cleaning as well

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
NerdyGriffin: I agree, feel free to commit that suggestion since I do not have write access to this branch

EjPlatzer (commented):
Finished reviewing all StudentApplication code. I have a couple more comments, but overall it's really good!

I still need to review the StaffMenu, but once that's done we should be ready to move forward.

src/views/ApartmentApp/components/StudentApplication/index.js
EjPlatzer: When you work on that, it might be worth considering `useReducer`. That hook is much more simple to use with complex state objects than `useState`

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -0,0 +1,174 @@
...
+        Applicants.forEach((applicant) => {
+          if (applicant.ApplicationID) {
+            applicantsForCsv.push(applicant);
+          } else {
+            applicantsForCsv.push({
+              ApplicationID: applicationDetails.ApplicationID,
+              ...applicant,
+            });
+          }
+        });
```
EjPlatzer, line 83: You can follow the code for appending apartment choices to remove the `if` block. 
```suggestion
        Applicants.forEach((applicant) =>
          applicantsForCsv.push({
            ApplicationID: applicant.ApplicationID ?? applicationDetails.ApplicationID,
            ...applicant,
          }),
        );
```

Additionally, because the spread operator will overwrite earlier properties only if the property exists on the object being spread, you could remove the explicit set of `applicant.ApplicationID`. However, I think this somewhat obfuscates the expected path of the code, so I left it the more obvious way above.

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * Attempt to load all existing applications for the current semester
+   */
+  const loadAllCurrentApplications = useCallback(async () => {
+    setLoading(true);
+    let applicationDetailsArray = await housing.getAllApartmentApplications();
+    setApplications(applicationDetailsArray ?? []);
+    setLoading(false);
+  }, []);
```
EjPlatzer, line 49: This is only called in the below `useEffect`, and it has 0 external dependencies, so I don't see a reason for it to be wrapped with `useCallback`. I think it would be much simpler to declare it within the `useEffect`.

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -0,0 +1,174 @@
...
+    let date = new Date();
+    let dateStr = DateTime.fromJSDate(date).toISODate();
+    setDateStr(dateStr);
+  }, [userProfile, loadAllCurrentApplications]);
+
+  /**
+   * Generate arrays of objects to be converted to a CSV
+   * @param {ApplicationDetails[]} applicationDetailsArray an array of ApplicationDetails objects
+   */
+  const generateCSVData = useCallback((applicationDetailsArray) => {
```
EjPlatzer, line 64: Similar to the above function, this is only called in the below `useEffect` and it also has no external dependencies. I don't see a need for it to use `useCallback`. You could put in inside the `useEffect` or just leave it externally, but either way `useCallback` isn't achieving much for you here.

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -0,0 +1,174 @@
...
+                  </Typography>
+                </Grid>
+              </Grid>
+              <Grid container justify="center" alignItems="center" spacing={2}>
+                <Grid item>
+                  <Button
+                    variant="contained"
+                    color="primary"
+                    startIcon={<GetAppIcon />}
+                    disabled={!authentication}
```
EjPlatzer, line 125: While it certainly doesn't hurt to have these, it isn't possible for this component to display if the user is unauthenticated, since the containing view will only render it if `authentication` is true.

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/index.js
```diff
@@ -0,0 +1,47 @@
...
+import ApplicantSubTable from './components/ApplicantSubTable';
+import HallSubTable from './components/HallSubTable';
+
+const ApplicationRow = ({ applicationDetails, labelId }) => {
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <React.Fragment>
+      <TableRow key={applicationDetails.ApplicationID} onClick={() => setOpen(!open)}>
+        <TableCell align="center" component="th" id={labelId} scope="row">
```
EjPlatzer, line 14: Is there a reason the `ApplicationID` cell should be a Table Header component? It seems to me like this should be a regular table data cell.

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/index.js
```diff
@@ -0,0 +1,47 @@
...
+          {applicationDetails.ApplicationID}
+        </TableCell>
+        <TableCell align="center">{applicationDetails.EditorUsername}</TableCell>
+        <TableCell align="center">{applicationDetails.Applicants.length}</TableCell>
```
EjPlatzer, line 18: I think we're fairly confident this should always be an array, but it might be useful to add some null safety to this:
```suggestion
        <TableCell align="center">{applicationDetails.Applicants?.length ?? 0}</TableCell>
```

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/components/ApplicantSubTable/index.js
```diff
@@ -0,0 +1,69 @@
...
+ * @param {ApartmentApplicant[]} props.applicants Array of applicant info
+ * @returns {JSX.Element} JSX Element for the applicant sub-table
+ */
+const ApplicantSubTable = ({ applicants }) => {
+  const [thisYear, setThisYear] = useState();
+
+  useEffect(() => {
+    const getYear = () => setThisYear(new Date().getFullYear());
+
+    getYear();
```
EjPlatzer, line 29: It really doesn't matter, but this could be simplified:
```suggestion
    setThisYear(new Date().getFullYear());
```

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/components/ApplicantSubTable/index.js
```diff
@@ -0,0 +1,69 @@
...
+            <TableCell>Probation</TableCell>
+            <TableCell align="right">Points</TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody className={'striped-table'}>
+          {applicants.map((applicantInfo) => (
+            <TableRow key={applicantInfo.Username}>
+              <TableCell component="th" scope="row">
+                {applicantInfo.Username}
+              </TableCell>
```
EjPlatzer, line 55: Again, maybe I'm missing something, but shouldn't this be a regular table data cell, not a table head cell?

EjPlatzer (commented):
The Staff Menu looks really good! I suggested a couple of small improvements, but honestly it's pretty great overall, not much to improve.

The table-based layout is not very responsive, but since this is an administrative page only used by the Housing administrator, I expect it will only be used in a desktop environment. And the mobile size isn't unusable, you just have to scroll the table because it can only shrink so much.

src/views/ApartmentApp/components/StaffMenu/index.js
EjPlatzer: Also, I forgot to mention this, but the CSV output for Applicants has the Profile as simply `[object Object]`, because the value for that column is an object.
image

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/components/ApplicantSubTable/index.js
NerdyGriffin: Thank you, I don't know I did it that way...
I approve this for commit

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/index.js
NerdyGriffin: That is a very good check to add, because I did have weird problems with it being null or undefined, even momentarily, causing needless error messages
Feel free to commit this as well

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: I think it may be the case that something else was supposed to go in that `disabled` prop, though I don't remember

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/index.js
NerdyGriffin: tbh, I did that because that is what they did in the examples in the Material-UI.com documentation, and I didn't fully understand what it meant

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: For this one, I think the idea was to have a button that could be clicked to reload the applications, but I think that should be equally accomplished by just refreshing the page. In that case, there is no reason to have it be separate like this

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: Correct, that is why I filter out the applicant objects from the Applications CSV array, so we should do the same with the Profile objects here

NerdyGriffin: That overwriting behavior of the spread operator is actually preferrable in this case, since we only want to set the ID if already null or undefined

src/views/ApartmentApp/components/StaffMenu/index.js
EjPlatzer: That actually makes a lot of sense, and would be worthwhile to keep I think! Refreshing the page does a LOT of extra work becuase the ENTIRE component tree has to be recreated from scratch. Having a button to only refresh the data we're actually concerned about being stale would be much better user experience I think.

src/views/ApartmentApp/components/StaffMenu/index.js
EjPlatzer: RE: Profile filtering - that should be easy, I'll implement and commit it.

RE: setting the ApplicationID - I suppose applicant.ApplicationID should never be different from applicationDetails.ApplicationID, right? Is it expected to have a case where the applicant object has a null/undefined ApplicationID?

NerdyGriffin: I did not actually know that the order of where you put the spread operator changes it's behavior like that, so that is very good to know. I should go through all the places I have used and make sure to take that into account

NerdyGriffin: It also looks like I should be filtering out the StudentID as well. I will add that to the branch I am currently working in

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: Resolved in upcoming PR

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: I think this was the same reason for doing that on the student page as well, since the act of changing the application editor should force it to save the application and then load the new data, to make sure it is displaying the latest info while also preventing the a non-editor from attempting to make changes 

Additionally, if the frontend changes the editor state but that action fails in the API, then the page would make it appear as if it succeeded by disabling the interactive parts of the application, when in reality the user is still the editor in the backend/database. This could be handled with other forms of try/catch stuff around the API request to change the editor, but I think it is best an most reliable to make the page reload the application data so that it ensures the data displays to the user is an accurate reflection of what is actually in the database

EjPlatzer (approved):
This is in a good state for testing! All of the major and obvious bugs seem to be resolved and the code is high quality and sensible. There is still room for more improvement in the coming days, but this is an excellent first release.

NerdyGriffin:
Sounds good, go ahead an merge when you are ready, more improvements are coming soon

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Interesting, I will read up on that right now

Work on this is in progress at #1072

src/views/ApartmentApp/index.js
```diff
@@ -0,0 +1,197 @@
...
+
+      user
+        .getProfileInfo()
+        .then((profileInfo) => {
+          if (isSubscribed) {
+            setUserProfile(profileInfo);
+            setIsUserStudent(profileInfo.PersonType.includes('stu'));
+          }
+        })
+        .catch(() => {});
```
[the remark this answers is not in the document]
NerdyGriffin: Now addressed in #1071

PR #927, opened 2020-12-07
Use simple wellness routes

EjPlatzer:
Refactor wellness service to use simple routes ('/wellness' instead of '/wellness/status'). These routes access the refactored and simpler API.


bennettforkner (approved):
I see this as an absolute win.

Everything looks clean and responsible.

PR #929, opened 2021-01-07
Extract Network State to Context Provider

### LINKED ISSUE #342: "Include Alumni" in People Search shows for current students
screen shot 2018-09-23 at 4 46 50 pm

I am getting this checkbox, and it requests that the backend include alumni, however, the backend is smart enough not to send it back (because of my viewertype).

(There is no breach* in privacy, but the checkbox should not be showing up if it is not functional.)

It is also weird that in the SQL Server, I searched for myself and I am considered an alum already...
iamanalumni_dominiqueadevai

EjPlatzer:
Since 360 is a Progressive Web App, it can be "installed" to the home screen of mobile devices and will, in many ways, imitate a native app. One of these is that the PWA will continue to function - albeit in a limited way - when the device is offline (i.e. not connected to the internet). While some pages continue to function as normal offline - mostly by clever use of caching - others are not available in offline mode. 

The typical method for handling this is to make pages that aren't available offline inaccessible when the device is offline. However, this requires many different components to track whether or not there is an active internet connection. Historically, this has been done by having the Service Worker watch for changes and send an event when the network state changes. Then, every component that is network-aware mounts it's own event listener and independently implements logic to track these network changes over that component's lifecycle.

This solution was working fine, but it was adding a significant amount of clutter to many components. This pull request remedies that issue by extracting the logic into a custom hook - useNetworkIsOnline - that returns a boolean which will be updated automatically. This hook is implemented using standard React hooks and the React Context API.  The NetworkContextProvider is included in the component tree at the top-level (within App.js), so that every component beneath it has the NetworkContext available to use the hook. This provider is the single component responsible for updating the network status, which will change the NetworkContext. This change to the context will cascade down the tree, only re-rendering components that implement the custom hook which watches the context.

The result is that any component that needs to be network-aware can simply use the custom hook. This reduces the complexity of these components exponentially, making the code much easier to maintain and expand. For example, compare [the Feedback component](https://github.com/gordon-cs/gordon-360-ui/pull/929/files#diff-898c235c2f390564da1330ea5ebca5484fc632c9b821319f080fe236b2ca6edc) before and after it's refactoring. It went from 112 lines of code down to 31, and now it is extremely readable and the content is almost all focused on the actual feedback form, rather than being 50% unrelated logic like network state and the offline card.

The main limitation is that - like all hooks - the custom hook is only usable by Functional react components. Thus, this pull request consists mainly of adding the useNetworkIsOnline custom hook (and it's prerequisite context provider component), as well as updating and - where necessary - refactoring all network-aware components to use this hook.

In the process of refactoring every network-aware component, I noticed that 7 different pages which are unavailable offline implement essentially the same card to notify the user that that page is unavailable. I factored this out into a stand-alone component - OfflinePanel - which takes the name of the component that is unavailable as a prop.

Refactoring also fixes #342


bennettforkner (dismissed):
Code looks responsible and thoughtfully written. Responsible modification of all network data values was used in refactoring isOnline to a boolean value. This merge will increase the efficiency as well as the responsibility of the code.

src/views/News/index.js
```diff
@@ -40,537 +44,319 @@ const styles = {
...
-              color="primary"
-              onClick={this.handlePostClick.bind(this)}
-              style={styles.fab}
-            >
-              <PostAddIcon />
-              Post Listing
-            </Fab>
-
-            <Grid container justify="center">
-              {/* Search */}
```
cpabbot, line 360: You removed search, but it should be working? If the refactoring broke it then it should be fixed before merging this.

cpabbot (changes requested):
What I see looks good for the most part; great work refactoring and making this code cleaner! I only skimmed some parts and do not have a great understanding of these types of changes, so I would ideally have another pair of eyes look this over.

I am requesting changes because a) the 'People' and 'Wellness' tabs in the header don't show active when clicked anymore (I didn't find the instance in the code but showed when running localhost). And b) I am uncertain as to why news search has been disabled. I also just realized that the entire news page on train is broken, so at least this fixes that!

src/views/News/index.js
EjPlatzer: If you look at the current Develop code (which can be seen on the 360 and 360train sites), search is not working in that code. I simply removed it until it can be fixed. See my other comment for where it is broken. It probably won't take much to fix, but that wasn't in the purview of this PR.

EjPlatzer: You're correct. I thought search wasn't working in develop, but it is. I think this is still in need of some fixing and currently no filtering is actually enabled but that's a separate feature. I will restore search to parity with develop and leave the full implementation to a separate pr.

EjPlatzer:
Restored News search and fixed issues from merging with new PeopleSearch UI.

PR #936, opened 2021-01-09
Load application

NerdyGriffin:
On the student application page, if the user is on an existing/saved application in the database, then the page will attempt to load the information of that application and display it (application editable only by the primary applicant)

- [ ] Await implementation of the LoadApplication endpoint in the backend
- [ ] Make sure the application information is correctly displayed on the ApplicationDataTable after an existing application is loaded
- [ ] (Note: This section already existing in the UI and has a header, and placeholder text where the agreements will go)


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -660,12 +684,18 @@ const StudentApplication = ({ userProfile, authentication }) => {
                   </Grid>
                 </Grid>
                 <Grid item xs={12} md={4}>
-                  <Card>
-                    <CardHeader title="Agreements" className="card-header" />
-                    <CardContent>
-                      <Typography variant="body1">Placeholder text</Typography>
-                    </CardContent>
-                  </Card>
+                  <Collapse
```
Nick-Noormand, line 687: Why did you change the agreements to be under this Collapse header?

Nick-Noormand (approved):
Overall the pull request looks good. The only comment that I had was why there was a change to the agreements in this pull request and not in a separate one.

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Oh, I think this is because it does not need to show the agreements if the current user is not the editor (And checking for whether the user is the editor is a part of the "load application" feature

NerdyGriffin: This comment/review should go in #963

Issue #882, opened 2020-09-22
CL&W Filter not working


Matt-Ramos:
results as 'no events to show' even though events that offer CLAW credit are in calendar

PR #938, opened 2021-01-12
Fix events

EjPlatzer:
This pr accompanies gordon-cs/gordon-360-api#442. The events API has been simplified and refactored for both performance and readability. This pr updates the UI to use the simplified API properly.  The following event filters are now defunct and have been removed from the UI:
- Fair
- Academics
- Calendar

In addition, the event.js service has been refactored for clarity and performance. All event logic is now handled within the event service  (rather than being split across multiple services), and the service functions use functional Array methods to filter, format, and sort events more concisely and efficiently.


src/services/event.js
```diff
@@ -5,98 +5,124 @@
...
+ */
+const getAllEvents = async () => {
+  const allEvents = await http.get('events/25Live/All');
+  return allEvents.map((e) => formatevent(e)).sort(sortEventsByTime);
+};
 
-const getAllCLAWEvents = () => http.get('/events/25Live/CLAW');
+/**
+ * Gets upcoming CL&W events and formats them for display
+ * TODO: Unused. Consider removing
```
cpabbot, line 55: Is this not used for the has CLW credit filter?

src/services/event.js
```diff
@@ -5,98 +5,124 @@
...
+  formattedEvent.title = event.Event_Title || event.Event_Name;
 
-  let title;
-  if (event.Event_Title === '') {
-    title = event.Event_Name;
-  } else {
-    title = event.Event_Title;
-  }
-  event.title = title;
+  formattedEvent.location = event.Occurrences[0].Location || 'No Location Listed';
```
cpabbot, line 102: This is outside of the `if (event.Occurrences[0]) {` block on line 93; will this throw a null pointer if there are no occurrences of the event? Or will this just use the second part of the OR? (a null location will result in 'no location listed' but will a null occurrence?) Love the simplicity here!

src/services/event.js
```diff
@@ -5,98 +5,124 @@
...
   if (
     filters.chapelOffice ||
     filters.art ||
     filters.cec ||
-    filters.calendar ||
     filters.admissions ||
     filters.sports ||
     filters.studentLife ||
-    filters.fair ||
-    filters.academics ||
```
cpabbot, line 99: Are we okay with losing these filters I assume?

src/services/user.js
```diff
@@ -392,68 +368,11 @@ const getLocalInfo = () => {
...
- * Get the number of cl&w credits aquired, and number of credits required.
- * @return {Promise.<CLWCredits>} An Object of their current and requiered number of CL&W events,
+ * Get the number of cl&w credits acquired, and number of credits required.
+ * @return {CLWCredits} An Object of their current and required number of CL&W events,
  */
-const getChapelCredits = async () => {
-  const termCode = session.getTermCode();
-  const attendedEvents = await getAttendedChapelEvents(termCode);
+const getChapelCredits = () => {
+  const attendedEvents = gordonEvent.getAttendedChapelEvents();
```
cpabbot, line 375: I assume this is okay not being async? Not really sure how this would effect other things but I assume you've thought it through.

src/views/Events/index.js
```diff
@@ -187,22 +169,19 @@ export default class Events extends Component {
...
         chapelOffice: false,
         studentLife: false,
       },
       async () => {
-        const events = await gordonEvent.getFilteredEvents(this.state);
-        this.setState({ filteredEvents: events, loading: false });
-        this.createURLParameters();
+        // Set includePast to true above so that it is "toggled" to false by the below method
+        // This will ensure we filter only future events
+        await this.togglePastEvents();
```
cpabbot, line 184: A comment or some slight code change might be helpful semantically understanding what this is doing (optional). On reading this, I would assume togglePastEvents would just toggle the includePast variable, but it also loads the appropriate events into the state which is why you were able to delete that old line 204. Works fine but might be slightly unintuitive - maybe at least just add a comment above 184 stating that this will also load the appropriate events into state?

src/services/event.js
```diff
@@ -5,98 +5,124 @@
...
+  return allGuest.map((e) => formatevent(e)).sort(sortEventsByTime);
+};
+
+/**
+ * Get chapel events attended by the user during the current term
+ * @return {AttendedEvent[]} all CL&W events attended by the user, formatted and sorted
+ */
+const getAttendedChapelEvents = async () => {
+  const termCode = session.getTermCode();
+  let attendedEvents = await http.get(`events/chapel/${termCode}`);
```
cpabbot, line 82: this HTTP route is very unintuitive to me since I think it implies that it will get all the chapel events for the given term, rather than those attended by the user. But it is what it is, no sense changing it right now.

src/services/event.js
```diff
@@ -5,98 +5,124 @@
...
+const getAllGuestEvents = async () => {
+  const allGuest = await http.get('events/25Live/Public');
+  return allGuest.map((e) => formatevent(e)).sort(sortEventsByTime);
+};
+
+/**
+ * Get chapel events attended by the user during the current term
+ * @return {AttendedEvent[]} all CL&W events attended by the user, formatted and sorted
+ */
+const getAttendedChapelEvents = async () => {
```
cpabbot, line 80: Should this be in event.js rather than user.js like you did? Curious what your thoughts were on this; I could understand the logic either way since it is user information but regarding events...

cpabbot (commented):
Looks good, appreciate the clean changes and removal of unnecessary fragments! Don't have a perfect understanding of this code at all, but left some comments and questions just to double check some things. Should be ready for merge though!

src/services/event.js
EjPlatzer: You would think so, but no it's not. Currently, the CLW filter is applied to the events that have already been fetched (e.g. by getAllEvents), using the HasCLAWCredit attribute (which is functionally the same as filtering for which of the events have category 85, but simpler because that work has already been done by the backend).

The question of whether filtering should be done in the backend or the front end is tricky. Usually, you want those kinds of resource intensive operations done by the backend so that the user's experience is fluid and quick. However, achieving the kind of granular filtering we have designed via a REST API is quite tricky and would result in a many many requests back and forth. For now, I've chosen to continue the already established pattern of filtering on the front end, but we might revisit this eventually.

src/services/event.js
EjPlatzer: It's certainly not ideal, but unfortunately we have no good means of replacing them, since they relied on event types that no longer exist. It might be possible to build something that looked for certain organizations, but it would introduce a lot of added complexity without ever being as reliable as the event types were. Additionally, I'm not positive that these filters were widely used when they did work.

src/services/event.js
EjPlatzer: Good catch! This would've thrown an error for any events with no Occurrences property or with an empty Occurrences array. I'm not sure there are any events with 0 Occurrences, but much better to build it robustly just in case. I've fixed this issue.

src/views/Events/index.js
EjPlatzer: It's true that this method is not at all clear in it's scope. Ideally, this would be fixed with a refactoring of the component, but that is beyond the scope of this pr. I'll revisit it if I have time.

src/services/event.js
EjPlatzer: If you're asking why I moved it from user.js to event.js, the simple explanation is that this keeps all event logic in the one service. Previously, formatevent had to be exposed outside the service so that user.js could reuse it, and the logic for sorting events was repeated in both services. This way, the event logic is better encapsulated, and any components or services  that  need to use events in some way, as user.js does to calculate CL&W credits, can use the event service and not have to repeat the logic.

src/services/user.js
EjPlatzer: No, it absolutely did need to be async. Thanks for noticing that! I've fixed it.

src/services/event.js
cpabbot: Got it, yeah that makes sense. Good to keep in mind.

src/services/event.js
cpabbot: That makes sense. Good change!

src/services/event.js
```diff
@@ -90,27 +90,27 @@ const getAttendedChapelEvents = async () => {
...
+    const beginTime = DateTime.fromISO(event.Occurrences[0].StartDate).toFormat('t');
+    const endTime = DateTime.fromISO(event.Occurrences[0].EndDate).toFormat('t');
     formattedEvent.timeRange = `${beginTime} - ${endTime}`;
     formattedEvent.date = DateTime.fromISO(event.Occurrences[0].StartDate).toFormat('LLL d, yyyy');
   }
 
   formattedEvent.title = event.Event_Title || event.Event_Name;
 
-  formattedEvent.location = event.Occurrences[0].Location || 'No Location Listed';
+  formattedEvent.location = event.Occurrences?.[0]?.Location || 'No Location Listed';
```
cpabbot, line 102: This is a neat operator, I had never seen it before but very useful!!

cpabbot (approved):
Changes look good!

EjPlatzer:
Fixed #882

PR #939, opened 2021-01-12
Added Advisors to Personal Profile Page

### LINKED ISSUE #876: Fix advisors
Just lmk if need help figuring out the status of this. Just wanted to put this here before we forget about it.

cpabbot:
## Advisors on Student Personal Profile
Thought we had finished this during the summer, but this PR allows a student to see his/her own advisors. They will not show on public profile (wouldn't hurt, but I thought it would be better and cleaner to keep private). Will need to test this using faculty/staff accounts and those without advisors, and will need to make sure that this doesn't affect anything using profile information, since I added advisors to the profile array for better usability.

Resolves #876


src/components/ProfileList/index.js
```diff
@@ -235,7 +234,13 @@ class ProfileList extends Component {
     let majors = createMajorsListItem(this.props.profile, rowWidths, { gridStyle });
 
     // Creates the Advisors List Item
-    let advisors = createAdvisorsListItem(this.props.profile, rowWidths, { gridStyle });
+    let advisors;
+    // only show on personal profile
+    if(this.props.myProf) {
+      advisors = createAdvisorsListItem(this.props.profile, rowWidths, { gridStyle });
+    } else {
+      advisors = <></>;
```
EjPlatzer, line 242: Consider returning null rather than an empty fragment. It is slightly more performant since React doesn't have to create the fragment object.

EjPlatzer (changes requested):
This looks like a great addition. Glad you noticed that it hadn't been finalized yet. The filter for persontype including "stu" seems to work decently. Interestingly, my profile in train currently has PersonType of "stualu", which I think is slightly inconsistent because it's train data. Graduates from earlier years no longer have the Stu type, including people who were previously students and are now employed. I have requested a couple of changes I think are necessary before it can go into production, but this is a great feature to have!

Other profile elements that are only visible on the user's personal profile are greyed out. This gives nice visual feedback that that info is private and can't be seen by regular users. We should do this for advisors as well since they're kept private.

It would be worthwhile to allow FacStaff to see the list of advisors on a student's profile, similar to how they have access to address info that's normally hidden on public profiles. This is certainly less sensitive info than addresses, but might be useful especially to faculty in some cases.

Currently, the text a user sees if they have no advisors is "(not assigned)". Could we change this to be more user friendly? Perhaps "None assigned" or something similar. The lowercase and parentheses make it look out of place, not like legitimate data.

src/components/AdvisorList/index.js
```diff
@@ -33,19 +33,15 @@ export default class Advisors extends Component {
...
           <div>
             <Typography>Advisor:</Typography>
           </div>
         );
-        content = (
-          <div>
-            <Typography>(not assigned)</Typography>
-          </div>
-        );
+        content = <Typography style={this.props.styles}>None assigned</Typography>;
```
EjPlatzer, line 44: Does removing the div here cause any CSS issues? I know sometimes empty divs can have unexpected results, so I wasn't sure if this one was important for spacing of something like that.

EjPlatzer (approved):
Looks good to me! Glad to see the private text style applied to advisors. I left one question about display logic since I noticed a slight inconsistency in what was receiving extra div wrappers. But otherwise, should be all set to merge!

src/components/AdvisorList/index.js
cpabbot: I checked this and could not find any issues that way; I preferred this consistency because you can see line 36 where content doesn't use a div like this, and they should appear formatted the same regardless of what that format is.

PR #941, opened 2021-01-13
Add Return to Campus form link

EjPlatzer:
Students are required to fill out the Return to Campus form before they can complete a COVID check in on 360. All on-campus students will have their status set to red before returning to campus, and this will stay until they fill out the form. 

This code can be removed once the return to campus process has been completed.

### DIFF
--- a/src/views/WellnessCheck/components/HealthStatus/index.js
+++ b/src/views/WellnessCheck/components/HealthStatus/index.js
@@ -70,6 +70,8 @@ const HealthStatus = ({ currentStatus, setCurrentStatus, username, image }) => {
                 alt={username}
               />
             </Card>
+            {/* TODO: Remove following code block after Spring 2021 move in is complete */}
+            {/* START */}
             {currentStatus === StatusColors.RED && (
               <Typography variant="h5">
                 Students must fill out{' '}
@@ -83,6 +85,7 @@ const HealthStatus = ({ currentStatus, setCurrentStatus, username, image }) => {
                 before checking in.
               </Typography>
             )}
+            {/* END */}
             <Grid spacing={2} className="wellness-status">
               <Card className={currentStatus}>
                 <CardContent className="status-box">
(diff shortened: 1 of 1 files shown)


src/views/WellnessCheck/components/HealthStatus/index.js
```diff
@@ -70,6 +70,19 @@ const HealthStatus = ({ currentStatus, setCurrentStatus, username, image }) => {
                 alt={username}
               />
             </Card>
+            {currentStatus === StatusColors.RED && (
```
cpabbot, line 73: I'm not exactly sure how this works, but wouldn't this then show the form anytime the status is red? So if someone tests positive won't they be shown the check in form, which wouldn't make sense? Also should probably include comment if this code should be deleted after the move in process is completed (or not if it has no effect).

cpabbot (commented):
Left one comment to check

src/views/WellnessCheck/components/HealthStatus/index.js
EjPlatzer: Yes, this will show the message to anyone who has a red status. If we had more time, we could get a more elegant solution where only people who haven't yet filled out the form would be shown the message. However, that requires building an extra integration for 360 to be aware of who that is. Instead, the plan is to set all student's status to red with no expiration, and then set up a one-way integration that will clear their status when they fill out the form. 

That way, the only people who will see the message without needing to fill out the form will be those with a red status for another reason. Presumably, that will be a small subset of people, since the red status is reserved for confirmed cases of exposure or positive tests, etc.

It's not the most elegant solution, but it works well and since the message is informative but not disruptive I think it's an acceptable compromise between functionality and speed.

cpabbot: That makes sense yeah, thanks for the explanation! Comment looks good and helpful as well!

cpabbot (approved):
Looks good!

PR #942, opened 2021-01-14
RELEASE: Wellness updates, events updates, misc fixes and linting

EjPlatzer:
This release contains significant updates to the Wellness system, modifications to the way events are fetched and displayed, a couple of minor bug fixes, and a lot of linting. 

The wellness system has been updated to use three possible statuses [green, yellow, or red]. It has also been refactored to display the question and the status appropriately, and to show the question on the wellness page (in addition to the home page). Furthermore, the UI has been updated to prevent incorrect results (i.e. people answering positively when they meant to answer negatively). 

The events feed has been edited to account for the fewer event types in use, mainly meaning less filters to use. 

The following misc changes:
- fixed warnings
- replaced blackboard link with canvas

And finally the linting from Prettier was applied globally, affecting many files with cosmetic changes.


src/views/Home/components/NewsCard/index.js
```diff
@@ -29,17 +29,17 @@ export default class DailyNews extends Component {
...
   async loadNews() {
     // let newsCategories = await NewsService.getCategories();
     let todaysNews = await NewsService.getTodaysNews();
 
-    this.setState({ 
-      // newsCategories: newsCategories, 
-      news: todaysNews
+    this.setState({
+      // newsCategories: newsCategories,
+      news: todaysNews,
```
cpabbot, line 42: this answers my previous question haha, for whatever reason the end commas bug me but ah well

src/views/Page404/components/ScottieDog/scottieDog.scss
```diff
@@ -3,112 +3,113 @@ $dog-body-width: 10em;
...
-    border-right: solid 1em transparent;
-    border-bottom: solid 2em $dog-color;
+  right: 0;
+  top: 0;
+  transform: translateY(-1.5em) rotate(15deg);
+  width: 0;
+  height: 0;
+  border-left: solid 1em transparent;
+  border-right: solid 1em transparent;
+  border-bottom: solid 2em $dog-color;
```
cpabbot, line 63: So little indentation sheesh, but okay we will live with this auto changes

cpabbot (approved):
LGTM, pretty sure I missed a lot though sorting through all the auto-generated changes. But the changes files look correct to me.

src/views/Home/components/NewsCard/index.js
EjPlatzer: Yes, it's common practice in JavaScript to end arrays, objects, etc. with a trailing comma (you can actually see in the first file changed ".prettierrc.json" that our linter is set to require that). The reasoning is that, since you don't have to add a comma to existing lines when you make an update, git can detect the difference more accurately. Notice, for example, that git has two lines marked as different in .prettierrc.json because that file doesn't have trailing commas, so the last line from the old file has a comma added, and then the new line is also added.

src/views/Page404/components/ScottieDog/scottieDog.scss
EjPlatzer: Yes, the original authors of the code decided on two spaces for indentation and now we are stuck with that forever.

src/views/Home/components/NewsCard/index.js
cpabbot: That makes sense yeah and is good to know, thanks!!

cpabbot (changes requested):
LGTM

cpabbot (dismissed):
Accidentally hit `request changes` but approving, looks good to me.

EjPlatzer:
Closing this PR to switch to a Release branch.

PR #943, opened 2021-01-19
Bug fix nav avatar

NerdyGriffin:
Fixed error caused by incorrect use of 'classes' and error caused by 'name' being null before page finishes loading.
See commit descriptions for more details about the errors in question


EjPlatzer (approved):
Good fix, thank you for correcting this!

PR #944, opened 2021-01-19
Merge 'develop' into 'apartapp-develop'

NerdyGriffin:
It appears the write access for the protected branch may be setup incorrectly. It should allow any member of the ApartApp team to write to the branch 'apartapp-develop', while also requiring approval to merge pull requests.


NerdyGriffin:
This is simply an automatic merge from 'develop' to 'apartapp-develop' to fix the merge conflict that is blocking #926 . Hopefully we will get the protected branch settings fixed in the future so that we can simply perform these sorts of upstream merges directly in apartapp-develop

glgnkim (approved):
LGTM. (Resolving the merge conflict)

PR #947, opened 2021-01-20
Fix profile login message error

EjPlatzer:
The profile page is perpetually displaying the message "Please login to view this profile", even when logged in. This was because of a small error merging master into develop for a release. The Authentication prop had been refactored into lowercase authentication by a change in develop. However, this change got overwritten by a change from master to this one reference to that prop. This meant that the check for authentication was always failing on the profile page because the component has no Authentication prop.

### DIFF
--- a/src/views/Profile/index.js
+++ b/src/views/Profile/index.js
@@ -189,7 +189,7 @@ export default class Profile extends Component {
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


bennettforkner (approved):
LGTM!

PR #948, opened 2021-01-21
RELEASE: Fix Profile Auth error, add advisors to My Profile

EjPlatzer:
This release is a hotfix to the (public) Profile view to fix an authentication error. Because of a merging oddity in a previous release, the Authentication prop had the wrong casing. In Develop, this prop had been refactored to be lower case, and was working correctly in all components. However, because of a commit that was in Master but not Develop that affected one instance of the prop in the Profile view, the casing was messed up. Because the uppercase Authentication prop no longer existed for the Profile view, the authentication check was failing every time (even when the user is logged in), which made it impossible to see other users' profiles. 

this release also includes the last code needed to add advisors to the My Profile view.


EjPlatzer:
Need to merge to master from a separate release branch. One of these days I'll remember that.

PR #950, opened 2021-01-22
People Search UI Visual Refresh

### LINKED ISSUE #867: Make People Search Great Again
The UI on the 360 People Search is kind of outdated. Could definitely be refreshed and be a lot simpler and more intuitive.

cpabbot:
## Visual Update to the People Search Page

This overhauls the People Search page with a UI refresh.
The three search categories drop down with options, and they highlight cyan when there is an active filter being applied (so that the user knows a filter is influencing their search).

This PR also fixes two bugs:
- Search by hall was broken due to extra whitespace in the database entries for halls, which causes the advanced search requests to fail. This was fixed with one line trimming the whitespace off before sending a request. This should be noted as this may cause unexpected issues in the future, but I did not want to change the actual database values right now without knowing the side effects it might have had.
- Search by some majors was broken due to colons and forward slashes not being translated properly through the request. This is resolved and corresponds with API PR #448 (see here https://github.com/gordon-cs/gordon-360-api/pull/448).

### BEFORE
image

### AFTER
image

Resolves #867


src/views/PeopleSearch/index.js
```diff
@@ -425,14 +427,27 @@ class PeopleSearch extends Component {
     this.setState((state) => ({ nameExpanded: !state.nameExpanded }));
   };
   handleAcademicsExpandClick = () => {
+    this.setState({
+      homeExpanded: false,
+      offDepExpanded: false,
+    });
     this.setState((state) => ({ academicsExpanded: !state.academicsExpanded }));
```
EjPlatzer, line 434: A small thing, but I see no reason why these should be split across two setStates. They should work fine (and be more easily readable) as one unified setState call.

src/views/PeopleSearch/index.js
```diff
@@ -425,14 +427,27 @@ class PeopleSearch extends Component {
...
   };
   handleHomeExpandClick = () => {
+    this.setState({
+      academicsExpanded: false,
+      offDepExpanded: false,
+    });
     this.setState((state) => ({ homeExpanded: !state.homeExpanded }));
```
EjPlatzer, line 441: Same as above.

src/views/PeopleSearch/index.js
```diff
@@ -425,14 +427,27 @@ class PeopleSearch extends Component {
...
   };
   handleOffDepExpandClick = () => {
+    this.setState({
+      academicsExpanded: false,
+      homeExpanded: false,
+    });
     this.setState((state) => ({ offDepExpanded: !state.offDepExpanded }));
```
EjPlatzer, line 448: Same as above.

EjPlatzer (changes requested):
A huge improvement! I have requested a couple of small changes, but I think this is great overall.

I think Yellow is the wrong color for dropdowns that have active filters. It looks more like a warning or something to me and draws too much attention. Maybe Gordon cyan is a better color?
image
Relatedly, when the filter category is active, the text color should switch to white. The navy text is hard to read on the yellow (or cyan) background. Currently it switches to white when the category is expanded, but it should stay that way when not expanded but still colored.

I think the Clear All button should be moved down to the same row as the search button. Putting it next to the expanded categories makes me think it will only clear those categories, but it also clears all the main fields above.

src/views/PeopleSearch/index.js
cpabbot: Totally hadn't noticed that, good spot!

src/views/PeopleSearch/index.js
```diff
@@ -1130,90 +1154,74 @@ class PeopleSearch extends Component {
...
-                          [classes.expandOpen]: this.state.additionalOpsExpanded,
-                        })}
-                        aria-expanded={this.state.additionalOpsExpanded}
-                        aria-label="Show more"
+                    {/* Clear All Button */}
+                    <Grid item xs={8} sm={'auto'}>
+                      <Button
+                        style={{backgroundColor: gordonColors.neutral.lightGray}}
+                        fullWidth
+                        disableElevation
```
EjPlatzer, line 1193: Is there a reason elevation is disabled here? To me at least, it makes it look less like a "clear" button and more like a disabled button that I can't interact with.

EjPlatzer (changes requested):
This looks great! Color updates look nice, and the new UI is very clean.

One more small but potentially important thing. Typically, a negative action button (such as cancel or - in this case - clear) sits to the left of the normal action, i.e. search. I think this would make our UI more intuitive for most people, and it would also make the Mobile view layout more intuitively. I got a second opinion from a co-worker who isn't familiar with the page and they agreed that left was more intuitive than right.

src/views/PeopleSearch/index.js
cpabbot: You're right. I originally had disabled it because it looked more consistent with the filter menu; I still prefer the flatter look but you're right and it should be elevated. I've rearranged the buttons as well. Thanks for pointing those things out! Let me know if you notice any other possible improvements!

EjPlatzer (approved):
Looks good to me! The new UI is now intuitive and nice to use for both desktop and mobile browsers.

PR #956, opened 2021-01-27
Apartapp functional components

NerdyGriffin:
Converted all ApartApp components from class components to functional components

### DIFF
--- a/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
+++ b/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
@@ -21,43 +21,38 @@ import user from '../../../../../../../../services/user';
 // Based off src/views/PeopleSearch/components/PeopleSearchResult
 // but using props.profile of type StudentProfileInfo
 // rather than using this.props.Person of type PeopleSearchResult
-const ApplicantListItem = (props) => {
+const ApplicantListItem = ({ profile, isApplicationEditor, onChangeEditor, onApplicantRemove }) => {
   const [avatar, setAvatar] = useState(null);
   const [anchorEl, setAnchorEl] = useState(null); // A HTML element, or a function that returns it. It's used to set the position of the menu.
-  const [hasNickName, setHasNickname] = useState(Boolean);
-  const [personClass, setPersonClass] = useState(props.profile.Class);
+  const [hasNickName, setHasNickname] = useState(false);
+  const [personClass, setPersonClass] = useState(profile.Class);
 
   useEffect(() => {
-    loadAvatar(props.profile);
-    createNickname(props.profile);
-    if (String(props.profile.PersonType).includes('stu') && props.profile.Class !== undefined) {
-      setPersonClass(props.profile.Class);
+    loadAvatar(profile);
+    createNickname(profile);
+    if (String(profile.PersonType).includes('stu') && profile.Class !== undefined) {
+      setPersonClass(profile.Class);
     } else {
       // Techincally, this case should never happen because the list does not allow the user to add a non-student to the applicant list
       setPersonClass('');
     }
-  }, [props.profile]);
+  }, [profile]);
 
   /**
    * Creates the Avatar image of the given user
    * @param {StudentProfileInfo} profile The StudentProfileInfo object for the student represented by this list item
    */
   const loadAvatar = async (profile) => {
-    const [{ def: defaultImage, pref: preferredImage }] = await Promise.all([
-      await user.getImage(profile.AD_Username),
-    ]);
-    let newAvatar = null;
-    if (profile.AD_Username) {
-      newAvatar = preferredImage || defaultImage;
-    }
-    if (newAvatar === null) {
-      newAvatar = (
+    try {
+      const { def: defaultImage, pref: preferredImage } = await user.getImage(profile.AD_Username);
+      setAvatar(preferredImage || defaultImage);
+    } catch {
+      setAvatar(
         <svg width="50" height="50" viewBox="0 0 50 50">
           <rect width="50" height="50" rx="10" ry="10" fill="#CCC" />
-        </svg>
+        </svg>,
       );
     }
-    setAvatar(newAvatar);
   };
 
   /**
@@ -80,28 +75,36 @@ const ApplicantListItem = (props) => {
 
   const handleChangeEditor = () => {
     // Make sure the chosen profile was not null
-    if (props.profile) {
+    if (profile) {
       // Send the selected profile to the parent component
-      props.onChangeEditor(props.profile);
+      onChangeEditor(profile);
       handleMenuClose();
     }
   };
 
   const handleRemove = () => {
     // Make sure the chosen profile was not null
-    if (props.profile) {
+    if (profile) {
       // Send the selected profile to the parent component
-      props.onApplicantRemove(props.profile);
+      onApplicantRemove(profile);
       handleMenuClose();
+    }
+  };
+
+  const displayName = () => {
+    if (hasNickName) {
+      return profile.fullName + ' (' + profile.NickName + ')';
+    } else {
+      return profile.fullName;
     }
   };
 
   return (
     <ListItem
-      key={props.profile.AD_Username}
+      key={profile.AD_Username}
       component={Link}
       target="_blank"
-      to={`/profile/${props.profile.AD_Username}`}
+      to={`/profile/${profile.AD_Username}`}
       className={'list-item'}
     >
       <ListItemAvatar>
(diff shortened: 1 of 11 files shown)


src/views/ApartmentApp/index.js
```diff
@@ -1,96 +1,105 @@
...
     try {
-      const profile = await user.getProfileInfo();
-      this.setState({ userProfile: profile });
-      this.setState({ isStu: String(profile.PersonType).includes('stu') });
-      this.setState({ loading: false });
+      user.getProfileInfo().then((data) => {
+        setUserProfile(data);
+        data.PersonType.includes('stu') ? setIsUserStudent(true) : setIsUserStudent(false);
+      });
     } catch (error) {
```
EjPlatzer, line 33: Using `promise().then(...)` syntax makes the Try Catch block meaningless sadly. It's probably best to switch it back to await since it's an async function.

src/views/ApartmentApp/index.js
```diff
@@ -1,96 +1,105 @@
...
+  const [canUseStaff, setCanUseStaff] = useState(false);
+  const [isUserStudent, setIsUserStudent] = useState(false);
 
-  componentDidMount() {
-    this.loadProfile();
-  }
+  useEffect(() => {
+    loadUserProfile();
+    checkHousingStaff();
+  });
```
EjPlatzer, line 21: Not passing a dependency array means that this useEffect gets run on every render. I think you probably mean for these to get called only after the first render, which would require an empty dependency array as the second argument of useEffect.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -14,171 +14,177 @@ import AlertDialogBox from '../../../../components/AlertDialogBox';
...
-      applicants.push(this.props.userProfile);
-      this.setState({ applicants });
-      this.setState({ loading: false });
-    } catch (error) {
-      // Do Nothing
-    }
-  }
+  useEffect(() => {
+    loadSavedApplication();
+  });
```
EjPlatzer, line 64: This is probably meant to be called only on the first render, right? If so, you want to pass it an empty dependency array.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -378,302 +375,283 @@ export default class StudentApplication extends Component {
...
-              <GordonLoader />
-            ) : (
-              <div className="apartment-application">
-                <Grid container justify="center" spacing={2}>
+  return (
+    <div>
+      {loading ? (
+        <GordonLoader />
+      ) : (
+        <div className="apartment-application">
```
EjPlatzer, line 460: Rather than wrapping in an empty div, could you just return the loader if loading, else return the application?
```suggestion
  if (loading) {
      return <GordonLoader />
  } else {
  return (
        <div className="apartment-application">
```

EjPlatzer (changes requested):
This looks great! Thank you for refactoring, I know it was probably frustrating and tedious but I hope you agree that the resulting code is of higher quality. Notice that despite implementing several new features you still ended deleting more code than you added.

I requested a couple of kinda nit-picky changes, but overall I would say this is nearly done!

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -17,178 +17,165 @@ import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
...
+    if (newAvatar === null) {
+      newAvatar = (
         <svg width="50" height="50" viewBox="0 0 50 50">
           <rect width="50" height="50" rx="10" ry="10" fill="#CCC" />
         </svg>
       );
     }
-    this.setState({ avatar });
-  }
+    setAvatar(newAvatar);
```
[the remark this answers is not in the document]
NerdyGriffin: Thank you, I will change it to something like that right now

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -17,178 +17,165 @@ import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
...
-import '../../../../../../../PeopleSearch/components/PeopleSearchResult/peopleSearchResult.css';
 
 // Based off src/views/PeopleSearch/components/PeopleSearchResult
-// but using this.props.profile of type StudentProfileInfo
+// but using props.profile of type StudentProfileInfo
 // rather than using this.props.Person of type PeopleSearchResult
-export default class ApplicantListItem extends Component {
-  constructor(props) {
-    super(props);
+const ApplicantListItem = (props) => {
```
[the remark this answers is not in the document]
NerdyGriffin: Yes, I noticed that going on in code such as 'Wellness' and I suspected that was what it meant. I figured i'd go back after and explore that, since I just needed to check the parent component to find out which components I had previously planned to use.

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -17,178 +17,165 @@ import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
...
+                    ')'
+                  : // If the user has a title and no nickname
+                    props.profile.Title + ' ' + props.profile.fullName
+                : // If the user doesn't have a title
+                hasNickName
+                ? // If the user doesn't have a title but has a nickname
+                  props.profile.fullName + ' (' + props.profile.NickName + ')'
+                : // If the user doesn't have a title or a nickname
+                  props.profile.fullName
+            }
```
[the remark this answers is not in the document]
NerdyGriffin: I agree, I have gone back-and-forth with different ways of doing this, trying to explore how to make it the least confusing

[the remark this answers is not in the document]
NerdyGriffin: I think the first half of that can be removed as well because the list items will never have profile.personType === 'fac' 
They can only ever be students,

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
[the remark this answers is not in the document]
NerdyGriffin: It was helpful for pointing out which props were being passed but not being used (likely due to losing track of things that were added and removed over time with different features)

src/views/ApartmentApp/index.js
NerdyGriffin: Thank you for clarifying that, I was unsure about how those dependencies worked and I meant to use it exactly the way you guessed it

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: I spotted this during my second review of the code, while making some simply bug fixes

NerdyGriffin:
I appreciate the nit-picky changes since I try to make maintain code that is consistent across the site and I appreciate the quality, clarity, conciseness, and readability of the code

Additionally, The switch to functional components made it easier to spot any redundant and unnecessary parts of sub-components that were already being handled in their parent components, and also created a great opportunity to fix other things that had been annoying me about the organization in the code.  The next challenge will be merging these changes into the two WIP branches that contain partial implementations of new features.

EjPlatzer (changes requested):
This looks great! The code is high quality and much easier to maintain and find bugs in.

I did notice one kinda bug, see below comment, but as soon as that's fixed this is ready!

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -14,171 +14,179 @@ import AlertDialogBox from '../../../../components/AlertDialogBox';
...
-      this.setState({ loading: false });
-    } catch (error) {
-      // Do Nothing
-    }
-  }
+  useEffect(() => {
+    loadSavedApplication();
+
+    // eslint-disable-next-line
+  }, []);
```
EjPlatzer, line 66: EsLint is trying to warn you about a legitimate bug, in this case. The loadSavedApplication function actually gets redefined on every render. This is not really impactful for performance but it does mean that the useEffect would have to rerun every render because the function keeps changing. The way you have it now, it will probably work 99% of the time but is not idiomatic React. The bug is that, if the component stayed mounted but the userProfile prop changed, this useEffect wouldn't reload so you wouldn't load the new application.

The simplest way to fix this is to define loadSavedApplication within the useEffect. Then, you can properly declare it's single dependency userProfile and the useEffect will run at the needed times (first render and whenever userProfile changes). There's another way to fix this using useCallback, but that is significantly more complex and is only necessary when the function (i.e. loadSavedApplication in this case) needs to be called elsewhere in the component outside the useEffect. Since you aren't doing that, I think the simpler fix is the way to go for you.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -14,171 +14,179 @@ import AlertDialogBox from '../../../../components/AlertDialogBox';
...
+    let newApplicationID = -1; // await housing.getApplicationID();
+    if (newApplicationID !== null && newApplicationID !== -1) {
+      setApplicationID(newApplicationID);
+      let applicationDetails = await housing.getApartmentApplication(newApplicationID);
+      if (applicationDetails) {
+        if (applicationDetails.Username) {
+          setEditorUsername(applicationDetails.Username);
+        }
+        if (applicationDetails.Applicants) {
+          setApplicants(applicationDetails.Username);
```
EjPlatzer, line 81: Is this supposed to be Username? My instinct would be that you would set applicants to applicationDetails.Applicants.

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: This function actually will need to be called elsewhere in the file for several features that are coming soon, so I will have to research that other solution today

EjPlatzer: I think I may have over stated it's complexity. You essentially just need to wrap the function declaration in the useCallback hook. The second argument of useCallback is the dependencies of the function (in this case, just userProfile I think). You also need to define the callback before you use it anywhere, so it would have to come before the useEffect call.

 i.e. `const loadSavedApplication = useCallback(async () => {...}, [userProfile]);`

NerdyGriffin: Thanks, I was confused when the EsLint used the phrase "wrap the function declaration in ....", but I didn't have a chance to look it up at the time.
After seeing an example like that it makes perfect sense

EjPlatzer (approved):
LGTM! Thanks for the quality refactoring!

PR #959, opened 2021-01-28
Fix shift rounding error when shift was > 5 minutes

EjPlatzer:
The math for rounding up student shifts was using strict equals but the hour difference variable was a string being compared to an integer, so it was always returning false. This way, we properly detect shifts that are less than 5 minutes and round them up to 5 minutes, which is the minimum length acceptable.


jason-whitehouse (changes requested):
For the first one (line 143) should we just change to 0.0 like lines 225 and 266 so it's all consistent? Also it looks like there is no need to put the quotes around 0.08 in line 144 if you can not do it and it works for lines 226 and 267? Or quote everything...I think just have it consistent all the way through so no one is confused about the difference later

jason-whitehouse (approved):
That's fine to use less than. It will always equal zero though with the rounding it does if hours worked is less than the min interval. (1/12th hour- 5minutes)

PR #960, opened 2021-01-29
RELEASE: Timesheets adjustment, new People Search UI, GHA CI/CD, Handshake Link

EjPlatzer:
This release contains a number of small features. It's being pushed before the semester starts because it contains an adjustment to timesheets that changes minimum shift lengths to 5 minutes. This adjustment is motivated by a job peculiar to the COVID-19 Pandemic in which students do extremely fast sanitizations of classrooms between classes. Typical shift lengths were less than the previous minimum of 15 minutes, and so the minimum has been changed.

This pull request also includes a refresh of the People Search UI, which makes it simpler and more appealing to use.

This pull request brings our new CI/CD solution with GitHub Actions into master for the first time.

Finally, this pull request adds Handshake links to Profiles.


jason-whitehouse (approved):
Timesheet adjustment is good.

PR #963, opened 2021-01-30
ApartApp Get Application ID

NerdyGriffin:
Successfully tested with [apartapp-get-id](https://github.com/gordon-cs/gordon-360-api/pull/440)

Adds support for getting application ID, checking for existing applications, checking if user is housing staff, and a few other endpoint that are still in development

- [x] Add support for endpoints to get an existing application ID
- [x] Use application ID to check if a student is on an existing application before adding them to the applicants list
- [x] Use application ID to check (during initial page load) whether the current user has an existing application save in the database
- [x] Disable the editable fields of the application if the current user is not the application editor (including when the current changes the editor)
- [x] Comments and documentation


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -122,57 +122,66 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+      // TODO: Implement this once save/load of application data has been implemented in the backend
+      setLoading(true);
+      // Check if the current user is on an application. Returns the application ID number if found
+      let newApplicationID = null;
+      try {
+        newApplicationID = await housing.getApplicationID();
+        console.log('Retrieved Application ID: ' + newApplicationID);
+        if (newApplicationID === null || newApplicationID === -1) {
+          // Intentionally trigger the 'catch'
+          throw new Error("Invalid value of 'newApplicationID' = " + newApplicationID);
```
Nick-Noormand, line 143: Is it necessary to have this error-checking/debugging merged?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -122,57 +122,66 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+          }
+          if (applicationDetails.DateModified) {
+            setDateModified(applicationDetails.DateModified);
+          }
+          if (applicationDetails.Username) {
+            setEditorUsername(applicationDetails.Username);
+          }
+          if (applicationDetails.Applicants) {
+            setApplicants(applicationDetails.Applicants);
+          }
```
Nick-Noormand, line 159: Is the way you are doing this if statement the most efficient way of setting everything?

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: This is not a debugging feature. As the comment above it says, the error also serves the purpose of forcing the `try` to go to its `catch`, which will make sure the variables are set to the correct default values. 
It is important that the code exits the `try` section if the `newApplicationID` is an invalid value so that it does not waste time calling the "load" API endpoint with a nonsensical ID value

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: This approach is for the sake of resilience/reliability at the expense of some efficiency. This is also needed because it is often possible for only some of the properties of `applicantionDetails` to have real values (Example: if you are loading an application that has applicants saved but no halls saved yet)
The `DateSubmitted` is also expected to be false or null or empty most of the time, so it is best to check the parts of the applicationDetails individually to make sure things always behave the way that they are supposed to

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -154,14 +154,20 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
   const loadSavedApplication = useCallback(async () => {
     // TODO: Implement this once save/load of application data has been implemented in the backend
     setLoading(true);
     // Check if the current user is on an application. Returns the application ID number if found
-    let newApplicationID = -1; // await housing.getApplicationID();
-    if (newApplicationID !== null && newApplicationID !== -1) {
+    let newApplicationID = null;
+    try {
+      newApplicationID = await housing.getApplicationID();
+      console.log('Retrieved Application ID: ' + newApplicationID);
```
NerdyGriffin, line 166:
```suggestion
      //! DEBUG
      console.log('Retrieved Application ID: ' + newApplicationID);
```

src/services/housing.js
```diff
@@ -30,6 +26,13 @@ import './user';
...
+/**
+ * @global
+ * @typedef ApplicantPartialInfo
+ * @property {String} Username The username of this applicant
+ * @property {String} OffCampusProgram The name of department of this applicant's off-campus program, or 'None'
+ */
+
 /**
  * Note: Properties 'HallRank' and 'HallName' must be capitalized to match the backend
  * @global
```
joshuaRogers24, line 38: Why is this Note written twice?

src/services/housing.js
```diff
@@ -55,12 +58,11 @@ import './user';
...
  */
 const checkHousingStaff = async () => {
-  return false; //! DEBUG
-  // try {
-  //   return await http.get(`housing/staff`);
-  // } catch {
-  //   return false;
-  // }
+  try {
+    return await http.get(`housing/staff`);
```
joshuaRogers24, line 62: The staff endpoint has not been made yet, has it?

joshuaRogers24: Oh, it's in 440. I will remember this while looking into authorization. I am not sure it fits the way authorization is done on the rest of the site.

src/services/housing.js
```diff
@@ -46,7 +49,7 @@ import './user';
  * @property {*} DateSubmitted The date the application was submitted, or null if not yet submitted
  * @property {*} DateModified The date the application was last modified
  * @property {String} Username Username of the primary applicant
- * @property {String[]} Applicants Array of student usernames
+ * @property {ApplicantPartialInfo[]} Applicants Array of objects with applicant username and off-campus program
```
joshuaRogers24, line 52: Should this object be named "Applicants" while your list of ApartmentApplicant objects in saveApartmentApplication is also named "applicants"?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -247,7 +264,6 @@ const StudentApplication = ({ userProfile, authentication }) => {
             ' because they are not a student.',
         );
         setSnackbarSeverity('warning');
-        setSnackbarOpen(true);
       } else if (newApplicantProfile.Gender && newApplicantProfile.Gender !== userProfile.Gender) {
```
joshuaRogers24, line 267: Did you mean to repeat the same value (newApplicantProfile.Gender)?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -227,18 +236,26 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
     try {
       // Get the profile of the selected user
-      let newApplicantProfile = await user.getProfileInfo(username);
+      const newApplicantProfile = await user.getProfileInfo(username);
       let newApplicantObject = { Profile: newApplicantProfile, OffCampusProgram: '' };
+
+      // Check if the selected user is already saved on an application in the database
+      let existingAppID = null;
+      try {
+        existingAppID = await housing.getApplicationID(username);
```
joshuaRogers24, line 250: Did we decide that it will be possible for a student to have their name on an application from a past session? If so we need to also check if the existingAppID is one from the current session or not.

joshuaRogers24 (changes requested):
Great work! See some comments.

src/services/housing.js
NerdyGriffin: probably a weird 'git merge', because I think I moved it from the top to the bottom or vice-versa. I will fix it rn

src/services/housing.js
NerdyGriffin: Yes. It is meant to match the variable named in the backend view model and the database because that typedef is used for the data returned by `getApartmentApplication` (the "load" endpoint, which is not yet implemented).
These typedefs exist mostly for documentation purposes, as I don't think the javascript does anything this enforce these variable typings, but I know it is parsed by the VSCode stuff to show the info in tooltips and it shows developers what properties they can expect to access from the data that is retrieved from the back.

This `Applicants` array is also meant to be the same as the `Applicants` array created within the `applicationDetails` object in `saveApartmentApplication`, whereas the lowercase `applicants` is a parameter of the function.

NerdyGriffin: Simply put, the `ApartmentApplicant` object type is used by the frontend because it is best to hold onto the full StudentProfileInfo object for each applicant 

The `ApplicantPartialInfo` object type is needed because we need to replace the StudentProfileInfo objects with just the usernames when we want to send/recieve data to/from the database

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: It is a common reliability practice that I see a lot in React code or JS in general, where you use `if (variable && variable === something)` rather than just `if (variable === something)`. This is done to explicitly check whether the thing happens to be null or false or some other invalid state before trying to make the comparison.

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: The backend part of this already handles that aspect.

The backend does all the session-code related stuff, the frontend doesn't need to send any additional information to contribute to the session-code check

src/services/housing.js
NerdyGriffin: Yes, this is a feature that is still WIP in the backend, which is why the call is also put into a `try ... catch`, 
(as well as to make sure the page defaults to the student page if something goes wrong under normal operation)

NerdyGriffin: More importantly, it is not the main feature of this branch, and it is only implemented as far as it is to allow the existing UI features to continue to work as expected. It will easily be changed in the future when that housing staff authorization aspect is actually ready

src/services/housing.js
joshuaRogers24: That makes sense. Thanks, Christian

src/services/housing.js
joshuaRogers24: Great, makes sense

src/views/ApartmentApp/components/StudentApplication/index.js
joshuaRogers24: Okay, interesting. Good thinking, then.

src/views/ApartmentApp/components/StudentApplication/index.js
joshuaRogers24: Oh, I see! Would it be more clear then to rename this service method getCurrentApplicationID or something similar?

src/services/housing.js
```diff
@@ -55,12 +58,11 @@ import './user';
...
-  //   return false;
-  // }
+  try {
+    return await http.get(`housing/staff`);
+  } catch {
+    return false;
+  }
 };
 
 /**
```
joshuaRogers24, line 67: Also, the comment on this service could use some clarification that it is getting the ID of an application _from this session_.

NerdyGriffin: Maybe "semester", since "session" doesn't seem to have much relevance on the frontend, as far as I've seen so far.

joshuaRogers24: Okay, sure :)

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Good idea, will make the change now

joshuaRogers24 (approved):

PR #967, opened 2021-02-05
Remove expo files

EjPlatzer:
The .expo folder and it's contents were added in a strange [Git manipulation commit](https://github.com/gordon-cs/gordon-360-ui/commits/develop/.expo), and aren't actually referenced or used anywhere else in the project. We don't have or use Expo in this project at all. Removing these helps get rid of unnecessary bloat and tidy up our top-level files so that the meaningful things can be found and used more easily.

### DIFF
--- a/.expo/packager-info.json
+++ b/.expo/packager-info.json
@@ -1,3 +1,0 @@
-{
-  "devToolsPort": 19002
-}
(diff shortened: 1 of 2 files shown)


Kindlewing (approved):
Looks good! If we don't need it, might as well get rid of it!

PR #969, opened 2021-02-05
Fix CL&W credits Required default

EjPlatzer:
The required CL&W Credits default was broken because it depended on AttendedChapelEvents, so we couldn't fetch the actual value until the student had attended some event in the current session. The API was fixed to add ChapelRequired and ChapelAttended to the StudentProfileInfo. This commit uses those values for the CLWCreditsDaysLeft wheel.

Fixes gordon-cs/gordon-360-api/issues/454

### DIFF
--- a/src/services/user.js
+++ b/src/services/user.js
@@ -11,7 +11,6 @@ import http from './http';
 import session from './session';
 import storage from './storage';
 import { socialMediaInfo } from '../socialMedia';
-import gordonEvent from './event';
 
 /**
  * @global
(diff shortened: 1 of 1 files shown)


cpabbot (approved):
Looks good, corresponds with API PR https://github.com/gordon-cs/gordon-360-api/pull/460

Side note: Should I adopt this `if (example)` convention with a space?

cpabbot (approved):
Looks good, corresponds with API PR https://github.com/gordon-cs/gordon-360-api/pull/460

Side note: Should I adopt this `if (example)` convention with a space?

PR #999, opened 2021-02-16
Inform how to check for nvm install

joshuaRogers24:
A new 360 developer changed the nvm environment variable for all CS-RDSH-02 users so that access to it was restricted. The issue was fixed. However, because they were likely following the instructions in the README without realizing that nvm was already installed, I thought a README change would prevent future such mishaps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
 
 This project was bootstrapped with [Create React App](https://github.com/facebookincubator/create-react-app). Read the user guide [here](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md).
 
-Make sure Node.js is set up on your machine. The following procedures will install nvm, a node.js version manager onto your machine. An npm installation may also work, but nvm will allow easier installation and use of this particular version.
+Make sure Node.js is set up on your machine. If you are on the CS-RDSH-02 virtual machine, it should be setup already. To check if it is on a machine, open a terminal and run the command `nvm`. If the output says `Running version...`, then Node.js is setup. If not, the following procedures will install nvm, a node.js version manager onto your machine. An npm installation may also work, but nvm will allow easier installation and use of this particular version.
 
 #### Windows:
 
--- a/README.md
+++ b/README.md
@@ -34,13 +34,17 @@ Make sure Node.js is set up on your machine. If you are on the CS-RDSH-02 virtua
 - Download the latest release of NVM (Node Version Manager) from [here](https://github.com/coreybutler/nvm-windows/releases). `Select nvm-setup.zip`.
   - Extract the file and run it.
   - The installer will ask you where to install NVM. It will display the path where it is currently set to install. Ensure that the path reflects the account you are logged into (example: if you're logged in as anthony.aardvark, make sure the path looks like `C:\Users\aanthony.aardvard\Program Files\etc`). If it says Node is already installed, proceed anyway.
-  - After the installation completes, open a terminal and run these commands in order:
-    - `nvm install 9.11.1`
-    - `nvm use 9.11.1`
+  - After this, simply run `nvm install`.
 
 #### Linux:
 
-Follow [these instructions](https://hackernoon.com/how-to-install-node-js-on-ubuntu-16-04-18-04-using-nvm-node-version-manager-668a7166b854) to install NVM. At the step where it asks you to run nvm install, run `nvm install 9.11.1` then run `nvm use 9.11.1`.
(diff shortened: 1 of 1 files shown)


joshuaRogers24:
I removed instructions to install and use nvm version 9.11.1 since Travis no longer holds us to that.

README.md
```diff
@@ -27,20 +27,24 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
...
   - The installer will ask you where to install NVM. It will display the path where it is currently set to install. Ensure that the path reflects the account you are logged into (example: if you're logged in as anthony.aardvark, make sure the path looks like `C:\Users\aanthony.aardvard\Program Files\etc`). If it says Node is already installed, proceed anyway.
-  - After the installation completes, open a terminal and run these commands in order:
-    - `nvm install 9.11.1`
-    - `nvm use 9.11.1`
+  - After this, run `nvm install --lts`, which installs the latest stable version of nvm.
 
 #### Linux:
 
-Follow [these instructions](https://hackernoon.com/how-to-install-node-js-on-ubuntu-16-04-18-04-using-nvm-node-version-manager-668a7166b854) to install NVM. At the step where it asks you to run nvm install, run `nvm install 9.11.1` then run `nvm use 9.11.1`.
+Follow [these instructions](https://hackernoon.com/how-to-install-node-js-on-ubuntu-16-04-18-04-using-nvm-node-version-manager-668a7166b854) to install NVM. At the step where it asks you to run nvm install, do so.
```
EjPlatzer, line 41: I'm not sure why we originally decided to use these instructions from hackernoon, but it seems like we shouldn't depend on an external website and should instead point directly to the nvm project itself, which has the correct version of these same instructions, as well as heplful troubleshooting for common errors.
```suggestion

- Follow [the nvm installation instructions](https://github.com/nvm-sh/nvm#install--update-script) to install NVM. 
- Once it is successfully installed, use nvm to install your desired Node version: `nvm install node`.
- Tell nvm that you want to use that verion: `nvm use node`. 

To use a specific Node version - usually the latest Long Term Support (LTS) version - replace `node` with the desired version or alias, e.g. `nvm install 14.15.4` for version 14.15.4 or `nvm install --lts` for the latest LTS.
```

README.md
```diff
@@ -27,20 +27,24 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
 
 This project was bootstrapped with [Create React App](https://github.com/facebookincubator/create-react-app). Read the user guide [here](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md).
 
-Make sure Node.js is set up on your machine. The following procedures will install nvm, a node.js version manager onto your machine. An npm installation may also work, but nvm will allow easier installation and use of this particular version.
+Make sure Node.js is set up on your machine. If you are on the CS-RDSH-02 virtual machine, it should be setup already. To check if it is on a machine, open a terminal and run the command `nvm`. If the output says `Running version...`, then Node.js is setup. If not, the following procedures will install nvm, a node.js version manager onto your machine. An npm installation may also work, but nvm will allow easier installation and use of this particular version.
```
EjPlatzer, line 30: > If the output says `Running version...`, then Node.js is setup.
This is only true for the `nvm-windows` package. On Linux and MacOS (or any other shell terminal), running nvm will give you a very  verbose description of the nvm CLI and how to use it. I think it might be clearer to say that if the terminal returns ` 'nvm' is not recognized...`, then it needs to be installed.

EjPlatzer (changes requested):
Thanks for clarifying! I requested a couple changes where I think things could be clearer/more helpful.

README.md
```diff
@@ -27,20 +27,24 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
...
 
 #### Windows:
 
 - Download the latest release of NVM (Node Version Manager) from [here](https://github.com/coreybutler/nvm-windows/releases). `Select nvm-setup.zip`.
   - Extract the file and run it.
   - The installer will ask you where to install NVM. It will display the path where it is currently set to install. Ensure that the path reflects the account you are logged into (example: if you're logged in as anthony.aardvark, make sure the path looks like `C:\Users\aanthony.aardvard\Program Files\etc`). If it says Node is already installed, proceed anyway.
-  - After the installation completes, open a terminal and run these commands in order:
-    - `nvm install 9.11.1`
-    - `nvm use 9.11.1`
+  - After this, run `nvm install --lts`, which installs the latest stable version of nvm.
```
EjPlatzer, line 37: This is actually incorrect for nvm-windows. It only works on the shell nvm. Nvm for windows only supports the alias `latest`, which will install the newest version (which might not necessarily be stable). To get the LTS version, a user would have to look at NodeJS's site and find the current latest LTS version, e.g. `14.15.4` as of writing this. Finding the latest LTS version is fairly easy (just go to [https://nodejs.org/en/](https://nodejs.org/en/) and look at the version number marked LTS in the download options).

EjPlatzer (approved):
These instructions are clear and, I think, easy to follow. This should help people get setup faster and prevent similar problems in the future. Thanks Josh!

PR #1000, opened 2021-02-17
Updated .gitignore to exclude the package-lock.json file

Kindlewing:
This file is generated when you run 'npm install' for the first time, and it causes errors if it already exists. Removing this from the repo should make it regenerate when it is cloned locally, fixing the issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ yarn-error.log*
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+
+#package-lock.json
+package-lock.json
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,42 +5,38 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.8.3",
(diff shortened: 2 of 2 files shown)


Kindlewing:
I should mention that the package-lock.json file should be deleted in the repo. I didn't update that in the commit, because I didn't want to delete my version of the file locally.

EjPlatzer:
The file `package-lock.json` is generated and maintained by npm and is intended to be included in source control.

> package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.
>
> This file is intended to be committed into source repositories, and serves various purposes

See the [`npm help` file for `package-lock.json`](https://github.com/npm/cli/blob/latest/docs/content/configuring-npm/package-lock-json.md) for further details.

It was not the source of your issue, since it simply describes the dependency tree for the packages used by our project. We should leave it in source control and continue to commit. One of the many reasons to include it is so that different users with different clones of the project can install the same dependencies. It also provides a mechanism for tracking changes to the dependency tree, which means if it breaks at some point we can restore a previous working version without having to commit the entire `node_modules` directory of >200MB worth of dependencies.

PR #1005, opened 2021-02-18
Reinstalled packages to use Node 14

Kindlewing:
Definitely verify this, but I think it should be using 14 at this point

I deleted node_modules and regenerated the package-lock json file.

Pull this down, and try running npm install using the current node version, than go and get nodejs version 14, and try running npm install again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,42 +5,38 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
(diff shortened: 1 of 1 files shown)


EjPlatzer:
This doesn't need to be it's own Pull Request. Regenerating the package-lock.json is not the same as updating Node. What we need is to verify that 360 still runs properly via Node >=14.

PR #1006, opened 2021-02-18
Include alum checkbox only shows for non-students

EjPlatzer:
The include alumni checkbox on people search was showing for people who had a composite persontype (e.g. 'stufac'). No it won't show for anyone whose persontype includes 'stu' for student.

### DIFF
--- a/src/views/PeopleSearch/index.js
+++ b/src/views/PeopleSearch/index.js
@@ -74,7 +74,7 @@ const styles = {
   colorChecked: {},
   icon: {
     color: gordonColors.neutral.grayShades[900],
-  }
+  },
 };
 
 const noResultsCard = (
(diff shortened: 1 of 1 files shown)


cpabbot (approved):
LGTM, thanks for the reformatting as well! Just out of curiosity and for future note, can you give me an example of composite users, like the 'stufac'?

EjPlatzer:
> LGTM, thanks for the reformatting as well! Just out of curiosity and for future note, can you give me an example of composite users, like the 'stufac'?

If you look at [ProfilesController](https://github.com/gordon-cs/gordon-360-api/blob/0186d36567e42d8567baa88c27f9db443047ffcc/Gordon360/ApiControllers/ProfilesController.cs#L52) in the API, you can see that we handle cases for all sorts of different personTypes. For example, because of the lag in data updates in Train, my personType in Train is currently `stualu`. Additionally, we have a not insignificant number of people who work full-time with us and are thus `fac` (meaning faculty or staff) but also take one or more classes on occasion and would therefore qualify as students.

PR #1007, opened 2021-02-18
Fix date formatting for timezone bug

EjPlatzer:
Sending dates as ISO strings to the backend caused an unforeseen problem in that dates were being parsed by the backend as UTC and stored in the database as such, but then compared to local time without adjusting for the timezone difference. Instead, we will send the date in a custom format that approximates ISO but is in the user's local time zone. This fixes the above error and preserves the fix from using toLocaleString that caused issues for user's with browsers that had different locales.

Depends on https://github.com/gordon-cs/gordon-360-api/pull/475

### DIFF
--- a/src/services/jobs.js
+++ b/src/services/jobs.js
@@ -5,6 +5,11 @@
 
 import http from './http';
 
+const dateFormat = new Intl.DateTimeFormat('en', {
+  timeStyle: 'medium',
+  dateStyle: 'short',
+});
+
 /**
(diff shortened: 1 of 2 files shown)


bennettforkner (approved):
Glad to see a lot of efficiency improvements within the code. Inline conditionals on the "canUseStaff" variable are much better than having duplicates of the same code written out and surrounded by if statements.

LGTM. I look forward to seeing this fix one of our major timesheets issues. Ready for merge into develop.

PR #1008, opened 2021-02-18
Add network context

EjPlatzer:
This commit introduces the NetworkContext, which uses React's Context API to track the online/offline status of the user's network. The goal is to have network-aware component's subscribe to this context provider so that the network is tracked in only one place, rather than having eventListeners in every component that needs to know the network status.

This may result in minimal performace gains (having only one event listener attached to the window instead of >10 at once), but I would guess that the rerendering takes significantly more processing than the eventlisteners themselves. The real advantage to this technique is the separation of concerns. Now, the NetworkContextProvider is the only component responsible for the logic of tracking network changes. Every ther network-aware component simply subscribes to these changes with the very minmial hook useNetworkStatus - e.g. `const isOnline = useNetworkStatus()`, rather than having a large block of logic duplicated across many components with a useEffect hook and window listeners that obscures the component's more natural responsibilities.

This first PR merely introduces the hook. Subsequent PRs will include refactors of network-aware components to make optimal use of the hook.


src/contexts/NetworkContext.js
```diff
@@ -0,0 +1,61 @@
...
+ * Custom hook to subscribe to network status.
+ *
+ * Value is retrieved from local storage initially and updated by service worker firing events
+ *
+ * Can be used by any functional component under the NetworkContextProvider in App.js
+ *
+ * @returns {boolean} true if connected to the network, false otherwise.
+ */
+export const useNetworkStatus = () => {
+  const context = useContext(NetworkConext);
```
joshuaRogers24, line 16: Is this missing a 't'?

src/contexts/NetworkContext.js
```diff
@@ -0,0 +1,61 @@
+import React, { createContext, useState, useEffect, useContext } from 'react';
+import storage from '../services/storage';
+
+const NetworkConext = createContext();
```
joshuaRogers24, line 4: Oh, this is also missing a 't' I think. I was wondering how you did not get any errors! XD

src/contexts/NetworkContext.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  return context;
+};
+
+const NetworkContextProvider = ({ children }) => {
+  const [isOnline, setIsOnline] = useState('online');
+
+  useEffect(() => {
+    // Retrieve network status from local storage or default to online
```
joshuaRogers24, line 28: Why should it default to online?

src/contexts/NetworkContext.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /* Used to re-render the page when the network connection changes.
+     * The origin of the message is checked to prevent cross-site scripting attacks
+     */
+    const updateNetworkStatus = (event) => {
+      setIsOnline((prevStatus) => {
+        if (
+          event.origin === window.location.origin &&
+          (event.data === 'online' || event.data === 'offline')
```
joshuaRogers24, line 44: Trying to understand this security check: Are we checking the event.data to make sure we are not getting any data other than those exact two states?

joshuaRogers24: If you do not mind explaining, too, is event.data basically supposed to mirror prevStatus? Or, if event.data is an indication of what the status should be changed to, why do we change it to online whether event.data is online OR offline?

src/contexts/NetworkContext.js
```diff
@@ -0,0 +1,61 @@
...
+        }
+      });
+    };
+
+    window.addEventListener('message', updateNetworkStatus);
+
+    return () => window.removeEventListener('message', updateNetworkStatus);
+  }, []);
+
+  return <NetworkConext.Provider value={isOnline}>{children}</NetworkConext.Provider>;
```
joshuaRogers24, line 58: The variable missing 't' is used again here.

src/contexts/NetworkContext.js
```diff
@@ -0,0 +1,61 @@
...
+          return event.data === 'online';
+        } else {
+          return prevStatus;
+        }
+      });
+    };
+
+    window.addEventListener('message', updateNetworkStatus);
+
+    return () => window.removeEventListener('message', updateNetworkStatus);
```
joshuaRogers24, line 55: With my meager understanding of event listeners, I cannot follow what is happening here. If you have the time to explain this, I would appreciate it.

src/contexts/NetworkContext.js
EjPlatzer: I did indeed misspell Context when creating the variable 🤦. Thanks for catching that! I've pushed a fix.

joshuaRogers24 (changes requested):
This is a great endeavor that should simplify things well. I would like to hear your explanation of the code, though, so I will message you.

src/contexts/NetworkContext.js
EjPlatzer: The short answer is that, if it defaults to offline, we risk showing a flash of the offline component tree to users who are actually online while we fetch the network status. Since ~95% of our users are exclusively using the app online, this is undesirable to say the least. And in the case that the app IS offline, the online component tree - which is essentially the default in users' mind - will be replaced with the offline version in a useful amount of time (usually imperceptibly quickly).

src/contexts/NetworkContext.js
EjPlatzer: The security aspect is mostly in the first half of the boolean, `event.origin === window.location.origin`. This ensures that we aren't accepting events from external sites (which can have arbitrary - aka malicious - code inside them).

src/contexts/NetworkContext.js
EjPlatzer: This is essentially the same as the existing network-status code (which will be refactored out of the ~30 components that repeat it in the near future hopefully). It is basically attaching an event listener callback (`updateNetworkStatus` in this case) to the `window`, aka the browser tab/window that 360 is running in. 

This event listener callback will be called whenever an event is sent to the browser that has the 'type' (which is a string describing the kind of event) of `message`. These `message` events are sent either by our serviceWorker or the browser itself (although any other JavaScript code could send such an event, which is why we have to make sure in the event listener callback that we only pay attention to events originating from within our site). The callback is passed the `event` that triggered the eventListener, which in this case should include data about the status of the network connection (i.e. 'online' or 'offline'). 

Line 55 is telling the `useEffect` hook to call that arrow function whenever the NetworkContextProvider component is unmounted. Since the provider is mounted at the root of our app, this is unlikely to ever happen, but is still best practice for eventListeners. The arrow function simply removes the event listener from the `window`, allowing the browser to forget about it. If you don't remove the event listener,  you're creating a memory leak where the browser maintains an ever-expanding list of event listeners that were used by components which no longer exist and are therefore useless but are taking up memory and slowing down the browser.

src/contexts/NetworkContext.js
EjPlatzer: `event.data` is a property of the event set by whatever code created the event. In this case, we're expecting the event to originate from our serviceWorker. The serviceWorker will send out an event of type `message` whenever the network connection status changes. This status is (I believe by browser API standards) defined as either 'online' or 'offline'. So this function is getting called for all `message` events (see my below comment for further details on events), and ensuring that the event is originating from within our code. It is also checking that the event contains data relevant to the network connection status. The serviceWorker could send `message` events about several other things that aren't relevant to this component, so we check if the data is one of the network-statuses to filter out the irrelevant events.

If you look again at the return when the event is legitimate and relevant (line 28), you'll notice that it's actually returning a `boolean` value, which is whether the event's data property is equal to 'online'. Since we already know the data is one of either 'online' or 'offline', this is equivalent to: 

```
if (event.data === 'online') {
    return true
} else if (event.data === 'offline') {
    return false
}
```

EjPlatzer:
Oh dear, I accidentally pushed the next set of changes, which weren't supposed to be included in this pr. Serves me right for reusing the same branch instead of making a new one from the base I wanted to commit. I'll close this PR and open a new one with the initial set of commits that should be reviewed first.

PR #1009, opened 2021-02-19
Create guest welcome message

Kindlewing:


Kindlewing:
Known issue: When the view is fullscreen, there's a white strip on the bottom of the page, that presumably gets larger as the screen size goes up, so I don't know what's causing that, but other than that, everything scales nicely. (Try testing it locally to see what I mean)

Kindlewing:
Moved out of the 'Needs review' column, because there's still some github related things to do.

PR #1012, opened 2021-02-23
Merged from develop and other requested improvements

NerdyGriffin:
Improvements to the code based on code review in #926


EjPlatzer (approved):
These look like good improvements! Thanks for keeping up to date with `develop`, that will make the eventual merge into `develop` much easier.

PR #1013, opened 2021-02-23
Moved custom alert dialog features to existing dialog box component

NerdyGriffin:
It was suggested to make these parts of AlertDialogBox optional like this to make the component more modular. However, the AlertDialogBox is simply a modified clone of the existing GordonDialogBox component, so it makes more sense to add the modular/optional features to that component and remove the now-redundant AlertDialogBox entirely


EjPlatzer (approved):
I think this is a helpful extension of our GordonDialogBox component API. You've tested other uses of the GordonDialogBox to ensure these changes don't break any of the existing code, right? I'm confident it won't because you're only including alerts and the cancel button if optional props are present, but I'm big believer in trust but verify. If these aren't breaking changes, than this is approved and good to merge!

PR #1014, opened 2021-02-23
ApartApp Staff Menu

NerdyGriffin:
Test with API branch apartapp-develop

- [x] "Download CSV" Card
- [x] Table of applications
  - [x] Sub-table of applicants and apartments
- [x] Add the probation as a prop of the ApplicantInfo 
- [x] Download three separate files: Applications, Applicants, and ApartmentChoice
- [x] ~~Finish the sorting feature for the table of applications~~


README.md
```diff
@@ -163,6 +163,10 @@ Links to the homepages of libraries used in this project, listed here for easy r
 
   Provides easy routing, allowing transitions between views with back button support and URL management.
 
+- [react-csv](https://github.com/react-csv/react-csv)
```
joshuaRogers24, line 166: Oh, good work remembering to add this!!

src/services/housing.js
```diff
@@ -21,49 +21,76 @@ import './user'; // Needed for typedef of StudentProfileInfo
...
  * @property {String} OffCampusProgram The name of department of this applicant's off-campus program, or 'None'
+ * @property {String} Probation Indicates whether the student has a disiplinary probation (visble only to housing admin)
+ * @property {Number} Points The number of application points for this student (only visible to housing admin)
  */
 
 /**
  * Note: Properties 'HallRank' and 'HallName' must be capitalized to match the backend
  * @global
  * @typedef ApartmentChoice
+ * @property {Number} [ApplicationID] Application ID number of this application
```
joshuaRogers24, line 44: Why is this in brackets?

src/services/housing.js
```diff
@@ -21,49 +21,76 @@ import './user'; // Needed for typedef of StudentProfileInfo
...
+ * @property {Number} AvgPoints The average application points per applicant
  */
 
 /**
  * Check if the current user is authorized to view the housing staff page for applications
  * @return {Promise.<Boolean>} True if the user is authorized to view the housing application staff page
  */
-const checkHousingStaff = async () => {
+const checkHousingAdmin = async () => {
+  // return true; //! DEBUG
```
joshuaRogers24, line 68: Can this debug be removed?
```suggestion
```

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/index.js
```diff
@@ -0,0 +1,48 @@
...
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <React.Fragment>
+      <TableRow key={applicationDetails.ApplicationID} onClick={() => setOpen(!open)}>
+        <TableCell align="center" component="th" id={labelId} scope="row">
+          {applicationDetails.ApplicationID}
+        </TableCell>
+        <TableCell align="center">{applicationDetails.EditorUsername}</TableCell>
+        <TableCell align="center">{applicationDetails.Applicants.length}</TableCell>
```
joshuaRogers24, line 19: Should this have a word identifying it `Applicants: 5`

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -1,35 +1,172 @@
...
+  /**
+   * @type {[ApartmentChoice[], React.Dispatch<React.SetStateAction<ApartmentChoice[]>>]} Array of apartment choice info, formatted for use with react-csv
+   */
+  const [apartmentChoiceJsonArray, setApartmentChoiceJsonArray] = useState([]);
+
+  const [dateStr, setDateStr] = useState('');
+  const filePrefix = 'apartapp';
+
+  /**
+   * Attempt to load an all existing application for the current semester
```
joshuaRogers24, line 44:
```suggestion
   * Attempt to load all existing applications for the current semester
```

joshuaRogers24 (changes requested):
Wow, the table really added a lot of code to the admin side. Nice work! See a few questions/suggestsions.

src/services/housing.js
NerdyGriffin: It's means it is optional,

Almost like how we can do `string?` in the C# view models to say that it is valid for that property to be entirely null

In the JS, it doesn't actually enforce those typing rules, so it is mostly just there for devs and documentation

It is optional in this case because the frontend doesn't need to store that value there when it does the save stuff, but when it is used to load an application it will expect to see the value filled in by the backend,

So it is masked as optional, because the custom object is equally valid with or without that property

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/index.js
NerdyGriffin: I'm not sure what you mean.
If you mean put the text in the cell with the number, then the answer is no, because the column header on the table tells you what the numbers in that column represent

joshuaRogers24: Oh, right, I saw that later on and should have removed this comment. Good!

src/services/housing.js
joshuaRogers24: Okay, got it,

joshuaRogers24 (approved):
nice.

PR #1015, opened 2021-02-23
First Attempt at New Event Filter Checkboxes

aidanperez2000:
I just attempted to create a multiple-select feature for the events on the events page.  There are still a lot of changes I need to make.  The checkboxes right now do not actually filter.  The same checkbox might appear multiple times.  The checkboxes also do not remain checked.  I am going to continue looking into these issues.  Let me know if you have suggestions as to how to fix these issues.


src/views/Events/index.js
```diff
@@ -10,15 +10,49 @@ import gordonEvent from './../../services/event';
...
+
+const campusEvents = [
+  "CL&W Credits",
+  "Admissions",
+  "Arts",
+  "Athletics",
+  "CEC",
+  "Chapel Office",
+  "Student Life"
+]
```
EjPlatzer, line 54: I don't think you're using this static data anywhere, but this would be almost the right way to create the filters object described below.

src/views/Events/index.js
```diff
@@ -254,68 +288,63 @@ export default class Events extends Component {
       content = <EventList events={this.state.filteredEvents} />;
     }
 
+    let eventsAtGordon = [];
```
EjPlatzer, line 291: You need to use state to store this variable. Otherwise, it's getting recalculated every time the component renders, and so will be almost constantly empty. Also, you could probably rename this to `activeFilters`, since that's what it represents.

src/views/Events/index.js
```diff
@@ -254,68 +288,63 @@ export default class Events extends Component {
...
+
+    const handleChange = (event) => {
```
EjPlatzer, line 293: One reason you might be having issues is because I don't think you bound this function anywhere. Because of some complicated stuff that gets into the worst parts of Javascript, methods in a React class component have to be bound to the class component. See lines 42-44 above for examples.

EjPlatzer (commented):
Added some comments that should hopefully point you in the right direction.

src/views/Events/index.js
```diff
@@ -10,15 +10,49 @@ import gordonEvent from './../../services/event';
 import EventList from '../../components/EventList';
 import GordonLoader from '../../components/Loader';
 import { gordonColors } from './../../theme';
+import clsx from 'clsx';
+import { makeStyles, useTheme } from '@material-ui/core/styles';
```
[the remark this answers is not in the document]
aidanperez2000: https://github.com/gordon-cs/gordon-360-ui/pull/1015/commits/6b0e68c082b6bd513b32c53f731740839fe4ffc2

src/views/Events/index.js
```diff
@@ -254,68 +288,63 @@ export default class Events extends Component {
...
-              control={
-                <Checkbox
-                  checked={this.state.chapelCredits}
-                  onChange={this.filterEvents('chapelCredits')}
-                />
-              }
-              label="CL&amp;W Credit"
-            />
+            <FormControl className={formControl}>
+              <InputLabel id = "event-checkboxes-label">Events</InputLabel>
```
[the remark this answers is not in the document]
aidanperez2000: https://github.com/gordon-cs/gordon-360-ui/pull/1015/commits/f7b09c3f2e3eb225d81114ccb90bb1da1a5b17f1

src/views/Events/index.js
```diff
@@ -254,68 +288,63 @@ export default class Events extends Component {
...
+        this.filterEvents("art");
+      } else if (event.target.value == "Athletics") {
+        this.filterEvents("sports");
+      } else if (event.target.value == "CEC") {
+        this.filterEvents("cec");
+      } else if (event.target.value == "Chapel Office") {
+        this.filterEvents("chapelOffice");
+      } else if (event.target.value == "Student Life") {
+        this.filterEvents("studentLife");
+      }
```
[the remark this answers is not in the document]
aidanperez2000: https://github.com/gordon-cs/gordon-360-ui/pull/1015/commits/bc2414ca16357c434e000e6664a67820df16f5ab

src/views/Events/index.js
aidanperez2000: https://github.com/gordon-cs/gordon-360-ui/pull/1015/commits/0990cff23f671494e1aa8835dfacbca3d2b2b1d6

src/views/Events/index.js
aidanperez2000: https://github.com/gordon-cs/gordon-360-ui/pull/1015/commits/0990cff23f671494e1aa8835dfacbca3d2b2b1d6

src/services/event.js
```diff
@@ -206,29 +206,28 @@ const getFutureEvents = (allEvents) => {
...
  */
-const getFilteredEvents = (filters) => {
-  const allEvents = filters.events;
+const getFilteredEvents = (filters, allEvents, search) => {
   let filteredEvents = [];
   let shownEvents = [];
   filteredEvents = filterbyCategory(filters, allEvents);
   if (filteredEvents === null) {
     filteredEvents = allEvents;
   }
```
bennettforkner, line 215: I may put this logic in the actual **filterByCategory** function. Then you can run with the assumption that you will always be working with an array as the return value.

EjPlatzer (changes requested):
This is great Aidan! Thanks for improving this UI. I think we need to polish the design a little bit more before it's ready to merge though.

Firstly, the `Select` component is currently too small and positioned weirdly. The `Include Past` checkbox overlaps it when it's empty:
image
Also, I think the label for the select should be `Filters`, not events, since you're selecting which filters to apply to the search:
image

Additionally, they are misaligned vertically. Notice how the select box is sitting flush against the events list below it but the check box is floating above it.
image
The select box and the checkbox should be on the same baseline so they look consistent.

Also, the `Chip`s for each filter should have a darker background so they stand out more clealry. 

Finally, I think it's worth considering whether we still need the dropdown for filters. When we had >10 individual check boxes, it made sense to keep them hidden unless the user wanted to use them. Now, though, we have one checkbox and a small Select that's empty by default. I think it would make sense to get rid of the `Filters` button and the dropdown, and instead put the `Attended Events` button, the `Filters` select, and the `Include Past` check box on the same line beneath the search bar.

src/views/Events/index.js
```diff
@@ -1,393 +1,238 @@
...
+      background: gordonColors.primary.cyan,
+      color: 'white',
+
+      attendedEvents: {
+        background: gordonColors.primary.cyan,
+        color: 'white',
+        marginLeft: '0.88rem',
+      },
+    },
+  };
```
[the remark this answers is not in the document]
aidanperez2000: In most of these, we have the button classes set in the javascript files.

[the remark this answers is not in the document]
aidanperez2000: https://github.com/gordon-cs/gordon-360-ui/pull/1015/commits/b4510e599b8021a9538d736d86bf684610a342e8

src/services/event.js
aidanperez2000: https://github.com/gordon-cs/gordon-360-ui/pull/1015/commits/d6fb428c5d788b0d39d7e59c72ea99d6434c12ac

Kindlewing (changes requested):
This is good! Nice work! One thing I would recommend, is putting an 'x'  on the filter itself to delete it rather than having the list of all the filters to click. I don't know how feasible that is, but it might be something to look into.

EjPlatzer:
> This is good! Nice work! One thing I would recommend, is putting an 'x' on the filter itself to delete it rather than having the list of all the filters to click. I don't know how feasible that is, but it might be something to look into.

It's a good idea, but it's actually impossible. We're using a `Select` component, which allows you to (unsurprisingly) select from a list of options. So when you click on the component, you open the list of valid options, and you can click the specific options to enable or disable them. What you're suggesting would be adding a delete button to the individual Chips within the select, but that would mean that when you interact with the select component, you open the list _unless_ you click specifically on one of those delete buttons. That is a very specific kind of interaction logic that would basically have to be recreated from scratch (which is more effort than it's worth), but it also is an anti-pattern with the standards for Selects. Imagine you were on a small mobile screen and you had selected some filters and then wanted to add another one. To do this, you would have to tap within the Select but not on one of the delete buttons. That would result in a really poor experience for a small but important subset of our users.

PR #1017, opened 2021-02-23
Disable save button unless there are unsaved changes

NerdyGriffin:


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -278,6 +281,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
       // The method is separated from callback because the housing API service must be handled inside an async method
       changeApplicationEditor(applicationID, newEditorProfile.AD_Username);
       handleCloseOkay();
+      setUnsavedChanges(true);
```
joshuaRogers24, line 284: Is it necessary to set it here when it is also set in changeApplicationEditor?

joshuaRogers24 (changes requested):
Nice! Just note that one comment.

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: That is actually a good point, especially because it would cause it to mark it as unsaved changes even if the `changeApplicationEditor` fails

NerdyGriffin: This change has now been committed

joshuaRogers24 (approved):
Nice.

PR #1018, opened 2021-02-24
Replacing favicons containing a white bg with those having a transpar…

josh-peters-2357:
…ent bg

I made a copy of the old favicons under old-images that can be deleted once we're ready to merge. Could I get feedback on whether this was the intended solution to the issue @EjPlatzer @cpabbot ? By the way, the new favicons are used on my laptop on firefox and phone on firefox, but I don't see it changing for my iphone safari even though I cleared the site data for the local network host.


josh-peters-2357:
@cpabbot @EjPlatzer Since no one commented on this in time, now there are too many conflicts to bother with resolving them. I'll bring it up in meeting tomorrow and we can restart fresh from development, which at this point will be the fastest work.

PR #1021, opened 2021-02-26
ApartApp - Add description and link to design doc

NerdyGriffin:
- [ ] make use of the markdown [link name](#section-name) syntax for referencing the other sections within the same README


README.md
```diff
@@ -406,3 +406,5 @@ In the unusual case that Train or Production have not been automatically deploye
 
 - The student timesheets project was done as a senior project by a trio of seniors Nathanial Rudenberg, Adam Princiotta, and Jacob Bradley. The handoff documentation for this project was a revision and update to the existing documentation which is in the [Connect Local Backend to React](#connect-local-backend-to-react) section of the README. The design document can be found [here](https://docs.google.com/document/d/1CepyCiMzBXQVM--JwKKstniU_H1TodzxHLuCqcHxLjk/edit?usp=sharing)
 - [Student Timesheets Final Presentation](https://docs.google.com/presentation/d/162V-DLuaEUyHDS2Diu09k5f4Tpo2iuoMtIDUyowa7eQ/edit?usp=sharing)
+
+- The apartment applications project ~was done as~ is a senior project by Josh Rogers, Christian Kunis, Gahngnin Kim, and Nick Noormand. The handoff documentation has not been created yet. The design document for the UI can be found [here](https://docs.google.com/document/d/16gvjNApyNMJbqjnwv2DSP0EvY4JJdjVZZAiST3MAZEo/edit?usp=sharing)
```
joshuaRogers24, line 410: Do you have the "was done as" there so that we can change it to that later? Also, We can get rid of the note about the handoff documentation, right? Or do you want to leave in case we add something?

joshuaRogers24 (changes requested):
A quick question.

README.md
NerdyGriffin: To answer the first question: Yes, its for end of semester
To answer the second question: We should replace it with something that says where to actually look for that stuff now. Which could be as simple as saying "look at the "getting started" at the top of this README, or whatever sections we need to refer to. Should make use of the markdown `[link name](#section-name)` syntax for referencing the other sections within the same README

joshuaRogers24: Oh, yeah, good point

README.md
```diff
@@ -415,3 +415,5 @@ In the unusual case that Train or Production have not been automatically deploye
...
```
joshuaRogers24, line 419:
```suggestion
- The apartment applications project ~was done as~ is a senior project by Josh Rogers, Christian Kunis, Gahngnin Kim, and Nick Noormand. The design document for the UI can be found [here](https://docs.google.com/document/d/16gvjNApyNMJbqjnwv2DSP0EvY4JJdjVZZAiST3MAZEo/edit?usp=sharing)
```

joshuaRogers24 (changes requested):
See suggestion.

joshuaRogers24 (approved):
Thanks!

PR #1022, opened 2021-02-26
Add network status hook

EjPlatzer:
This commit introduces the NetworkContext, which uses React's Context API to track the online/offline status of the user's network. The goal is to have network-aware component's subscribe to this context provider so that the network is tracked in only one place, rather than having eventListeners in every component that needs to know the network status.

This may result in minimal performace gains (having only one event listener attached to the window instead of >10 at once), but I would guess that the rerendering takes significantly more processing than the eventlisteners themselves. The real advantage to this technique is the separation of concerns. Now, the NetworkContextProvider is the only component responsible for the logic of tracking network changes. Every ther network-aware component simply subscribes to these changes with the very minmial hook useNetworkStatus - e.g. const isOnline = useNetworkStatus(), rather than having a large block of logic duplicated across many components with a useEffect hook and window listeners that obscures the component's more natural responsibilities.

This first PR merely introduces the hook. Subsequent PRs will include refactors of network-aware components to make optimal use of the hook.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -7,12 +7,10 @@ import MomentUtils from '@date-io/moment';
 import './app.css';
 import analytics from './services/analytics';
 import { isAuthenticated } from './services/auth';
-//import GordonError from './components/Error';
+import NetworkContextProvider from './contexts/NetworkContext';
 import GordonHeader from './components/Header';
 import GordonNav from './components/Nav';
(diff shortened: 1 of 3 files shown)


joshuaRogers24 (approved):
Good work! I am looking forward to seeing all the redundant network code replaced.

PR #1023, opened 2021-03-02
Refactor memberships list

EjPlatzer:
The Involvements component was designed to display a list of a user's memberships, either on the MyProfile view or on the Profile view. It accepted props for the array of memberships to display and whether it was displaying for MyProfile or Profile. It would use the (poorly named) MyProfileActivityList and ProfileActivityList components (which were actually cards for one specific membership).

These have all been deeply refactored, reducing the extensive code duplication and over-engineering. The result is much shorter, much simpler, much easier to read, and much less bug-prone.


src/components/MembershipsList/index.js
```diff
@@ -0,0 +1,127 @@
...
+
+  useEffect(() => {
+    async function loadMemberships() {
+      setLoading(true);
+      if (myProf) {
+        const myMemberships = await userService.getMembershipsAlphabetically(user);
+        await Promise.all(
+          myMemberships.map(async (membership) => {
+            const involvement = await activity.get(membership.ActivityCode);
+            membership.IsInvolvementPrivate = involvement.Privacy;
```
joshuaRogers24, line 31: Was this occurrence of the term "involvement" difficult to change to membership, or did you think the term actually makes sense here?

src/components/MembershipsList/index.js
```diff
@@ -0,0 +1,127 @@
...
+        const publicMemberships = await userService.getPublicMemberships(user);
+        setMemberships(publicMemberships);
+      }
+      setLoading(false);
+    }
+    loadMemberships();
+  }, [myProf, user]);
+
+  const MembershipsList = () => {
+    if (memberships.length === 0) {
```
joshuaRogers24, line 45: Wow, nice simplification here by removing the redundant code that gets public memberships.

src/components/MembershipsList/index.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+  };
+
+  const toggleMembershipPrivacy = async (membership) => {
+    try {
+      await membershipService.toggleMembershipPrivacy(membership);
+      createSnackbar(membership.Privacy ? 'Membership Shown' : 'Membership Hidden', 'success');
+      setMemberships(
+        memberships.map((m) => {
+          if (m.MembershipID === membership.MembershipID) m.Privacy = !m.Privacy;
```
joshuaRogers24, line 71: Do you think it is a little hard to see that ` m.Privacy = !m.Privacy` is the body of the if?

src/components/MembershipsList/index.js
joshuaRogers24: Actually, another question about this: does `involvement.Privacy` the same no matter the user? (Are there just some involvements that are set to private by default?) Or is `involvement.Privacy` different depending on the user?

src/components/MembershipsList/components/MembershipInfoCard/index.js
```diff
@@ -0,0 +1,106 @@
...
+import { Link } from 'react-router-dom';
+import LockIcon from '@material-ui/icons/Lock';
+import useNetworkStatus from '../../../../hooks/useNetworkStatus';
+import './index.css';
+
+const MembershipInfoCard = ({ myProf, membership, onTogglePrivacy }) => {
+  const isOnline = useNetworkStatus();
+
+  const OnlineOnlyLink = ({ children }) => {
+    const linkClass = classnames({
```
joshuaRogers24, line 13: This style class is a neat, helpful strategy!

src/components/MembershipsList/components/MembershipInfoCard/index.scss
```diff
@@ -42,7 +54,7 @@
         border-left: 2px solid $primary-blue;
       }
     }
-    &-info-two {
+    &-picture {
       max-width: 100% !important;
```
joshuaRogers24, line 58: Is this classname used?

src/components/MembershipsList/index.js
EjPlatzer: So you're suggesting 
```suggestion
          if (m.MembershipID === membership.MembershipID) {
            m.Privacy = !m.Privacy;
          };
```
might be clearer? I personally find them equally clear but if you feel differently then I'm happy to change it.

EjPlatzer: Actually after looking at it in context, I'm inclined to agree that brackets clarify the code. I'll update it.

joshuaRogers24 (approved):
Thanks for those changes. Looks good!

PR #1024, opened 2021-03-03
Use automated tool to update MUI imports

EjPlatzer:
The Material UI library allows you to import component as a default from their particular sub-folder, OR as a named import from the core folder. e.g. `import Component from '@material-ui/core/Component'` vs `import { Component } from '@material-ui/core'`.

The former, default import used to be standard in older versions of the material-ui library because it imported exlcusively that component's code, whereas the latter named import syntax would actually import the entire library.

This is *no longer true*, as of webpack version 4.x (our react-scripts dep currently runs webpack v4.3). The named imports are properly handled and only import the necessary code.

Furthermore, the named imports have the advantage of being more concise and easier to add to/remove from/maintain when multiple components are imported. e.g.
```
import { Component1, Component2, Component3 } from '@material-ui/core'
```

vs.

```
import Component1 from '@material-ui/core/Component1';
import Component2 from '@material-ui/core/Component2';
import Component3 from '@material-ui/core/Component3';
```

This pull request uses the [`codemod`](https://github.com/mui-org/material-ui/blob/master/packages/material-ui-codemod/README.md#top-level-imports) tool provided by the Material-ui team to automatically switch all default imports to named imports.

### DIFF
--- a/src/components/Error/index.js
+++ b/src/components/Error/index.js
@@ -1,9 +1,9 @@
-import Grid from '@material-ui/core/Grid';
-import Typography from '@material-ui/core/Typography';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 
 import './error.css';
+
+import { Grid, Typography } from '@material-ui/core';
 
 export default class GordonError extends Component {
   render() {
--- a/src/components/AdvisorList/index.js
+++ b/src/components/AdvisorList/index.js
@@ -1,8 +1,5 @@
-import Divider from '@material-ui/core/Divider';
 import React, { Component } from 'react';
-import ListItem from '@material-ui/core/ListItem';
-import Typography from '@material-ui/core/Typography';
-import Grid from '@material-ui/core/Grid';
+import { Divider, ListItem, Typography, Grid } from '@material-ui/core';
 
 export default class Advisors extends Component {
   constructor(props) {
--- a/src/components/EventList/components/CollapsableEventItem/index.js
+++ b/src/components/EventList/components/CollapsableEventItem/index.js
@@ -1,10 +1,8 @@
 import PropTypes from 'prop-types';
-import Typography from '@material-ui/core/Typography';
 import React, { Component } from 'react';
-import CardContent from '@material-ui/core/CardContent';
-import Collapse from '@material-ui/core/Collapse';
-import Grid from '@material-ui/core/Grid';
 import './collapsable-event-item.css';
+
+import { Typography, CardContent, Collapse, Grid } from '@material-ui/core';
 
 //Switched to table rows
 export default class GordonCollapsableEventItem extends Component {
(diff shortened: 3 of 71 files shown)


src/components/Error/index.js
```diff
@@ -1,10 +1,10 @@
-import Grid from '@material-ui/core/Grid';
-import Typography from '@material-ui/core/Typography';
 import PropTypes from 'prop-types';
```
josh-peters-2357, line 1: What's the difference between `import PropTypes from 'prop-types';` and `import { PropTypes } from 'prop-types';`? I see a similar thing on the next line, where React is not in braces.

josh-peters-2357:
Hey @EjPlatzer, what is the interaction between the [codemod](https://github.com/mui-org/material-ui/blob/master/packages/material-ui-codemod/README.md#top-level-imports) tool and Prettier, our code formatting tool? The changes all look fine to me; have you just used Material's tool on our files and then Prettier will do its own work after you save the files in your editor? If so, I think all is well.

src/components/Error/index.js
EjPlatzer: Good question! It's a counterpart to the two ways  things can be exported. The ES6 import/export syntax allows you to export code from a file in two ways: `export variable` and `export default variable`.

The former can be repeated multiple times, so you can export many things from one file. The latter, default export can only be done once in a file. The default export allows you to set some default value for the file that is exported, unsurprisingly, _by default_. 

When you call `import Component from 'moduleName/Path/To/Component'`, you're importing whatever is exported by default from that file. The other option, `import { Component } from 'moduleName'`, you're importing a specific (non-default) export from that file. The biggest difference is that the default export can be named anything and can be imported by any name, e.g. `import ARBITRARY_NAME_THAT_COULD_BE_ANYTHING from 'moduleName/Path/To/Component'`. The non-default imports require that the name of the imported variable match the name of the exported variable. So if you have `export const SpecificComponentName = ...`, then you have to have `import { SpecificComponentName } from 'moduleName'`. 

In this specific case, the developers of the material-ui library have supported both ways (default and named) for importing their components. To use the default import, you import the exact component you want from a file named for that component (e.g. `import Grid from '@material-ui/core/Grid'`). They only later started supporting named imports where you can import whichever set of components you need from the top-level folder (e.g. `import { Grid, Typography } from '@material-ui/core'`). This used to have some performance implications because of what was essentially a bug that meant the entire library was being imported, but that is no longer the case.

EjPlatzer:
> Hey @EjPlatzer, what is the interaction between the [codemod](https://github.com/mui-org/material-ui/blob/master/packages/material-ui-codemod/README.md#top-level-imports) tool and Prettier, our code formatting tool? The changes all look fine to me; have you just used Material's tool on our files and then Prettier will do its own work after you save the files in your editor? If so, I think all is well.

That's a great question! We have the repo configured to use Prettier as a pre-commit hook, which means that when you stage files for a commit, they are linted by prettier (with some rare edge-case exceptions). Also, the `codemod` tool happened to output imports in a format that prettier has no complaints about, so it doesn't seem to matter in this case.

josh-peters-2357 (approved):
Everything looks good to me! Thanks for the answers to my two questions as well.

PR #1027, opened 2021-03-08
RELEASE: timesheets bug fix, Involvements padding fix, Color and Import Refactorings, README updates

EjPlatzer:
This release includes three functional changes, as well as a significant amount of refactoring.

The functional changes are:

1. Making use of the simplified timesheets endpoint for getting jobs and changing how dates are formatted to fix the existing timezone bugs.
2. Fixing the padding for membership requests and involvement cards on the involvement page so that both wrap properly and sit inside their containing card.
3. A minor bug fix to ensure that the "Include Alumni" checkbox for people search will not display for anyone who has 'stu' in their personType, rather than excluding people whose personType was exactly 'stu'

The refactoring, apart from README updates, comes in two parts:

1. Updating the Material-UI imports to use named imports universally, since the bundle size issues are now fixed and named imports are simpler to understand and maintain and at least as concise but often more so than default imports.
2. Unifying grayscale colors across all our .scss files to make use of the global variables in _vars.scss. This means that our colors are all defined in one place. This means that we no longer have several tiny variations on the same color in different places, and also that if we want to change a color in the future, it can be done from one place, rather than having to hunt down every redundant declaration of the same color.


README.md
```diff
@@ -19,28 +19,37 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
...
   - Extract the file and run it.
-  - The installer will ask you where to install NVM. It will display the path where it is currently set to install. Ensure that the path reflects the account you are logged into (example: if you're logged in as anthony.aardvark, make sure the path looks like `C:\Users\aanthony.aardvard\Program Files\etc`). If it says Node is already installed, proceed anyway.
-  - After the installation completes, open a terminal and run these commands in order:
-    - `nvm install 9.11.1`
-    - `nvm use 9.11.1`
+  - The installer will ask you where to install nvm. It will display the path where it is currently set to install. Ensure that the path reflects the account you are logged into (example: if you're logged in as anthony.aardvark, make sure the path looks like `C:\Users\aanthony.aardvard\Program Files\etc`). If it says Node is already installed, proceed anyway.
+  - After this, go to https://nodejs.org/en/ and look for the version labeled "lts" (which indicates the latest stable version). 
+  - Finally, run `nvm install <version>` where <version> is the version you found.
```
joshuaRogers24, line 39:
```suggestion
  - Finally, run `nvm install <version>` where `<version>` is the version you found.
```
< and > by themselves actually make whatever is inside invisible

README.md
```diff
@@ -19,28 +19,37 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
...
+- Once it is successfully installed, use it to install a version of Node.js: `nvm install node`.
+- Then, tell nvm that you want to use that verion: `nvm use node`. 
+To use a specific Node version - usually the latest Long Term Support (LTS) version - replace `node` in the above commands with the desired version or alias, e.g. `nvm install 14.15.4` for version 14.15.4 or `nvm install --lts` for the latest LTS.
+
+#### Troubleshooting NVM
+
+- If a developer accidentally follows the above Windows instructions on CS-RDSH-02, the environment variable for NVM will be set by their installation to a path within their user directory. Thus, many if not all other users on the machine will lose access to NVM. To fix this, any user can open Powershell as administrator and run `choco install -y nvm`. (The `-y` option answers `yes` to any prompts that occur during the installation.) Then, in a Powershell terminal not running as admin, run `nvm install --lts`. Users should log out and back in to see the fix take effect. As usual, you can check if this worked by opening a terminal and running the command `nvm`. If the output says `Running version...`, then it is all set.
```
joshuaRogers24, line 50:
```suggestion
- If a developer accidentally follows the above Windows instructions on CS-RDSH-02, the environment variable for nvm will be set by their installation to a path within their user directory. Thus, many if not all other users on the machine will lose access to nvm. To fix this, any user can open Powershell as administrator and run `choco install -y nvm`. (The `-y` option answers `yes` to any prompts that occur during the installation.) Then, in a Powershell terminal not running as admin, run `nvm install --lts`. Users should log out and back in to see the fix take effect. As usual, you can check if this worked by opening a terminal and running the command `nvm`. If the output says `Running version...`, then it is all set.
```
Keeping it consistent

src/_vars.scss
```diff
@@ -15,10 +15,26 @@ $secondary-yellow: #fdb913;
...
-$neutral-gray: #cccccb;
+$neutral-light-gray2: #dddddd;
+$neutral-gray: #808080;
+$neutral-gray2: #cccccb;
+$neutral-gray3: #adadad;
 $neutral-dark-gray: #31342b;
+$neutral-dark-gray2: #444444;
+$neutral-dark-gray3: #555555;
+
+$neutral-black: #000000;
```
NerdyGriffin, line 30: Thank you for updating this part, I had been wondering why there weren't more "global" theming variables for these commonly used style variations. I will make a note to incorporated these new theme options into the ApartApp code wherever it is appropriate.

package.json
```diff
@@ -4,66 +4,66 @@
...
+    "downshift": "^4.1.0",
     "history": "^4.10.1",
     "http-proxy-middleware": "^0.20.0",
-    "interweave": "^12.1.1",
-    "jwt-decode": ">=2.2.0",
-    "lodash": ">=4.17.11",
-    "luxon": "^1.21.3",
-    "moment": ">=2.24.0",
+    "interweave": "^12.7.0",
+    "jwt-decode": "^3.1.2",
```
joshuaRogers24, line 21: I just want to point out the change in major version here and want to make sure there were no breaking changes...

package.json
```diff
@@ -4,66 +4,66 @@
...
+    "moment": "^2.29.1",
     "prop-types": ">=15.6.0",
-    "react": "^16.12.0",
-    "react-big-calendar": ">=0.21.0",
-    "react-chartjs-2": ">=2.7.4",
-    "react-cropper": ">=1.0.1",
+    "react": "^16.14.0",
+    "react-big-calendar": "^0.32.0",
+    "react-chartjs-2": "^2.11.1",
+    "react-cropper": "^2.1.7",
```
joshuaRogers24, line 29: Also a change in major version here. We might want to test that ID uploader still works fine.

package.json
```diff
@@ -4,66 +4,66 @@
...
-    "react-dropzone": ">=10.1.5",
-    "react-ga": ">=2.3.5",
-    "react-graceful-image": ">=1.2.9",
-    "react-icons": ">=2.2.7",
-    "react-image-gallery": "^1.0.5",
+    "react-dom": "^17.0.1",
+    "react-dropzone": "^11.3.1",
+    "react-ga": "^3.3.0",
+    "react-graceful-image": "^1.5.0",
+    "react-icons": "^4.2.0",
```
joshuaRogers24, line 35: Four ones here also, not sure what react-dom and react-ga are used for, but the others look like they could affect ID Uploader and our icons

package.json
```diff
@@ -4,66 +4,66 @@
...
+    "react-image-gallery": "^1.0.9",
     "react-media": ">=1.8.0",
-    "react-responsive-carousel": "^3.1.51",
+    "react-responsive-carousel": "^3.2.12",
     "react-router": "~5.0.1",
     "react-router-dom": "~5.0.1",
-    "react-timeago": ">=4.4.0",
-    "typescript": ">=3.5.3",
+    "react-timeago": "^5.2.0",
+    "typescript": "^4.2.2",
```
joshuaRogers24, line 42: And these 2 had major updates

package.json
```diff
@@ -4,66 +4,66 @@
...
-    "eslint-plugin-import": "^2.20.0",
-    "eslint-plugin-jsx-a11y": ">=5.1.1",
-    "eslint-plugin-react": "^7.18.0",
-    "husky": "^4.2.3",
-    "jasmine-core": "^3.5.0",
-    "jsdoc": ">=3.6.2",
+    "eslint-config-react-app": "^5.2.1",
+    "eslint-plugin-flowtype": "^4.7.0",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
```
joshuaRogers24, line 49: Another one

package.json
```diff
@@ -4,66 +4,66 @@
...
     "lint-staged": "^8.0.0",
-    "node-sass-chokidar": "^1.4.0",
+    "node-sass-chokidar": "^1.5.0",
     "npm-run-all": ">=4.1.2",
-    "prettier": ">=1.9.2",
-    "react-scripts": "^3.3.0",
-    "stylelint": ">=10.1.0",
+    "prettier": "^2.2.1",
+    "react-scripts": "^3.4.4",
+    "stylelint": "^13.11.0",
```
joshuaRogers24, line 64: Another two

package.json
EjPlatzer: The only breaking change from v2 -> v3 for jwt-decode was that a deprecaterd way of importing the package (`import * as jwt_decode from 'jwt-decode'`) has been removed. We were already using the preferred import syntax in the only place we use this library, namely in [`user.js`](https://github.com/gordon-cs/gordon-360-ui/blob/develop/src/services/user.js#L7), so this should not cause any problems. Considering that the decoding is still successful (i.e. the [getLocalInfo()](https://github.com/gordon-cs/gordon-360-ui/blob/develop/src/services/user.js#L362) method and all it's uses), this is a clean update.

package.json
EjPlatzer: `react-dom` is part of the core React functionality. React v17.x (and by extension `react-dom`) contains virtually no breaking changes, it was a purely internal update from them. You can read more about it in [the React team's official release announcement](https://reactjs.org/blog/2020/10/20/react-v17.html).

EjPlatzer: `react-ga` is used for Google Analytics. The major update was simply to indicate that they are following semver strictly now, and contained no breaking changes.

package.json
EjPlatzer: Timeago is used in our SchedulePanel, which was already broken and needs a deep restructuring. We do not use TypeScript so the update won't affect us.

package.json
EjPlatzer: This is a devDependency so it's guaranteed to not break our code unless it affects our build. eslint plugins specifically only affect what warnings eslint will provide when linting our code for correctness.

package.json
EjPlatzer: These are also devDependencies and cannot result in breaking changes.

README.md
EjPlatzer: The README changes are coming from `develop`, which is the version actually shown in our documentation because `develop` is our default branch. Any changes should be made into a separate issue/pull request into develop.

README.md
EjPlatzer: See the above comment about README changes on a release.

package.json
EjPlatzer: `react-icons` updated by two major versions. [v2.x->v3.x](https://github.com/react-icons/react-icons#migrating-from-version-2---3) had one breaking change which was changing the import path. We already use the new import path. 

[v3.x->v4.x](https://github.com/react-icons/react-icons/releases/tag/v4.0.0) had no breaking changes, but rather introduced a new package for projects configured differently from ours, which will not affect us. 

Therefore the `react-icons` update is clean.

src/services/jobs.js
```diff
@@ -16,14 +21,20 @@ const getStaffPageForUser = async () => {
...
  * Get active jobs for current user
  * @param {Boolean} canUseStaff Whether user can use staff timesheets
- * @param {String} details The user's details
+ * @param {Date} shiftStart The start of the shift to get jobs for
+ * @param {Date} shiftEnd The end of the shift to get jobs for
  * @return {Promise.<String>} User's active jobs
  */
-const getActiveJobsForUser = (canUseStaff, details) => {
+const getJobs = (canUseStaff, shiftStart, shiftEnd) => {
+  const urlParams = `?shiftStart=${dateFormat.format(shiftStart)}&shiftEnd=${dateFormat.format(
```
joshuaRogers24, line 29: With the apartment application page, we handle multi-parameter route calls this way:  In the API, we create a view model that contains a class variable for each parameter and replace all parameters in the controller with one of the created view model. Then, the frontend just needs to pass a JSON string with a similar format to the API view model.
I do not know if this is common across the whole API, but it may be better and more readable than this URL syntax.

src/views/Home/components/NewsCard/index.js
```diff
@@ -5,12 +5,8 @@
 // May be a potential future feature, but not sure
 
 import React, { Component } from 'react';
-import Grid from '@material-ui/core/Grid';
-import CardHeader from '@material-ui/core/CardHeader';
-import Card from '@material-ui/core/Card';
-import Typography from '@material-ui/core/Typography';
 import { CardContent } from '@material-ui/core';
```
joshuaRogers24, line 8:
```suggestion
```

src/views/Home/components/NewsCard/index.js
```diff
@@ -5,12 +5,8 @@
...
-import { Button } from '@material-ui/core';
+import { Button, Grid, CardHeader, Card, Typography } from '@material-ui/core';
```
joshuaRogers24, line 9:
```suggestion
import { Button, Grid, CardHeader, Card, CardContent, Typography } from '@material-ui/core';
```

src/services/jobs.js
EjPlatzer: I agree that this interpolated string is not the clearest thing in the world, but I think it is readable and the query parameter names provide enough context to make the rest fairly obvious. The main reason we can't use a View Model and send a complex type in the reuqest is that this is a GET method, which means it should never have a body (or more specifically, by HTTP specification, any body in a GET request should be ignored). See the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) for reference. The right way to send request parameters in a GET method is using query parameters, which is exactly what is being done here. This is the standard and accepted way of interacting with a REST API.

The old, deprecated route for getting a user's jobs (`HTTP POST /api/jobs/getJobs`) included a `details` object in the parameter. However, as you can see this was perhaps even more confusing and broke REST conventions by using a POST method to request information without creating any information, and also it had a DIFFERENT HTTP method specified in the URL route, which was both unconventional and confusingly contradictory. I'm inclined to believe the revised approach using query parameters is clearer overall.

joshuaRogers24 (changes requested):
All great refactoring and additions! Just a couple of small questions and suggestions.

src/services/jobs.js
joshuaRogers24: I see! Definitely keep it that way then.

src/views/Home/components/NewsCard/index.js
EjPlatzer: I will refactor these in a new pr to develop. Since this is purely stylistic though, I don't think it should hold up our push to master.

joshuaRogers24 (approved):
Looks good!

PR #1029, opened 2021-03-09
Fix npm vulnerabilities

EjPlatzer:
We had several packages which were outdated/unmaintained and were introducing vulnerabilities into our node_modules dependency tree. I have upgraded the culprits as noted below and we are now vulnerability free (for the time being). Note that the `package-lock.json` has also been updated to use lockfile version, which is now standard for npm v7.x

React-scripts was updated to version 4. The old version depended on a version of http-proxy that had vulnerabilities. The only breaking change was the way that svg imports is handled by our bundler. SVGs can no longer be included as relative imports in an `img` tag's `src` attribute, as was being done in the VictoryPromiseDisplay component on the MyProfile view. The new way to use svgs is to import them as React components (which is handled automagically by SVGR via react-scripts' bundler). Then they can be added to the component tree in the normal way components are used in react. Note that this revealed a pre-existing issue in our images, namely that all of our svgs had `clipPath` tags whose `id` attirbute was identical. Every element in our DOM should have a unqiue `id` attribute. In this case, it was causing all the effected svgs to use the first one's clipPath value, which meant the others were not drawing beyond the width of the first tag (i.e. getting cut off horizontally). This has been fixed so that each svg has a nuique `id` attribute on it's `clipPath` tag, and they now draw properly. 

The unused testing dependencies (Jasmine, Karma, etc.) were removed because they are currently totally unused and karma had vulneable dependencies. Once we decide on the best testing solution in the future and begin work on actually implementing tests, restoring these dependencies is trivial.

The url-regex package had a vulnerability which didn't affect our site but was polluting the results of npm audit. That package is no longer maintained, but url-regex-safe is an update that fixes the vulnerability and improves the package's API.

http-proxy-middleware was updated because it's old version depended on a version of http-proxy which had high severity vulnerabilities. The new http-proxy-middleware has a slightly changed API that requires us to call it slightly differently. setupProxy.js has been adjusted accordingly.

react was updated to fit the rest of the dependencies properly. The major version 17 has virtually no breaking changes and no new features. It is mostly an internal update preparing for future changes in coming major versions.

date-fns and @date-io/date-fns were updated as well.


bennettforkner (approved):
Glad to see the move towards using SVGs instead of IMGs went well after some difficulty. Thanks for updating React-scripts to V4 as well!

Ready for merge into develop.

PR #1030, opened 2021-03-13
Fixed hall issue caused by changed variable/prop name

NerdyGriffin:
Fixed incorrect variable name,
But the old bu has returned which causes  the hall dropdowns to not always update the displayed text right after the user makes a selection. That issue will require a much more thorough investigation later


joshuaRogers24 (approved):
Good catch!

Issue #919, opened 2020-11-13
ApartApp (5) Make student feature for entering the department of an applicant's off-campus program


joshuaRogers24:
The result of which is saved with the applicants list, associating  each listed department with the applicant for whom it was entered (maybe through a dictionary of sorts where 'applicant' is the key and 'department' is the value)

[Student Page Mockup for reference](https://docs.google.com/drawings/d/11ySYKYrAWUhx8zvqkH5X15JeJ2TDmo3dy-ih8VinVV0/edit?usp=sharing)

- [x] Create a branch off of one of the existing branches that already contains starter-code for this feature
- [x] Change the `applicants` array from an array of `StudentProfileInfo` to an array of JSON Objects, { StudentProfile: ' ', OffCampusProgDepartment: ' ' }
- [ ] Create custom List and ListItem components for displaying the off-campus prog info (much like the implementation of ApplicantList and HallList)
- [ ] In off-campus prog card, each list item should have the name of the student and a dropdown menu of valid department names to choose from, or the option "None" (See HallListItem and/or the PeopleSearch page for examples of how to implement the dropdown)
- [ ] Add callback to update the state variables whenever a dropdown selection is changed
- [ ] Add support for this off-campus info to the "save" feature (this requires changes in `src/views/ApartmentApp/components/StudentApplication/index.js` and in `src/services/housing.js`
- [ ] Comments and documentation

PR #1031, opened 2021-03-14
Apartapp develop off campus

Nick-Noormand:
Have had issues with testing it but I think my code is alright. This pull request should be able to have base off-campus section functionality working.


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -673,7 +748,15 @@ const StudentApplication = ({ userProfile, authentication }) => {
                     <Card>
                       <CardHeader title="Off-Campus Work Study" className="card-header" />
                       <CardContent>
-                        <Typography variant="body1">Placeholder text</Typography>
+                        <offCampusSection
+                          offCampusApplicantList={offCampusApplicants}
+                          availableApplicants={applicants}
```
NerdyGriffin, line 751: You are using this `offCampusApplicantList` to do exactly what the `applicants` state variable was designed to do. There is already support for this data structure built into the frontend and the backend, so why create another array variable to hold that same information? 
We should be thinking of the applicants as being almost like a key-value pair, where the StudentProfileInfo object is the unique key and the name of the off campus program department is one of the values

src/views/ApartmentApp/components/StudentApplication/components/OffCampusSection/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <Card>
+      <CardHeader title="Off Campus Program Applicants" className="card-header" />
+      <CardContent>
+        <Grid container justify="space-between" spacing={2}>
+          <Grid item xs={12}>
+            <List className="off-Campus-list" aria-label="apartment applicants off campus programs" disablePadding>
+              {offCampusApplicantList ? (
+                offCampusApplicantList.map((memberInfo, index) => (
```
NerdyGriffin, line 58: You could easily use an array map on the original applicants array rather than creating this additional array with duplicate information. Either way, you have to make sure that the state variable `applicants` in `StudentApplication/index.js` is being updated to include the off campus selection, which does not appear to be the case

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: If you make sure to merge in the latest changes from `apartapp-develop`, I think you should be able to check the `@typedef` in `services/housing.js` for the properties that have already been assigned to the objects in the `applicants` array

NerdyGriffin:
Since the off-campus entries need to be mapped one-to-one to each applicant, the most reliable way to implement this feature will be to match our original mock-up by having one row for each student already in the applicants array, rather than mimicking the already-buggy method used in the hall list. 
I would also highly recommend merging apartapp-develop into this branch soon, since there are many bug fixes downstream that are not yet merged into this branch

NerdyGriffin:
> Since the off-campus entries need to be mapped one-to-one to each applicant, the most reliable way to implement this feature will be to match our original mock-up by having one row for each student already in the applicants array, rather than mimicking the already-buggy method used in the hall list.
> I would also highly recommend merging apartapp-develop into this branch soon, since there are many bug fixes downstream that are not yet merged into this branch

If you open the console, you will see that you have a dependency loop creating a constant stream of error/warning messages from the list items.

src/views/ApartmentApp/components/StudentApplication/index.js
Nick-Noormand: Ohhhhh. That makes sense. So it should be less of having a separate list of off-campus applicants and more of checking that applicants list for which ones program values are being updated.

.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,16 @@
...
+    "\\src\\views\\ApartmentApp\\components",
+    "\\src\\views\\ApartmentApp\\components\\StudentApplication",
+    "\\src\\views\\ApartmentApp\\components\\StudentApplication\\components",
+    "\\src\\views\\ApartmentApp\\components\\StudentApplication\\components\\HallSelection",
+    "\\src\\views\\ApartmentApp\\components\\StudentApplication\\components\\HallSelection\\components",
+    "\\src\\views\\ApartmentApp\\components\\StudentApplication\\components\\HallSelection\\components\\HallListItem"
+  ],
+  "SelectedNode": "\\src\\views\\ApartmentApp\\components\\StudentApplication\\index.js",
+  "PreviewInSolutionExplorer": true
+}
```
NerdyGriffin, line 16: I looks like these files in .vs should be excluded from the repo, since they are local Visual Studio config files.
You can use something like `git update-index --skip-worktree ./.vs/` to make them locally ignored if you plan to continue using Visual Studio instead of VSCode, which might still require you to manually remove them via something like `git rm -r --cached ./.vs/`, which will remove them from the commit but should not actually delete the files from your computer

src/views/ApartmentApp/components/StudentApplication/components/OffCampusSection/components/ProgramListItem/index.js
```diff
@@ -0,0 +1,120 @@
...
+      setDepartmentValue(availableApplicants[index].major);
+    };
+
+    getMembersFromProps();
+  });
+
+  const previousInputs = useRef();
+  useEffect(() => {
+    previousInputs.current = [index, availableApplicants];
+  });
```
NerdyGriffin, line 43:
```suggestion
const ProgramListItem = ({
  disabled,
  index,
  applicantName,
  departmentName,
  availableDepartments,
  onOffCampusChanged,
}) => {
  const [applicantDepartmentValue, setDepartmentValue] = useState(''); // Major drop-down menu value

  useEffect(() => {
    setDepartmentValue(departmentName);
  }, [departmentName);
```
If you eliminate the dropdown of the applicants and just have it be a fixed map of the applicants already in the list, then you can use the applicant info directly from the props. Additionally, it is more reliable to pass the applicant and department values directly in props rather than trying to get them out of the applicants array, thus eliminating the complexity of trying the use the isEqual to perform a deep comparison, which was actually a flawed workaround that I tried to use before I figured out what was actually going wrong with the hall array

.gitignore
```diff
@@ -32,3 +32,4 @@ yarn-error.log*
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+.vs/slnx.sqlite
```
NerdyGriffin, line 35: Since VS is not recommended editor for this project, these .vs config files should not be included at all. I recommend removing this line and removing the .vs/ folder from the commit entirely
```suggestion
```

src/services/housing.js
```diff
@@ -189,42 +167,162 @@ const changeApartmentAppEditor = async (applicationID, newEditorUsername) => {
...
-  } catch (err) {
-    if (err?.status === 404 || err?.name?.includes('NotFound')) {
-      console.log(
-        'Received 404 indicates that the requested application was not found in the database',
-      );
-    } else {
-      throw err;
-    }
-    return null;
-  }
```
NerdyGriffin, line 203: Things like this indicate there are some pretty big problems with the recent merge, as it is trying to remove things that you did not even modify. This merge mistake will have to be fixed before I can continue the review, since it is very difficult to tell which things are meant to be here. Also merging it in this state would likely break most of ApartApp

NerdyGriffin (changes requested):
Merge problems may break the apartapp-develop build, I will attempt a rebase of the branch to see if that can fix it before I continue my review

src/views/ApartmentApp/components/StudentApplication/components/OffCampusSection/components/ProgramListItem/index.js
```diff
@@ -0,0 +1,68 @@
...
+  Grid,
+  Input,
+  InputLabel,
+  ListItem,
+  ListItemText,
+  MenuItem,
+  Select,
+} from '@material-ui/core';
+
+const ProgramListItem = ({
```
NerdyGriffin, line 14: Please use a more consistent naming scheme for these components, such as `OffCampusList` and `OffCampusListItem`, to match both the code style of the 360 project and to match the official terminology of the ApartApp project

src/views/ApartmentApp/components/StudentApplication/components/OffCampusSection/components/ProgramListItem/index.js
```diff
@@ -0,0 +1,68 @@
...
+  applicantProfile,
+  offCampusProgram,
+  departments,
+  onOffCampusInputChange,
+}) => {
+  const [applicantDepartmentValue, setDepartmentValue] = useState(''); // Department drop-down menu value
+
+  useEffect(() => {
+    setDepartmentValue(offCampusProgram ?? '');
+  }, [offCampusProgram]);
```
NerdyGriffin, line 26: This use of the state variable is not necessary, as you can simply pass the prop to the `Select` component directly. Evan gave a great explanation of this is another PR, but I do not know which one and I was unable to find it
```suggestion
```

src/views/ApartmentApp/components/StudentApplication/components/OffCampusSection/components/ProgramListItem/index.js
```diff
@@ -0,0 +1,68 @@
...
+        <Grid container alignItems="flex-end" spacing={1}>
+          <Grid item xs={12} sm={4}>
+            <ListItemText primary={applicantProfile.fullName} className={'list-item'} />
+          </Grid>
+          <Grid item xs={12} sm={8}>
+            <FormControl fullWidth>
+              <InputLabel>Department</InputLabel>
+              <Select
+                disabled={disabled}
+                value={applicantDepartmentValue}
```
NerdyGriffin, line 53: Here is where you use the prop instead of the state variable
```suggestion
                value={offCampusProgram}
```

src/views/ApartmentApp/components/StudentApplication/components/OffCampusSection/components/ProgramListItem/index.js
NerdyGriffin: You can also see the newest version of the Hall components, which have been greatly improved since the last time you worked on this feature

NerdyGriffin (changes requested):

NerdyGriffin:
I have to stop for food now, but I will come back and continue reviewing after dinner

NerdyGriffin:
Applicant name is no longer displayed on the list items 
image

NerdyGriffin:
> Applicant name is no longer displayed on the list items
> image

Nevermind, this problem is not your code, I think it is caused by an API change that hasn't been merged yet. Do not worry about this, it will be in the near future

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -345,6 +338,37 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+   * @param {Number} index The index of the applicant in the list
+   */
+  const handleOffCampusInputChange = (offCampusProgramValue, index) => {
+    if (index !== null && index >= 0) {
+      // Get the profile of the selected user
+      const applicant = applicants[index];
+      let newApplicant = {
+        ...applicant,
+        OffCampusProgram: offCampusProgramValue ?? applicant.OffCampusProgram,
+      };
```
NerdyGriffin, line 353:
```suggestion
      let newApplicant = {
        ...applicants[index],
        OffCampusProgram: offCampusProgramValue ?? applicant.OffCampusProgram,
      };
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -345,6 +338,37 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+      // Error checking on the hallNameValue before modifying the newHallInfo object
+      setApplicants((prevApplicants) =>
+        // replace the element at index with the new hall info object
+        prevApplicants.map((prevApplicant, j) => {
+          if (j === index) {
+            return newApplicant;
+          } else {
+            return prevApplicant;
+          }
+        }),
```
NerdyGriffin, line 363: You can simplify this using the inline conditional syntax
```suggestion
        prevApplicants.map((prevApplicant, j) =>
          j === index ? newApplicant : prevApplicant,
        ),
```

NerdyGriffin (changes requested):
Some more suggestions for simplification

NerdyGriffin (approved):
I fixed the problems that you re-introduced in your last merge, because you tried merging your local copy of the previously broken version with the working version from upstream, thus bringing back all errors from the first "revert merge" attempt. 

@Nick-Noormand Please avoid using the "revert" command to undo a merge in the future, because the revert command will actually revert all of the commits that were brought in by the merge, rather than simply "undo" the merge. This means that if you revert a merge commit, then when the pull request gets merged it would have also reverted all of those commits inside apartapp-develop. 

In other words, it will undo all of the work that other people had pushed into apartapp-develop branch, going all the way back to the last time your branch did a merge or, if there were no previous merges, to the time when your branch was created.

If you absolutely **_need_** to undo a merge, then you should uses `git reset --hard [commit you want to reset to]` to reset your local branch to the commit that you did just before the merge, followed by a `git push --force`. **This should only be done if you know that no one else has pushed anything to the same branch on Github**, because it will overwrite anything that is not already recorded in your local commit history.

NerdyGriffin:
Linked issue automation:
closes #919 
closes #1040

PR #1025, opened 2021-03-04
Color Consistency: Black, White, Gray

iamphg97:
Many places still use "RGBA" and "Hexadecimal" color standard.
The inconsistent colors were changed to standard color of Gordon (declared color on [_vars.scss](url))
At this point, I worked on black, white and gray color.

PR #1032, opened 2021-03-16
Color Consistency: All of chromatic color

iamphg97:
Last change: All of inconsistent white, black and gray colors(achromatic color).
(See Color Consistency: Black, White, Gray #1025)

Now, I have worked on all of inconsistent chromatic colors.
There's only few colors I need to change.

Thing I have changed:

1. src/_vars.scss
- Add $primary-red: #ff0000
- Delete $neutral-black2: #111111

2. src/components/PWAInstructions/index.scss
- #ce392e -> $secondary-red

BEFORE: #ce392e 
Select Device Type

AFTER: $secondary-red
Select Device Type 2

- #0078da -> $primary-cyan

BEFORE: #0078da
other Chrome-based browsers such as Brave but Google Chrome is

AFTER: $primary-cyan
other Chrome-based browsers such as Brave but Google Chrome is

- #014983 -> $primary-blue

  in _vars.scss $primary-blue is already declared (Hexadecimal Color #014983).

- #31342b -> $neutral-dark-gray

  in _vars.scss $neutral-dark-gray is already declared (Hexadecimal Color #31342b).

3. \src\views\ActivitiesAll\components\Requests\components\RequestsReceived\requests.scss

- #fa3e3e -> $primary-red

NOTE:
I realized that some features (such as rbc) were no longer used in \src\components\SchedulePanel\components\ScheduleCalendar\schedulecalendar.scss. I have already checked the file to import schedulecalendar.scss and the only file to import this file is \gordon-360-ui\src\components\SchedulePanel\components\ScheduleCalendar\index.js. This file does not use rbc features at all. Someone should work on this problem.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,6 +17,6 @@
     // "editor.formatOnSave": true
   },
   "extensions.ignoreRecommendations": true,
-  "files.autoSave": "off",
+  "files.autoSave": "afterDelay",
   "debug.allowBreakpointsEverywhere": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,6 +17,6 @@
     // "editor.formatOnSave": true
   },
   "extensions.ignoreRecommendations": true,
-  "files.autoSave": "afterDelay",
+  "files.autoSave": "off",
   "debug.allowBreakpointsEverywhere": true
 }
--- a/src/_vars.scss
+++ b/src/_vars.scss
@@ -30,7 +30,6 @@ $neutral-dark-gray2: #444444;
 $neutral-dark-gray3: #555555;
 
 $neutral-black: #000000;
-$neutral-black2: #111111;
 $neutral-black-opacity90: #000000e6;
 $neutral-black-opacity55: #0000008c;
 $neutral-black-opacity20: #00000033;
--- a/src/_vars.scss
+++ b/src/_vars.scss
@@ -9,13 +9,12 @@
 
 $primary-blue: #014983;
 $primary-cyan: #00aeef;
-
-$primary-red: #ff0000;
-
 $secondary-green: #b2bb1c;
 $secondary-yellow: #fdb913;
 $secondary-orange: #de571f;
 $secondary-red: #b53228;
+
+$alternative-red: #ff0000;
 
 $neutral-white: #ffffff;
 $neutral-white-opacity50: #ffffff80;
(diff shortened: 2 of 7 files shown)


src/_vars.scss
```diff
@@ -10,6 +10,8 @@
 $primary-blue: #014983;
 $primary-cyan: #00aeef;
 
+$primary-red: #ff0000;
```
EjPlatzer, line 13: I think this variable needs to be renamed. The `primary` and `secondary` names refer specifically to [Gordon's brand design standards](https://gordonedu.sharepoint.com/WebLinks/Shared%20Documents/Design%20Center/Gordon%20Colors.pdf?originalPath=aHR0cHM6Ly9nb3Jkb25lZHUuc2hhcmVwb2ludC5jb20vOmI6L2cvV2ViTGlua3MvRWFyQVZ1YmM2Y3hNcGRRNFBFMXM0YXNCRUxtTFgwVUFHU0tLWmI4MHZfTVp6QT9ydGltZT1HLWVqQUx2bzJFZw), and so to call this `primary-red` is misleading because it isn't one of our primary brand colors.

.vscode/settings.json
```diff
@@ -17,6 +17,6 @@
     // "editor.formatOnSave": true
   },
   "extensions.ignoreRecommendations": true,
-  "files.autoSave": "off",
+  "files.autoSave": "afterDelay",
```
EjPlatzer, line 20: Was there a reason you set this for the project? I think perhaps this should be a user setting, because we don't want to adjust everyone else's settings unless there's a specific reason to.

EjPlatzer (changes requested):
These are great updates! Thanks for the good work refactoring. I left a couple small comments on minor things I think should be changed, but overall it looks great!

src/_vars.scss
iamphg97: Ahhh..! I did not know that at all! Thanks! :)

.vscode/settings.json
iamphg97: That was my mistake. I should not upload this file.

iamphg97:
I already shared my thought with Josh that I realized it is impossible to reduce the number of colors. This is because, I thought some adjacent colors can be replace to the nearby colors (for example, neutral-gray-2 -> neutral-light-gray), however still I can see the big difference. So I think it is better to keep all of the colors. I checked all of the colors.

I think this is enough for the color consistency. I will move on to the next one after merging the latest version to the develop (Of course, after someone review it).

EjPlatzer (approved):
Looks good to me! Great job making these consistent, this will make the site easier to maintain and extend!

PR #1034, opened 2021-03-18
Ran prettier code formatter on `src/` to maintain consistent formatting

NerdyGriffin:
The title should be pretty self-explanatory. All the changes in this branch were made by running the Prettier formatter, in an attempt to maintain or clean-up the consistency of the formatting across the project


.vscode/settings.json
```diff
@@ -18,5 +18,6 @@
   },
   "extensions.ignoreRecommendations": true,
   "files.autoSave": "off",
-  "debug.allowBreakpointsEverywhere": true
+  "debug.allowBreakpointsEverywhere": true,
+  "gitdoc.enabled": false
```
EjPlatzer, line 22: I see that you disabled gitdoc in the project settings. I'm not sure what gitdoc is (maybe it's this VSCode extension?), but it isn't something that we have as a default in the project so I don't think we need a project-level setting to disable it. Should this be a user setting instead?

EjPlatzer (changes requested):
Thanks for running prettier! It's surprising that these aren't being caught by the pre-commit hook (apart from the .svg files, which we don't normally lint, but there's no harm in doing so).

I noticed one change to the project-wide settings that doesn't seem necessary, but otherwise this all looks good to me!

.vscode/settings.json
NerdyGriffin: I think I do have this disabled in my user settings, but it automatically adds that every time I open the code unless I globally disable the extension (i.e. disable from the extensions menu, not just setting `gitdoc.enabled` to false. I'll check on that further

NerdyGriffin, line 22:
```suggestion
```

EjPlatzer (approved):
LGTM! Good formatting changes!

PR #1036, opened 2021-03-23
Update Return to Campus link & text for Easter

EjPlatzer:
Students are required to fill out another, different Return to Campus form after returning from Easter Break in the Spring of 2021. This updates the link to point to the correct form, and changes the text to clarify the new form.


bennettforkner (approved):
LGTM

PR #1037, opened 2021-03-23
Guest welcome fix

Kindlewing:


aidanperez2000 (changes requested):
Overall changes look good.  I think you should make the links that the user doesn't have access to when they are on the guest view not be there unless they are logged in.  For example, the links to people and wellness in the header should not be there when the user is not logged in.  I think that would probably be better than letting the guest click links they do not have access to.  Everything else looks great though!

aidanperez2000 (approved):
Changes look good!  i think they're ready to merge!  This will be grat to have on the site!

PR #1041, opened 2021-03-24
Add error handling for bad shift times

EjPlatzer:
The API has been updated to detect bad shift times when they are saved. This introduces some preliminary error handling on the front end to notify the user more concretely on why their shift failed to save.


jason-whitehouse (approved):
Looks good

PR #1042, opened 2021-03-24
Fix & Restore timesheets

EjPlatzer:
Timesheets was taken down temporarily because of an issue where shifts were being submitted with no Time information. It turns out that there were options on Intl.DateTimeFormat JavaScript built-in we were using (specifically `dateStyle` and `timeStyle`) that aren't supported by Safari. This was causing safari users to lose the time of the their shift start and end DateTimes.

### DIFF
--- a/src/services/jobs.js
+++ b/src/services/jobs.js
@@ -5,9 +5,12 @@
 
 import http from './http';
 
-const dateFormat = new Intl.DateTimeFormat('en', {
-  timeStyle: 'medium',
-  dateStyle: 'short',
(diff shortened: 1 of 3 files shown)


bennettforkner (approved):
Seems responsible to me. I have faith that this more-general formatting functionality will fix our Safari compatibility issues.

jason-whitehouse (approved):
All looks good.  If submit doesn't work we can look at just passing shift.SHIFT_END_TIME since it's already coming from the DB as a datetime format at that point.

PR #1046, opened 2021-03-25
Apartapp develop merge fixes

NerdyGriffin:
Branch with merge fixes and bug fixes that I forgot I made long ago


src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
```diff
@@ -25,6 +27,12 @@ const HallListItem = ({
   const [hallRankValue, setHallRankValue] = useState(1); // Rank drop-down menu value
```
joshuaRogers24, line 27: Does this make every hall item start with rank 1?

joshuaRogers24 (approved):
I have some questions, but nothing I think would keep us from merging. Good fixes!

src/services/housing.js
```diff
@@ -71,12 +71,23 @@ const checkHousingStaff = async () => {
...
+    }
+  } catch (e) {
+    // handle thrown 404 errors
+    if (e.status !== 404) throw e;
+    err = e;
+    applicationID = false;
+    console.log('A 404 code indicates that an application was not found for this applicant');
   }
-  return applicationID;
+  // handle "unhandled" requests, `this.status = 404`s, and 404 errors
```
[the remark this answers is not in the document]
NerdyGriffin: The goal it to detect 404 error/responses, and allow the frontend to continue

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
NerdyGriffin: Technically yes, but that is only as a fail-safe, to prevent the page from throwing an error before it has a chance to set the true value.
The JavaScript likes jump the gun with null errors if you accidentally forget to give things a valid initially placeholder values at moment the component is initialized

joshuaRogers24: Huh, okay!

PR #1048, opened 2021-03-25
Merged develop into apartapp-develop

NerdyGriffin:


src/setupProxy.js
```diff
@@ -1,13 +1,13 @@
-const proxy = require('http-proxy-middleware');
+const {createProxyMiddleware} = require('http-proxy-middleware');
```
joshuaRogers24, line 1: I know this change may cause us trouble if we do not update the http-proxy-middleware dependency. It was a breaking change.

joshuaRogers24: We need to let the whole team know that, once we merge these changes in, they will definitely have to delete node_modules and reinstall

joshuaRogers24 (changes requested):
Nice. We have some changes to make due to some updates. Can you make sure our Snackbar use matches how it works now?

src/components/Snackbar/index.js
```diff
@@ -1,15 +1,27 @@
...
+  CheckCircleOutlineOutlined,
+  ErrorOutline,
+  InfoOutlined,
+  ReportProblemOutlined,
+} from '@material-ui/icons/';
 import MuiAlert from '@material-ui/lab/Alert';
 
-const Alert = (props) => {
-  return <MuiAlert elevation={6} variant="filled" {...props} />;
+const defaultAlertIconMapping = {
```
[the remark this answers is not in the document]
NerdyGriffin: What do you mean "the old version"? 
It is a site-wide component, I'm not sure what you are actually trying to ask.
If you are worried about compatibility, it looks like the outward "interface" of the component is functionally identical to how it was before, it is just improvements to the implementation, probably more consistent styling implementation, and taking better advantage of some React-related features

src/setupProxy.js
NerdyGriffin: Note by the title that this is a merge of stuff from develop into our branch. So these were changes already reviewed and approved before they got merged into develop, and not written by me

NerdyGriffin: Ah, I now understand what you mean in the second comment, that is definitely a good point, that we would expect the existing 'npm install' of our local clones of the repo to break once this branch gets merged into apartapp-develop.

src/components/Snackbar/index.js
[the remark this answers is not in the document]
joshuaRogers24: Okay, that is what I was wondering, thank you

src/setupProxy.js
joshuaRogers24: right, thanks for trying to understand!

joshuaRogers24 (approved):

joshuaRogers24 (approved):
Nice.

PR #1049, opened 2021-03-25
Fix timesheets rerender issue

EjPlatzer:
There was an issue with Timesheets where the page would rerender every time a user typed in the `End Time` field. This was making the page nigh-unusable for people who chose not to use (or didn't know about) the Picker ui for selecting their shift end time.

The heart of the issue was that the UI was attempting to fetch the user's jobs every time the selectedDateIn or selectedDateOut value changed - as long as neither was null - even when one wasn't a date. In this case, a user manually typing the date their shift ended would start typing, which would update selectedDateOut to be non-null, and the UI would try to fetch jobs. However, the user had JUST started typing, so the shift end date was not a valid date yet (e.g. '03' instead of '03/24/2021 4:30 PM'). This would cause when trying to format the date for transmission to the API. That error would force the component to rerender, and thus the user was never able to enter the end of their shift.

The best fix for this issue involved a decent amount of refactoring, because the  best solution is to have a useEffect callback watching selectedDateIn and selectedDateOut that will try to fetch jobs ONLY if they are both valid dates (and don't violate any of the other business logic for shift times such as not being in the future). Doing it this way allowed for the existing code to be drastically simplified and stream-lined, because the state-update logic for the date values can be cleanly separated from the validation logic for the date values. 

In testing this, I also exposed a co-existing issue where the user's profile was being fetched on every render because of an improper dependency array for useEffect. This was the second layer of the issue, which meant that when typing in End Time forced the page to rerender, it would also refetch their personType and erroneously display the Staff view.

While I was working in Timesheets, I streamlined and eliminated other dead code to make detecting and fixing bugs easier in the future.


src/views/Timesheets/index.js
```diff
@@ -28,6 +29,13 @@ import GordonLoader from '../../components/Loader';
 import { makeStyles } from '@material-ui/core/styles';
 import SimpleSnackbar from '../../components/Snackbar';
 import user from '../../services/user';
+import useNetworkStatus from '../../hooks/useNetworkStatus';
+
+const MINIMUM_SHIFT_LENGTH = 0.08; // Minimum length for a shift if 5 minutes, 1/12 hour
```
bennettforkner, line 34:
```suggestion
const MINIMUM_SHIFT_LENGTH = 0.08; // Minimum length for a shift is 5 minutes, 1/12 hour
```

bennettforkner (commented):
LGTM. Waiting for @jason-whitehouse for approval

jason-whitehouse (approved):
Looks good! Great job!

PR #1050, opened 2021-03-26
Change image imports for react-scripts update

EjPlatzer:
Some images were being included in an old-fashioned way (using JavaScript's `require()` function), and that way is no longer supported as of `react-scripts v4.x`. The reason they were failing is because the paths were relative to the file structure of the unbundled project, so when the bundler built the site, the images were no longer available at that URL.

This pull request updates the method for including the affected images to use ES6 `import` syntax so the bundler can track the component's dependency on the image file and update the relative path of the import. These import paths are mapped to a `static/...` path where the elided portion is generated at build time.


src/views/Feedback/index.js
```diff
@@ -78,10 +78,7 @@ export default class Feedback extends Component {
                     marginRight: 'auto',
                   }}
                 >
-                  <img
-                    src={require(`${'../../NoConnection.svg'}`)}
-                    alt="Internet Connection Lost"
-                  />
+                  <NoConnection />
```
bennettforkner, line 81: Perhaps change to NoConnectionSVG for readability
```suggestion
                  <NoConnectionSVG />
```

bennettforkner (changes requested):

src/views/Feedback/index.js
bennettforkner: And the same for all of the others :)

bennettforkner (approved):
LGTM. Much more efficient to have these SVGs as components. Thanks Evan!

PR #1053, opened 2021-03-27
Improved styling for apartapp staff page

NerdyGriffin:


src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/applicationRow.scss
```diff
@@ -1,6 +1,12 @@
...
+.MuiTableCell-root.collapsible-row {
+  padding-top: 0;
+  padding-bottom: 0;
+}
+
+.MuiTable-root.sub-table {
+  border-left: 2px solid $neutral-black-opacity10;
+  border-right: 2px solid $neutral-black-opacity10;
+  border-bottom: 1px solid $neutral-black-opacity10;
 }
```
EjPlatzer, line 12: This looks like a huge improvement! My one worry now is that these class names are pretty general, and I could see someone reusing `sub-table` elsewhere on the site - especially inside a MuiTable - and then being confused why these extra styles were being applied to it. 

You might want to re-add the `apartment` prefix to ensure these stay scoped to only the components you're targeting. Alternatively, you could wrap both these rules in a selector targeting the top of the Apartment view, with a class name like '360-apartment-application-root' or something. That way these rules are guaranteed to only ever affect components within the apartment application. That would make your styling something like this:
```suggestion
.360-apartment-application {
  & .MuiTableCell-root.collapsible-row {
    padding-top: 0;
    padding-bottom: 0;
  }
  
  & .MuiTable-root.sub-table {
    border-left: 2px solid $neutral-black-opacity10;
    border-right: 2px solid $neutral-black-opacity10;
    border-bottom: 1px solid $neutral-black-opacity10;
  }
}
```

Then you would just have to add the `360-apartment-application` class to the root of the view or something else high up in the tree. This approach also has the benefit of being reusable across all of your styling rules.

EjPlatzer (changes requested):
These look like very clean and responsible changes! I suggested one extra thing to prevent possible css rule clashes in the future, but other than that this looks great!

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/applicationRow.scss
NerdyGriffin: One rebuttal I have to this is the fact that this css only applies to the files that import it manually or to child components of the thing that imports the css.
So the only way for it to accidentally apply would be if someone were working within the exact scope of this component sub-tree

This also means that the existing behavior already accomplish the same limited scoping behavior as what you have suggested with manually adding a className `360-apartment-application` to the root-ish of the ApartApp component tree

NerdyGriffin:
I think this accidentally merged since the stuff from here was also in the branch of #1051 

That was not supposed to happen, meaning I forgot to update the base branch in the #1051 PR

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationRow/applicationRow.scss
EjPlatzer: You would think that's how an imported css file would behave, but actually that isn't the case at all. If you look at the create-react-app documentation (which in this case is simply explaining the behavior of webpack), what actually happens in production builds is that every css file that's used anywhere in the build gets concatenated together into one long file. So styles defined anywhere on the site will end up affecting every element in the html that fits their specifiers.

NerdyGriffin: Ah, so the debug build's behavior isn't the same as the production build...
That sounds quite problematic

NerdyGriffin: I looks like I already have classNames assigned to the root of the apartment application stuff, which I can use for this purpose.

PR #1054, opened 2021-03-28
Styling updates

NerdyGriffin:
An update to the styling inspired by [this review](https://github.com/gordon-cs/gordon-360-ui/pull/1053#discussion_r602734096)


src/views/ApartmentApp/apartmentApp.scss
```diff
@@ -1,84 +1,87 @@
 @import '../../vars';
 
-strong.over-emphasized {
-  text-transform: uppercase;
-  text-decoration: underline;
-}
+.student-apartment-application,
+.apartment-application {
```
EjPlatzer, line 4: Is there a reason that you specified both of these classes? Could it not just be `.student-apartment-application`?

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -46,7 +46,7 @@ const StaffMenu = ({ userProfile, authentication }) => {
     setLoading(true);
     let applicationDetailsArray = await housing.getAllApartmentApplications();
     if (applicationDetailsArray) {
-      setApplications(applicationDetailsArray);
+      setApplications(applicationDetailsArray ?? []);
     }
```
EjPlatzer, line 50: This is fairly redundant, because `null` is falsy in JavaScript, so the `if` case ensures that `setApplications(...)` will never get called if `applicationDetailsArray` is `null`. Perhaps you meant to remove the `if` so that it will be set to an empty array if `applicationDetailsArray` is falsy?

src/views/ApartmentApp/apartmentApp.scss
```diff
@@ -1,84 +1,87 @@
 @import '../../vars';
```
EjPlatzer, line 1: This file (apartmentApp.scss) ends up getting referenced three times in your JavaScript. `apartmentApp.css` is imported from both `src/views/ApartmentApp/components/StaffMenu/index.js` and `src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js`. Furthermore, `apartmentApp.scss` is imported in `src/views/ApartmentApp/index.js`. Note that the last one is the `scss` file, not the `css` file. Since `src/views/ApartmentApp/index.js` is above both of the other components, you could simplify this by importing the actual `css` file in `src/views/ApartmentApp/index.js` and then removing the (now redundant) imports in the other two files.

EjPlatzer (changes requested):
I noticed a few cases where some clean up might be needed, but beyond those small things this looks great! Once they're addressed, I think this is ready.

src/services/housing.js
```diff
@@ -174,7 +174,87 @@ const getApartmentApplication = async (applicationID) => {
  * @return {Promise.<ApplicationDetails>[]} Application details
  */
 const getAllApartmentApplications = async () => {
-  let applicationDetailsArray = await http.get(`housing/admin/apartment/applications/`);
+  // let applicationDetailsArray = await http.get(`housing/admin/apartment/applications/`);
+  let applicationDetailsArray = [
```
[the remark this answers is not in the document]
NerdyGriffin: It is more debug data so that I don't have to insert a bunch of data into the database tables just for a test, but yes, same idea.
The API is functional now

I intend to remove that debug data before merging this, but I figured I would leave it there until any suggested changes have been address, so that I don't have to manually add it back if I need to test other styling changes

src/views/ApartmentApp/apartmentApp.scss
```diff
@@ -1,84 +1,87 @@
...
-  background-color: $primary-blue;
-  color: $neutral-white;
-  padding: '10px';
-  .MuiTypography-colorTextSecondary {
-    color: $neutral-light-gray;
-    font-size: 1.25rem;
+  & .apartment-card-header {
+    background-color: $primary-blue;
+    color: $neutral-white;
+    padding: '10px';
```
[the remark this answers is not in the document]
NerdyGriffin: Ah, that is probably an old copy-and-paste mistake,

It also makes me wonder why I ever put that there in the first place, since clearly I never even noticed that it isn't doing anything during recent testing, so why did past-me think we needed to add that padding...

src/views/ApartmentApp/apartmentApp.scss
NerdyGriffin: the other one occurs in a different place in the tree, and I don't remember why... I'll have to explore later the figure that out, because yes it should be fine with just `student-apartment-application`

EjPlatzer: I'm fairly certain that `apartment-application` is applied to an otherwise empty `div` wrapping all other contents of the StudentApplication component.

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: I think I did mean to remove the `if`

I think I realized it probably should just be in a try-catch or using the `.then` stuff, since the null case is just the case where there was an error, including the case where there are no applications at all in the database

I also don't think there is any reason to put this in the useCallback instead of just putting it directly in the useEffect, unless we plan to add a button to call that function, which seems silly if it is mostly equivalent to just refreshing the page

It seems like both the `loadAllCurrentApplications` and the `generateCSVData` might be better to just put inside one `useEffect`

src/views/ApartmentApp/apartmentApp.scss
NerdyGriffin: Actually that sounds like it is entirely a typo and unintended mistakes that simply went unnoticed, it was only supposed to be imported in `src/views/ApartmentApp/index.js`

src/views/ApartmentApp/apartmentApp.scss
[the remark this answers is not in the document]
NerdyGriffin:
```suggestion
```
If the page looks fine the way it is now, then I would suggest just removing that padding because I don't see why that padding would need to be added

src/views/ApartmentApp/apartmentApp.scss
NerdyGriffin: I think so too, I think it was an old leftover from many months ago

NerdyGriffin: That div also wraps up the snackbar, which is best to keep it outside the Grid otherwise it will make the Grid items sometimes align incorrectly

EjPlatzer (approved):
LGTM!

PR #1063, opened 2021-04-03
ApartApp code quality improvements

NerdyGriffin:
- [ ] Test out the "fixed" sorting feature on the admin page


EjPlatzer (commented):
This looks good! Using the lodash function is a huge improvement, always nice to not reinvent the wheel.

It seems like the actual functionality of the sorting is not yet used though. I can click on the column headings and the arrow representing ordering will change appropriately, but the contents of the table do not change at all.

Before:
image

After:
image

This seems to be true for all columns:
image
image


Are we expecting sorting to be functional?

NerdyGriffin:
@EjPlatzer  Yes, the sorting should be functional, I followed the official Material-UI examples/guide, but it did not work and I guess it still isn't working with the lodash sorting instead. 
I have spent a good chunk of time trying to figure out why it's not sorting, but it was not very urgent so I haven't really given it any deep debugging yet. 
It should be okay to merge it even without the sorting, and I added a comment to the known issues in the README for now until I find a fix. It might be worth it for me to leave it as-is and deal with it later in a separate PR, unless you have any ideas about quick things I could try to make the sorting work. 

I think it is best to prioritize the other core features first since we aim to do user testing Friday and the sorting is merely a convenience rather than a necessity

EjPlatzer (approved):
Good improvements! Since the non-functional sorting is a known issue, this is all set to merge!

PR #1065, opened 2021-04-06
Use Webpack Aliases to import via absolute path from src directory

EjPlatzer:
Previously, importing something from within our project required a relative path. So if you wanted to import `src/services/user.js` from `src/components/Identification/Components/LinksDialog.index.js`, the import path would have to be `../../services/user.js`. In some cases, we had much more deeply nested files, and so the relative imports were large, repetitive, and confusing.

Thanks to a (new? Or at least newly discovered) Webpack feature, we can define a `baseUrl` for imports as a compiler option, and when Webpack builds our project it will use that `baseUrl` to resolve absolute imports (and then check `node_modules` if it can't be found there).

So, I have defined `src` as our `baseUrl`, which means any relative import can now be replaced with an absolute one that starts in `src`. For example, the above import would now be simply `services/user.js`.

I have also replaced all such relative imports with their aboslute counterpart, with a couple of exceptions. Specifically, I left relative imports when the importer and imported were in the same nested sub-directory, so going from `src` down was actually longer and less straightforward than the relative path.

For example, the CategorizedNews component imports the NewsItem component, and both are  sub-components of the NewsCard component, which is itself a sub-component of the Home component. Hence, the relative path is simply `../NewsItem`, whereas the absolute path would be `views/Home/components/NewsCard/components/NewsItem`

Because this is an option for our 'compiler', any errors are detected while building the project, which means all of the updated imports have imported successfully. I also verified this while running it locally to be sure everything continued to function correctly.


bennettforkner (approved):
LGTM

PR #1066, opened 2021-04-07
Removed points calculation and added handling of 404 not found

NerdyGriffin:


EjPlatzer (dismissed):
Made some comments with tiny possible improvements, but they are not very significant. Overall, this looks good to me!

EjPlatzer (approved):
Looks good to me!

PR #1067, opened 2021-04-07
ApartApps Fix application load bug and other changes from review

EjPlatzer:
The StudentApplication component was loading apartment applications in a useEffect that depended on unnecessary state and so was being run too oten and getting cut short, causing state inconsistencies. 

The useEffect has been simplified to only depend on `userProfile`, so that it only runs when the component is initially rendered.

### DIFF
--- a/src/views/ApartmentApp/components/StaffMenu/index.js
+++ b/src/views/ApartmentApp/components/StaffMenu/index.js
@@ -71,16 +71,12 @@ const StaffMenu = ({ userProfile, authentication }) => {
         let { Applicants, ApartmentChoices, ...filteredApplicationDetails } = applicationDetails;
         applicationsForCsv.push(filteredApplicationDetails);
 
-        Applicants.forEach((applicant) => {
-          if (applicant.ApplicationID) {
-            applicantsForCsv.push(applicant);
-          } else {
-            applicantsForCsv.push({
-              ApplicationID: applicationDetails.ApplicationID,
-              ...applicant,
-            });
-          }
-        });
+        Applicants.forEach((applicant) =>
+          applicantsForCsv.push({
+            ApplicationID: applicant.ApplicationID ?? applicationDetails.ApplicationID,
+            ...applicant,
+          }),
+        );
 
         ApartmentChoices.forEach((apartmentChoice) => {
           apartmentChoicesForCsv.push({
--- a/src/views/ApartmentApp/components/StaffMenu/index.js
+++ b/src/views/ApartmentApp/components/StaffMenu/index.js
@@ -65,13 +65,12 @@ const StaffMenu = ({ userProfile, authentication }) => {
     let applicationsForCsv = [];
     let applicantsForCsv = [];
     let apartmentChoicesForCsv = [];
-    applicationDetailsArray.forEach((applicationDetails) => {
+    applicationDetailsArray.forEach(({ Applicants, ApartmentChoices, ...applicationDetails }) => {
       // Only add the applications that have been submitted
       if (applicationDetails.DateSubmitted) {
-        let { Applicants, ApartmentChoices, ...filteredApplicationDetails } = applicationDetails;
-        applicationsForCsv.push(filteredApplicationDetails);
+        applicationsForCsv.push(applicationDetails);
 
-        Applicants.forEach((applicant) =>
+        Applicants.forEach(({ Profile, ...applicant }) =>
           applicantsForCsv.push({
             ApplicationID: applicant.ApplicationID ?? applicationDetails.ApplicationID,
             ...applicant,
--- a/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
+++ b/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
@@ -67,9 +67,7 @@ const ApplicantListItem = ({
    * @param {StudentProfileInfo} profile The StudentProfileInfo object for the student represented by this list item
    */
   const createNickname = (profile) => {
-    let Name = String(profile.fullName);
-    let FirstName = Name.split(' ')[0];
-    setHasNickname(FirstName !== profile.NickName && profile.NickName !== '');
+    setHasNickname(profile.FirstName !== profile.NickName && profile.NickName !== '');
   };
 
   /**
@@ -97,8 +95,8 @@ const ApplicantListItem = ({
   };
 
   const displayName = hasNickName
-    ? profile.fullName + ' (' + profile.NickName + ')'
-    : profile.fullName;
+    ? `${profile.FirstName} ${profile.LastName} (${profile.NickName})`
+    : `${profile.FirstName} ${profile.LastName}`;
 
   return (
     <React.Fragment>
--- a/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
+++ b/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
@@ -36,7 +36,7 @@ const ApplicantListItem = ({
 
   useEffect(() => {
     loadAvatar(profile);
-    setHasNickname(profile.FirstName === profile.NickName && profile.NickName !== '');
+    setHasNickname(profile.FirstName !== profile.NickName && profile.NickName !== '');
     if (String(profile.PersonType).includes('stu') && profile.Class !== undefined) {
       setPersonClass(profile.Class);
     } else {
--- a/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
+++ b/src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
@@ -36,7 +36,7 @@ const ApplicantListItem = ({
 
   useEffect(() => {
     loadAvatar(profile);
-    createNickname(profile);
+    setHasNickname(profile.FirstName === profile.NickName && profile.NickName !== '');
     if (String(profile.PersonType).includes('stu') && profile.Class !== undefined) {
       setPersonClass(profile.Class);
     } else {
@@ -60,14 +60,6 @@ const ApplicantListItem = ({
         </svg>,
       );
     }
-  };
-
-  /**
-   * Saves the nickname of the given user if available
-   * @param {StudentProfileInfo} profile The StudentProfileInfo object for the student represented by this list item
-   */
-  const createNickname = (profile) => {
-    setHasNickname(profile.FirstName !== profile.NickName && profile.NickName !== '');
   };
 
   /**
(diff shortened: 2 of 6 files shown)


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -90,68 +88,45 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
           }
-        })
-        .catch((error) => {
-          handleError(error);
-        });
-      setLoading(false);
-    }
+        }
+      } catch (error) {
+        handleError(error);
```
NerdyGriffin, line 114: ~~Since this now only needs to be done in one place, is there any reason to keep it as separate helper function, rather than just putting the contents inside this `catch { ... }` directly?~~
I now this it would be better to change the name of this function, and have it called directly inside the `if (newApplicationID === null)` rather than attempting to trigger it by manually throwing an error.
The new name also helps clarify the true purpose of this function

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -90,68 +88,45 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
-          } else if (isSubscribed) {
-            setApplicationID(newApplicationID);
+    const loadApplication = async () => {
+      try {
+        setLoading(true);
+        // Check if the current user is on an application. Returns the application ID number if found
+        const newApplicationID = await housing.getCurrentApplicationID();
+        if (newApplicationID === null || newApplicationID === -1) {
+          // Intentionally trigger the 'catch'
+          throw new Error("Invalid value of 'newApplicationID' = " + newApplicationID);
```
NerdyGriffin, line 98: Maybe instead of using an error to trigger the catch, we should just call the `handleError` function inside this `if`, especially since there is nothing after the else anyways

NerdyGriffin, line 98:
```suggestion
        if (newApplicationID === null) {
          initializeNewApplication();
```

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin:
```suggestion
        initializeNewApplication();
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -77,8 +77,6 @@ const StudentApplication = ({ userProfile, authentication }) => {
    * Loads the user's saved apartment application, if one exists
    */
   useEffect(() => {
-    let isSubscribed = true;
-
     const handleError = (_error) => {
```
NerdyGriffin, line 80:
```suggestion
    const initializeNewApplication = () => {
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -90,68 +88,45 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
-            // Intentionally trigger the 'catch'
-            throw new Error("Invalid value of 'newApplicationID' = " + newApplicationID);
-          } else if (isSubscribed) {
-            setApplicationID(newApplicationID);
+    const loadApplication = async () => {
+      try {
+        setLoading(true);
+        // Check if the current user is on an application. Returns the application ID number if found
+        const newApplicationID = await housing.getCurrentApplicationID();
+        if (newApplicationID === null || newApplicationID === -1) {
```
NerdyGriffin, line 96: This actually should not be working right now, because I accidentally set `applicationID` to `false` in the housing service instead of `null`. It should be `null` if the application is not found, so I will also make that change inside `src/services/housing.js`

NerdyGriffin (changes requested):

NerdyGriffin:
Ah I forgot to put a comment in the review, I meant to say that it looks good, and most of my suggestions are to improve flaws that were leftover from the code prior to your changes (so essential my own past mistakes 😆 )

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -36,7 +36,7 @@ const ApplicantListItem = ({
 
   useEffect(() => {
     loadAvatar(profile);
-    createNickname(profile);
+    setHasNickname(profile.FirstName === profile.NickName && profile.NickName !== '');
```
NerdyGriffin, line 39:
```suggestion
    setHasNickname(profile.FirstName !== profile.NickName && profile.NickName !== '');
```

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
```diff
@@ -102,7 +94,7 @@ const HallListItem = ({
               <InputLabel>Rank</InputLabel>
               <Select
                 disabled={disabled}
-                value={hallRankValue}
+                value={hallRank}
```
NerdyGriffin, line 97: Does that actually work, to use the prop directly? I thought that didn't work when I tried testing it in the past

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -77,9 +77,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
    * Loads the user's saved apartment application, if one exists
    */
   useEffect(() => {
-    const handleError = (_error) => {
-      // No existing application was found in the database,
-      // or an error occurred while attempting to load the application
+    const initializeNewApplication = () => {
       setApplicationID(null);
       setEditorUsername((prevEditorUsername) => prevEditorUsername ?? userProfile.AD_Username);
```
NerdyGriffin, line 82: There should probably also be a `setApplicationDetails`, since that contains all the other variables describing the state of the application

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
EjPlatzer: Ahh right, good catch. That's what I get for copying it manually instead of cutting and pasting.

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
EjPlatzer: I've been using it all day during my testing and it worked perfectly.

src/views/ApartmentApp/components/StudentApplication/index.js
EjPlatzer: I'm guessing the reason you didn't include that originally was because `applicationDetails` has a default value? But you're right, I think this is the cleaner way to do it.

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
EjPlatzer: In case it isn't obvious why that works, I guess I should clarify. The real source of truth for the hall choices are stored as state on the StudentApplication component. When the user changes a hall rank or name, you call that component's method to handle changes and changing that component's state accordingly. This state change results in new props cascading through the HallSelection component and down to this component. The new props cause the component to re-render, and thus the displayed values update accordingly.

I'm actually not totally clear on why your previous way was working. Either the new props were re-calculating the state every time, or else I would expect that the component's state never updated at all, since you never called the setState functions after the initial useEffect.

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -65,22 +65,17 @@ const StaffMenu = ({ userProfile, authentication }) => {
     let applicationsForCsv = [];
     let applicantsForCsv = [];
     let apartmentChoicesForCsv = [];
-    applicationDetailsArray.forEach((applicationDetails) => {
+    applicationDetailsArray.forEach(({ Applicants, ApartmentChoices, ...applicationDetails }) => {
```
NerdyGriffin, line 68: Oh this is a very satisfying improvement

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -65,22 +65,17 @@ const StaffMenu = ({ userProfile, authentication }) => {
...
-          if (applicant.ApplicationID) {
-            applicantsForCsv.push(applicant);
-          } else {
-            applicantsForCsv.push({
-              ApplicationID: applicationDetails.ApplicationID,
-              ...applicant,
-            });
-          }
-        });
+        applicationsForCsv.push(applicationDetails);
```
NerdyGriffin, line 71: Would it be better to do this via something like this instead of pushing to an array, followed by using that array in the `setApplicationJsonArray()`?

```
setApplicationJsonArray(applicationDetailsArray.filter((applicationDetails) => (
    applicationDetails?.DateSubmitted)
  )
  .map(({ Applicants, ApartmentChoices, ...applicationDetails }) => {
    // do the stuff with the applicants and apartmentchoices here
    return applicationDetails;
  }
)
```

NerdyGriffin: I have begun drafting this in a new branch, to test it out

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
NerdyGriffin: The previous way worked because having the props explicitly listed in the dependency array tells the useEffect to re-run every time those props get changed.

I think the real reason I thought I needed that was because of a problem in the `handleHallInputChange` in the `StudentApplication` component, where I was using the setState things incorrectly and I thought the problem that in the list items not at the top of the component tree. I tried many ridiculous variation in my attempts to fix that bug, which obvious did not succeed because I didn't realize that problem was in `StudentApplication` and the setState things behaved differently than how I thought they did at the time

NerdyGriffin: I also think I used to have the state variable in this component so that the `handleNameInputChange` and such could do a `setState` for the thing that the user selected, and then after that I added in the functionality of having it mapped to the state array from higher up

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -21,7 +21,17 @@ import Agreements from './components/Agreements';
...
 const MAX_NUM_APPLICANTS = 8;
+const BLANK_APPLICATION_DETAILS = {
+  ApplicationID: null,
+  DateSubmitted: null,
+  DateModified: null,
+  EditorUsername: null,
+  EditorEmail: null,
+  Applicants: [],
+  ApartmentChoices: [],
+};
```
NerdyGriffin, line 34: I love this solution, this is exactly the kind of redundant/repeated code that had been bothering me

NerdyGriffin (approved):

PR #1069, opened 2021-04-08
Refactor nav buttons and header

EjPlatzer:
The Header and Nav buttons have been refactored to make use of the useNetworkStatus hook.

In the process, I discovered that a significant amount of code was being repeated for the nav buttons between the rigth corner and the mobile drawer. The repeated logic was streamlined and extracted to a new reusable component, NavButton. 

I also fixed several miscellaneous errors and warnings, mostly from Material-ui props being used incorrectly or updates to dependencies.


josh-peters-2357 (approved):
Looks good to me! I did some "frontend testing" which was just me using a local server to see if it everything still worked the same, which it does :).

PR #1070, opened 2021-04-08
Apartapp simplified CSV generation and refresh data button

NerdyGriffin:
This branch implements several simplifications of the process of generating the arrays for the CSV files, based on ideas and suggestions from #1067 
It also adds a button that allows the user to refresh or reload the application data that is used in the data table and CSV files, which is more efficient than refreshing the entire page within the browser. The styling and position of this button have not yet been finalized, but the functionality of the button should be working. 
Feel free to offer styling/placement suggestions for the button, but the primary goal is to get it functional and merged in time for the user testing tomorrow; the aesthetic styling can be cleaned up next week in a separate PR

### DIFF
--- a/src/views/ApartmentApp/components/StaffMenu/index.js
+++ b/src/views/ApartmentApp/components/StaffMenu/index.js
@@ -2,6 +2,7 @@ import React, { useState, useEffect, useCallback } from 'react';
 import { CSVLink } from 'react-csv';
 import { Grid, Card, CardHeader, CardContent, Button, Typography } from '@material-ui/core/';
 import GetAppIcon from '@material-ui/icons/GetApp';
+import RefreshIcon from '@material-ui/icons/Refresh';
 import { DateTime } from 'luxon';
 import GordonLoader from '../../../../components/Loader';
 import housing from '../../../../services/housing';
@@ -143,6 +144,24 @@ const StaffMenu = ({ userProfile, authentication }) => {
                     target="_blank"
                   >
                     Download Apartment Hall Choices
+                  </Button>
+                </Grid>
+              </Grid>
+            </CardContent>
+          </Card>
+        </Grid>
+        <Grid item xs={12} sm={9} md={6} lg={3}>
+          <Card>
+            <CardContent>
+              <Grid container direction="row" spacing={2} padded>
+                <Grid item xs={12}>
+                  <Button
+                    variant="filled"
+                    color="primary"
+                    startIcon={<RefreshIcon />}
+                    onClick={loadAllCurrentApplications}
+                  >
+                    Refresh Application Data
                   </Button>
                 </Grid>
               </Grid>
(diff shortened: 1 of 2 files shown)


src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -117,7 +109,7 @@ const StaffMenu = ({ userProfile, authentication }) => {
                     variant="contained"
                     color="primary"
                     startIcon={<GetAppIcon />}
-                    disabled={!authentication}
+                    disabled={applicationJsonArray?.length < 1} // This check works correctly for both Number and null
```
EjPlatzer, line 112: I think the comment doesn't need to stay here long-term. It's good to test it, but since that's a feature of JavaScript it doesn't necessarily require a comment.

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -57,42 +58,33 @@ const StaffMenu = ({ userProfile, authentication }) => {
...
-  }, []);
-
   useEffect(() => {
-    generateCSVData(applications);
-  }, [applications, generateCSVData]);
+    setApplicationJsonArray(
+      applications
+        ?.filter((applicationDetails) => applicationDetails?.DateSubmitted) // Only add the applications that have been submitted
+        ?.map(({ Applicants, ApartmentChoices, ...applicationDetails }) => {
+          const applicationID = applicationDetails.ApplicationID;
```
EjPlatzer, line 66: It might be more clear to use `applicationDetails.ApplicationID` directly in the two spots below, rather than storing it as this `applicationID` variable. That way, you can clearly see the distinction between the ID coming from the application versus the ID coming from the ApartmentChoice/Applicant

I'm still wondering though - shouldn't it be impossible to have a mismatch between these? Like, the application ID DEFINES the relationship between these entities, so an Applicant with no application ID should not belong to an application, right?

EjPlatzer (approved):
This looks good! I haven't looked at the styling yet but functionally this is a nice improvement. I did leave a couple of comments for possible future improvements but in the meantime this is ready to merge.

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: Agreed

NerdyGriffin:
```suggestion
                    disabled={applicationJsonArray?.length < 1}
```

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: That is true, I think that was because the Applicant object on the frontend didn't originally contain the ApplicationID, which A.) should be something that I add to the StudentApplication code and B.) would not effect the StaffMenu anyway, so I think you are correct that this is no longer needed

I think another reason was that the Applicants array is an array within the ApartmentApplication, so it must always correspond to the same ApplicationID, and I wasn't clear until now that we would include the ApplicationID within the Applicant object received from the API. In their combine/nested form, the ApplicationID in the Applicant objects is redundant because it is nested within the ApartmentApplication object, but the ApplicationID is a necessity when converting it to a CSV

PR #1072, opened 2021-04-08
Apartapp - Change over to combined ApplicationDetails obj in state

NerdyGriffin:
- [x] Merge 'apartapp-style-tweak' into this branch for the latest improvements
  - [x] Change the base branch of this PR to 'apartapp-style-tweak'
- [x] ~Investigate `useReducer` for handling `applicationDetails` rather than `useState`~
The research into `useReducer` will have to wait until later, we need this update to be fully functional within the next few days


Nick-Noormand (commented):
Besides the one comment that I left of a bit of confusion the code looks good.

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -57,26 +57,29 @@ const StaffMenu = ({ userProfile, authentication }) => {
...
+    setApplicationsForCSV(
       applications
         ?.filter((applicationDetails) => applicationDetails?.DateSubmitted) // Only add the applications that have been submitted
         ?.map(({ Applicants, ApartmentChoices, ...applicationDetails }) => {
-          const applicationID = applicationDetails.ApplicationID;
-
-          setApplicantJsonArray((prevApplicantsJsonArray) => [
-            ...prevApplicantsJsonArray,
-            ...Applicants.map(({ Profile, StudentID, ...applicant }) => ({
-              ApplicationID: applicant.ApplicationID ?? applicationID,
```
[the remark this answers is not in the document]
NerdyGriffin: The ApplicationID is built into the objects when they are sent from the backend, so it is impossible for the API to add the applicant objects to this array without also giving them the ApplicationID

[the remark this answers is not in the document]
NerdyGriffin: Basically what this line does is set `applicant.ApplicationID = applicant.ApplicantID` (unchanged), unless `applicant.ApplicantID` is null or undefined, in which case it sets it to the thing on the right side of the `??`, which is the applicationID that was extracted from the `applicationDetails`

This is called the "Nullish coalescing operator", it is one of my new favorite things 😆 
Here is a place you can read and see examples <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator>

src/views/ApartmentApp/components/StaffMenu/components/ApplicationTable/components/ApplicationTableHead/index.js
```diff
@@ -3,16 +3,16 @@ import PropTypes from 'prop-types';
...
-  { id: 'hall', numeric: false, disablePadding: true, label: 'First Hall Choice' },
-  { id: 'total-pts', numeric: true, disablePadding: false, label: 'Total Points' },
-  { id: 'avg-pts', numeric: true, disablePadding: false, label: 'Avg. Points' },
+  { id: 'ApplicationID', numeric: true, disablePadding: false, label: 'Application ID' },
+  { id: 'EditorUsername', numeric: false, disablePadding: true, label: 'Editor' },
+  { id: 'NumApplicants', numeric: true, disablePadding: false, label: '# of Applicants' },
+  { id: 'Gender', numeric: false, disablePadding: true, label: 'Gender' },
+  { id: 'FirstHall', numeric: false, disablePadding: true, label: 'First Hall Choice' },
+  { id: 'TotalPoints', numeric: true, disablePadding: false, label: 'Total Points' },
+  { id: 'AvgPoints', numeric: true, disablePadding: false, label: 'Avg. Points' },
```
NerdyGriffin, line 12: This is the reason why the sort feature was not working on the staff table! 
🤦

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
```diff
@@ -40,37 +39,13 @@ const ApplicantList = ({
...
       // Send the selected username to the parent component
-      onSearchSubmit(theChosenOne);
-    }
-  };
-
-  /**
-   * Callback for changing the application editor
-   * @param {StudentProfileInfo} profile The StudentProfileInfo object for the person who is to be made the application editor
-   */
-  const handleChangeEditor = (profile) => {
```
joshuaRogers24, line 57: Where did this and the function below go?

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
```diff
@@ -26,7 +26,7 @@ import ClearIcon from '@material-ui/icons/Clear';
  * @param {Number} props.index The index of this list item
  * @param {Number} props.hallRank The rank assigned to this hall by the user
  * @param {String} props.hallName The name of the apartment hall
- * @param {ApartmentChoice[]} props.preferredHalls Array of apartment choices
+ * @param {ApartmentChoice[]} props.apartmentChoices Array of apartment choices
```
joshuaRogers24, line 29: Actually, I think HALLChoices would be more accurate. Of course, this is what we call it in the API, isn't it...

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -490,9 +528,23 @@ const StudentApplication = ({ userProfile, authentication }) => {
     handleCloseOkay();
   };
 
+  /**
+   * Submit the current application as completed
+   *
+   * @async
+   * @function changeApplicationEditor
```
joshuaRogers24, line 535:
```suggestion
   * @function submitApplication
```

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
joshuaRogers24: Oh, are these now just handled in StudentApplication/index.js?

joshuaRogers24 (changes requested):
Good work. A few suggestions and questions.

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
NerdyGriffin: Basically, all it did was set the onChange___ of the component to this function, and then this function would call onChangeEditor without adding or doing anything else, so you can just "cut out the middle man" and pass the callback from the ApplicantList prop directly to the callback in the ApplicantListItem component below

Evan pointed this out in other places, so I figured I should clean up any others that I noticed later

NerdyGriffin: When commenting on parts like this, it helps to select also the lines after the function name, so that it is easy to view the full context of what your comment is referring to

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
NerdyGriffin: actually there are a few reasons why we use apartmentChoice instead of hallChoice, but of course I forgot since it was a while ago. It has something to do with being specific to the apartment part and differentiating between future housing-related projects. 
ALso because halls include all sorts of things that are not even dorms, so it is best to keep it specific.

NerdyGriffin: ApartmentHalls have ApartmentChoices, 

DormHalls might have DormChoices some day 😆

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
joshuaRogers24: Oh, gotcha! I will try to remember that next time,

src/views/ApartmentApp/components/StudentApplication/components/HallSelection/components/HallListItem/index.js
joshuaRogers24: Okay, yeah, makes sense!

joshuaRogers24 (approved):

PR #1073, opened 2021-04-11
Added link to credits earned portion

aidanperez2000:
I just added a link to the credits earned portion that links to the attended page.  Right now, the days finished portion also links to this page.  I will work on making it so only the credits earned portion links to that page.  Let me know if you have any suggestions.


josh-peters-2357:
@aidanperez2000 Could you explain to me what this is supposed to do? I've looked at it for quite a while now and still don't understand. The doughnuts aren't links for me.

aidanperez2000:
@josh-peters-2357 The whole doughnut does not link to anything.  Only the CL&W credits earned portion links to the attended page.

josh-peters-2357 (changes requested):
Two issues @aidanperez2000 @joshuaRogers24 .

First, I'd really suggest having the mouse change style upon hover. The "More Credits" button in this frame does this so people know it goes somewhere. But a computer user using this might not ever try to click it. We should probably change that before merging.

Second, on mobile, I get an issue using my local server. I've attached a screenshot. That should get resolved.

Overall, it's interesting and good work. But one of the most obvious issues I saw outside technical things was that a mobile user can only see numbers that the doughnut sections would display by *tapping* on them- the user can just hover and they appear. To be sure the credits earned appears as a number below the doughnut too, so they don't have to tap the circle to see. But if they want to see the number as it hovers over the doughnut when they tap the right section, too bad, because that's impossible now. To me, that sounds like a problem, and I wonder if the easiest way to solve that would be to incorporate a link to credits earned into this component as a button like the More Credits. Just an idea.

20210415_030757000_iOS

src/views/Home/components/CLWCreditsDaysLeft/index.js
```diff
@@ -187,14 +187,23 @@ export default class CLWCreditsDaysLeft extends Component {
       <Card>
         <CardContent>
           <Grid container direction="row" alignItems="center">
-            <Grid item xs={7} align="left">
+            <Grid item xs={4} align="left">
+              <Button
+                variant="contained"
+                style={style.button}
+                onClick={(event) => (window.location.href = '/attended')}
```
EjPlatzer, line 194: Rather than doing this browser navigation, using the browser's `window` object to change the URL, 360 uses a routing package called `react-router`. The main advantage is that, while browser navigation reloads the entire page from scratch, `react-router` just changes which component is displayed at the view level - i.e. instead of `Home` we render `Events`. This means we aren't reloading the entire site from scratch, so the user has a much quicker, much smoother experience.

If you import the `Link` component from the `react-router-dom` package at the top of this file with the other imports, then you can use it for the behavior of the button:
```suggestion
                component={Link}
                to="/attended"
```

src/views/Home/components/CLWCreditsDaysLeft/index.js
```diff
@@ -187,14 +187,23 @@ export default class CLWCreditsDaysLeft extends Component {
...
+            <Grid item xs={4} align="center">
               <CardHeader title={this.state.currSessionDescription} />
             </Grid>
-            <Grid item xs={5} align="right">
+            <Grid item xs={4} align="right">
               <Button
                 variant="contained"
                 style={style.button}
-                onClick={(event) => (window.location.href = '/events?CLW')}
+                onClick={(event) => (window.location.href = '/events?CLW%20Credits')}
```
EjPlatzer, line 206: Just like above, we can make the routing much smoother and quicker using the `Link` component from `react-router-dom`:
```suggestion
                component={Link}
                to="/events?CLW%20Credits
```

EjPlatzer (changes requested):
This is an excellent upgrade to the CLWCreditsDaysLeft component!

I know it must've been frustrating figuring out the intricacies of linking from the donut only to abandon that approach, but I think having another button is ultimately a better solution.

There is still one more thing that needs to be updated, though. Since this is a React application, we want to avoid using the `window` object for navigation, because that causes the entire page to reload, which is very slow and looks weird to the user. Instead, we use `react-router`, and it's browser implementation `react-router-dom` to update the component tree programatically, without ever reloading the page. The result is much much faster and looks smoother to the user.

EjPlatzer (dismissed):
LGTM! The new link is useful for people, and fixing the link to CLW Credit events is desperately needed!

PR #1074, opened 2021-04-11
ApartApp - More user-friendly date display

NerdyGriffin:


src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -53,9 +53,7 @@ const StaffMenu = ({ userProfile, authentication }) => {
     loadAllCurrentApplications();
 
     // Generate string of today's date in ISO format for use in CSV filename
-    let date = new Date();
-    let dateStr = DateTime.fromJSDate(date).toISODate();
-    setDateStr(dateStr);
+    setDateStr(DateTime.now().toISODate());
```
EjPlatzer, line 56: You probably don't need the timezone offset in the ISO Date string output:
```suggestion
    setDateStr(DateTime.now().toISODate({ includeOffset: false }));
```

src/views/ApartmentApp/components/StudentApplication/components/Agreements/index.js
```diff
@@ -19,51 +20,63 @@ import {
...
-      checked: false,
-      label:
-        'We certify that all information provided on this application is accurate, to the best of our knowledge',
-    },
-  ]);
+  const [checkboxes, setCheckboxes] = useState([]);
+
+  useEffect(() => {
+    const loadAgreements = async () => {
+      const currentYear = Number(new Date().getFullYear);
```
EjPlatzer, line 27: I don't think you need the explicit conversion to a number here, but you do definitely need to call the `getFullYear` method rather than simply referencing it:
```suggestion
      const currentYear = new Date().getFullYear();
```

src/views/ApartmentApp/components/StudentApplication/components/InstructionsCard/index.js
```diff
@@ -23,12 +24,11 @@ const InstructionsCard = () => {
 
   useEffect(() => {
     const getYear = () => setThisYear(new Date().getFullYear());
+    const loadApartmentSelectionDate = async () =>
+      setApartmentSelectionDate(await housing.getApartmentSelectionDate());
 
     getYear();
```
EjPlatzer, line 30:
```suggestion
    setThisYear(new Date().getFullYear());
    setApartmentSelectionDate(await housing.getApartmentSelectionDate();
```

EjPlatzer (commented):
Just left a few code quality suggestions.

src/views/ApartmentApp/components/StaffMenu/index.js
NerdyGriffin: Ah yes, thank you, I was trying to figure out how to efficiently format that

src/views/ApartmentApp/components/StudentApplication/components/Agreements/index.js
NerdyGriffin: I think I added that conversion because I didn't know why it VSCode was detecting it as the wrong thing. Now I know that it must have been because I forgot the `()` after the method call

src/views/ApartmentApp/components/StudentApplication/components/InstructionsCard/index.js
NerdyGriffin: This caused a build error due to the use of `await` outside of an async function. 
Should the selection date be returned to a helper function, or it is possible to do something like 
```
useEffect(async () => {
  // the stuff inside
}, []);
```

EjPlatzer: Oh right, I forgot about the needed `async` context. The callback passed to useEffect unfortunately cannot be async directly. The two workarounds are to define an async function and call it within the callback (which you were already doing), or use `Promise.then()` syntax. It's your call which you think is cleaner in this context.

src/views/ApartmentApp/components/StudentApplication/components/InstructionsCard/index.js
```diff
@@ -22,13 +23,10 @@ const InstructionsCard = () => {
...
-
-    getYear();
-
-    // let selectionDate = housing.getApartmentSelectionDate();
-    let selectionDate = 'Apr. 27'; // API endpoint for this is planned for the future
-    setApartmentSelectionDate(selectionDate);
+    setThisYear(new Date().getFullYear());
+    housing
+      .getApartmentSelectionDate()
+      .then((selectionDate) => setApartmentSelectionDate(selectionDate));
```
EjPlatzer, line 29: Fun fact about `Promise.then` and other functions that expect a callback - when you simply want to call a function on the parameters, you can pass that function directly, rather than passing an arrow function that does the same thing:
```suggestion
      .then(setApartmentSelectionDate);
```
In this case, I think it actually obscures the code more than simplifying it, but it's a neat thing to keep in mind.

It can be especially helpful defining the `onClick` prop of components. For example, rather than having an input with
`onClick={(event) => setInputValue(event.target.value)}`
you can simply do
`onClick={setInputValue}`

EjPlatzer (approved):
These changes look good to me!

I mentioned one JavaScript quirk that is nice to know about, but doesn't actually improve the code in this context in my opinion.

src/views/ApartmentApp/components/StudentApplication/components/InstructionsCard/index.js
NerdyGriffin: I would want to double check the `.then` documentation, because I know that there can be unexpected behavior if there are any optional secondary or tertiary parameters involved. 
I suspect this will change again in the future when we actually make the endpoint on the API side
But I agree in general, and in this case the simplified form might make it look a little more confusing to read, so i'll leave it as-is for now

PR #1075, opened 2021-04-11
ApartApp - UI Style Tweaks

NerdyGriffin:
Tweaks and improvements to UI styling, partially inspired by user feedback / user testing

I rearranged the Grid component structure to make better use of the responsive scaling features in the way they were intended to work. The entire group of components is now nested inside an outer Grid item which allow for the total page width to be consistently adjusted, in a way that automatically makes the inner nested Grid items display with the correct widths more reliably. 
The end result is a page that behaves just as good as before on smaller devices (mobile scaling), while making the layout respond much better on large and extra-large displays (which used to have weird layout problems, unnaturally stretched out components, or Cards not lined up correctly)


src/views/ApartmentApp/apartmentApp.scss
```diff
@@ -70,15 +70,16 @@
     font-size: medium;
   }
 
-  & .save-bar {
+  & .MuiGrid-item.sticky-page-bottom-bar {
```
EjPlatzer, line 73: I like that this makes the bottom bar for starting/editing/viewing your application sticky as well. However, with that one sticky, the top one seems extremely redundant:
image
Is there a reason to keep the top bar around that I'm not aware of?

src/views/ApartmentApp/components/StudentApplication/components/ApartmentHeader/index.js
```diff
@@ -8,68 +8,83 @@ const ApartmentHeader = ({
...
-                  View your application
-                </Button>
-              </Grid>
-            </React.Fragment>
-          )}
+          {!applicationID
+            ? newApplicationContent
+            : userProfile.AD_Username === editorUsername
+            ? editApplicationContent
+            : viewApplicationContent}
```
EjPlatzer, line 87: All three of these variables are literally the same, except for a different `<Typography>` content and button label. I think this component would end up much smaller and much more readable if you made those two pieces into variables instead.
```suggestion
          <Grid item xs={6} sm={8}>
            <Typography variant="body1">{text}</Typography>
          </Grid>
          <Grid item xs={6} sm={4}>
            <Button
              variant="contained"
              onClick={onShowApplication}
              color="primary"
              fullWidth
              disabled={applicationCardsOpen}
            >
              {buttonLabel}
            </Button>
          </Grid>
```

Then you could replace all the logic above the return with simply:
```
let text, buttonLabel;

  if (!applicationID) {
    text = (
      <>
        You are not on any applications for the current semester.
        <br />
        Click the button to begin a new application.
      </>
    );
    buttonLabel = 'Create a new application';
  } else if (userProfile.AD_Username === editorUsername) {
    text = 'You are on an existing application for this semester.';
    buttonLabel = 'Edit your application';
  } else {
    text = (
      <>
        You are on an existing application for this semester, but you are not assigned as the
        editor.
        <br />
        Only the application editor may edit the application.
      </>
    );
    buttonLabel = 'View your application';
  }
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -439,7 +444,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
       result = false;
     }
     console.log('result of saving: ' + result); //! DEBUG
-    if (result !== null && result !== false && result !== -1 && typeof result === 'number') {
+    if (result && typeof result === 'number') {
```
EjPlatzer, line 447: I assume you've changed the logic in the backend to no longer return `-1` for a bad result, since `-1` is a truthy value in JavaScript?

EjPlatzer (changes requested):
A couple of minor improvements, but overall this looks like a great upgrade!

src/views/ApartmentApp/apartmentApp.scss
NerdyGriffin: I think no, there is no reason...

src/views/ApartmentApp/components/StudentApplication/components/ApartmentHeader/index.js
NerdyGriffin: Oh that's a really good point, that was annoying me trying to figure out how to simplify that when there was so much similar between them

NerdyGriffin: Also, I never knew that syntax was a thing:
```
<>
  stuff
</>
```

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Correct, we phased out the -1 a while ago, but it was left it for a bit until all the various features had a chance to be reviewed and merged

src/views/ApartmentApp/components/StudentApplication/components/ApartmentHeader/index.js
NerdyGriffin: I just pushed an change based on this

src/views/ApartmentApp/apartmentApp.scss
NerdyGriffin: I will definitely remove one of them, but I am not sure whether to have it on top or on the bottom. It will remain sticky in both cases.
I think on the bottom makes it easier to read the instructions card, and there is the nice visual aspect of it keeping the view/edit/new buttons in the same place as the save/submit, so it provides a sense of continuity for the user,

Let me know your thoughts, and any suggestions for renaming the `ApartmentHeader` if it gets moved to the bottom (or, I could possibly just migrate the features to the `BottomBar` component, which currently holds the save/submit stuff, since it is just a boolean prop that determines which one gets displayed)

EjPlatzer: Hmm, well I'm not privy to all the requirements and use cases of the app, but I would think the bottom is nicer because, as you mentioned, it's easier to read the instructions. Keeping it at the bottom would also mean that people have to at least notice the instructions before they create a new application.

Reusing `BottomBar` would make sense to me.

src/views/ApartmentApp/components/StudentApplication/components/ApartmentHeader/index.js
EjPlatzer: Yes, `<>SomeJSX</>` is shorthand for `<React.Fragment>SomeJSX</React.Fragment>`. Being shorthand, it's less obvious/explicit, but I think it's at least as readable and still conveys the intent of wrapping the elements together but not providing any wrapping functionality.

src/views/ApartmentApp/components/StudentApplication/components/ApartmentHeader/index.js
```diff
@@ -8,83 +8,47 @@ const ApartmentHeader = ({
...
-        </Button>
-      </Grid>
-    </React.Fragment>
-  );
+  const headerContent = !applicationID
+    ? {
+        primaryText: 'You are not on any applications for the current semester.',
+        secondaryText: 'Click the button to begin a new application.',
+        buttonLabel: 'Create a new application',
+      }
```
EjPlatzer, line 16: Now that I look at it this way, `primaryText`, `secondaryText`, and `buttonLabel` seem more like props. Perhaps you could extract the content logic and make this component purely presentational? That might also make it possible to meld into the `BottomBar` more smoothly.

Nick-Noormand (commented):
Overall the branch changes look good, but there are still a few questions that I left on the review to try to understand whats up better.

src/views/ApartmentApp/apartmentApp.scss
EjPlatzer: A common practice, especially since the headings on the app are our primary (gordon blue), would be to set the button to our secondary (cyan). This would visually distinguish it from the rest of the page without cluttering the site's colors.

src/views/ApartmentApp/components/StaffMenu/index.js
```diff
@@ -9,19 +9,18 @@ import housing from '../../../../services/housing';
...
 
 /**
  * Renders the page for the apartment application housing staff menu
  * @param {Object} props The React component props
- * @param {*} props.authentication The user authentication
  * @param {StudentProfileInfo} props.userProfile The student profile info of the current user
  * @returns {JSX.Element} JSX Element for the staff menu web page
  */
-const StaffMenu = ({ userProfile, authentication }) => {
+const StaffMenu = ({ userProfile }) => {
```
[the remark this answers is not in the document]
NerdyGriffin: It was never actually used inside this component. It is used in the parent components, and it is only a boolean, so if you are even viewing this component at all, then the `authentication` prop must have been `true` in the parent component, thus making it needlessly redundant in this scope

src/views/ApartmentApp/components/StudentApplication/components/BottomBar/index.js
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import { Backdrop, Button, Card, CardContent, Grid, Typography } from '@material-ui/core/';
+import GordonLoader from '../../../../../../components/Loader';
+import GordonDialogBox from '../../../../../../components/GordonDialogBox';
+import SaveButton from './components/SaveButton';
+
+const submitAlertText = (
+  <span>
+    This feature is not yet implemented.
```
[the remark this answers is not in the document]
NerdyGriffin: As discussed with @EjPlatzer , I will be moving the `ApartmentHeader` stuff into this component, for various logistical and stylistic reasons. I guess I will do that right now, because the "Delete" option will likely need to go in here as well

src/views/ApartmentApp/components/StudentApplication/components/ApartmentHeader/index.js
NerdyGriffin: On the other hand, doing so would add a lot more lines of logic in the parent component, it becomes a tradeoff between abstraction vs. modularity
In it's current concept, the StudentApplication tells the bar what the state of the application is using a few simple abstracted props, and then this component display whatever is appropriate for that state, so the component structure of the parent (StudentApplication) is more simplified and easier to understand

src/views/ApartmentApp/components/StudentApplication/components/BottomBar/index.js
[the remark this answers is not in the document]
NerdyGriffin: Also the key detail is that in the picture, that was not the bottom bar, that was the ApartmentHeader object but pinned at the bottom. A merged version will be pushed in about 15 minutes or less

src/views/ApartmentApp/apartmentApp.scss
NerdyGriffin: Nice, I'll play around with that right now

EjPlatzer (changes requested):
I think this could be made more readable by using the ternary operator less, but otherwise it's looking good to me!

src/views/ApartmentApp/components/StudentApplication/components/BottomBar/index.js
```diff
@@ -0,0 +1,136 @@
...
+  onCloseDialog,
+  onCloseOkay,
+  onSaveButtonClick,
+  onShowApplication,
+  onSubmitAppAccepted,
+  onSubmitButtonClick,
+}) => {
+  const dynamicContent = !applicationCardsOpen
+    ? !applicationID
+      ? {
```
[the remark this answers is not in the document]
NerdyGriffin: Yeah, I switched it there are back and then left it like this cause I was in a hurry, but it was confusing me as well

src/views/ApartmentApp/components/StudentApplication/components/BottomBar/index.js
```diff
@@ -0,0 +1,136 @@
...
+              >
+                {dynamicContent.buttonLabel}
+              </Button>
+            </Grid>
+          ) : (
+            <Grid container item xs={12} sm={6} lg={4} spacing={2}>
+              <Grid item xs>
+                <SaveButton
+                  saving={saving}
+                  onClick={canEditApplication && onSaveButtonClick}
```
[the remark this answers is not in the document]
NerdyGriffin: I figured if it would for abstract objects, JSX components, and every other thing I have tried so far, then it would probably work for this too, so I tried it out. My original plan was a `canEditApplication ? onSaveButtonClick : null` which I knew was valid syntax with functions from past things and examples in documentation, so it seemed like a reasonable guess

src/views/ApartmentApp/components/StudentApplication/components/BottomBar/index.js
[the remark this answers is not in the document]
NerdyGriffin: A version of this has now been pushed

src/views/ApartmentApp/components/StudentApplication/components/BottomBar/index.js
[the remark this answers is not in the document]
EjPlatzer: Right, that makes sense because of course `&&` is a Logical AND, not a Boolean AND in JavaScript, so it's value resolves to the first falsy value or the last value. That is, when `canEditApplication` is false, `&&` returns false, but when `canEditApplication` is true, `&&` will evaluate onSaveButtonClick (which is a truthy value) and return that. If at some point `canEditApplication` were `undefined`, then `&&` would actually return `undefined` which would be fine in this case, but might be unexpected for people like me who think of it as a Boolean operator rather than a 'Logical Operator'.

EjPlatzer (approved):
This looks good to me!

PR #1077, opened 2021-04-12
ApartApp - Bug fix for date submitted saying "Invalid DateTime"

NerdyGriffin:
This bug was revealed during the in-class demo


joshuaRogers24 (approved):
Nice.

PR #1080, opened 2021-04-15
Replace white background pictures with transparent bg

josh-peters-2357:
I didn't do the splashes; I think those are meant to have a solid background?

Anyways, here is a new pull request. The old one was ignored for so long, the development branch was ages ahead of my pull request and so it would've been too much work to resolve it I think.


Kindlewing (approved):
LGTM!

PR #1081, opened 2021-04-15
Refactor MyProfile and PublicProfile

EjPlatzer:
This pull request represents a refactoring and major simplification of the code related to Profiles (both `PublicProfile` and `MyProfile`). These two views have a whole lot of common code, but the differences between them are vital because they usually relate to privacy.

The majority of these refactorings consist of decreasing repetitive code, especially by extracting repetitious code into reusable components. Additionally, there was a decent amount of redundant layout components (e.g. overly nested Material-UI grids), and state that was more complicated than necessary.

In particular:

- The Personal Info and Office Info lists now make use of the generalized component ProfileInfoListItem. This component encapsulates all of the layout and display logic so that the other two can focus on formatting the data and preserving privacy.
- the public `Profile` component has been renamed to `PublicProfile` for more clear contrast with `MyProfile`. It has also been refactored into a functional, hooks-based component. This refactoring revealed that a lot of dead code had been left in it after the cards it shared with `MyProfile` were extracted last summer. This dead code has been pruned.
- A new `Profile` component has been created which encapsulates the layout of the `PublicProfile` and `MyProfile` pages. It takes a prop `myProf` which is true if it is displaying on `MyProfile` and false if it is displaying on `PublicProfile`. The code between these two pages was nearly identical, with only an additional card for the VictoryPromise and a slightly different layout differentiating them. This is mostly possible because of work done last summer encapsulating the cards on the profile pages.
- The components for the cards on the profile pages (e.g. `Identification`, `MembershipsList`, `PersonalInfoList`, etc.) have been moved from `src/components` to `src/components/Profile/components`, since they are all only referenced once within the new `Profile` component. This has the added benefit of decluttering the `src/components` folder somewhat.
- The social media links in the `Identification` component have been restructured and streamlined. `src/socialMedia.js` now contains - in addition to the link prefixes and icons for each supported social media - a new list of supported platforms. This means that the platforms can be operated over algorithmically instead of repeatedly being individually specified. For example, instead of having state variables `facebookLink`, `twitterLink`, etc., we now have a single state object `links` that can be updated per-platform using the platforms list in `socialMedia.js`

There are still two things needed to complete the refactoring of all profile-related components:

1. The photo display and update logic in `Identification` should be extracted to it's own component. This is to keep the components manageable in size (instead of e.g. the 720 lines of dense code that `Identification` is currently), and also to aid in identifying a possible reusable `photoCropper` component in the future
2. The `SchedulePanel` and all of its sub-components are in need of a deep refactor. They are not only out-dated, but also bug-ridden - either because they were never polished properly or because the dependencies have outstripped us.


EjPlatzer:
This pull request entered an untenable state because of a merge. The same changes, but with a clean merge, will be applied by #1092.

Issue #742, opened 2020-06-04
Involvements shows up too many times


aridospassos:
Could look a lot better if we made a way for the involvements to be combined into one year instead of by semesters.

cpabbot:
Not necessarily by year either but by involvement. Would be much neater than having 8 cards of the same thing for something a student participated in for all 8 semesters. (instead would just be one card with dates or # of semesters)

speroJay:
It appears that (at least one reason) each semester is listed, instead of all the semesters as one membership entry (as in the actual printout transcript), is so that each membership can be accessed for the specified time period.  For example, a student who was part of IMT for two semesters can go back and see who is the leader for each semester, as the link leads to a page with the corresponding year and information.  Furthermore, if a student had a different role a different year, this allows them to see the activity page for that year as that role.  It also allows a leader in a group to deny or approve a student for a previous semester/allows a student to ask to be a part of a group so it will show up on their transcript (maybe they didn't know about 360 when they were part of the group and now want it to show up on the prepared transcript).  A drop down menu for each activity might be a solution, or some other way to still access specified activity for a specified semester.  The question is, is it something that will look good and be useful (and of course just flat out work).

bennettforkner:
@speroJay , I like your note that it is good to be able to view a historical involvement as well as the one for the current semester. Still, I am definitely for de-cluttering students' involvement history on their profile.

To that extent, the clearest to improve this is likely to combine involvements on a student's profile. If we do something like the very bare-bones example below, we could compile a student's history with an involvement into one card per.
![image](https://user-images.githubusercontent.com/43764277/110731196-de334180-81ef-11eb-9c39-5fd65b7ef2d6.jpeg)
That would make it more similar to a very nicely formatted resumé and would show the involvement data related to a student on that student's profile.

Then, involvement homepage would be a good place to show the history of the club / activity. A term selector would be one way to show a timeline as you said. It could probably go on the top right of the involvement page on desktop and on the right side of the 'Members' label on mobile.

It may be difficult to finish this one before we start to wrap up this semester, so keep that in mind as you consider how you want to proceed with your changes.

speroJay:
I like the way that works!  
The biggest issue I see with the previous timeline for the involvement's homepage.  As it currently stands, if a student was a member of a club in say Fall 2019, they can only see the members of the club for Fall 2019.  For all other semesters, they can only see the information made for the group that was public.  
[...]

PR #1083, opened 2021-04-22
My profile involvments clutter

speroJay:
Current Output:
   360 MyProfile page will show only one entry for each activity.  Each entry shows a list of sessions(dates) with corresponding membership role(guest,  leader, etc.).  Image for the activity shows.
   Each entry acts as a link to the earliest date's activity page.  Privacy toggles don't entirely work.  

image

Desired Output:
  See conversation at: https://github.com/gordon-cs/gordon-360-ui/issues/742

==================================================================

* Create an array of activities where each element is an array of elements with the same activity code.
    (gordon-360-ui\src\services\membership.js)

image

* Display the title for each activity by displaying the title of the first element in every array
    (src\components\MembershipsList\index.js)

*  Display a list of sessions/membership status by accessing each array's elements individually 
      (src\components\MembershipsList\components\MembershipInfoCard\index.js)

*Auxiliary*
Get activities by activity code:  getMembershipsByActivityCode (src\services\user.js)


speroJay:
Github issue: https://github.com/gordon-cs/gordon-360-ui/pull/1083

As of 10 / 13 / 21:

- Each involvement has its own `<Accordion>` and image, with the dropdown result being a list of the session/position (s) the user has been involved in, and a privacy toggle for **each** item on the list.  

- The privacy toggle will change the backend data accordingly, and the snackbar will appear.
- Changing the privacy will cause the ReactComponent `Membership-List` to refresh, which causes all accordions to close.  The toggle does not change appearance.
- To re-open the accordion, the first click times out, and the second click is successful. However, as mentioned earlier, the privacy toggle does not change appearance.
- When the whole page is refreshed, the toggle reflects the proper privacy setting, and accordions can be opened as normal.
image
image
image


Potential Solutions:

- Hold off the re-render until the privacy toggle changes.  It’s possible Membership-List refreshes too soon, the accordion closes and the toggle can’t update.  Changing the render order seemed the most likely solution, however I tried leaving the `<Accordion>` open constantly, but it still didn’t work.  It could be  `<Accordion>`  “refreshed” but didn’t change appearance.
- Turn the toggle into its own component (Kevin's idea).  This would work (theoretically), but it could really muddy the files.  Is a component necessary for a toggle?
- Potentially keys?  It could solve how every list item is updated, but I’m not so sure it would actually change the order of how the nested components update.


Update 11 / 12 / 21:

- https://reactjs.org/docs/hooks-custom.html seems to use a similar example.  Hooks are already used instead of component classes (that explains a lot).  The toggle itself (MUI) must have an on-off, but should each individual membership have it’s own privacy on/off `useState` for privacy?   A lot of things seem to update when privacy updates, and it's not certain that it would stop just if privacy toggle had its own state.  
-  To coincide with these changes, perhaps `onprivacytoggle()` should be moved to each involvement card, as it seems to be more closely associated with it.


Relevant Locations: 
\gordon-360-ui\src\components\Profile\components\MembershipsList\index.js
src\components\Profile\components\MembershipsList\components\MembershipInfoCard\index.js

speroJay:
Moved the privacy toggles to their own component/hook so that each toggle keeps track of its own state independently of the whole list.  Although the accordions still snap closed after the privacy toggled is clicked and require two clicks to re-open (as no error occurs, I'm assuming it is because they think they are still open ), the slider changes visually without a page refresh.  Lock icons have also returned for activities with forced privacy.

EjPlatzer:
Superceded by #1918

Issue #962, opened 2021-01-29
Refactor deprecated method `componentWillMount`


EjPlatzer:
The React component method `componentWillMount` has been deprecated. In a class component, it should be replaced with `componentDidMount`, which is virtually identical.

Thanks to a generous contribution from the GitHub community, [branch removeWillMount](https://github.com/gordon-cs/gordon-360-ui/tree/removeWillMount) has the first step of this process complete. The next step is to clean up components with duplicate calls of componentDidMount, and then testing a local run to ensure the updated components function correctly.

dezy-x2:
Hi, I would like to take this on. Just to clarify I just need to swap out `componentWillMount()` with `componentDidMount` correct? Thank you

josh-peters-2357:
@dezy-x2 @EjPlatzer Maybe we could link #964 to this issue so it can be closed?

EjPlatzer:
That was a good start to the issue, but as  you can see from the issue description, there is still more to be done. At this point, there's some cleanup that needs to be done and then a lot of testing the system to ensure the changes haven't broken anything.

PR #1088, opened 2021-04-26
Remove deprecated lifecycle method componentWillMount

EjPlatzer:
The React lifecycle method `componentWillMount` has been deprecated since React 16.3 because of it's potential for misunderstanding and misuse leading to common bugs. React now produces a console warning about `componentWillMount`, and it will be removed in a coming update.

For this reason, it has been refactored - most commonly to use `componentDidMount` in it's place. In some cases (e.g. loading authentication state in app.js), all that is needed is loading the initial state, which is done in the constructor. 

Solves #962

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -47,7 +47,7 @@ export default class App extends Component {
 
     this.setState({ error, errorInfo });
   }
-  componentWillMount() {
+  componentDidMount() {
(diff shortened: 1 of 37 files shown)


bennettforkner (approved):
LGTM

PR #1089, opened 2021-04-26
ApartApp - Submit application feature

NerdyGriffin:


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -535,15 +535,33 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+            // loadApplication(); //? Coming soon to a feature near you
+          }
+        })
+        .catch((err) => {
+          showGenericSubmitError();
+        });
     } else {
-      console.log('Not all applicants are valid'); //! DEBUG:
+      showGenericSubmitError();
     }
```
joshuaRogers24, line 565: If this case occurs when the application ID has not yet been set, then should we not tell them to make sure they save their application first?

joshuaRogers24 (changes requested):
Nice. Question, though...

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: The UI disables to button entirely in that case, so it should never get to that code without having an applicationID, unless something is actually broken

joshuaRogers24: I see, this tests for the "actually broken" case.

joshuaRogers24 (approved):

PR #1090, opened 2021-04-26
ApartApp - Urgent bug fixes

NerdyGriffin:
- [ ] Applicant student/staff check failed during demo
- [ ] Applicant gender check failed during demo
- [ ] Halls sometimes not loading when running locally

### DIFF
--- a/src/views/ApartmentApp/components/StudentApplication/index.js
+++ b/src/views/ApartmentApp/components/StudentApplication/index.js
@@ -110,20 +110,24 @@ const StudentApplication = ({ userProfile, authentication }) => {
     loadApplication();
   }, [userProfile]);
 
+  //! DEBUG
   useEffect(() => {
-    // setUnsavedChanges(true);
     //! DEBUG
     console.debug('Array state variable. Printing contents:');
+    //! DEBUG
     console.debug('EditorUsername:');
     console.debug(applicationDetails?.EditorProfile?.AD_Username);
+    //! DEBUG
     console.debug('Applicants:');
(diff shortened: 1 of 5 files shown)


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -144,12 +144,11 @@ const StudentApplication = ({ userProfile, authentication }) => {
   /**
    * Check whether a given applicant is valid for this current application
    *
-   * @async
    * @function isApplicantValid
    * @param {ApartmentApplicant} applicant The applicant to be checked
    * @return {Boolean} True if valid, otherwise false
    */
-  const isApplicantValid = async (applicant) => {
+  const isApplicantValid = (applicant) => {
```
NerdyGriffin, line 151: @EjPlatzer  I am guessing that the async behavior might have caused problems on the line with 
``` javascript
... else if (isApplicantValid(newApplicantObject)) { ...
```
Which was being handled as "true" in cases that clear should have been false,

I am not entirely familiar with the details of how async functions behave, so let me know your thoughts

EjPlatzer: Yes, if you were to call it like you mentioned, then that would definitely cause issues. An `async` function always returns a Promise, which will resolve to the value returned from the function once the function finishes executing. So in the case you mentioned, `if (isApplicantValid(newApplicantObject))` is equivalent to `if (Promise)` which evaluates to `if (true)`.

joshuaRogers24 (approved):
Nice. SO glad those were simple fixes.

PR #1093, opened 2021-04-29
Apartapp develop

joshuaRogers24:


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -535,15 +557,33 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
    */
   const submitApplication = async () => {
-    if (applicationDetails.Applicants.every((applicant) => isApplicantValid(applicant))) {
+    const showGenericSubmitError = () => {
+      setSnackbarText('Something went wrong while trying to submit the application.');
+      setSnackbarSeverity('error');
+      setSnackbarOpen(true);
+    };
+
+    if (!applicationDetails.Applicants.every((applicant) => isApplicantValid(applicant))) {
```
EjPlatzer, line 566: If `isApplicantValid` is still an `async` function, then this won't behave as expected, because calling `isApplicantValid(applicant)` will return a `Promise`, which is true because it's an `object`.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -535,15 +557,33 @@ const StudentApplication = ({ userProfile, authentication }) => {
    * @function submitApplication
    */
   const submitApplication = async () => {
-    if (applicationDetails.Applicants.every((applicant) => isApplicantValid(applicant))) {
+    const showGenericSubmitError = () => {
+      setSnackbarText('Something went wrong while trying to submit the application.');
+      setSnackbarSeverity('error');
+      setSnackbarOpen(true);
+    };
```
EjPlatzer, line 564: It would be a quick and easy update to convert this to a general function `createSnackbar(message, severity)` that can be used throughout the entire application.

EjPlatzer (changes requested):
These look like good updates! I did notice one bug where you're evaluating the truthiness of a Promise, rather than the value that the Promise resolves to.

src/views/ApartmentApp/components/StudentApplication/index.js
EjPlatzer: This is fixed in #1094 .

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Addresses in #1096

NerdyGriffin:
Now should be ready for review

Known issues: Attempting to view an application for which you are not the editor will return either a 401 or a 403 (I forgot which one)

NerdyGriffin:
See #1104 for final bug fixes

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -152,59 +171,67 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
     if (applicationDetails.Applicants.length >= MAX_NUM_APPLICANTS) {
       // Display an error if the user try to add an applicant when the list is full
-      setSnackbarText(`You cannot have more than ${MAX_NUM_APPLICANTS} applicants'`);
-      setSnackbarSeverity('warning');
+      createSnackbar(`You cannot have more than ${MAX_NUM_APPLICANTS} applicants`, 'warning');
       return false;
     }
 
+    applicant.Profile.fullName ??
+      (applicant.Profile.fullName = `${applicant.Profile.FirstName}  ${applicant.Profile.LastName}`);
```
EjPlatzer, line 192: This would be another opportunity for the nullish assignment operator:
```suggestion
    applicant.Profile.fullName ??= `${applicant.Profile.FirstName}  ${applicant.Profile.LastName}`;
```

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -55,76 +57,93 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+        throw e;
       }
-    };
-
-    loadApplication();
+    } finally {
+      setNewEditorProfile(null);
+      return result;
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
EjPlatzer: When testing locally, `eslint` wasn't warning me about missing dependencies. Was this left over from testing? It is almost never a good idea to disable that warning because it warns about some potential bugs that are VERY hard to track down otherwise.

NerdyGriffin:
@EjPlatzer All incoming changes have been merged (aside from a small change in #1106 ) so this branch is now ready for review

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -230,35 +262,41 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
-    if (profile) {
+    if (canEditApplication && profile) {
       if (
         applicationDetails.Applicants.some(
           (applicant) => applicant.Profile.AD_Username === profile.AD_Username,
         )
       ) {
+        if (unsavedChanges) {
+          saveApartmentApplication(applicationDetails);
+        }
```
EjPlatzer, line 299: Does this mean that all changes get saved without the user knowing when they switch editors?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -287,37 +330,54 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
-      // loadApplication(); //? Coming soon to a feature near you
-    } catch {
-      setSnackbarText('Something went wrong while trying to save the new application editor.');
-      setSnackbarSeverity('error');
-      setSnackbarOpen(true);
-      setSaving('failed');
+      if (result) {
+        try {
+          const result = loadApplication();
+          if (!result) {
```
EjPlatzer, line 346: It's a little bit confusing to shadow the `result` variable declared in the outer `try-catch` block by redeclaring `result` in this inner `try-catch` block. Could you use a different name to distinguish them?

src/services/housing.js
```diff
@@ -35,7 +35,7 @@ import './user'; // Needed for typedef of StudentProfileInfo
  * @property {DateTime} [BirthDate] The birthday of this applicant (only visible to housing admin)
  * @property {Number} [Age] The age of the student (in years) (only visible to housing admin)
  * @property {String} [Class] Class
- * @property {String} OffCampusProgram The name of department of this applicant's off-campus program, or 'None'
+ * @property {String} OffCampusProgram The name of department of this applicant's off-campus program, or '' (empty string)
```
NerdyGriffin, line 38: This confused me just now, I thought it was a typo, so maybe this will make it easier to read?
```suggestion
 * @property {String} OffCampusProgram The name of department of this applicant's off-campus program, or ( '' ) (i.e., an empty string)
```

EjPlatzer (approved):
These are excellent changes! I had a few questions and stylistic qualms, but overall this a is good improvement!

src/services/housing.js
EjPlatzer: I personally find the parentheses around `''` more confusing. Maybe this would be the clearest?:
```suggestion
 * @property {String} OffCampusProgram The name of department of this applicant's off-campus program, or  an empty string
```

NerdyGriffin (commented):
Some self-review things

src/views/ApartmentApp/components/StudentApplication/index.js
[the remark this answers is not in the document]
NerdyGriffin: The eslint was simply catching the call to `debugPrintApplicationDetails`, which was intentionally excluded

If I have since removed the call to `debugPrintApplicationDetails`, then I must have forgotten to remove this `// eslint-disable-next-line`

src/services/housing.js
```diff
@@ -198,6 +178,48 @@ const changeApartmentAppEditor = async (applicationID, newEditorUsername) => {
...
+  user.setFullname(applicant.Profile);
+  user.setClass(applicant.Profile);
+
+  // The following 'Class' property is needed for the staff page
+  if (applicant.Class === null || Number(applicant.Class)) {
+    // Use converted Class from number ('1', '2', '3', ...) to words ('Freshman', 'Sophomore', ...)
+    applicant.Class = applicant.Profile.Class;
+  }
+
+  applicant.OffCampusProgram ?? (applicant.OffCampusProgram = '');
```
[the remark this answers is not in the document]
NerdyGriffin: This is the intent. However, I made an issue about a babel compile error, there may be a dependency issue or a module config issue preventing the `??=` from compiling

I used the `a ?? (a = b)` as a workaround because the documentation for `a ??= b` explained that it is functionally equivalent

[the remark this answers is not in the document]
NerdyGriffin: I described the issue in #1101

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Yes, those should be a different name

NerdyGriffin: Yes, those should be a different name

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Yes, that is true. I think it may have been a workaround for something, I would want to do a lot more extensive testing both with and without that line before removing that

EjPlatzer: Presumably it's because any unsaved changes would be lost otherwise, since the user would no longer be the editor and so couldn't change anything. I would just want them to be told, explicitly, that changing editors will save all unsaved changes.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -287,37 +330,54 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+            throw new Error('Failed to load apartment application.');
+          }
+        } catch {
```
NerdyGriffin, line 349:
```suggestion
        try {
          const loadingResult = loadApplication();
          if (!loadingResult ) {
            throw new Error('Failed to load apartment application.');
          }
        } catch {
```
https://github.com/gordon-cs/gordon-360-ui/pull/1093/files#r625092241

src/services/housing.js
NerdyGriffin: Feel free to commit your suggestion here

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: I meant that it was an issue with causing the entire changeEditor operation to fail in some way, I forget the details of what the error was.

If you look at the handleChangeEditorAccepted, you will see that the intent is to use `saveApplication` instead of `changeApplicationEditor`, so this save behavior would happen anyways

NerdyGriffin: I agree we should have the UI text explicitly state whether the unsaved changes will be saved or not saved when changing the editor

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -342,8 +342,8 @@ const StudentApplication = ({ userProfile, authentication }) => {
       );
       if (result) {
         try {
-          const result = loadApplication();
-          if (!result) {
+          const loadingResult = loadApplication();
```
NerdyGriffin, line 345: Shouldn't this be using an `await` as well?

EjPlatzer: Yes, since `loadApplication()` is `async`, this should absolutely be using await, otherwise `loadingResult` will always evaluate to `true` since it's a `Promise`.

EjPlatzer (approved):
LGTM!

NerdyGriffin:
Go ahead and merge when ready, Then these changes will be usable during our presentation/demo

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Coming soon-ish in another branch, it is not as urgent as the larger bug fixes

PR #1094, opened 2021-04-29
Apartapp fix isApplicantValid

EjPlatzer:
There were a couple of insidious bugs related to the function `isApplicantValid`.

1. Assigning an applicant's full name to the `applicant.Profile` property instead of `applicant.Profile.fullName`, accidentally overwriting all other Profile info
2. Calling `user.formatName(...)`, which is a function in `user.js` that wasn't exported (as therefore could not be called). 
3. Making it an async function means that calling it elsewhere is inconvenient, especially when you only want to extract the boolean result for a one-time comparison. I converted it to a synchronous version using `Promise.then()` syntax. This should fix present errors in places it was being used as-if it were synchronous, and also help prevent future bugs of a similar nature.

### DIFF
--- a/src/services/housing.js
+++ b/src/services/housing.js
@@ -65,6 +65,8 @@ import user from './user';
  * @property {Number} AvgPoints The average application points per applicant
  */
 
+// const customErrorHandler = async () => {};
+
 /**
  * Check if the current user is authorized to view the housing staff page for applications
  *
@@ -76,6 +78,10 @@ const checkHousingAdmin = async () => {
   try {
     return await http.get(`housing/admin`);
   } catch (err) {
+    console.log(err);
+    console.log(err.response);
+    console.log(err.response.status);
+    console.log(err.status);
     // handle thrown 404 errors
     if (err.status === 404 || err.name.includes('NotFound')) {
       console.log('A 404 code indicates that current user was not found on the list of admins');
--- a/src/services/housing.js
+++ b/src/services/housing.js
@@ -201,18 +201,21 @@ const changeApartmentAppEditor = async (applicationID, newEditorUsername) => {
 /**
  * Helper function to fill in any missing properties of an applicant object's Profile, OffCampusProgram, etc.
  *
- * @async
  * @function setApplicantInfo
  * @param {ApartmentApplicant} applicant an object representing an apartment applicant
  * @return {ApartmentApplicant} Application details
  */
-const setApplicantInfo = async (applicant) => {
-  //! DEBUG: Temporary workaround for 'Profile.PersonType' being undefined
+const setApplicantInfo = (applicant) => {
+  //! DEBUG: Temporary workaround for an API bug that causes 'Profile.PersonType' to be undefined
   user.getProfileInfo(applicant.Username ?? applicant.Profile.Username).then((profile) => {
     applicant.Profile = profile;
   });
-  
-  //? Ideal solution, requires more testing once 'PersonType' issue is fixed
+
+  /**
+   * The following commented out commands are implicitly handled by `user.getProfileInfo()`,
+   * so these lines are not needed while the above workaround is still in place
+   */
+  //? This is the ideal solution. Requires more testing after the 'PersonType' issue is fixed in the API
   // applicant.Profile = user.setFullname(applicant.Profile);
   // applicant.Profile = user.setClass(applicant.Profile);
 
@@ -227,7 +230,9 @@ const setApplicationDetails = async (applicationDetails) => {
     NumApplicants: applicationDetails.Applicants?.length ?? 0,
     FirstHall: applicationDetails.ApartmentChoices[0]?.HallName ?? '',
   };
-  applicationDetails.Applicants = applicationDetails.Applicants.map((applicant) => setApplicantInfo(applicant));
+  applicationDetails.Applicants = applicationDetails.Applicants.map((applicant) =>
+    setApplicantInfo(applicant),
+  );
   return applicationDetails;
 };
 
--- a/src/services/housing.js
+++ b/src/services/housing.js
@@ -222,8 +222,8 @@ function setApplicantInfo(applicant) {
    * so these lines are not needed while the above workaround is still in place
    */
   //? This is the ideal solution. Requires more testing after the 'PersonType' issue is fixed in the API
-  // applicant.Profile = user.setFullname(applicant.Profile);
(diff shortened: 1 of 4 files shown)


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -163,7 +163,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
     }
 
     if (applicant.Profile.fullName === null) {
-      applicant.Profile = user.formatName(applicant.Profile);
+      applicant.Profile.fullName = user.setFullname(applicant.Profile);
```
NerdyGriffin, line 166: According to the definition of `setFullname` in the user.js file, this function appears to return a modified profile object with the `fullName` property added to it, rather than returning just the `fullName` as you have used it here. If that is not working as intended, then perhaps there is an addition problem in `user.js` that requires attention

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -199,8 +199,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
     }
 
     // Check if the selected user is already saved on an application in the database
-    try {
-      let existingAppID = await housing.getCurrentApplicationID(applicant.Profile.AD_Username);
+    housing.getCurrentApplicationID(applicant.Profile.AD_Username).then((existingAppID) => {
```
NerdyGriffin, line 202: ~Just want to make sure, how does the `.then` respond to receiving error codes? Because we expect to receive errors like 403 and 404 in certain cases~
I looked up the documentation, and I believe the behavior here is exactly what I intended

NerdyGriffin (commented):
Looks good, though I have some concerns about the `fullName`

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Alternatively, we should build this call of `setFullname` (or the above suggestion) into that part of the code that loads applications from the API, since any `Profile` that comes from `user.getProfileInfo` will already have the `fullName` property set by the `user.js`, and it is only the versions of `Profile` that come from `housing.js` that will be missing the `fullName` property. 

So maybe we should instead do this by default in the `getApartmentApplication` of `services/housing.js`

EjPlatzer: You are correct, `setFullname` does indeed return the entire profile, so the previous way was more correct (although that function wasn't exported before).

I think adding it by default in `getApartmentApplication` is the most robust solution, certainly. I was just trying to make this work locally.

src/views/ApartmentApp/components/StudentApplication/index.js
EjPlatzer: I believe the housing service actually catches the 404 and returns null. It would be a good idea to include a `.catch()` statement here if you need to handle any errors (including a 403).

NerdyGriffin:
Please check #1095 , then re-request review of this PR when ready

NerdyGriffin (approved):
LGTM, though I would recommend cross-examining the changes I contributed here

src/services/housing.js
```diff
@@ -208,7 +257,9 @@ const changeApartmentAppEditor = async (applicationID, newEditorUsername) => {
  */
 const getApartmentApplication = async (applicationID) => {
   try {
-    return await http.get(`housing/apartment/applications/${applicationID}/`);
+    let applicationResult = await http.get(`housing/apartment/applications/${applicationID}/`);
+    setApplicationDetails(applicationResult);
```
joshuaRogers24, line 261: Does this do anything if you do not catch its output?

joshuaRogers24: (Is applicationDetails defined outside the context  of setApplicationDetails?)

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -151,7 +151,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
    * @param {ApartmentApplicant} applicant The applicant to be checked
    * @return {Boolean} True if valid, otherwise false
    */
-  const isApplicantValid = async (applicant) => {
+  const isApplicantValid = (applicant) => {
```
joshuaRogers24, line 154: I understand that using `async` led to issues of False statements being read as true (or vice versa), and I thought using `await` fixed that. Why are we getting rid of both now?

joshuaRogers24 (changes requested):
Nice. I have some questions...

src/services/housing.js
```diff
@@ -65,6 +65,8 @@ import './user'; // Needed for typedef of StudentProfileInfo
  * @property {Number} AvgPoints The average application points per applicant
  */
 
+// const customErrorHandler = async () => {};
```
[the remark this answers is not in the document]
NerdyGriffin: Nope, the error handling has already been overhauled in #1100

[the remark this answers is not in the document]
NerdyGriffin:
```suggestion
```

src/services/housing.js
NerdyGriffin: No it is not, which is actually intentional in this case

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: There is no need to have it be async because all the internal parts of the function are now done in a synchronous manner, so it helps clean things up (because having the unneeded `await ...` can also lead to unwanted unexcepted behaviors if it is not used correctly)

It just seems to be more reliable if you don't use async unless it makes sense with the whole context

NerdyGriffin: Because of what this function does and what it is meant to represent, is it actually preferable to not have it be async, since its sole purpose is to block the execution of the things that come after it. So it is by design a synchronous concept.

src/services/housing.js
NerdyGriffin: I have done a lot more testing of this feature in #1100  and have been able to verify its effectiveness

src/views/ApartmentApp/components/StudentApplication/index.js
joshuaRogers24: Oh, good point!

joshuaRogers24 (approved):
Looks good!

PR #1095, opened 2021-04-29
ApartApp - Update/filter applicant data during first load from API

NerdyGriffin:


src/services/housing.js
```diff
@@ -208,7 +250,11 @@ const changeApartmentAppEditor = async (applicationID, newEditorUsername) => {
  */
 const getApartmentApplication = async (applicationID) => {
   try {
-    return await http.get(`housing/apartment/applications/${applicationID}/`);
+    let applicationResult = await http.get(`housing/apartment/applications/${applicationID}/`);
+    if (applicationResult) {
+      await setApplicationDetails(applicationResult);
```
NerdyGriffin, line 255: For some reason, the Profile objects from this endpoint are lacking the `PersonType` prop, but all the other expected props appear to be there...

This breaks several parts of the page

EjPlatzer: Ah yes, apparently that property is added by the `ProfilesController.Get()` method and isn't part of the actual view models. I'm not sure what the best solution for this is, outside of getting profiles from the ProfilesController instead of creating them in the backend.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -224,7 +216,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
   const addApplicant = async (newApplicantUsername) => {
     try {
       // Get the profile of the selected user
-      const newApplicantProfile = await user.getProfileInfo(newApplicantUsername);
+      let newApplicantProfile = await user.getProfileInfo(newApplicantUsername);
```
EjPlatzer, line 219: Is there a reason this was changed from `const` to `let`? It doesn't seem like you mutate it later on, so presumably `const` is the correct declaration.

EjPlatzer (changes requested):
This is a good upgrade to the data flow and integrity of the apartment application. It is concerning that we aren't getting `PersonType` from the backend, since that's not a simple thing to fix. 

Also, it seems like you changed a constant variable from `const` to `let` unnecessarily.

src/views/ApartmentApp/components/StudentApplication/index.js
EjPlatzer:
```suggestion
      const newApplicantProfile = await user.getProfileInfo(newApplicantUsername);
```

EjPlatzer (approved):
The issue of not having `Profile.PersonType` is a bigger one than can easily be solved now, and is arguably outside the scope of this pull request. As such, I'm approving and merging this one so as to not slow y'all down, and you can fix the PersonType issue in a later pr.

src/services/housing.js
NerdyGriffin: If we have the HousingController.GetApartmentApplication send only the usernames, and then have this file do a user.getProfileInfo for each, but this would mean a whole bunch of extra back-and-forth between the UI and API just to load a single application, and the API code needs to get the student profiles during the loading process regardless, so it would seem extremely inefficient. 
This sounds like something to be best addressed in the API side, I will make a note

PR #1096, opened 2021-04-30
Added the createSnackbar helper function

NerdyGriffin:
Suggested by @EjPlatzer [here](https://github.com/gordon-cs/gordon-360-ui/pull/1093#discussion_r623272075)

Style of createSnackbar was inspired by the elegant implementation used in src/components/MembershipList

### DIFF
--- a/src/views/ApartmentApp/components/StudentApplication/index.js
+++ b/src/views/ApartmentApp/components/StudentApplication/index.js
@@ -56,9 +56,6 @@ const StudentApplication = ({ userProfile, authentication }) => {
   const [changeEditorDialogOpen, setChangeEditorDialogOpen] = useState(false);
   const [submitDialogOpen, setSubmitDialogOpen] = useState(false);
   const [snackbar, setSnackbar] = useState({ message: '', severity: '', open: false });
-  const [snackbarOpen, setSnackbarOpen] = useState(false);
-  const [snackbarText, setSnackbarText] = useState('');
-  const [snackbarSeverity, setSnackbarSeverity] = useState('info');
   const [saveButtonAlertTimeout, setSaveButtonAlertTimeout] = useState(null);
 
   /**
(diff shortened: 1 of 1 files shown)


EjPlatzer (commented):
This is a really nice change! One day in the future, I hope to have one single Snackbar component at the `app.js` level and simply make `createSnackbar` available as a hook to any component that needs to give user feedback. In the meantime, it's great to have this simplified and streamlined component api implemented at the view level. 

It looks like you forgot to remove the old snackbar state (snackbarOpen, snackbarText, etc.). That is causing the CI to fail because it treats warnings as errors. Once that's fixed, I feel great about approving this.

NerdyGriffin:
> This is a really nice change! One day in the future, I hope to have one single Snackbar component at the `app.js` level and simply make `createSnackbar` available as a hook to any component that needs to give user feedback.... 

I was thinking this as well, though I noticed that one (and maybe more) pages also modify the snackbar to utilize the positioning props and it would also be great to have the stacking snackbars like what is demonstrated [here](https://material-ui.com/components/snackbars/#notistack)

Since the customize snackbar used across the site is already a component in `src/components/Snackbar`, it should be relatively simple to refactor the existing stuff to have that single component in the `app.js` level

EjPlatzer:
> > This is a really nice change! One day in the future, I hope to have one single Snackbar component at the `app.js` level and simply make `createSnackbar` available as a hook to any component that needs to give user feedback....
> 
> I was thinking this as well, though I noticed that one (and maybe more) pages also modify the snackbar to utilize the positioning props and it would also be great to have the stacking snackbars like what is demonstrated [here](https://material-ui.com/components/snackbars/#notistack)
> 
> Since the customize snackbar used across the site is already a component in `src/components/Snackbar`, it should be relatively simple to refactor the existing stuff to have that single component in the `app.js` level

EjPlatzer (approved):
LGTM!

PR #1097, opened 2021-04-30
Fixed the editor state not updating after changing the editor

NerdyGriffin:


joshuaRogers24 (changes requested):
Ah, that's better! Still a little issue: the star does not get filled in for the editor. It stays only filled in for the old editor.

NerdyGriffin:
> hat's better! Still a little issue: the star does not get filled in for the editor. It stays only fille

What! I thought I fixed that one two days ago 🤦 
That's an easy fix (should be...)

NerdyGriffin:
> > hat's better! Still a little issue: the star does not get filled in for the editor. It stays only fille
> 
> What! I thought I fixed that one two days ago 🤦
> That's an easy fix (should be...)

It was not easy, I am not sure what it wrong there. 

There are other underlying bugs that need to be addressed first, and they may be related or the cause

src/services/housing.js
```diff
@@ -171,10 +179,28 @@ const saveApartmentApplication = async (applicationDetails) => {
...
+      console.log('Received 401 (Unauthorized)');
+    } else if (err?.status === 404 || err?.name?.includes('NotFound')) {
+      console.log(
+        'Received 404 indicates that the requested application was not found in the database',
+      );
+    } else {
+      throw err;
+    }
+  } finally {
+    return result;
```
NerdyGriffin, line 203: Many of these endpoints need to use the exact same structure of 
``` javascript
let result = null
try {
   result = await DoSomething
} catch (err) {
  if (it's a 401 error) {
    console.log('Message that 401')
  } else if (it's a 404 error) {
    console.log('Custom message about meaning of 404)
  } else {
    throw err
  }
} finally {
  return result;
}
```

Since much of that `try ... catch ... finally` needs to be identical across multiple functions in housing.js, is there a way to put that in a function/wrapper to reduce the repetitive lines of code?

The tricky part is somehow changing the API request and log text while leaving the rest the same (the text is trival, just pass a string. It is the API requests that I am not sure about, since we need `http.get`, `http.put`, `http.post`, etc.)

NerdyGriffin: I suspect callbacks might be a solution

src/services/housing.js
```diff
@@ -207,21 +248,14 @@ const changeApartmentAppEditor = async (applicationID, newEditorUsername) => {
...
-    applicant.OffCampusProgram = '';
-  }
+  let profile = applicant.Profile;
+  user.setFullname(profile);
+  user.setClass(profile);
 
+  applicant.Profile = profile;
+  applicant.Username = profile.AD_Username;
+  applicant.Class = profile.Class; // This should be after the conversion from class = number to class = word
+  applicant.OffCampusProgram = applicant.OffCampusProgram ?? '';
```
NerdyGriffin, line 258: This will need to change or be rolled-back,

src/services/housing.js
NerdyGriffin: Improvements coming in #1100

NerdyGriffin:
@joshuaRogers24 Re-review with the assumption that the change editor button bug that you described will be worked on in an upcoming PR

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -752,7 +752,11 @@ const StudentApplication = ({ userProfile, authentication }) => {
                   !applicationCardsOpen ||
                   !agreements ||
                   !(applicationDetails?.Applicants?.length > 0) ||
-                  !(applicationDetails?.ApartmentChoices?.length > 0)
+                  !(
+                    applicationDetails?.ApartmentChoices?.filter(
+                      (apartmentChoice) => apartmentChoice.HallName,
+                    )?.length > 0
+                  )
```
joshuaRogers24: Could you explain what this was doing before vs now?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -728,8 +730,12 @@ const StudentApplication = ({ userProfile, authentication }) => {
                 disableSubmit={
                   !applicationCardsOpen ||
                   !agreements ||
-                  !(applicationDetails.Applicants.length > 0) ||
-                  !(applicationDetails.ApartmentChoices.length > 0)
+                  !(applicationDetails?.Applicants?.length > 0) ||
+                  !(
+                    applicationDetails?.ApartmentChoices?.filter(
+                      (apartmentChoice) => apartmentChoice.HallName,
+                    )?.length > 0
```
joshuaRogers24, line 737: Can you explain what this did before vs now?

joshuaRogers24 (changes requested):
It looks pretty good. I am just confused about the empty apartment choice fix.

src/views/ApartmentApp/components/StudentApplication/index.js
joshuaRogers24: Oh, that did save. Weird. Well, you do not have to answer the question twice.

NerdyGriffin: Before, it was a check for whether the array of ApartmentChoices was empty.

Now, it uses the `.filter` to get the subset of array which have a nonempty HallName, and checks that
In other words, `ApartmentChoices.filter((apartmentChoice) => apartmentChoice.HallName)` returns an array of ApartmentChoices containing only the elements that have a real value selected for the `HallName`.

So basically, it filters out any of the Hall dropdown menus that are blank (i.e., the ones that don't have a hall selected yet), because before you could hit the "Add Hall" button but leave the dropdown blank, and it would still allow you to submit press submit even though there were technically no halls selected. 

Additionally, using the `?.` (such as in `ApartmentChoices?.filter()?.length` ) means that if the array itself is null, then it shortcuts the thing after the `?.` and returns null, instead of trying to perform `null.length` which would give an error. And then the `null > 0` returns false, so the `array?.length > 0` returns false if the array is empty or if the array variable is null, so we get the desired behavior of `(array?.length > 0) === true` whenever the array is not null and not empty.

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: See [this reply](https://github.com/gordon-cs/gordon-360-ui/pull/1097#discussion_r624556360) for the same explanation

> Before, it was a check for whether the array of ApartmentChoices was empty.
>
> Now, it uses the `.filter` to get the subset of array which have a nonempty HallName, and checks that
> In other words, `ApartmentChoices.filter((apartmentChoice) => apartmentChoice.HallName)` returns an array of ApartmentChoices containing only the elements that have a real value selected for the `HallName`.
> 
> So basically, it filters out any of the Hall dropdown menus that are blank (i.e., the ones that don't have a hall selected yet), because before you could hit the "Add Hall" button but leave the dropdown blank, and it would still allow you to submit press submit even though there were technically no halls selected.
> 
> Additionally, using the `?.` (such as in `ApartmentChoices?.filter()?.length` ) means that if the array itself is null, then it shortcuts the thing after the `?.` and returns null, instead of trying to perform `null.length` which would give an error. And then the `null > 0` returns false, so the `array?.length > 0` returns false if the array is empty or if the array variable is null, so we get the desired behavior of `(array?.length > 0) === true` whenever the array is not null and not empty.

src/views/ApartmentApp/components/StudentApplication/index.js
joshuaRogers24: Okay, sweet, that makes plenty of sense.

joshuaRogers24 (approved):

joshuaRogers24:
Approved, but only under the assumption that adding someone who is on an existing application is fixed in the child branch.

NerdyGriffin:
> Approved, but only under the assumption that adding someone who is on an existing application is fixed in the child branch.

This fix has just been pushed to #1100

PR #1099, opened 2021-04-30
Update and refactor About page design

EjPlatzer:
The About page benefitted greatly from a refactor, and in the process was redesigned for better parity with the rest of the site. The list of contributors is now a `.json` which can be easily edited and added to without having to deeply understand the structure of the page (and the repeated layout components are no longer repeated, and instead mapped to).

The previous design:
image

The new design:
image

The main elements of the redesign were to put the content on a card and make consistent use of Material UI design.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,25449 @@
 {
   "name": "gordon-360-ui",
   "version": "0.0.0",
-  "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


bennettforkner (approved):
LGTM. About page looks a lot better with these changes.

PR #1106, opened 2021-05-03
Update submit button to disable when application is already submitted

NerdyGriffin:


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -880,6 +880,7 @@ const StudentApplication = ({ userProfile, authentication }) => {
                 deleteDialogOpen={deleteDialogOpen}
                 deleting={deleting}
                 disableSubmit={
+                  applicationDetails?.DateSubmitted ||
```
joshuaRogers24, line 883: Will this allow them to submit again, though, if they realized they made a mistake before the deadline?

joshuaRogers24 (changes requested):
Nice. Quick question:

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: If you submit once, we currently do not un-submit it when you make changes to the application

NerdyGriffin: So you can save new changes and it will not affect the `dateSubmitted` value in the database at all.

If it is changed in the future to maybe un-submit when there are new changes, then it would be done by setting the `DateSubmitted` in the database back to `null`, which would thereby re-enable to submit button

joshuaRogers24: Oh, right! They only need to submit once and whatever changes they make will also be "submitted".

joshuaRogers24 (approved):

PR #1107, opened 2021-05-03
ApartApp - Improve user feedback when changing editor

NerdyGriffin:


src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -294,9 +294,6 @@ const StudentApplication = ({ userProfile, authentication }) => {
           (applicant) => applicant.Profile.AD_Username === profile.AD_Username,
         )
       ) {
-        if (unsavedChanges) {
-          saveApartmentApplication(applicationDetails);
-        }
```
joshuaRogers24, line 299: Where in the editor change process do we save instead?

joshuaRogers24 (changes requested):
Question:

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Look at `handleChangeEditorAccepted`

NerdyGriffin: In `handleChangeEditorAccepted()` it does `saveApartmentApplication({ ...applicationDetails, EditorProfile: newEditorProfile })`
(that may be approximate, I am guesstimating it from memory)

NerdyGriffin: Which is also a good policy, because then it will save after the user clicks "accept", rather than silently saving when the star icon is clicked

joshuaRogers24: Oh, yes, agreed.

joshuaRogers24 (approved):
Nice.

PR #1108, opened 2021-05-03
Apartapp - Update and refactor callback functions and more

NerdyGriffin:


src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -112,7 +89,11 @@ const ApplicantListItem = ({
         </ListItemAvatar>
         <Grid container alignItems="center" spacing={1}>
           <Grid item xs={8} sm>
-            <ListItemText primary={displayName} secondary={personClass} className={'list-item'} />
+            <ListItemText
+              primary={displayName}
+              secondary={profile.Class ?? ''}
+              className={'list-item'}
```
EjPlatzer, line 95: Technically you could pass this as a simple string:
```suggestion
              className="list-item"
```
but it really doesn't matter very much.

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
```diff
@@ -121,7 +102,7 @@ const ApplicantListItem = ({
               <IconButton
                 aria-label={isApplicationEditor ? 'current-editor' : 'set-new-editor'}
                 disabled={isApplicationEditor || disabled}
-                onClick={handleChangeEditor}
+                onClick={() => profile && onChangeEditor?.(profile)}
```
EjPlatzer, line 105: So are there cases where the `onChangeEditor` prop is `null`? That seems like an odd design choice.

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
```diff
@@ -22,33 +22,33 @@ import GordonPeopleSearch from 'components/Header/components/PeopleSearch';
...
+ * @typedef { import('services/housing').ApplicationDetails } ApplicationDetails
  * @typedef { import('services/user').StudentProfileInfo } StudentProfileInfo
  */
 
-// Create a list of applicants, displayed by name, username, and class standing.
+/**
+ * Renders the list of applicants, displayed by name, username, and class standing.
+ * @param {Object} props The React component props
+ * @param {Boolean} props.disabled Boolean to disable the interactive elements of this list item
+ * @param {*} props.authentication The user authentication
```
EjPlatzer, line 33: `authentication` is a simple Boolean representing where the user is authenticated (I didn't name it).

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
```diff
@@ -120,10 +120,12 @@ const ApplicantList = ({
...
               <GordonPeopleSearch
                 disableLink
-                disabled={disabled || applicationDetails.Applicants?.length > maxNumApplicants}
+                disabled={disabled || Applicants?.length > maxNumApplicants}
                 icon={<GroupAddIcon />}
                 customPlaceholderText={'Add Applicant'}
-                onSearchSubmit={handleSelection}
+                onSearchSubmit={(selectedUsername) =>
+                  selectedUsername && onSearchSubmit(selectedUsername)
+                }
```
EjPlatzer, line 128: Yet again - are there times when `onSearchSubmit` is expected to be `null`?

src/views/ApartmentApp/components/StudentApplication/components/HallChoiceList/components/HallListItem/index.js
```diff
@@ -102,8 +68,11 @@ const HallListItem = ({
               <InputLabel>Rank</InputLabel>
               <Select
                 disabled={disabled}
-                value={hallRank}
-                onChange={handleRankInputChange}
+                value={HallRank}
+                onChange={(event) =>
+                  event.target.value !== null &&
+                  onHallInputChange?.(String(event.target.value), HallName, index)
+                }
```
EjPlatzer, line 75: Were you getting instances where `event.target.value` was null?

src/views/ApartmentApp/components/StudentApplication/components/HallChoiceList/index.js
```diff
@@ -2,16 +2,26 @@ import React, { useState, useEffect } from 'react';
...
  * @typedef { import('services/housing').ApartmentChoice } ApartmentChoice
  */
 
-// Create a list of selection boxes to choosing preferred halls
-const HallSelection = ({
+/**
+ * Renders a list of selection boxes to choosing preferred halls
+ * @param {Object} props The React component props
+ * @param {Boolean} props.disabled Boolean to disable the interactive elements of this list item
+ * @param {*} props.authentication The user authentication
```
EjPlatzer, line 17: Again, `authentication` is a Boolean.

src/views/ApartmentApp/components/StudentApplication/components/OffCampusList/index.js
```diff
@@ -4,8 +4,21 @@ import { Grid, Card, CardHeader, CardContent, List } from '@material-ui/core';
...
+/**
+ * @typedef { import('services/housing').ApartmentApplicant } ApartmentApplicant
+ * @typedef { import('services/user').StudentProfileInfo } StudentProfileInfo
+ */
+
+/**
+ * Renders the list of selection boxes to choosing which applicants are doing off campus programs.
+ * @param {Object} props The React component props
+ * @param {Boolean} props.disabled Boolean to disable the interactive elements of this list item
+ * @param {*} props.authentication The user authentication
```
EjPlatzer, line 16: `authentication` is a Boolean. Just leaving this note so you can update it everywhere.

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -640,29 +612,14 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
-  };
-
   /**
    * Submit the current application as completed
    *
    * @async
    * @function submitApplication
    */
   const submitApplication = async () => {
+    handleCloseDialog();
```
EjPlatzer, line 622: It's a little strange to have the dialog behavior coupled this tightly with the data handling. Would it be reasonable to have the `handleCloseDialog()` call in an anonymous callback that also calls `submitApplication()`?

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -894,14 +823,14 @@ const StudentApplication = ({ userProfile, authentication }) => {
...
+                onCloseOkay={handleCloseDialog}
+                onDeleteAppAccepted={() => deleteApartmentApplication()}
                 onDeleteButtonClick={() => setDeleteDialogOpen(true)}
-                onSaveButtonClick={handleSaveButtonClick}
-                onShowApplication={handleShowApplication}
-                onSubmitAppAccepted={handleSubmitAppAccepted}
-                onSubmitButtonClick={handleSubmitButtonClick}
+                onSaveButtonClick={() => saveApartmentApplication(applicationDetails)}
+                onShowApplication={() => setApplicationCardsOpen(true)}
+                onSubmitAppAccepted={() => submitApplication()}
```
EjPlatzer, line 832: For example, here you're already defining an anonymous function. Realistically this could simply be:
```suggestion
                onSubmitAppAccepted={submitApplication}
```

However, I think it would be much cleaner to have the dialog behavior attached to the dialog itself, rather than the data handling callbacks it calls:
```suggestion
                onSubmitAppAccepted={() => {
                    handleCloseDialog();
                    submitApplication();
                }
```

EjPlatzer (commented):
These are some very clean updates! Everything seems to work well, but I did have a few questions (especially related to having null callbacks or invalid values for `event.target.value`).

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
NerdyGriffin: This is done to allow you to omit the onChangeEditor from the props entirely and still have the component display correctly. In those cases, it is specifically handled as `undefined`, so I did some research and learned that you can use the `?.` for functions, and I did a bunch of experimentation in CodePen to test the behavior in a bunch of possible situations

This "feature" is mostly used for when the user is not the editor of an application, so they are allowed to view all the stuff but all the interactive things are disabled. Thus there is no need to pass the functions, and having them undefined helps prevent things from accidentally triggering when they are not supposed to.

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
NerdyGriffin: I have actually been wondering why much of the site uses `className={'example-name'}` instead of `className="example-name"`
I saw it nearly everywhere and didn't understand the distinction back in the fall, so I just assumed there might be some reason

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
NerdyGriffin: I have also been going back and forth about removing it from these subcomponents because it is impossible to even view the student or staff page without being authenticated. The top-level `ApartmentApp` component handles that, so it is redundant in it's subcomponents

src/views/ApartmentApp/components/StudentApplication/components/HallChoiceList/components/HallListItem/index.js
NerdyGriffin: I think so? 
I remember there are weird edge-cases where things do strange behaviors as the page is loading,...

Though now that I think about it, the errors with callbacks getting triggered on declaration was something that happened when using class components, unless you put some extra stuff in place, 
and I think the way the functional components handle things prevents those unwanted behaviors, so I might be no longer an issue

src/views/ApartmentApp/components/StudentApplication/components/OffCampusList/components/OffCampusListItem/index.js
```diff
@@ -21,75 +21,63 @@ import {
...
  * @param {CallbackFcn} props.onOffCampusInputChange Callback for dropdown menu change
- * @returns {JSX.Element} JSX Element for the hall list item
+ * @returns {JSX.Element} JSX Element for the off-campus program list
  */
 const OffCampusListItem = ({
   disabled,
   index,
-  applicantProfile,
-  offCampusProgram,
+  applicant: { Profile, OffCampusProgram },
```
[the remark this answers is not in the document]
NerdyGriffin: No particular reason, I was trying out ideas at the time

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: It's used to be in a callback, which I shorten and made in-line, 
So there are several ways of moving it back

NerdyGriffin: I meant to convey that I do agree, it feels inappropriate to lump the dialog stuff into the operation of "submitting" or "saving" or whatever else

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/index.js
EjPlatzer: I think removing it is a totally valid option then!

One of the eventual goals is to use the React Context API to handle Auth state more robustly and with a much neater solution for deeply nested components that need to be aware of auth state.

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Since the instance of the `GordonDialogBox` within the `BottomBar` are do nothing other than directly apply the props passed to the `BottomBar`, I was just thinking it would probably be better to just put those dialog boxes directly in the `StudentApplication`, so because we are just passing all these callbacks through the a subcomponent for no real reason.

I think it would remove a lot of confusion and clutter from the `BottomBar` component

src/components/GordonDialogBox/index.js
```diff
@@ -65,14 +64,14 @@ export default class GordonDialogBox extends Component {
...
             <Button variant="contained" onClick={this.props.cancelButtonClicked} color="primary">
-              {this.props.cancelButtonName}
+              {this.props.cancelButtonName ?? 'Cancel'}
             </Button>
           )}
-          {this.props.buttonName && (
+          {this.props.buttonClicked !== undefined && (
             <Button variant="contained" onClick={this.props.buttonClicked} color="primary">
-              {this.props.buttonName}
+              {this.props.buttonName ?? 'Okay'}
```
NerdyGriffin, line 74: Literally every other page that uses this component currently sets the button name to 'Okay', so it was an intuitive default choice

src/views/ApartmentApp/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.js
NerdyGriffin: This is universally true for the classNames? Because if so, I will do it all at once in all the ApartApp files with a regex statement

EjPlatzer: Technically it's true of all JSX props, yes. If the value you're passing is a string, you can pass a string literal directly: `<Component prop="some string">` is equivalent to `<Component prop={'some string'}>`

EjPlatzer: See the 'Props in JSX' section of this page for more details: https://reactjs.org/docs/jsx-in-depth.html#props-in-jsx

NerdyGriffin: That's what I figured based on the meaning of the `{ }` in the JSX content, which is also why I was confused that most things I looked at were consistently using `{'class-name'}`

src/views/ApartmentApp/components/StudentApplication/components/HallChoiceList/components/HallListItem/index.js
NerdyGriffin: @NerdyGriffin  Note to self: test this out later

src/views/ApartmentApp/components/StudentApplication/index.js
```diff
@@ -26,6 +26,12 @@ const BLANK_APPLICATION_DETAILS = {
   Applicants: [],
   ApartmentChoices: [],
 };
+const INITIAL_DIALOG_PROPS = {
+  labelledby: 'default-dialog',
+  describedby: 'dialog-undefined',
+  title: 'How did you get here?',
+  text: 'This text should not be displayed.',
+};
```
EjPlatzer, line 34: I think you could simplify the dialog usage further by defining all the static dialog props here:
```suggestion
const DIALOG_PROPS = {
  default: {
    labelledby: 'default-dialog',
    describedby: 'dialog-undefined',
    title: 'How did you get here?',
    text: 'This text should not be displayed.',
  },
  ...,
  submit: {
    labelledby: 'submit-application-dialog',
    describedby: 'submit-application',
    title: 'Submit apartment application?',
    text: 'Please confirm that all the information you have entered is valid\nClick "Accept" below to submit this application',
  },
};
```

Then, instead of having two separate state variables, you could just call `setDialogProps(DIALOG_PROPS.submit)` for example.

EjPlatzer (commented):
This is a really nice clean up! I did have one suggestion for simplifying the Dialog handling further.

src/views/ApartmentApp/components/StudentApplication/index.js
NerdyGriffin: Oh nice! That is far more convenient
I was specifically trying to find a better way to handle the static stuff like this, so thanks!

One problem I was having, which inspired the behavior of leaving it unchanged on default, for example, was that if you close accept or close, you would see the default text showup inside the dialog a split-second before the dialog box closed, which just felt a bit sloppy to look at, so I tried to have it leave the props unchanged when it closes the dialog. Though, I guess my comment here is more about the behavior of the switch-case.
Really the problem was that the state updates faster than the DialogBox realizes that its `open` prop changed to false and then it performs the closing animation, so it's a race condition in state change vs. visual animation.

EjPlatzer: Oh yes, I remember encountering the same issue when trying to simplify the API for snackbars. The way I handled it was to only update the `open` boolean to false when the dialog is closed, and to update the content and set `open` to true when a new dialog is opened.

NerdyGriffin: Oh right! the snackbars do the `setSnackbar((s) => {...s, open: false})` or something like that, I remember copying that idea from another 360 component somewhere and thinking that it was a very elegant solution 😆

EjPlatzer (approved):
These are excellent updates! Vast improvements to code quality and readability.

PR #1109, opened 2021-05-04
ApartApp - Exclude [object] from CSV

NerdyGriffin:
I spotted during the presentation demo that the CSV still includes `EditorProfile` with cells containing `[object]` so I made this quick update to fix that


glgnkim (approved):
Good catch! LGTM

PR #1110, opened 2021-05-04
Refactor Help Page & Adjust Messaging

bennettforkner:
The Help page, similar to the About page, has needed a refactor for design and readability when users are looking for easily-accessible assistance.

Current page:
image

After changes:
image

Similar to #1099 , "The main elements of the redesign were to put the content on a card and make consistent use of Material UI design."

Tagging @ChrisImagineer for wording updates.

Changes from #1099 have been incorporated into this pull request.

PR #1111, opened 2021-05-04
Refactor and Update Content on Help Page

bennettforkner:
Opened as a new version of #1110 because of merging issues.

The Help page, similar to the About page, has needed a refactor for design and readability when users are looking for easily-accessible assistance.

Current page:
image

After changes:
image

Similar to #1099 , "The main elements of the redesign were to put the content on a card and make consistent use of Material UI design."

### DIFF
--- a/src/views/Help/help.scss
+++ b/src/views/Help/help.scss
@@ -4,11 +4,6 @@
 
 .help {
   
-  strong {
-    font-weight: bold;
-    margin-top: 20px;
-  }
-  
   margin: 1rem;
 
   li {
--- a/src/views/Help/help.scss
+++ b/src/views/Help/help.scss
@@ -5,3 +5,34 @@ strong {
   font-weight: bold;
   margin-top: 20px;
 }
+
+.help {
+  margin: 1rem;
+
+  li {
+    list-style-type: '⮞ ';
+  }
+
+  & .MuiCard-root {
+    box-shadow: none;
+  }
+
+  &-title {
+    text-align: center;
+    padding: 1rem;
+  }
+
+  &-contributors.MuiTypography-root {
+    white-space: pre-wrap;
+    margin-left: 1rem;
+  }
+
+  &-header {
+    background-color: $primary-blue;
+    color: $neutral-white;
+  }
+  //TODO Issue #1098 - remove when timeshets.scss stops centering grid items
+  &-content {
+    text-align: start;
+  }
+}
--- a/src/views/Help/help.scss
+++ b/src/views/Help/help.scss
@@ -1,7 +1,6 @@
 @import '../../vars';
 
 strong {
-  font-size: 15px;
   font-weight: bold;
   margin-top: 20px;
 }
@@ -13,8 +12,13 @@ strong {
     list-style-type: '⮞ ';
   }
 
-  & .MuiCard-root {
-    box-shadow: none;
+  li li {
+    list-style-type: '⮚ ';
+  }
+
+  //TODO: Remove this when timesheets bugs are resolved
+  ul {
+    text-align: start;
   }
 
   &-title {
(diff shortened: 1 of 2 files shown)


src/views/Help/help.scss
```diff
@@ -1,7 +1,36 @@
...
-  font-weight: bold;
-  margin-top: 20px;
+
+
+.help {
+  
+  strong {
+    font-weight: bold;
+    margin-top: 20px;
+  }
```
EjPlatzer, line 10: Normally, we would need to check every instance of `<strong>` across the rest of the site to ensure this didn't break any other styling. However, I just discovered that this exact same styling is added at the top-level `app.scss` file (which is intended to be site-wide):
```
strong {
  font-size: 15px;
  font-weight: bold;
  margin-top: 20px;
}
```
So this particular styling is totally redundant, and should probably be removed.

src/views/Help/help.scss
```diff
@@ -1,7 +1,36 @@
...
+  
+  margin: 1rem;
+
+  li {
+    list-style-type: '⮞ ';
+  }
+
+  li li {
+    list-style-type: '⮚ ';
+  }
```
EjPlatzer, line 20: Why do you have both of these?

EjPlatzer (changes requested):
This looks good! Just a few small things that still need updating.

src/views/Help/help.scss
bennettforkner: The second one is the indented list item. See attached photos under 'Home:'

src/views/Help/help.scss
bennettforkner: I see. I'm not sure I understood exactly before. Should we revert this until we can resolve that separate issue?

src/views/Help/index.js
```diff
@@ -2,256 +2,279 @@ import React, { Component } from 'react';
...
+                        <Typography variant="body1" component="ul">
+                          <li>
+                            <strong>Student:</strong> Includes your Dining “Meal Wheel’ tracker synced to days remaining in the
+                            semester and your Christian Life & Worship (CL&W) credits. You can click on the
+                            CL&W tracker to see which events are recorded for you.
+                          </li>
+                          <li>
+                            <strong>Faculty/Staff:</strong> Includes days remaining wheel and a separate Dining Dollars tracker.
+                            Also view your pending join requests for Involvements you’ve requested, as well
+                            as the requests that are pending for you approve if you are a Group Leader/Advisor.
```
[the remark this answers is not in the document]
bennettforkner: I wasn't sure about that. That was directly from CC's changes, but I have never looked at the homepage as a group advisor. Just remove this part you think?

src/views/Help/help.scss
EjPlatzer: Well, now that we know this is (intentionally) a site-wide style from `app.scss`, this particular repeat of it is irrelevant and never affects anything. So for this pull request, I think you can just delete it.

src/views/Help/index.js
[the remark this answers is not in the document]
EjPlatzer: The code of the home page no longer includes that section, so it would never show up for anyone. Perhaps you can move the explanation to the Involvements page (since that is where the requests component now lives)?

src/views/Help/help.scss
EjPlatzer: Oooh, I see now that those are different characters. I hate that we have to manually define the nesting styles for this case, but it makes sense at least.

src/views/Help/index.js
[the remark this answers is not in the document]
bennettforkner: sounds good

src/views/Help/help.scss
bennettforkner: great.

EjPlatzer (approved):
LGTM!

PR #1114, opened 2021-05-06
Split Occurrences of the Same Event

bennettforkner:
For some time, events that happen on multiple different dates have only had their first occurrence show up on the Events view. This means that once the first occurrence has passed, the future occurrences no longer appear.

I have addressed this issue by writing a function that takes in the array of events and loops through it, splitting the ones that have multiple occurrences into one 'event' for each occurrence. It spits out an array of an almost identical format, allowing us to keep the same logic to process, filter, and display the events.

See below for example with "3d lacrosse"

Current search results (production):
image

Search results with proposed changes:
image

One thing to consider is that events like semester breaks (spring, winter, fall, etc.) will now show up with one occurrence for each 'day off'. (see below):
image
A future development that we might consider would be to find a way to compress these into one row that says something like "Spring Break | [location] | Mar 5-13, 2022 | [blank]"

Additionally, this pull request adds a filter option for showing recurring events or not.

image


src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+          date: thisOccurrenceDate,
+          location: (thisOccurrence.location ? thisOccurrence.location : thisEvent.location),
+          timeRange: thisOccurrenceTimeRange,
+          title: thisEvent.title,
+          startDateTime: thisOccurrenceStartDateTime,
+          allOccurrences: allOccurrences,
+          recurring: true
+        };
+
+        splitEvents.push(thisEventOccurrence);
```
EjPlatzer, line 326: You're still using `var` here. Can we change this to `const`? Also, technically the intermediate variable isn't necessary, since you're only pushing it onto `splitEvents`. You could just push the object literal.

Also, this could be made a lot more concise using the spread operator:
```suggestion
        splitEvents.push( {
          ...thisEvent,
          Event_ID: ("" + thisEvent.Event_ID + "-" + j),
          date: thisOccurrenceDate,
          location: (thisOccurrence.location || thisEvent.location),
          timeRange: thisOccurrenceTimeRange,
          startDateTime: thisOccurrenceStartDateTime,
          allOccurrences: allOccurrences,
          recurring: true
        };);
```

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+/**
+ * Splits up multiple occurrences of events into individual events
+ * and re-sorts the array
+ *
+ * @param {Object[]} events the array of events to process
+ * @returns {Object[]} the flattened array of events after splitting multiple occurrences
+*/
+const processMultipleOccurences = (events) => {
+
+  var splitEvents = [];
```
EjPlatzer, line 289: You used `var` here.
```suggestion
  const splitEvents = [];
```

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+
+  for (let i = 0; i < events.length; i++) {
+
+    let thisEvent = events[i];
```
EjPlatzer, line 293: You can use a `for...of` loop here for more concise syntax:
```suggestion
  for (const event of events) {
```

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+
+    // if the event has more than 1 occurrences, loop over them and create new events for each; otherwise add the event itself
+    if (thisEvent.Occurrences && thisEvent.Occurrences.length > 1) {
+
+      for (let j = 0; thisEvent.Occurrences && j < thisEvent.Occurrences.length; j++) {
+
+        const thisOccurrence = thisEvent.Occurrences[j];
```
EjPlatzer, line 300: A `for...of` loop could also be used here:
```suggestion
      for (const occurrence of event.Occurrences) {
```

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+        const thisOccurrenceStartDateTime = DateTime.fromISO(thisOccurrence.StartDate).toJSDate();
+        const thisOccurrenceEndDateTime = DateTime.fromISO(thisOccurrence.EndDate).toJSDate();
+        const thisOccurrenceDate = moment(thisOccurrenceStartDateTime).format('MMMM D, YYYY');
+        const thisOccurrenceTimeRange = ("" + moment(thisOccurrenceStartDateTime).format('h:mm A') + " - " + moment(thisOccurrenceEndDateTime).format('h:mm A'));
+        const allOccurrences = thisEvent.Occurrences.map((occurrence) => {return {start: occurrence.StartDate,
```
EjPlatzer, line 305: Do we need `allOccurrences` on the objects output from this function? Surely, once we split events into their occurrences, there's no need to store all the occurrences on every occurrence?

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+    // if the event has more than 1 occurrences, loop over them and create new events for each; otherwise add the event itself
+    if (thisEvent.Occurrences && thisEvent.Occurrences.length > 1) {
+
+      for (let j = 0; thisEvent.Occurrences && j < thisEvent.Occurrences.length; j++) {
+
+        const thisOccurrence = thisEvent.Occurrences[j];
+        const thisOccurrenceStartDateTime = DateTime.fromISO(thisOccurrence.StartDate).toJSDate();
+        const thisOccurrenceEndDateTime = DateTime.fromISO(thisOccurrence.EndDate).toJSDate();
+        const thisOccurrenceDate = moment(thisOccurrenceStartDateTime).format('MMMM D, YYYY');
+        const thisOccurrenceTimeRange = ("" + moment(thisOccurrenceStartDateTime).format('h:mm A') + " - " + moment(thisOccurrenceEndDateTime).format('h:mm A'));
```
EjPlatzer, line 304: Was there a reason to use two different date libraries here? You're using `DateTime` from luxon but then converting those into JavaScript dates and then converting those into moment `moment`s for formatting. I think luxon has it's own formatting solution that would probably streamline the process.

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+/**
+* Removes recurring events from the events list
+* 
+* @param {Object[]} events the list of events to be filtered
+* @returns {Object[]} the list of non-recurring events
+*/
+const removeRecurring = (events) => {
+  console.log(events.filter((event) => !event.recurring));
+  let filteredEvents = events.filter((event) => (!event.recurring));
+  return filteredEvents;
```
EjPlatzer, line 346: This is a minor thing, but was there a reason to store `filteredEvents` as in intermediate variable, rather than simply returning it directly?
```suggestion
  return events.filter((event) => (!event.recurring));
```

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
   return false;
 };
 
+/**
+ * Splits up multiple occurrences of events into individual events
+ * and re-sorts the array
+ *
+ * @param {Object[]} events the array of events to process
+ * @returns {Object[]} the flattened array of events after splitting multiple occurrences
```
EjPlatzer, line 285: Giving the param and return value type `Object[]` is not very helpful. We have type definitions for events and event occurrences in this file. Maybe you should use/update those? Having accurate type definitions is especially helpful because it informs VSCode's intellisense and code completion.

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
+  }
+
+  return splitEvents;
+}
+
+/**
+* Removes recurring events from the events list
+* 
+* @param {Object[]} events the list of events to be filtered
+* @returns {Object[]} the list of non-recurring events
```
EjPlatzer, line 342: Could we clarify the typing of these as well?

src/services/event.js
```diff
@@ -273,11 +277,83 @@ const makeMatchesFilters = (filters) => (event) => {
...
 };
 
+/**
+ * Splits up multiple occurrences of events into individual events
+ * and re-sorts the array
+ *
+ * @param {Object[]} events the array of events to process
+ * @returns {Object[]} the flattened array of events after splitting multiple occurrences
+*/
+const processMultipleOccurences = (events) => {
```
EjPlatzer, line 287: I think that instead of having this as a separate function called by all the HTTP functions, you could make this a helper function of `formatEvent` and tweak it just slightly to process occurrences for a single event. That way, the processing is included in the function that is already used to format the events in all other ways.

EjPlatzer (changes requested):
This is a great update! Excited to see this feature finally being added (it's been on the to-do list in one form or another for almost 3 years).

I have a few suggestions to clean things up further, but overall this looks really good!

src/services/event.js
bennettforkner: I'm not sure that is possible because formatEvent is supposed to return one event. This function would still have to return multiple events (for each occurrence). Thoughts?

src/services/event.js
bennettforkner: I did not intend to use it now, I thought that it may be useful to have in the future if we want to reference the other instances of this event from it.

bennettforkner: I guess whoever does that can just do that work themselves 👍

src/services/event.js
EjPlatzer: Hmm, that's a good point.

There is a solution, but it would mean returning an array of events from `formatEvent` and then `flatmap`ping over `formatEvent` instead of `map`ping. That would work perfectly, but it might be opaque to people who aren't as familiar with the process.

src/components/EventList/components/EventItem/event-item.scss
```diff
@@ -16,3 +16,7 @@
 .event-item:hover .descriptionText {
   color: $neutral-white;
 }
+
+.event-details {
+  width:100%;
+}
```
EjPlatzer, line 22: Did you add this to center the event details?
With width 100%:
Screen Shot 2021-05-07 at 9 12 39 AM

Without width 100%
Screen Shot 2021-05-07 at 9 12 56 AM

It's a decent update, but wouldn't the better solution be to left-align the text for readability?
Screen Shot 2021-05-07 at 9 14 15 AM

Also, we should consider getting rid of the "Description:" label. It is pretty redundant, since people are clicking on the event to expand further details.

src/services/event.js
EjPlatzer: I'm realizing now that, in reality, we should probably be doing this work in the backend (this is prime backend kind of data normalization). Especially because processing large amounts of data in a uniform way is not something that end user devices are good at.

EjPlatzer (changes requested):
So I had a couple thoughts. The biggest one (I know it's a little late for this, but I didn't realize until now) is that splitting occurrences out for display is really a backend kind of task.

src/services/event.js
bennettforkner: Hmmm. So you think hold off on this one then @EjPlatzer ?

bennettforkner:
We have determined that this issue needs back-end work before we can submit changes to the front-end. I am closing this pull request with plans to open a new one when the time comes.

PR #1117, opened 2021-05-11
Fix unscoped css rules affecting other components unintentionally

### LINKED ISSUE #1098: Fix unscoped css rules
Some css rules have been written for a component's css that target an html tag or a Material UI class and have no scoping. Because these rules are not scoped (e.g. by nesting them under a page specific class), they unintentionally affect all instances of the targeted component site wide (unless there is a more specific rule cancelling the site-wide one out). Any css rules affecting the whole site should only be in app.scss and should be intentionally added as a default setting. These rules were likely intended to affect only certain parts of the component they are included by, and have unintentionally affected other portions of the site.

The identified culprits so far are:
  1. In `src/views/Timesheets/timesheets.scss` tageting `.MuiGrid-item` with `text-align: center` which will center all text across the site for any children of a Material UI grid (which is almost all text).
  2. In `src/views/BannerSubmission/banner.scss` targeting `<p>` with `margin-left: 15px`. This will affect all paragraphs across the site, giving the appearance of indentation effectively.
  3. In `src/views/Help/help.scss` targeting `<strong>` with `font-size: 15px; font-weight: bold; margin-top: 20px`. This will affect all strong tags across the site, hard-coding their font size (which is bad for accessibility) and potentially muddying their margin.

The process for fixing these rules will involve properly scoping  them to affect only the intended targets, and then checking the rest of the site to ensure that there aren't other components which (unknowingly) depended on that styling and no longer look good without it.

EjPlatzer:
We had a few instances of css rules that were not "component scoped" (e.g. having the selectors nested under a component-level class specific to that component). These rules would be imported in the site-wide stylesheet and affect any html element that fit the selector. For example, `timesheet.scss` included a rule that selected all `MuiGrid-item` elements and set `text-align: center`. This selected every material ui Grid item across the site and aligned the text of their children to center. This broke the layout of many pages, and also affected the development of future pages because text would be centered "by default" without the designer knowing why.

This pr fixes #1098. The issues listed there were already affecting different components across the site. This pr also includes updates to css rules that aren't yet affecting other components but could very easily. For example, rules selecting classes like `heading` or `subtitle` that could innocently be used in other components and unknowingly be affected by those rules.


src/views/Home/components/DiningBalance/index.js
```diff
@@ -27,10 +27,9 @@ export default class DiningBalance extends Component {
   }
 
   async loadData() {
-    const diningInfoPromise = user.getDiningInfo();
-    const daysLeftPromise = session.getDaysLeft();
-    const daysLeft = await daysLeftPromise;
-    const diningInfo = await diningInfoPromise;
+    const diningInfo = await user.getDiningInfo();
+    const daysLeft = await session.getDaysLeft();
+    console.log(diningInfo);
```
bennettforkner, line 32:
```suggestion
```

src/views/Login/index.js
```diff
@@ -322,49 +256,7 @@ export default class Login extends Component {
...
-          onClose={this.handleCloseSnackbar.bind(this)}
-          ContentProps={{
-            'aria-describedby': 'message-id',
-            style: { backgroundColor: gordonColors.primary.cyan },
-          }}
-          message={
-            <span id="message-id">
-              Whoops! It looks like you're using Edge. Unfortunately, Gordon 360 doesn't support
-              Edge yet, but we hope to soon! In the meantime, please try Chrome or Firefox.
+              doesn't support IE. Please use a modern browser like Chrome or Firefox.
```
bennettforkner, line 259:
```suggestion
              doesn't support IE. Please use a modern browser like Chrome, Firefox, or Edge.
```

bennettforkner (changes requested):
Requesting a few changes. LGTM otherwise.

src/views/Home/components/DiningBalance/index.js
EjPlatzer: Oh good catch, thank you!

src/views/Login/index.js
EjPlatzer: I suppose this is acceptable since anyone running IE must be on Windows. The one thing is that we have instructions included for installing Chrome and Firefox, but not Edge. It's installed by default on new Windows 10 systems, but if we do have any users using IE, they won't be on Windows 10.

bennettforkner: Hmmm. Well, I dont think we should advertise that we support it without putting in instructions for installing it on Edge. I think we should just include it here and write instructions in order to cover all bounds. What do you think?

EjPlatzer: I'm not sure what you mean by "installing it on Edge". I was referring to the fact that this dialogue includes links to install Chrome and Firefox (as an alternative to using IE). Since anyone reading the dialogue will be running Windows (since they're using IE), they should have Edge available (although they might not have it by default since they probably won't be on Windows 10). So I guess we need to add a link to the Edge downloads page, and/or instructions for accessing it if you already have it installed.

That seems like it deserves a separate issue though, and doesn't necessarily fit this pull request. Realistically, I shouldn't have edited the login page as much as I did, but a lot of it was outdated/leftover from now solved bugs.

bennettforkner: I see what you're saying. Up to you on what to do here.

bennettforkner (approved):
LGTM.

PR #1121, opened 2021-05-17
Update Class Years of Developers

bennettforkner:
A few of the students who we just added to the about page as the 2020-21 SD group had incorrect class years. This pull request resolves that.


cpabbot (approved):
LGTM, thanks for the fix!

PR #1123, opened 2021-05-17
Removed IE Message, Refactored Login and Cleaned Guest Welcome

### LINKED ISSUE #1119: Add Edge to list of "Modern Browsers" in Login dialog for IE
The dialog that pops up when someone tries to access the site via Internet Explorer lists Chrome and Firefox as "modern browser" alternatives to IE. Now that Edge is chromium-based and therefore fully supported, we should consider adding it to that list, especially since it will probably be taking over some of IE's market share as the default Windows browser moving forward.

cpabbot:
### Remove React Message for Internet Explorer + Refactored Login + Cleaned Up Guest Welcome
Likely due to dependency upgrades, internet explorer is no longer supported at all by 360 (including the message we used to give them to use a 'modern browser'). Faced with this, we had to decide whether to drop support entirely, or manually browser check (with vanilla JS) to provide a message. Google Analytics reported one user on IE in the last year and 52 in the past three years, so we decided to drop support, as it is cleaner and somewhat more performant than browser checking every load.

**_Note:_** Last commit is just formatting so this PR is much smaller than it looks. (ignore `Help/index.js` changes)
**_Question:_** Should we remove `errorlog.js` manual browser checking in `parseNavigator` entirely since this would need to be implemented in vanilla JS for most unsupported browsers? It is used by the IDUploader, though I'm not sure if this is necessary.

'Guest Welcome' / login page cleanup:
image
image

--------
#### Old Description for Reference:
Status update: Currently we cannot load any of the browser warning code, likely because of a new dependency update or tool being used that breaks on IE. There was not a clear indicator of a culprit or way to track it down since all React code won't run. I wrote a script to detect browsers that we could use to redirect / inform users, but not sure if this is worth the potential efficiency loss of checking a user's browser every time. We will have to consider (perhaps check with @russtuck) whether this is worth pursuing or if we should just scrap all of the unsupported browser code.

Quick PR that cleans up a few things and adds one:
- capitalized "lts" to "LTS" to distinguish better from the word "its"
- removed known issue of Edge not working (it does)
- adjusted wording
- added Edge to list of "modern browsers" and added button for installation, but this needs to be tested out before uncommented

360 Site does not load for me at all on IE so I cannot test this change fully; will wait to merge until someone has tested.

Resolves #1119


bennettforkner:
IE does not load for me as well. I think there may be a problem with the site recognizing that a user is using Internet Explorer. That is a bigger issue that we will probably want to address before we merge this. @EjPlatzer

src/views/Login/index.js
```diff
@@ -134,92 +83,86 @@ export default class Login extends Component {
...
+          </Typography>
+          <Grid container justify="center">
+            <Button
+              variant="contained"
+              className="submit-button"
+              type="submit"
+              color="primary"
+              disabled={!username || !password || loading}
+            >
+              {loading ? <CircularProgress size={24} /> : 'Log in'}
```
EjPlatzer, line 131: Is there a good reason to not use `GordonLoader` here?

src/views/Login/index.js
```diff
@@ -134,92 +83,86 @@ export default class Login extends Component {
...
-              <Fab variant="extended" color="primary">
-                <GetAppIcon />
-                <Typography variant="subtitle1">Install Gordon 360</Typography>
-              </Fab>
-            </Grid>
+  };
+
+  return (
+    <div className="login">
+      <DocumentTitle title={`Login | ${projectName}`} />
```
EjPlatzer, line 95: This dependency is largely irrelevant now, since with hooks you can achieve the same functionality with simply:
```
useEffect(() => {
  document.title = `Login | ${projectName}`
}, [])
```

src/views/Login/index.js
```diff
@@ -134,92 +83,86 @@ export default class Login extends Component {
...
+      <Grid container direction="column" className="container">
+        <img className="login-img" src={GordonLogoVerticalWhite} alt={`${projectName}`} />
+        <form onSubmit={(e) => logIn(e)}>
```
EjPlatzer, line 98: Since you're only passing on the argument, you can just pass `logIn` directly, instead of creating an arrow function callback that passes on the argument to `logIn`:
```suggestion
        <form onSubmit={logIn}>
```

EjPlatzer (changes requested):
This looks great! I had a couple suggestions for ways it might be improved, but overall I think trimming away the IE code and refactoring Login is a huge win.

src/views/Login/index.js
cpabbot: Are you suggesting I delete `DocumentTitle` and create that hook? Unless there's some performance difference, this seems more consistent with all of the other document title setting in `Header/index.js`

EjPlatzer: Yes, because the`react-document-title` dependency isn't well-maintained (it hasn't been updated since 2018 and depends on a package that still uses the deprecated `componentWillMount`) and is nearly useless. So gradually phasing it out is a huge win in my book.

src/components/GordonUnauthorized/index.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { Grid, Card, CardContent, Button } from '@material-ui/core/';
+
+/**
+ * @param {Object} props the component props
+ * @param  {Boolean} props.open Boolean that determines if the dialog should be visible or not
+ *
+ * @returns {JSX.Element} A dialog box with the passed in content
```
EjPlatzer, line 8: This isn't a dialog box, it's a card with an unauthorized message

src/components/GordonUnauthorized/index.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { Grid, Card, CardContent, Button } from '@material-ui/core/';
+
+/**
+ * @param {Object} props the component props
+ * @param  {Boolean} props.open Boolean that determines if the dialog should be visible or not
```
EjPlatzer, line 6: These props are also incorrect.

cpabbot: Thanks for catching these, was residual from copy/paste

src/views/Admin/index.js
```diff
@@ -11,7 +12,11 @@ const Admin = ({ authentication }) => {
   const isOnline = useNetworkStatus();
 
   useEffect(() => {
-    setIsAdmin(user.getLocalInfo().college_role === 'god');
+    try {
+      setIsAdmin(user.getLocalInfo().college_role === 'god');
+    } catch (error) {
+      // Unauthorized exception
+    }
```
EjPlatzer, line 19: You should specifically catch an unauthorized error here, and throw any other errors.
```suggestion
    } catch (error) {
      if (error instanceof AuthError) {
        // Unauthorized exception expected when user not authenticated
      } else {
        throw error;
      }
    }
```

src/views/Home/components/GuestWelcome/index.js
```diff
@@ -1,32 +1,110 @@
...
-import React from 'react';
-import Login from 'views/Login';
-import './guest-welcome.css';
+import React, { useEffect, useState } from 'react';
+import { Typography, Grid, Fab, Card, CardHeader, CardContent } from '@material-ui/core';
+import useNetworkStatus from 'hooks/useNetworkStatus';
+import { projectName } from 'project-name';
+import PropTypes from 'prop-types';
+import PWAInstructions from 'components/PWAInstructions/index';
+import LoginDialogue from '../../../../components/LoginDialogue';
```
EjPlatzer, line 7: You can use an absolute import here for consistency
```suggestion
import LoginDialogue from 'components/LoginDialogue';
```

src/views/IDUploader/index.js
```diff
@@ -22,7 +22,7 @@ import 'cropperjs/dist/cropper.css';
 import './IDUploader.css';
 import user from 'services/user';
 import errorLog from 'services/errorLog';
-import Login from 'views/Login';
+import Login from '../../components/LoginDialogue';
```
EjPlatzer, line 25: Same as above
```suggestion
import Login from 'components/LoginDialogue';
```

src/views/WellnessCheck/index.js
```diff
@@ -3,7 +3,7 @@ import React, { useState, useEffect } from 'react';
 import GordonLoader from 'components/Loader';
 import WellnessQuestion from 'components/WellnessQuestion';
 import HealthStatus from './components/HealthStatus';
-import Login from 'views/Login';
+import Login from '../../components/LoginDialogue';
```
EjPlatzer, line 6: Absolute import
```suggestion
import Login from 'components/LoginDialogue';
```

EjPlatzer (changes requested):
This is looking great! A huge improvement. Just a few minor cleanups and then I think this'll be ready.

src/components/LoginDialogue/index.js
```diff
@@ -0,0 +1,91 @@
...
+        <TextField
+          id="password"
+          label="Password"
+          type="password"
+          autoComplete="current-password"
+          value={password}
+          onChange={(e) => setPassword(e.target.value)}
+          margin="normal"
+          fullWidth
+        />
```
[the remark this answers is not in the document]
cpabbot: I'm not totally enthralled visually by using filled text boxes here, but it's growing on me and Dr. Senning agreed with the study results haha, so here we have it.

EjPlatzer (approved):
LGTM!

PR #1125, opened 2021-05-18
Fix error displaying Activity Profile when Advisors or GroupAdmins is null

EjPlatzer:
The Advisors and GroupContacts components were not returning from `render()` when they had no data to display. React requires every component to return, so this was causing an error and getting the stuck in a perpetual loading loop.

I have fixed it so that these components return null when they have no data, and I have also refactored them, removing a great deal of unnecessary code for loading that didn't actually load any data.


cpabbot (approved):
LGTM, code looks cleaner and minimized.

PR #1128, opened 2021-05-18
Involvements UI Improvements and Bug Fixes

### LINKED ISSUE #1126: Involvements messes with browser back/forward arrows
This was an issue I wrote up, then removed when I realized it was working in 360 prod, and am now putting back after remembering that train is our updated version that will eventually be pushed. I'm going to leave the original issue I wrote below, but you can see the difference in execution between 360 prod and 360 train. This must have been changed for some consistency reason, which is probably good but the browser history might need to be manually modified to preserve browsing consistency.

EDIT: Related issue is that on page load, we reset the term to the current term so we cannot refresh or share a page with a selected term in the URL. I will attempt to fix this in PR #1128.

-------------------------------------------------------------------------------------------

This is a general issue with execution of storing data within URL (the data stored like a GET fetch with `?session=202101` appended to end of URL), and I have not investigated the full extent of potential issues, none of which I think are very major or urgent.

Rather than try to explain, here is a sample issue demo:
- When clicking on "involvements" tab, we append data to the URL for the current session. The problem is that the way we do this adds to the browser history, so clicking 'back' first simply removes that data (which does nothing until it's not the default session). This makes sense from a developer's perspective, but the user just sees that they have to click 'back' twice when they only went 'forward' once.

### LINKED ISSUE #1127: Cancelling involvement request does not actually cancel it?
If I click join, it creates a request. When I click cancel, it hides temporarily but does not appear to change database as the request returns upon refresh.

Join should also change to "Leave" once joined on activity profile.

One should also be able to click on the request itself to link straight to the activity profile.

cpabbot:
### Language:
- Changed "session" to "term" because this makes more sense to a college student, for whom this is primarily targeting
- Changed "type of involvement" to "type" for the same reasons we do not say "search involvements" or "term/session of involvement" (to reduce redundancy, improve consistency, and also to prevent the label from overflowing the input dropdown field)
- Change "no involvements for `term name here (e.g. Spring 20-21 Academic Year) `" to simply "no involvements for this term" because the header already specifies the appropriate term and so this is redundant.
- Changed "All Involvements" header to specify the term, like we had done with the personal involvements

### Membership requests:
_Note: this is not the most elegant code and should probably be improved in future, but I think the direction is good_
- The average student on the average day will not have an membership requests sent nor received, so the page now hides the card entirely if there are no membership requests.
- For students who are not leading any activities, they now see a collapsible card (default open) of their current sent requests. This is better UX than before when they would see a "Membership Requests" header and inside a seemingly redundant "Requests Sent" header (redundant to the user that generally isn't aware of "Requests Received").
- For students who _are_ leading activities, they see the same card as before with two dropdowns - one for requests sent and one for requests received.

### Bug Fixes:
- Before the selectedSession was loaded properly, the session text for the headers was left undefined; this was changed to default to blank text so the user potentially sees for example "Involvements" -> "Current Involvements" instead of "undefined Involvements" -> "Current Involvements"
- Issue #1126 explains a bug that occurred in using browser back/forward buttons. This happened essentially because we used a UseEffect that had dependencies [selectedSession, history]. When a session was selected, this function pushed it onto the history, which then essentially called the function again since history was changed. This second call then messed up the browser history the same way that if you hit 'undo' and then type something, you can no longer 'redo'. Instead, this was changed to a select handler function rather than UseEffect.
- Issue #1127 explains a bug where membership requests were not being cancelled successfully. This ended up being a simple error where `member.RequestID` needed a capital `R`

### Notes for further improvement:
- I'm not sure if it looks good to have the involvements search area 'fixed' because a material design background should never overlap the foregrounded card, and also especially since this is inconsistent with the events page. That being said, I can understand the functionality and will leave it for now. Created issue #1129.
- There is a separate cancel icon depending if a request has been approved, which makes sense functionally but might be poor visual design. Created issue #1130.
image

Resolves #1126, 
Resolves #1127

### DIFF
--- a/src/views/InvolvementsAll/components/Requests/components/RequestSent/index.js
+++ b/src/views/InvolvementsAll/components/Requests/components/RequestSent/index.js
@@ -7,7 +7,7 @@ import membership from 'services/membership';
 
 const RequestSent = ({ member, onCancel }) => {
   const handleCancel = () => {
-    membership.cancelRequest(member.requestID);
+    membership.cancelRequest(member.RequestID);
     onCancel(member); // Updates state of parent component to cause rerender
   };
 
--- a/src/views/InvolvementsAll/components/Requests/index.js
+++ b/src/views/InvolvementsAll/components/Requests/index.js
@@ -95,7 +95,7 @@ const Requests = () => {
     return (
       <Grid item xs={12} lg={8}>
         <Card className="requests">
-          <Accordion defaultExpanded="true">
+          <Accordion defaultExpanded>
             <AccordionSummary
               aria-controls="received-requests-content"
               expandIcon={<ExpandMore style={{ color: 'white' }} />}
@@ -131,7 +131,7 @@ const Requests = () => {
       </Grid>
     );
   }
-  // otherwise hide since we have no requests
+  // otherwise hide component entirely since we have no requests
   return null;
 };
 
--- a/src/views/InvolvementsAll/components/Requests/index.js
+++ b/src/views/InvolvementsAll/components/Requests/index.js
@@ -33,59 +33,106 @@ const Requests = () => {
     setRequestsSent((prevRequestsSent) => prevRequestsSent.filter((r) => r !== request));
   };
 
-  return (
-    <Card className="requests">
-      <CardHeader title="Membership Requests" className="requests-header" />
-
-      <CardContent>
-        {involvementsLeading?.length > 0 && (
-          <Accordion>
+  // if leading involvements, show two dropdowns for requests received/sent
+  if(involvementsLeading?.length > 0) {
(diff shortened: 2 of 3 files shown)


bennettforkner (approved):
LGTM

bennettforkner:
Just get the "Lint & Build" check to run before merging of course. It may take a bit for GitHub actions to come back online.

EjPlatzer (approved):
LGTM! Good touch with conditionally hiding the requests card, that's very nice!

We'll have to look into styling the search bar appropriately so it doesn't look odd hovering over the cards, but that's a separate issue. 

Thanks for tracking down some insidious loading and routing bugs!

PR #1136, opened 2021-05-21
People Search URL Work: Partial URL Queries

### LINKED ISSUE #802: Update People tab URL parameters

cpabbot:
- People Search now only updates the parts of the URL that are not blank. Likewise it also accepts partial URLs. For example, entering the last name 'Abbot' now appends `?lastName=Abbot` rather than `https://360.gordon.edu/people?firstName=&lastName=abbot&major=&minor=&hall=&classType=&homeCity=&state=&country=&department=&building=&includeAlumni=false`
- URL is now kept much more in sync with actual search results. For example, clearing filters clears the URL as well.
- Refactor: Search (when user presses enter or search) and "GoBackPage" (also used for loading from URL and searching) had a lot of duplicated code. There is now a single universal `search` function that performs the actual people search work, with more specific methods `updateURL` and `loadPageFromURL` that are called depending along with the search (depending on the method of search).
- All of the search values are now stored in a single object rather than several state variables, which makes everything cleaner (though might cause unnecessary updates to state/url).

Opportunities for Improvement:
- When this is presumably one day refactored to use React Hooks, the URL can be kept in sync with the actual search results much more cleanly. One issue right now is that if you click the 'People Search' tab while having some of the input fields entered, it will update the URL to remove all queries (because it resets the route) but it won't clear the filters/queries themselves. With React Hooks, we could probably do this with a UseEffect with url/input dependencies, whereas right now I think the only way would be through the ComponentDidUpdate method which would be highly inefficient. (might be a good way but UseEffect makes sense to use here I think)

Resolves #802


src/views/PeopleSearch/index.js
```diff
@@ -89,13 +89,63 @@ const noResultsCard = (
...
+  </Media>
+);
+
 class PeopleSearch extends Component {
   constructor(props) {
     super(props);
 
     this.state = {
       personType: '',
-      nameExpanded: true,
```
cpabbot, line 98: unused

src/views/PeopleSearch/index.js
```diff
@@ -134,182 +182,30 @@ class PeopleSearch extends Component {
...
-
   componentDidUpdate() {
     window.onpopstate = () => {
       if (!window.location.href.includes('?')) {
-        window.location.reload();
-      } else {
-        this.goBackPage();
+        this.setState({ header: '', peopleSearchResults: null });
       }
+      this.loadSearchParamsFromURL();
```
cpabbot, line 190: People search is significantly faster with loadSearchParamsFromURL than actually reloading

src/views/PeopleSearch/index.js
```diff
@@ -508,20 +346,7 @@ class PeopleSearch extends Component {
...
-    minor,
-    hall,
-    classType,
-    homeCity,
-    state,
-    country,
-    department,
-    building,
-  ) {
+  async search() {
```
cpabbot, line 349: Removed parameters because these parameters were being called by the identical state properties anyway

src/views/PeopleSearch/index.js
```diff
@@ -577,7 +395,7 @@ class PeopleSearch extends Component {
       } else {
         this.setState({
           peopleSearchResults: (
-            <Media query="(min-width: 1025px)">
+            <Media query="(min-width: 960px)">
```
cpabbot, line 398: The two old search methods had an inconsistency here. Should be 960px as the medium breakpoint.

src/views/PeopleSearch/index.js
```diff
@@ -589,15 +407,49 @@ class PeopleSearch extends Component {
...
-          header: this.makeHeader(),
+          header: peopleSearchHeader,
         });
       }
     }
   }
 
-  getDate = () => {
-    return new Date();
-  };
```
cpabbot, line 600: Was confusing, this is now done right with the April Fools code (which it is for)

src/views/PeopleSearch/index.js
```diff
@@ -589,15 +407,49 @@ class PeopleSearch extends Component {
...
+    searchParameters += this.state.includeAlumniSearchValue
+      ? `includeAlumni=${this.state.includeAlumniSearchValue}&`
+      : '';
+
+    if (searchParameters[searchParameters.length - 1] === '&') {
+      searchParameters = searchParameters.substring(0, searchParameters.length - 1);
+    }
+
+    this.props.history.push(searchParameters);
+  }
```
cpabbot, line 452: This method only adds the parameters that are not blank. The last few lines of this simply remove the unnecessary last `&` at the end of the url

src/views/PeopleSearch/index.js
```diff
@@ -1200,6 +1054,7 @@ class PeopleSearch extends Component {
                             header: '',
                             peopleSearchResults: null,
                           });
+                          this.props.history.push('?');
```
cpabbot, line 1057: The `Clear All` Button now clears the URL as well

src/components/Header/index.js
```diff
@@ -122,7 +122,7 @@ const GordonHeader = ({ authentication, onDrawerToggle, onSignOut }) => {
     }
   };
 
-  const disablableTab = (name, icon) => {
+  const requiresAuthTab = (name, icon) => {
```
cpabbot, line 125: Renamed for clarity

src/views/PeopleSearch/index.js
```diff
@@ -325,102 +222,44 @@ class PeopleSearch extends Component {
...
+    let lastName = urlParams.get('lastName')?.trim() || '';
+    let major = urlParams.get('major')?.trim() || '';
+    let minor = urlParams.get('minor')?.trim() || '';
+    let hall = urlParams.get('hall')?.trim() || '';
+    let classType = urlParams.get('classType')?.trim() || '';
+    let homeCity = urlParams.get('homeCity')?.trim() || '';
+    let state = urlParams.get('state')?.trim() || '';
+    let country = urlParams.get('country')?.trim() || '';
+    let department = urlParams.get('department')?.trim() || '';
+    let building = urlParams.get('building')?.trim() || '';
```
cpabbot, line 243: Needs `?.trim()` instead of `.trim()` in case a particular URL query does not exist

src/views/PeopleSearch/index.js
```diff
@@ -589,15 +408,49 @@ class PeopleSearch extends Component {
...
+          header: peopleSearchHeader,
         });
       }
     }
   }
 
-  getDate = () => {
-    return new Date();
-  };
+  async updateURL() {
```
EjPlatzer, line 417: The verbosity of this method suggests that we could probably improve the code a lot by storing search values in an object of the form `{ firstName: value, lastName: otherValue, ...}`. Then, rather than repeating the same process for every searchValue, we could simply map over the values: 
```
const searchParameters = Object.entries(this.state.searchValues)
  .map(([key, value]) => value ? `${key}=value` : '') // [ 'firstName=value', 'state=texas']
  .join('&') // 'firstName=value&state=texas' - note the lack of trailing &

if (searchParameters) {
  this.props.history.push(`?${searchParameters}`)
}
```

src/views/PeopleSearch/index.js
EjPlatzer: Is the `'?'` necessary here? Typically, that marks the beginning of a query string, so it's a little odd to have just that.

EjPlatzer (changes requested):
This is a huge improvement! I did suggest one possible way to clean up the handling of search parameters further.

src/views/PeopleSearch/index.js
cpabbot: Okay so the problem with this approach that I am now realizing is that this will mess up setStates because in order to update the search values, we would have to create an entirely new object. For example, in `handleFirstNameInputChange` we cannot just write
```
this.setState({
      searchValues.firstNameSearchValue: e.target.value,
    });
```
I think this would be an issue to some extent because then anytime a value is changed, everything updates (including URL maybe). Not sure how bad this would be in practice but certainly doesn't seem ideal.

EjPlatzer: You can just do `this.setState({searchValues: {...searchValues, firstName: e.target.value}})` right?

cpabbot: First of all, yes and I had no idea that ("spread operator") was a thing. But secondly, the concern is still just that the state will get set every time. We lose the feature of setState where it only changes the value if the value has changed. This might have no effect, especially since we're only setting this when an onChange is called, but there might be a concern with "oversetting".

EjPlatzer: Ah, right, I get what you mean. Because `this.setState` only does shallow object comparison for `state`, having a nested object that gets overwritten will cause a re-render every time. I think that shouldn't affect performance negatively though, since there's so few of them. That's something that can be solved when this page is refactored to use hooks.

src/views/PeopleSearch/index.js
```diff
@@ -287,83 +281,70 @@ class PeopleSearch extends Component {
...
-      this.state.hallSearchValue === '' &&
-      this.state.classTypeSearchValue === '' &&
-      this.state.homeCitySearchValue === '' &&
-      this.state.stateSearchValue === '' &&
-      this.state.countrySearchValue === '' &&
-      this.state.departmentSearchValue === '' &&
-      this.state.buildingSearchValue === ''
-    ) {
-      // do not search
+    if (!Object.values(this.state.searchValues).some((x) => x !== '' && x !== false)) {
```
EjPlatzer, line 346: While this is a good explicit check, `''` and `false` are both falsy values, so this would work just as well as `some((x) => x)

src/views/PeopleSearch/index.js
```diff
@@ -415,59 +382,19 @@ class PeopleSearch extends Component {
...
-      : '';
-    searchParameters += this.state.buildingSearchValue
-      ? `building=${this.state.buildingSearchValue}&`
-      : '';
-    searchParameters += this.state.includeAlumniSearchValue
-      ? `includeAlumni=${this.state.includeAlumniSearchValue}&`
-      : '';
+    const searchParameters = Object.entries(this.state.searchValues)
+      .map(([key, value]) => (value ? `${key}=${value}` : '')) // [ 'firstName=value', 'state=texas']
+      .filter((n) => n) // removes empty strings
```
EjPlatzer, line 387: It might be slightly more performant to filter and then map, although with an array of size 12 it's unlikely to matter

src/views/PeopleSearch/index.js
```diff
@@ -415,59 +382,19 @@ class PeopleSearch extends Component {
...
+      .join('&'); // 'firstName=value&state=texas' - note the lack of trailing &
```
EjPlatzer, line 388: I'm not sure the comment is necessary. It would only really make sense to someone who knew the old code I think.

src/views/PeopleSearch/index.js
EjPlatzer: Also, the comment is somewhat redundant in my opinion.

EjPlatzer (approved):
LGTM! This is an excellent update! Much cleaner code and urls! A true win-win.

src/views/PeopleSearch/index.js
cpabbot: That makes sense, I'll change that. Might leave the comment just because I'm not familiar enough with falsy values etc. for this to be intuitive, so I'm guessing other young developers might not understand its purpose just by reading it.

EjPlatzer (approved):
LGTM!

PR #1137, opened 2021-05-21
Adjusted people search results alignment, changed type to description

cpabbot:
Adjusted the people search results layout sizing so that way every column has an equal width; class/job title originally had a wider column because faculty/staff have longer titles but this is arguably not worth the inconsistency. Also changed type to description honestly just because it looked bad with a word as short as 'type' which also doesn't have very much significance anyway (type of what, is this telling me what type of person they are?).

Before:
image

After:
image


EjPlatzer (approved):
LGTM! A small but nice update to something that was fairly ugly on certain screen sizes.

PR #1142, opened 2021-05-27
Created UseQueryState Hook

cpabbot:
Extracts logic of events URL updating into a separate UseQueryState hook that can be used for other instances where we want to sync state with URL query parameters. All the information needed for this PR is documented within the code itself.

Also fixes some layout issues with events that you can see in `EventsList/index.js`

Minor issue that we can fix if need be: currently when you remove the last filter it collapses the whole menu area, rather than waiting for you to click 'clear filters'. I'm not too concerned about this mostly since 'Clear Filters' acting as a show/hide toggle isn't really intuitive anyways.

Other issue: setting "include past" takes forever to render. Pagination would be nice goal, though even without it this was faster before, so this should probably be fixed (or at least understood why this is slower) before merging.

### DIFF
--- a/src/hooks/useQueryState.js
+++ b/src/hooks/useQueryState.js
@@ -1,6 +1,17 @@
 import { useState, useCallback, useEffect, useLayoutEffect } from 'react';
 import { useHistory } from 'react-router-dom';
 
+export const types = {
+  SingleValue: 'SingleValue',
+  Array: 'Array',
+  Boolean: 'Boolean',
+};
+const defaultValues = {
+  SingleValue: '',
+  Array: [],
+  Boolean: false,
+};
+
 /**
  * Custom hook to useState synced with URL query parameters.
  * Inspiration from François Best https://github.com/47ng/next-usequerystate/tree/next/src
@@ -26,24 +37,13 @@ import { useHistory } from 'react-router-dom';
  * we cannot set to an initial value or else we might load (and overwrite the state) from the url
  * before we can set the url
  *
- * @param {String} keyParam - key of the QueryState variable
+ * @param {String} key - key of the QueryState variable
  * @param {String} typeString - the type of variable (must correspond with a types enum value)
  * @param {String} initial - initial value of the QueryState variable (optional)
  *          *** DOES NOT currently work because of loading from URL before URL gets set
  * @returns {[var, Callback]} - the state variable, the setQueryState callback function
  */
-function useQueryState(keyParam, typeString, initial) {
-  const key = keyParam; // we should never change the particular key once we start using it
-  const types = {
-    SingleValue: 'SingleValue',
-    Array: 'Array',
-    Boolean: 'Boolean',
-  };
-  const defaultValues = {
-    SingleValue: '',
-    Array: [],
-    Boolean: false,
-  };
+function useQueryState(key, typeString, initial) {
   const type = types[typeString];
   if (type === undefined)
     throw new Error(`type must be one of the given types: ${Object.values(types)}`);
--- a/src/hooks/useQueryState.js
+++ b/src/hooks/useQueryState.js
@@ -39,7 +39,7 @@ const defaultValues = {
  *
  * @param {String} key - key of the QueryState variable
  * @param {String} typeString - the type of variable (must correspond with a types enum value)
- * @param {String} initial - initial value of the QueryState variable (optional)
+ * @param {String} [initial] - initial value of the QueryState variable
  *          *** DOES NOT currently work because of loading from URL before URL gets set
  * @returns {[var, Callback]} - the state variable, the setQueryState callback function
  */
@@ -52,11 +52,11 @@ function useQueryState(key, typeString, initial) {
   const history = useHistory();
 
   /**
-   * Anytime the state changes, setURLParam is invoked
+   * Anytime the state changes, setURLFromState is invoked
    * This function updates the URL with the state (if changed)
    */
   useEffect(() => {
-    const setURLParam = async () => {
+    const setURLFromState = async () => {
       let urlParams = new URLSearchParams(history.location.search);
 
       // if the url value is already the same as the state value -> don't set the url again
@@ -80,21 +80,22 @@ function useQueryState(key, typeString, initial) {
 
       history.push('?' + urlParams);
     };
-    setURLParam();
+    setURLFromState();
     /*********************** ESLint Disable Justification: *******************************/
-    // ESLint wants to include `defaultVal, key, history` as dependencies but to prevent
+    // ESLint wants to include `history` as dependencies but to prevent
     // extra loops of ~ url setting state setting the url ~ we do not include.
     // The only time we want to set the URL manually is when the state changes
+    // (key and defaultVal are constant)
     /*************************************************************************************/
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [state]);
+  }, [state, key, defaultVal]);
 
   /**
-   * Anytime the url is changed, loadURLParams is invoked
+   * Anytime the url is changed, loadStateFromURL is invoked
    * This function pulls the value of the key from the url
    *   and updates the state with it (if changed)
    *
-   * This is a useLayoutEffect rather than useEffect in order to prioritize loadURLParams;
+   * This is a useLayoutEffect rather than useEffect in order to prioritize loadStateFromURL;
    *   this is necessary for preventing the race condition where (for ex.) the user clicks 'back'
    *   twice quickly and the first 'back' sets the URL params before the second 'back' can
    *   load the accurate new URL param. This is still not totally perfect but only breaks when
@@ -102,7 +103,7 @@ function useQueryState(key, typeString, initial) {
    * Comparison of the two effects: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect
    */
   useLayoutEffect(() => {
-    const loadURLParams = async () => {
+    const loadStateFromURL = async () => {
       const urlParams = new URLSearchParams(history.location.search);
       let urlValue = urlParams.get(key);
 
@@ -118,8 +119,7 @@ function useQueryState(key, typeString, initial) {
           // if the same value (or unset and state array is empty) -> do nothing
           // this check is necessary because [] -> [] are different memory objects
           if (urlValue === state.toString() || (!urlValue && state.length === 0)) return;
-          else if (!urlValue) setState(defaultVal);
-          else setState(urlValue.split(','));
+          else setState(urlValue?.split(',') ?? defaultVal);
           break;
 
         case types.SingleValue:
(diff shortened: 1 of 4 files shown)


src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+   *   this is necessary for preventing the race condition where (for ex.) the user clicks 'back'
+   *   twice quickly and the first 'back' sets the URL params before the second 'back' can
+   *   load the accurate new URL param. This is still not totally perfect but only breaks when
+   *   a user really messes with the navigation significantly and quickly.
+   * Comparison of the two effects: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect
+   */
+  useLayoutEffect(() => {
+    const loadURLParams = async () => {
+      const urlParams = new URLSearchParams(history.location.search);
+      let urlValue = urlParams.get(key);
```
EjPlatzer, line 107: Does this load array values properly? I only ask because I know that there's also `urlParams.getAll(key)`.

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+ * - UseLayoutEffect is not a perfect solution to the race condition issue. Symptoms of this
+ * possibly poor design are evident when the browser 'back'/'forward' arrows are tested heavily
+ * which can skip queryStates (likely not a practical use case), as well as the fact that
+ * we cannot set to an initial value or else we might load (and overwrite the state) from the url
+ * before we can set the url
+ *
+ * @param {String} keyParam - key of the QueryState variable
+ * @param {String} typeString - the type of variable (must correspond with a types enum value)
+ * @param {String} initial - initial value of the QueryState variable (optional)
+ *          *** DOES NOT currently work because of loading from URL before URL gets set
```
EjPlatzer, line 32: One way that could possibly fix this would be to have a useEffect that runs the first time this hook is called (i.e. it's dependencies are simply `[]`) that will set the URL to `initial`, and also set a separate state value like `isInitialLoad` to false. Then, the other useEffects would only run once `isInitialLoad` is false. It's a lot of extra code for a small gain, but considering the current API of this hook is broken, it might be worth trying.

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+      urlParams = urlParams
+        .toString()
+        .replace(/=true|=$/g, '')
+        .replace(/=&/, '&');
+
+      history.push('?' + urlParams);
+    };
+    setURLParam();
+    /*********************** ESLint Disable Justification: *******************************/
+    // ESLint wants to include `defaultVal, key, history` as dependencies but to prevent
```
EjPlatzer, line 85: Presumably, including `defaultVal` and `key` would not do any harm, since those are stable for the lifetime of the hook. Including them would also make it more clear that it's _specifically_ `history` that we do not want to depend on.

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+          return;
+
+        // different logic if we are working with an array than with single value
+        // because setState will not do deep array comparison
+        case types.Array:
+          // if the same value (or unset and state array is empty) -> do nothing
+          // this check is necessary because [] -> [] are different memory objects
+          if (urlValue === state.toString() || (!urlValue && state.length === 0)) return;
+          else if (!urlValue) setState(defaultVal);
+          else setState(urlValue.split(','));
```
EjPlatzer, line 122: Would it be clearer to combine these?:
```suggestion
          else setState(urlValue ? urlValue.split(',') : defaultVal);
```
Or since `urlValue` is `null` when `key` is not present in the url, you could even do:
```suggestion
          else setState(urlValue?.split(',') ?? defaultVal);
```

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+   *
+   * This is a useLayoutEffect rather than useEffect in order to prioritize loadURLParams;
+   *   this is necessary for preventing the race condition where (for ex.) the user clicks 'back'
+   *   twice quickly and the first 'back' sets the URL params before the second 'back' can
+   *   load the accurate new URL param. This is still not totally perfect but only breaks when
+   *   a user really messes with the navigation significantly and quickly.
+   * Comparison of the two effects: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect
+   */
+  useLayoutEffect(() => {
+    const loadURLParams = async () => {
```
EjPlatzer, line 105: It might be clearer to call this function `loadStateFromURL`

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+    // (ex. 'filters' array has a few different possible filter values)
+    if (Array.isArray(defaultVal) && !JSON.stringify(state) === JSON.stringify(val)) return;
+    setState(val);
+    /*********************** ESLint Disable Justification: *******************************/
+    // We do not want the defaultVal or state dependencies here because they were simply added to
+    // prevent unecessarily setting the value array. We do not want any dependencies that might
+    // change the state after the initial setQueryState is called from the user function.
+    /*************************************************************************************/
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
EjPlatzer, line 159: I think you are misunderstanding how `useCallback` works. Unlike `useEffect`, we do not call the callback of `useCallback` when its dependencies change. Rather, what happens is the callback returned by `useCallback` has its definition updated. The way you have this defined currently, `useCallback` will create the `setQueryState` with `state` = `initial ?? defaultVal` (the value returned by `useState` above). That will never change. So even as `state` gets updated over time, `setQueryState` will still be comparing `val` to that first state of `initial ?? defaultVal`. I'm fairly certain the intended behavior is that `setQueryState` is redefined whenever `state` changes so that it is comparing against the most recent value of `state`.

Also, since `defaultVal` is stable, there should be no issues including it in the dependency array.

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Callback function to allow user component to setQueryState just like one would use setState
+   */
+  const setQueryState = useCallback((val) => {
+    // this check prevents unnecessary array setting
+    // ([] -> [] are technically different memory objects)
+    // this is safe to deep compare because the array value should not have very many values
+    // (ex. 'filters' array has a few different possible filter values)
+    if (Array.isArray(defaultVal) && !JSON.stringify(state) === JSON.stringify(val)) return;
```
EjPlatzer, line 151: Is there a reason you're checking the type of `defaultVal` instead of referencing `type` like you do elsewhere in this hook?

Also, `JSON.stringify()` will work but it's one of the slowest options. Perhaps `Array.every` is better here?
```suggestion
    if (type === types.Array && !state.every((elem, index) => elem === val[index])) return;
```

src/views/Events/index.js
```diff
@@ -14,22 +13,25 @@ import {
...
+  const [includePast, setIncludePast] = useQueryState('past', 'Boolean');
   const [loading, setLoading] = useState(true);
-  const [filters, setFilters] = useState([]);
+  const [filters, setFilters] = useQueryState('filters', 'Array');
   const futureEvents = useMemo(() => gordonEvent.getFutureEvents(allEvents), [allEvents]);
 
   useEffect(() => {
-    const loadEvents = async () => {
+    // Single use - Loads events data on page load / authentication
+    const loadEventsPage = async () => {
```
EjPlatzer, line 34: If you rename the function `loadAllEvents`, the comment is probably redundant.

src/views/Events/index.js
```diff
@@ -38,44 +40,24 @@ const Events = (props) => {
...
-
       setLoading(false);
     };
+    loadEventsPage();
+  }, [props.authentication]);
 
-    loadEvents();
-  }, [props.authentication, props.location.search]);
-
+  // When the actual events data changes update the events
```
EjPlatzer, line 48: I'm not sure what this comment means. What is "the actual events data"? Also, I think this `useEffect` is small enough that it speaks for itself.

src/views/Events/index.js
```diff
@@ -38,44 +40,24 @@ const Events = (props) => {
...
   useEffect(() => {
     setEvents(includePast ? allEvents : futureEvents);
   }, [includePast, allEvents, futureEvents]);
 
+  // Apply filters to the events data
```
EjPlatzer, line 53: Again, I think the code is self-documenting here.

EjPlatzer (changes requested):
The updates to the events page are great! I'm still not totally convinced that `useQueryState` is the right way to go here, but I can see that you've made it work well. I left several comments about ways it might be improved slightly or things that weren't totally clear to me.

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+    Boolean: 'Boolean',
+  };
+  const defaultValues = {
+    SingleValue: '',
+    Array: [],
+    Boolean: false,
+  };
+  const type = types[typeString];
+  if (type === undefined)
+    throw new Error(`type must be one of the given types: ${Object.values(types)}`);
```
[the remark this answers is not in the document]
cpabbot: This makes sense about exporting `types`! I thought there must be some way to do that but I couldn't figure it out; I can definitely do that.

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+ * before we can set the url
+ *
+ * @param {String} keyParam - key of the QueryState variable
+ * @param {String} typeString - the type of variable (must correspond with a types enum value)
+ * @param {String} initial - initial value of the QueryState variable (optional)
+ *          *** DOES NOT currently work because of loading from URL before URL gets set
+ * @returns {[var, Callback]} - the state variable, the setQueryState callback function
+ */
+function useQueryState(keyParam, typeString, initial) {
+  const key = keyParam; // we should never change the particular key once we start using it
```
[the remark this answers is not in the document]
EjPlatzer: That makes sense, except that a function's argument is necessarily a constant within the scope of the function's execution. In my mind, redeclaring the argument as a constant variable is just obscuring where the argument actually comes from, without gaining much other clarity.

src/hooks/useQueryState.js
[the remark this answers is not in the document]
EjPlatzer: Yeah, it should be as simple as extracting `types` outside the hook declaration and then changing `const types` to `export const types`. It won't be the default export, but that's actually ideal.

src/hooks/useQueryState.js
cpabbot: Yes. get(key) returns the first value, which there should only be one. getAll(key) would return an array which is actually not what we want. This is because we want to encode an array (like filters) like `?arrayKey=value1%2Cvalue2%2Cvalue3` rather than `?arrayKey=value1&arrayKey=value2&arrayKey=value3`. That's actually the whole point of why there is all the array checking and default values etc.

src/hooks/useQueryState.js
```diff
@@ -0,0 +1,163 @@
...
+      if (
+        (urlParams.get(key) ?? defaultVal.toString()) === state.toString() ||
+        (state === true && urlParams.get(key) === '')
+      )
+        return;
+
+      // key with no value OR boolean key set to false -> absent from url, push empty
+      if (!state || state === '' || state.length === 0 || state === false) urlParams.delete(key);
+      // boolean or value key is present in url with its current value
+      else urlParams.set(key, state.toString());
```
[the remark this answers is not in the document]
EjPlatzer: Right, that makes sense. What I meant was that we're using `Array.toString()` to set the value of the array in the query string. There are at least three ways of storing arrays in the query string:

1. `arrayKey=value1&arrayKey=value2&arrayKey=value3...`
2. `arrayKey[]=value1&arrayKey[]=value2&arrayKey=value3...`
3. `arrayKey=value1,value2,value3...`

Different frameworks accept different versions, because the query string is not standardized. There's no single agreed upon way to do it. If this version works, then that's good enough.

src/hooks/useQueryState.js
EjPlatzer: Ah, I see. Both ways are actually equally valid methods for storing an array in the query string, since there is no agreed upon standard. In fact, if it would be simpler to do your latter example, we should probably do that.

src/hooks/useQueryState.js
cpabbot: Yep, totally missed that when I changed to using the types enum (before I just had a defaultVal set) so thank you! And okay, I can change that.

src/views/Events/index.js
cpabbot: Good point, and I just was so tired out from getting setQueryState to work that I didn't properly clean those comments up. I meant this as referring the the events data itself as opposed to the "events" variable which is confusingly set to "allEvents" or "futureEvents"

src/hooks/useQueryState.js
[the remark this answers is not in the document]
cpabbot: That makes sense, and I thought about all of those to some degree actually (except the second one, I've never seen something like that). The first one doesn't look as nice in terms of distinguishing the array of filters from other query parameters in the url (for example). The third one is apparently less stable as commas aren't advised to include in URLs (though I believe they usually work), and the URLSearchParams function automatically converts/cleans these to the hex values.

src/hooks/useQueryState.js
cpabbot: We could probably debate this for a while, but I think that it's cleaner to not save every array value separately, because then nothing distinguishes those values from any other query parameter value. I suppose it's somewhat subjective but to my eyes, the other way looks quite messy in the URL.

EjPlatzer:
The updates to the events page are now out of date. Also, we still do not have a use case for the `useQueryState` hook elsewhere on the site, so I don't currently see the utility in extracting this behavior.

For those reasons, I'm going to close this for now.

PR #1146, opened 2021-05-28
Fix vulnerabilities and update packages

EjPlatzer:
We had a severe vulnerability from `lint-staged`, which we were 3 versions (>2 years) behind on. I have updated most of our packages and fixed all but two outstanding vulnerabilities. Both of the remaining vulnerabilities require packages we depend on to update, and neither represent a serious threat to the site because we do not use the package in such a way as to expose the vulnerability.

Updating packages included Major Version bumps to:
- downshift: introduced a new hooks-based API but didn't break existing `<Downshift>` component. No chnages needed
- http-proxy-middleware: chore updating to make Node 10 or lower obsolete
- husky: updated to v6, switched to using `.husky` repo for pre-commit hooks
- lint-staged: upgraded to v11, removed redundant `git add` step

Packages not updated were:
- @date-io/date-fns: v2 is incompatible with @material-ui/pickers v3. This will be solved once @material-ui v5 is released and the pickers library is obsolete
- chart.js and react-chartjs-2: It is not clear whether react-chartjs-2 is being maintained at this point. chart.js recently released version 3, and a new version of react-chartjs-2 is out that supports it but the details and extent of breaking changes are undocumented and potentially large. It might be worth investigating a different chart.js wrapper.
- history: history v5 is not supported until react-router v6, which just had its first beta release and is not yet stable.


src/services/session.js
```diff
@@ -73,10 +73,12 @@ const getTermCode = () => {
   return `${year.toString().substr(-2)}${term}`;
 };
 
-export default {
+const sessionUtils = {
```
cpabbot, line 76: Reason for calling this sessionUtils rather than sessionService (which is the convention)?

EjPlatzer: Well actually this one I should have called `sessionService`, I'll update that. The other `...Utils` names were because they didn't call out to an API or anything, they were just helper functions. See e.g. `storageUtils` and `httpUtils`.

cpabbot (approved):
LGTM, thanks for updating these!

Issue #1150, opened 2021-05-29
People Search Include Alumni checkbox not centered


EjPlatzer:
Somehow, the include alumni checkbox for people search is no longer centered:
<img width="941" alt="Screen Shot 2021-05-29 at 9 55 37 AM" src="https://user-images.githubusercontent.com/35319956/120072907-08183980-c064-11eb-8705-27dda35814ee.png">

cpabbot:
Note that this checkbox is only seen from fac/staff accounts, so to test this use the 360.facultytest or 360.stafftest accounts. This likely broke due to the same Timesheets scss global styling removal.

PR #1155, opened 2021-06-03
Centering include alumni textbox

silasw2:
Attempt to fix issue #1150. I am not sure if this is the way it is supposed to be done, or if there is another way that display problems are supposed to be handled, but as for me, this aligns the checkbox to the center of the card it is in.

### DIFF
--- a/src/views/PeopleSearch/index.js
+++ b/src/views/PeopleSearch/index.js
@@ -283,7 +283,10 @@ class PeopleSearch extends Component {
   };
   handleChangeIncludeAlumni() {
     this.setState({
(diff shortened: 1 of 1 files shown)


cpabbot (approved):
LGTM, thanks for centering this again! I included another quick fix so be sure to check that before you merge this change.

Noting that you'll want to install the Prettier VSCode extension as well as ESLint (and I believe stylelint) as guided in the ReadMe; this is done for consistent formatting between different developers.

silasw2:
Checkbox toggle works for me! I missed that on my testing when centering the check, thanks for the catch!

PR #1157, opened 2021-06-03
Disabled Legend on Days Remaining for Staff View

### LINKED ISSUE #1149: Legend has reappeared on DaysLeft Doughnut
The graph legend was purposefully disabled on the DaysLeft doughnut, but after the most recent release it has reappeared:
Screen Shot 2021-05-29 at 9 53 26 AM

Davideo37:
I managed to disable the legend on the staff view of the days remaining doughnut.

Added an "options" var to the Daysleft index.js with legend set to false, which disables the legend but keeps the hover text.

Resolved #1149

### DIFF
--- a/src/views/Home/components/DaysLeft/index.js
+++ b/src/views/Home/components/DaysLeft/index.js
@@ -55,6 +55,9 @@ export default class DaysLeft extends Component {
         datasets: [{ data: [pastDays, daysleft], backgroundColor: [gordonColors.primary.blue] }],
         labels: ['Days Finished', 'Days Remaining'],
       };
+      const options = {
+        legend: false,
+      };
       content = (
         <div>
           <Grid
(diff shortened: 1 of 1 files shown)


src/views/Home/components/DaysLeft/index.js
```diff
@@ -55,6 +55,9 @@ export default class DaysLeft extends Component {
         datasets: [{ data: [pastDays, daysleft], backgroundColor: [gordonColors.primary.blue] }],
         labels: ['Days Finished', 'Days Remaining'],
       };
+      const options = {
```
silasw2, line 58: Possibly change options to something more descript? Looks fine other than that though!

silasw2 (approved):
Looks good to me, only suggested a slight style change. The legend no longer appears for me. Thanks for fixing this!

EjPlatzer:
Well the situation is that the component had code for several different versions of `react-chart-js`. The original way of removing the legend was the `defaults.global.legend.display = false;` that got removed because it no longer changes anything.

The `options` is what currently disables the legend. However, this will need to be updated in the future when `react-chart-js` updates to use the recently released `chart-js` v3, which has many breaking changes including in this `options` interface.

PR #1153, opened 2021-06-01
Clarify nvm's Linux instructions, add Mac

russtuck:
Mac is like Linux.  Made LTS clearer and more clearly preferred.

PR #1159, opened 2021-06-03
S21 roddy

roddy3:
Just adjusted  the page title formatting for long tiles. "Please let me know if that looks good"


src/views/About/origins.json
```diff
@@ -0,0 +1,18 @@
...
+    },
+    {
+        "title": "Product Deployment and Support",
+        "body": "Emmanuel Roussel '17 and Stephanie Powers '17\nChris Hansen, Jason Whitehouse '99, Information Systems Group"
+    },
+    {
+        "title": "Benefactors and Contributors",
+        "body": "We’re continually grateful to the many visionary contributors and benefactors who support both the Summer Practicum in Computer Science and the ongoing efforts of GoCo Tech Labs. Thanks to all of our mentors and coaches in development and deployment for your guidance, advice, and encouragement to try new things."
+    }
+]
```
cpabbot, line 18: Can you add a newline at the end of this file? Not a big deal for our project, but a good practice. https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/729795#729795

README.md
```diff
@@ -54,6 +54,7 @@ Make sure Node.js is set up on your machine. If you are on the CS-RDSH-02 virtua
 
 
 
+
```
cpabbot, line 57: This also isn't necessary here. Prettier might ensure consistency here (so be sure to install that extension) but might not with newlines, not sure. But good practice to only change files intentionally!

cpabbot (changes requested):

README.md
```diff
@@ -45,6 +45,16 @@ Make sure Node.js is set up on your machine. If you are on the CS-RDSH-02 virtua
 - Then, tell nvm that you want to use that verion: `nvm use node`.
   To use a specific Node version - usually the latest Long Term Support (LTS) version - replace `node` in the above commands with the desired version or alias, e.g. `nvm install 14.15.4` for version 14.15.4 or `nvm install --lts` for the latest LTS.
 
+#### MacOS:
+
+- Follow the linux installations. 
+- After installing nvm make sure you do quit terminal and then open it back. 
+- To check if the `nvm` is install in your macOS type this command `nvm -v`.
+- Next you can install the latest version of the nvm you can do it by typing this command nvm `nvm install --lts` for the most recent stable version which means you will not worry to update it for a while.
```
EjPlatzer, line 53: I think this addition is redundant after #1153. Since the Linux and Mac instructions are the same (and have been updated to state as much), this isn't needed.

src/components/Header/index.js
```diff
@@ -178,7 +178,7 @@ const GordonHeader = ({ authentication, onDrawerToggle, onSignOut }) => {
             <MenuIcon className="menu-button-icon" />
           </IconButton>
 
-          <Typography className="title disable-select" variant="h6" color="inherit">
+          <Typography className="header disable-select" variant="h6" color="inherit">
```
EjPlatzer, line 181: What was the goal/reason for this change? It doesn't seem related to the other changes. Also, there doesn't appear to be any styles that target the `header` class anywhere in the project. There were styles targeting `title` that would've been applied to this element (see `src/components/Header/header.scss` lines 95-99).

src/views/About/index.js
```diff
@@ -61,7 +62,25 @@ const About = () => {
             </Card>
 
             <Card>
-              <CardHeader className="about-header" title="Origins" />
+              <CardHeader className="about-header" title="origins" />
```
EjPlatzer, line 65: Can we capitalize 'origins' for consistency with the other card headers?
```suggestion
              <CardHeader className="about-header" title="Origins" />
```

EjPlatzer (changes requested):
This looks like a nice update! Thanks for keeping the credits up to date and well organized.

I noticed a couple small things that didn't make much sense to me.

src/components/Header/index.js
roddy3: @EjPlatzer, The reason was to change the the spacing in my profile here. 

Here is the before picture. 



and Here is the after picture.

Screen Shot 2021-06-07 at 11 28 15 AM

That is what I changed in here. I hope that is clear let me know if you have any other question @EjPlatzer

EjPlatzer: Thanks for the clarification! That makes sense that you're trying to fix the spacing on the header. 

I don't think this is the best way to achieve that goal though. The `title` class that was applied to this component was used to give it this styling (found on lines 95-99 of `src/components/Header/header.scss`):
```css
.title {
    flex: 1;
    margin-right: 10px;
    font-weight: bold;
}
```
I think your intent was to remove the `font-weight: bold` property, since a bold font takes up more space than a regular font. However, by removing the entire `title` class, you also removed the other properties - most notably `flex: 1`. That has some unintended effects, especially when you view the page on a mobile or tablet sized screen (less than 960 pixels wide):
Screen Shot 2021-06-07 at 3 13 50 PM

I think the best approach would be to use the original `title` class, and update `header.scss` so that `title` is not bold.

Also, this doesn't seem to be relevant to the other changes you made in this pull request. Maybe you should move them to a separate branch and submit a different pull request focused on this issue?

cpabbot: Moved to branch `pageTitle`

EjPlatzer (approved):
LGTM!

cpabbot (approved):
LGTM, thanks for working on this! Looks a lot cleaner!

PR #1160, opened 2021-06-03
Refactor Admin view to functional components

EjPlatzer:
I have refactored the Admin view and its subcomponents to be functional components using hooks. 

I have also simplified the code wherever possible and renamed things for clarity and consistency. The admin list no longer refreshes the entire page when the state updates. Instead, it will re-render the affected components only. 

Also, I have used the Material-UI `List` component and it's helper components to display lists of information, rather than using `Grid`s to poorly imitate it as we were doing before. This has resulted in minor changes to the UI - which I think are for the best:

Previous UI
Screen Shot 2021-06-03 at 5 27 40 PM

New UI
Screen Shot 2021-06-03 at 5 28 08 PM

(P.S. - when reviewing this pull request, view all but the last commit. The last commit consists only of renaming some folders - but since git doesn't track renames, it thinks that I deleted the previous files and added brand new ones. If you view all but that last commit, you can compare what actually changed between the files that got renamed.)


josh-peters-2357 (approved):
LGTM! By the way, I see the difference made here by using MaterialUI List and not Grid, but I'm not sure if I completely understand why using List components is *better*; could you say a little on that? Anyways, looks great!

EjPlatzer:
> LGTM! By the way, I see the difference made here by using MaterialUI List and not Grid, but I'm not sure if I completely understand why using List components is _better_; could you say a little on that? Anyways, looks great!

Good question! It's not necessarily universally better, but there are reasons to prefer List over Grid in most cases. 

1. Firstly, using Lists is more idiomatic - it's the natural way of representing a list of items. This can also sometimes be very important to screen readers, because they might read the information differently if it's marked up in a list versus in some divs that are styled to imitate a list.
2. Secondly, using lists gets us a lot of styling for free - MUI lists automatically do much of the styling that we were previously implementing manually using nested grids and other markup. For example, the `InvolvementsStatusList` when using Grids had to force spacing by adding extra grid item wrappers - but this didn't add spacing between the header and the first item properly. You can see this in the screenshot above where the image for the first activity overlaps the header slightly. Other things that we get for free with lists include automatic dividers, centering content within the list, responsively laid out action buttons, etc.

That was my reasoning for preferring List to Grid. The other reason I think it's useful is because the code itself now makes clear that we're laying out a list of items, and those items include some text, maybe and avatar, and an action button. The code is more concise and - I think - more readable.

PR #1158, opened 2021-06-03
fixed Wellness Check Color Border #1122

Kevin-Chu58:
I centered the header with a phone number and expanded the border of the wellness question.

PR #1161, opened 2021-06-04
Recreation: Fixed Wellness Check Color Border #1122

### LINKED ISSUE #1122: Wellness Check Color Border
It appears that some of the styling of the Wellness Check page was dependent on the un-scoped CSS rules that @EjPlatzer recently cleaned up. These changes have not been copied over into production yet, so it would be good to catch them before then.

Below, you can see the correct, production, version of the page on the right, and the changes that we want to avoid (currently in train) on the left.

image

Kevin-Chu58:
This is a new PR that recreated this closed PR: https://github.com/gordon-cs/gordon-360-ui/pull/1158

Fixed:

- Wellness check color border expanded
- Center aligned the header with a phone number and "Wellness Check"

Resolves #1122


cpabbot (approved):
LGTM, thanks for fixing this!

PR #1162, opened 2021-06-04
Fixed a bug: People Search Advanced filtering expansion buttons

### LINKED ISSUE #1151: People Search Advanced filtering expansion buttons not changing color
The advanced filtering expansion buttons (i.e. `Academic Info`, `Home Info` and `Office Info`) are intended to change color when they have a valid search value versus when each of their search categories are empty. However, this recently stopped happening. Notice how in the following screenshots, `Academic Info` is the same color when all three search dropdowns are empty as when a `Major` search value has been selected:

Screen Shot 2021-05-29 at 10 00 56 AM

Screen Shot 2021-05-29 at 10 01 28 AM

Whereas the intent is that the buttons are `primary` colored (e.g. the Gordon navy blue) until they have a search value. This is to help the user quickly distinguish which ones have a search value and which ones don't, even when they're closed
:

Screen Shot 2021-05-29 at 10 04 51 AM

Screen Shot 2021-05-29 at 10 06 57 AM

Screen Shot 2021-05-29 at 10 07 37 AM

Screen Shot 2021-05-29 at 10 08 33 AM

iamphg97:
Fixed #1151 .


src/views/PeopleSearch/index.js
```diff
@@ -671,9 +671,15 @@ class PeopleSearch extends Component {
...
-                          this.state.classTypeSearchValue !== ''
+                          this.state.searchValues.major === '' &&
+                          this.state.searchValues.minor === '' &&
+                          this.state.searchValues.classType === ''
+                            ? {
+                                backgroundColor: gordonColors.primary.contrastDefaultColor,
+                              }
+                            : this.state.searchValues.major !== '' ||
+                              this.state.searchValues.minor !== '' ||
+                              this.state.searchValues.classType !== ''
```
cpabbot, line 682: Nice catch changing the variable names! It looks like this might be redundant with double ternary operators? I might suggest using the original logic with the fixed variable names (e.g. `this.state.searchValues.major`). Also it would be good to make this logic clearer by wrapping the OR expressions in parentheses.

src/views/PeopleSearch/index.js
```diff
@@ -691,9 +697,15 @@ class PeopleSearch extends Component {
...
-                          this.state.countrySearchValue !== ''
+                          this.state.searchValues.homeCity === '' &&
+                          this.state.searchValues.state === '' &&
+                          this.state.searchValues.country === ''
+                            ? {
+                                backgroundColor: gordonColors.primary.contrastDefaultColor,
+                              }
+                            : this.state.searchValues.homeCity !== '' ||
+                              this.state.searchValues.state !== '' ||
+                              this.state.searchValues.country !== ''
```
cpabbot, line 708: Same here as above

cpabbot (changes requested):
Thanks for fixing the variable change issue! Left a couple comments that would be good to clean up the code a little, let me know if clarification would be helpful!

cpabbot (approved):
LGTM, thanks for catching this!

PR #1164, opened 2021-06-07
(S21) Fixed two small People Search issues and combined mobile/desktop files

### LINKED ISSUE #1163: Fix #1136 side-effects: e.g. &&&& issue

cpabbot:
Combined peopleSearchResult and mobilePeopleSearchResult into one file to eliminate redundant code. Could definitely use more work but that is beyond of my current scope with this PR.

Quick fix of two issues:
- Changes from #1137 got overwritten by changes in #1136 (merged afterwards)
- Empty URL query params issue was broken by swapping the order of `.filter` and `.map`. Now this has been done correctly.

Resolves #1163


src/views/PeopleSearch/index.js
```diff
@@ -386,8 +386,8 @@ class PeopleSearch extends Component {
 
   async updateURL() {
     const searchParameters = Object.entries(this.state.searchValues)
-      .filter((n) => n) // removes empty strings
       .map(([key, value]) => (value ? `${key}=${value}` : '')) // [ 'firstName=value', 'state=texas']
+      .filter((n) => n) // removes empty strings
```
EjPlatzer, line 390: We can do the filtering first, you just have to access  the value part of the [key, value] tuple:
```suggestion
      .filter(([key, value]) => value)
      .map(([key, value]) => (value ? `${key}=${value}` : ''))
```

That would also allow you to simplify the `map` call, since all the empty searchValues have already been removed:
```suggestion
      .filter(([key, value]) => value)
      .map(([key, value]) => (`${key}=${value}`))
```

EjPlatzer (approved):
LGTM!

PR #1166, opened 2021-06-07
Update README: clarify env vars, remove old issues

EjPlatzer:
The documentation was a little bit confusing and misleading when discussing our `.env` files and how to use them with a local backend. I have corrected, updated, and clarified that wording.

I also removed a couple of "known issues" that no longer apply:
- Edge is now a chromium-based browser and runs 360 just fine.
- The login "hanging" has been fixed
- Filtering on the housing admin view of apartment applications is fully implemented and working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ This project is the frontend of Gordon 360 in React. [The retired frontend](http
 
 - [Getting Started](#getting-started)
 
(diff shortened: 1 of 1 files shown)


cpabbot (approved):
LGTM, thanks for cleaning this up!

PR #1167, opened 2021-06-07
Fixed notification when people change the mobile phone number privacy option

iamphg97:
Realized we have a small bug when people change the mobile phone number privacy option...

When people change their privacy option, the website shows us a wrong notification message.
Screen Shot 2021-06-07 at 6 11 59 PM
The private option --> "Mobile Phone Visible" (Wrong!!)
Screen Shot 2021-06-07 at 6 12 09 PM
The public option --> "Mobile Phone Hidden" (Wrong!!)

I just fixed the bug... Now, it should be all set.
Screen Shot 2021-06-07 at 6 16 37 PM
Screen Shot 2021-06-07 at 6 16 44 PM


EjPlatzer (approved):
Oh, excellent catch, thank you! LGTM!

PR #1169, opened 2021-06-08
Clifton Strengths UI

### LINKED ISSUE #884: Add Top 5 Signature Themes (CliftonStrengths) to My Profile page, and Public Profile
Table has been created for student/staff Signature Themes. Add the Five to My Profile page, and Public Profile page. When ready - ask for hex colors assigned to each Theme to display in one of the four Clifton Colors.

cpabbot:
### Clifton Strengths (UI)
Displays Clifton Strengths for a user if they are set.
Note that these may be viewed on a user's public profile by all users; we decided to do so for the time being until there is an issue or we can create a privacy toggle option under a customization menu (see #884)

Also have plans for different UX ideas to display this that I have created a new issue for (#1188), but I wanted to keep this PR as rudimentary as possible.

Resolves #884

image


src/services/user.js
```diff
@@ -423,15 +423,25 @@ const getProfile = (username) => {
...
+    // should just return null or empty
   }
+};
+
+async function setAdvisors(profile) {
+  profile.Advisors = await getAdvisors(profile.AD_Username);
+}
+
+async function setCliftonStrengths(profile) {
+  profile.CliftonStrengths = await getCliftonStrengths(profile.AD_Username);
```
EjPlatzer, line 444: Would it make more sense to have the `CliftonStrengths` property on the profile actually be the array of strengths, rather than an object with the property `Strengths` that is the array of strengths?
```suggestion
  const strengths = await getCliftonStrengths(profile.AD_Username);
  profile.CliftonStrengths = strengths.Strengths;
```

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -183,6 +184,13 @@ const PersonalInfoList = ({
     />
   ) : null;
 
+  const cliftonStrengths = CliftonStrengths?.Strengths ? (
+    <ProfileInfoListItem
+      title="Clifton Strengths:"
+      contentText={CliftonStrengths.Strengths.join(', ')}
+    />
+  ) : '';
```
EjPlatzer, line 192: I'm not sure that it matters, but usually `null` is a more common way to represent a non-existent element than the empty string.
```suggestion
  ) : null;
```

EjPlatzer (dismissed):
This looks good! I left a couple minor suggestions, but they don't matter much, I would say this code is ready to merge as-is.

src/services/user.js
cpabbot: Yes! Though it feels silly that we currently have a CliftonStrengths object with a single array in it, but it makes sense for if we expand it ever in the future and add info to the view model.

EjPlatzer (approved):
LGTM!

Issue #1168, opened 2021-06-08
Develop New/Improved Styling Strategy


cpabbot:
Note: This issue does not require styling to be actually implemented. This only regards to researching and testing various methods.

Approaches to Explore:
- **CSS Modules (with or without scss)**
    - https://glenmaddern.com/articles/css-modules
    - https://css-tricks.com/css-modules-part-1-need/
    - https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/
    - https://www.javascriptstuff.com/what-are-css-modules/
    - _**Example:**_ #1174 
    - _**Advantages:**_ Easy to migrate by just running a replace script, can still use scss just as before
    - _**Disadvantages:**_ 
- JSS aka CSS in JS
  - https://www.smashingmagazine.com/2020/05/styling-components-react/#styled-components
  - MUI implements JSS, so this is very similar. If we took the JSS approach, we would probably just use MUI. At very least, the advantages/disadvantages and migration difference between them appears to be negligible.
- MUI CSS (uses JSS)
  - https://material-ui.com/styles/basics/#why-use-material-uis-styling-solutionhttps://material-ui.com/guides/interoperability/#react-jss
  - _**Example:**_ #1183
  - _**Advantages:**_ Everything is kept very tightly in one JS file, good for small components
  - _**Disadvantages:**_ Larger components will become much less readable as the CSS is all coupled with the JS

Other References:
- https://github.com/oliviertassinari/a-journey-toward-better-style

cpabbot:
Issue is resolved as we have determined to migrate to CSS Modules (simplest and effective).

PR #1174, opened 2021-06-08
Styling Example - CSS Modules

cpabbot:
### Styling Example - CSS Modules
`styling-example-scss` is base branch that follows our current styling methodology. `styling-example-cssmodules` demonstrates what this would look like to implement CSS Modules, using HealthStatus as the example component.
#1168 

_Note: Drafted for demo purposes only (do not merge)_


src/views/WellnessCheck/components/HealthStatus/index.js
```diff
@@ -60,11 +61,11 @@ const HealthStatus = ({ currentStatus, setCurrentStatus, username, image }) => {
   return (
     <Grid container justify="center" spacing={2}>
       <Grid item xs={12} md={8}>
-        <Card className="wellness-check">
+        <Card className={styles['wellness-check']}>
```
EjPlatzer, line 64: This is not really the most idiomatic way too use css modules. The string index access is a clunky syntax to use when regular object property access works just great. The only thing is that class names that have `-` will need to be converted to use `_`:
```suggestion
        <Card className={styles.wellness_check}>
```
In addition to being more natural/idiomatic javascript, this is also less of a stringly-typed syntax, so mistakes may be easier to catch.

src/views/WellnessCheck/components/HealthStatus/index.js
```diff
@@ -60,11 +61,11 @@ const HealthStatus = ({ currentStatus, setCurrentStatus, username, image }) => {
...
           <CardHeader title={username} />
           <CardContent>
             <img
-              className="rounded-corners user-image"
+              className={styles['rounded-corners user-image']}
```
EjPlatzer, line 68: This syntax does not work for adding multiple classes. You would need to split them up:
```suggestion
              className={`${styles['user-image']} rounded-corners`}
```

EjPlatzer (changes requested):
There's a more idiomatic and syntactically pleasing way to use css modules that I think should be our example.

EjPlatzer:
Gotcha, that makes sense. I guess it's a question of what we want to compare. In my mind, it makes more sense to compare the most idiomatic versions of each solution so we can see which one suits 360 best moving forward. This way I guess you're comparing which is easiest to adopt, but you might be missing which is actually healthiest for our codebase in the long run.

EjPlatzer:
Closing this because the decision has been made and the initial migration to CSS Modules is already complete.

Issue #1131, opened 2021-05-19
Adjust Page Title Formatting for Long Titles


cpabbot:
Not sure if this has a duplicate issue, but we could probably afford to make the allotted page title space wider. Furthermore, when it is beyond the width and has to wrap text, the line height could probably be made narrower so the text does not look like it is trying to escape the header.

![image](https://user-images.githubusercontent.com/17221652/118873376-b16a6d00-b8b7-11eb-903b-1d8074fc0cf7.png)

The 404 Page also has this to the extreme. This could either be changed to something short, or actually maybe it should just be blank. However, it could also be fixed simply by allotting a wider page title space and decreasing the line height for multi-line titles (probably harder than you'd think).

![image](https://user-images.githubusercontent.com/17221652/119161228-12648300-ba27-11eb-99b1-53cccbcaef7b.png)

iamphg97:
Closed by #1177.

PR #1177, opened 2021-06-09
S21 page title

roddy3:
Did some redundant clean up in the `people_search.scss file` as well as changing the flex property in the `header.scss file`. 

------
The Page Title Formatting for Long Titles #1131, the padding in the header file has some issue that when I was trying to divide the flex property that was used in 3 groups see image below

Screen Shot 2021-06-09 at 11 23 17 AM

right now we only the header is only divided into two part:

Screen Shot 2021-06-09 at 11 33 35 AM
I could not exactly make it even because of some more issues, My guess is the overall padding of the header. 

This is another issue that was found while working on the page title it is not a major issue but something to consider.

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -221,14 +221,12 @@ const GordonHeader = ({ authentication, onDrawerToggle, onSignOut }) => {
           </div>
 
           {authentication ? <GordonPeopleSearch authentication={authentication} /> : loginButton}
-
           <GordonNavAvatarRightCorner
             onSignOut={onSignOut}
             authentication={authentication}
             onClick={handleOpenMenu}
             menuOpened={isMenuOpen}
           />
-
           <GordonNavButtonsRightCorner
             open={isMenuOpen}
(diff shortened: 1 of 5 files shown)


src/components/Header/index.js
```diff
@@ -237,7 +235,6 @@ const GordonHeader = ({ authentication, onDrawerToggle, onSignOut }) => {
             anchorEl={anchorElement}
             onClose={handleCloseMenu}
           />
-
```
cpabbot, line 240: Need to run prettier on this file or check out origin/develop version

cpabbot (commented):
LGTM after the one formatting change, but someone else should review since I worked closely with this

src/components/Header/index.js
roddy3: Done

silasw2 (approved):
LGTM! With my testing the header always seems to format correctly, even on longer page titles and different page sizes.
Good catch with removing the redundant code as well!

Issue #1172, opened 2021-06-08
Update Phone Number (Student)


russtuck:
Update on 6/25/2021: Allow only student, not FacStaff, to update their phone number.

Allow everyone to update their phone and address.  (Already approved by the registrar for students; Need HR to approve for employees (Chris Jones);  need alumni affairs approval for alumni (Amy France).)  Ideally, add cell phones for employees and alumni.

iamphg97:
Duplicate issue #1190

PR #1178, opened 2021-06-09
Added Emergency Info ui

### LINKED ISSUE #897: People search a complete replacement for go's people search
Converted from a note Dr. Tuck left. 
We should be pretty close, it seems like the only thing people search needs is the Home Zip and Room Number.

Kevin-Chu58:
This PR allows Gordon police accounts to see emergency contact information on profiles along with more disclaimers

Added:

- Emergency contact Infomation UI (only accessible by police)

 Made progress on #897 (Only the UI part)

### DIFF
--- a/src/components/Profile/components/PersonalInfoList/index.js
+++ b/src/components/Profile/components/PersonalInfoList/index.js
@@ -437,16 +437,9 @@ const PersonalInfoList = ({
     ) : null;
 
     const disclaimer2 =
-    !myProf &&
-    (isHomePhonePrivate ||
-      isAddressPrivate ||
-      isMobilePhonePrivate ||
-      isCampusLocationPrivate ||
-      isSpousePrivate) ? (
-      <Typography align="left" className="disclaimer">
+    <Typography align="left" className="disclaimer">
         Private by default, visible only to gordon police
       </Typography>
-    ) : null;
 
   return (
     <Grid item xs={12}>
--- a/src/components/Profile/components/PersonalInfoList/index.js
+++ b/src/components/Profile/components/PersonalInfoList/index.js
@@ -34,6 +34,16 @@ const PersonalInfoList = ({
     Advisors,
     BuildingDescription,
     Country,
+    EmergencyContact1,
+    EmergencyRelationship1,
+    EmergencyHomePhone1,
+    EmergencyCellPhone1,
+    EmergencyWorkPhone1,
+    EmergencyContact2,
+    EmergencyRelationship2,
+    EmergencyHomePhone2,
+    EmergencyCellPhone2,
+    EmergencyWorkPhone2,
     Hall,
     HomeCity,
     HomePhone,
@@ -59,6 +69,7 @@ const PersonalInfoList = ({
   const isOnline = useNetworkStatus();
   const isStudent = PersonType?.includes('stu');
   const isFacStaff = PersonType?.includes('fac');
+  const isPolice = (user.getLocalInfo().college_role === 'gordon police') ? true : false;
 
   // KeepPrivate has different values for Students and FacStaff.
   // Students: null for public, 'S' for semi-private (visible to other students, some info redacted)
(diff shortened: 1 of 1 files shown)


src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -297,6 +436,18 @@ const PersonalInfoList = ({
...
 
+    const disclaimer2 =
+    !myProf &&
+    (isHomePhonePrivate ||
+      isAddressPrivate ||
+      isMobilePhonePrivate ||
+      isCampusLocationPrivate ||
+      isSpousePrivate) ? (
+      <Typography align="left" className="disclaimer">
+        Private by request, visible only to gordon police
```
iamphg97, line 447: I think all of emergency contact information should be kept private by default, not by request.

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -318,8 +469,19 @@ const PersonalInfoList = ({
             {studentID}
             {home}
             {spouse}
+            {disclaimer1}
```
iamphg97, line 472: In my opinion, the order should be like this...

```suggestion
            {note}
            {disclaimer1}
            {emergencyContact1}
            {emergencyRelationship1}
            {emergencyHomePhone1}
            {emergencyCellPhone1}
            {emergencyWorkPhone1}
            {emergencyContact2}
            {emergencyRelationship2}
            {emergencyHomePhone2}
            {emergencyCellPhone2}
            {emergencyWorkPhone2}
            {disclaimer2}
```

iamphg97 (commented):
Overall, LGTM!

src/components/Profile/components/PersonalInfoList/index.js
Kevin-Chu58: Thank you, that's a good idea!

Davideo37 (commented):
LGTM, Although I agree with what Hyungyu mentioned about privacy and consistency.

cpabbot (changes requested):
This is looking good! That being said, a few things I'm thinking about:
- Disclaimer: Assuming we go with this style, all of the entries would need to be colored red according to the disclaimer. `gordon police` should also be capitalized. Also, since this information must _always_ be private, it should probably be more like `Private: Visible only to Gordon Police`.
- Emergency contact listings can be condensed to `Emergency Contact One: name (relationship) - phoneNumber` or something like that
- Also wondering about security-wise, if this is the best way to handle it or if this means that all public profiles are now returning emergency contact information (just not displaying it). This might be okay since I think it's how we show phone numbers, but not sure
- Wondering if we can/should extract this police information out somehow. This might not be a good idea, but it just seems like a lot of cluttered content that is only visible by GoPo. Not sure...

Also note: in general I would advise trying to have specific branch names (though obviously difficult sometimes when you don't know the scope of your project), since `s21-people-search` could mean a lot of things. Just note for future.

Kevin-Chu58:
> * Emergency contact listings can be condensed to `Emergency Contact One: name (relationship) - phoneNumber` or something like that

For the Emergency contact listings, we plan to update the content after we finished #1172. The current content is the exact same content we found on go.gordon.

EjPlatzer:
In response to Cameron's concern about Emergency Contact Info being visible on PublicStudentProfile objects returned from the database - I think we can handle this the same way we handle other private information. The API will be responsible for determining whether or not the logged in user is authorized to view the information (which will only be true if they are police or are viewing themselves). If they are not authorized, the API will return a Public[Student/FacultyStaff]Profile where the private fields are replaced by 'Kept Private'. That way, none of the private information ever leaves the API.

Kevin-Chu58:
> The API will be responsible for determining whether or not the logged in user is authorized to view the information (which will only be true if they are police or are viewing themselves).
I used `const isPolice = (user.getLocalInfo().college_role === 'gordon police') ? true : false;` and `const emergencyContact1 = isPolice ? ......` to check whether the user is a Gordon Police. I guess this solves the above concern.

> That said, conceptually it might make sense to have this be a separate data point, not a part of Profile info. It is just as easy to create a new ViewModel just for emergency contacts and only fetch that when the logged in user has permission to see it.
For this suggestion, Hyungyu and I were thinking about it yesterday. The ideal design we came up with is making a small button either at the right bottom corner of the Personal Information or below it, when we click a dialog with a table shows up (similar to the dialog when we click the avatar then the Links).

For now, Dr. Tuck suggests just adding those attributes onto the Personal Information to check whether the backend we are working on is correct. After that, we will work on a better solution on visual quality.

Kevin-Chu58:
> The API will be responsible for determining whether or not the logged in user is authorized to view the information (which will only be true if they are police or are viewing themselves).
I think ``

> That said, conceptually it might make sense to have this be a separate data point, not a part of Profile info. It is just as easy to create a new ViewModel just for emergency contacts and only fetch that when the logged in user has permission to see it. That has the advantage of not bloating Profile any more than it already is. I therefore agree with Cameron that emergency contact probably shouldn't be included on Profiles, although not because it would necessarily be a security issue.

PR #1180, opened 2021-06-09
Updating Class Components to Functional Components

silasw2:
All of the pages work for me right now, but I am not sure if the the attended events page works completely as it does not say I have attended any events on 360Train server.


src/components/Error/index.js
```diff
@@ -1,41 +1,44 @@
...
-  render() {
-    const subject = 'Gordon 360 Error';
-    return (
-      <Grid className="gordon-error" container justify="center" alignItems="center">
-        <Grid item xs={11} md={7} lg={5}>
-          <Typography color="primary" variant="h3">
-            Something went wrong.
+const GordonError = ({ error, errorInfo }) => {
+  const [subject, setSubject] = useState(null);
+  setSubject('Gordon 360 Error');
```
cpabbot, line 10:
```suggestion
  const [subject, setSubject] = useState('Gordon 360 Error');
```

src/views/EventsAttended/index.js
```diff
@@ -7,66 +7,38 @@ import { gordonColors } from 'theme';
...
-    };
-  }
-  componentDidMount() {
-    if (this.props.authentication) {
-      this.loadEvents();
-    }
-  }
+const EventsAttended = (props) => {
+  const [events, setEvents] = useState([]);
+  const [loading, setLoading] = useState(null);
```
cpabbot, line 12:
```suggestion
  const [loading, setLoading] = useState(true);
```

cpabbot (changes requested):
Not a thorough review (@EjPlatzer should approve final since he has most refactor experience), but left a couple things to change first.

EjPlatzer (changes requested):
This looks like a great improvement! Thanks for working on it! I did have a couple of small notes, but overall it looks great!

src/components/Error/index.js
```diff
@@ -1,41 +1,43 @@
...
-export default class GordonError extends Component {
-  render() {
-    const subject = 'Gordon 360 Error';
-    return (
-      <Grid className="gordon-error" container justify="center" alignItems="center">
-        <Grid item xs={11} md={7} lg={5}>
-          <Typography color="primary" variant="h3">
-            Something went wrong.
+const GordonError = ({ error, errorInfo }) => {
+  const [subject, setSubject] = useState('Gordon 360 Error');
```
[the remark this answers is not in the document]
silasw2: Thanks for the catch! I was wondering whether I should use state or not, thanks for clearing that up for me!

src/views/EventsAttended/index.js
```diff
@@ -33,47 +33,9 @@ const EventsAttended = (props) => {
...
-          >
-            Need More Chapel Credits?
-          </Button>
-        </Grid>
-      );
-    }
-  } else {
+  if (loading === true) {
+    content = <GordonLoader />;
+  } else if (!props.authentication) {
```
silasw2, line 38: I did not incorporate the exact change that Evan mentioned as I wasn't sure exactly he was referring too, but this should clear up some of the logic for this section. Let me know if it looks good!

EjPlatzer (dismissed):
LGTM!

src/views/Home/components/CLWCreditsDaysLeft/index.js
```diff
@@ -11,194 +11,192 @@ import './CLWChart.css';
...
-      loading: true,
-      currSessionDescription: '',
+const CLWCreditsDaysLeft = () => {
+  const [state, setState] = useState({
+    daysLeft: [],
+    chapelCredits: {},
+    error: null,
+    loading: true,
+    currSessionDescription: '',
+  });
```
EjPlatzer, line 21: This doesn't work because of one of the key differences between `React.Component.setState` and `useState`. 

The former did shallow object merging, so if `state` was an object with properties 
```
{
    daysLeft: [],
    chapelCredits: {},
    error: null,
    loading: true,
    currSessionDescription: '',
}
```
and you called `setState({ loading: false })`, then it would replace `state.loading` with the new value `true`, leaving `state` as 
```
{
    daysLeft: [],
    chapelCredits: {},
    error: null,
    loading: false,
    currSessionDescription: '',
}
```


The new hook `useState` doesn't do any kind of merging - it will replace the previous state with the new state. What that means is that if you had the same starting `state` of 
```
{
    daysLeft: [],
    chapelCredits: {},
    error: null,
    loading: true,
    currSessionDescription: '',
}
``` 
and then called the `setState` from the new hook as `setState({ loading: false })`, then your new state will simply be
```
{
    loading: false
}
```

The fix for this is to separate the different parts of the component's state into separate variables:
```
const [daysLeft, setDaysLeft] = useState([])
const [chapelCredits, setChapelCredits] = useState({})
...
```

src/views/Home/components/CLWCreditsDaysLeft/index.js
```diff
@@ -11,194 +11,192 @@ import './CLWChart.css';
...
+  useEffect(() => {
+    const loadData = async () => {
+      setState({ loading: true });
+      try {
+        const daysLeftPromise = session.getDaysLeft();
+        const chapelCreditsPromise = user.getChapelCredits();
+        const currSessionPromise = session.getCurrent();
+        const daysLeft = await daysLeftPromise;
+        const chapelCredits = await chapelCreditsPromise;
+        const currSession = await currSessionPromise;
```
EjPlatzer, line 32: We can streamline this data fetching code. It was written when `async await` syntax was still relatively new and our developers were unfamiliar with it. The more idiomatic (and clear) syntax would simply be:
```suggestion
const daysLeft = await session.getDaysLeft();
const chapelCredits = await user.getChapelCredits();
const currSession = await session.getCurrent();
```

src/views/Home/components/CLWCreditsDaysLeft/index.js
```diff
@@ -11,194 +11,192 @@ import './CLWChart.css';
...
-      const currSessionDescription = currSession.SessionDescription.replace(
-        /(Academic Year)|(Grad)/gm,
-        '',
-      );
-      this.setState({ loading: false, daysLeft, chapelCredits, currSessionDescription });
-    } catch (error) {
-      this.setState({ error });
-    }
+  if (state.error) {
+    throw state.error;
```
EjPlatzer, line 51: This is an anti-pattern that we should try to remove from the code. The only reason to catch an error is to handle it in a particular way. Instead, what we're doing is catching an error and setting it as part of the state, then checking state and re-throwing the error. That's actually worse than never catching the error at all. IF you don't catch the error, it will continue to propagate up the call stack until it gets caught elsewhere. However, since we're catching it, storing it in state, and then rethrowing it, we have gotten rid of the call stack info showing where exactly the error came from.

src/views/Home/components/Carousel/index.js
```diff
@@ -1,76 +1,142 @@
...
+    },
+    [carouselContent, imageGallery],
+  );
 
-  handleClickSlide() {
-    if (this.state.carouselContent[this._imageGallery.getCurrentIndex()].ActionLink !== '') {
-      window.location = this.state.carouselContent[this._imageGallery.getCurrentIndex()].ActionLink;
-    }
+  if (error) {
+    throw error;
```
EjPlatzer, line 36: As above, if our only way of handling the error is to rethrow it, it's better to not catch it at all.

src/views/Home/components/Carousel/index.js
```diff
@@ -1,76 +1,142 @@
...
+      <ImageGallery
+        ref={(i) => { setImageGallery(i); }}
+        showThumbnails={false}
+        showFullscreenButton={false}
+        showPlayButton={false}
+        showBullets={true}
+        autoPlay={true}
+        showNav={false}
+        slideInterval={5000}
+        items={images}
```
EjPlatzer, line 62: Since `images` only gets used once, you can increase the performance slightly by passing the result of `map` directly rather than storing it in an intermediate variable:
```suggestion
    content = (
      <ImageGallery
        ref={(i) => { setImageGallery(i); }}
        showThumbnails={false}
        showFullscreenButton={false}
        showPlayButton={false}
        showBullets={true}
        autoPlay={true}
        showNav={false}
        slideInterval={5000}
        items={carouselContent.map((slide) =>
             ({
                original: slide.ImagePath,
                originalAlt: slide.AltTag,
                originalTitle: slide.Title,
              })
          );
        }
```

src/views/Home/components/Carousel/index.js
```diff
@@ -1,76 +1,142 @@
...
+        onClick={handleClickSlide}
+      />
+    );
   }
-}
+  return <div>{content}</div>;
```
EjPlatzer, line 67: Wrapping the content with an empty `div` is a bit of an anti-pattern, since it adds an extra div to the html that isn't accomplishing anything. You can instead return content directly:
```suggestion
return content;
```

Or better yet, where you have `content = ...` above, you can simply do `return ...`

src/views/Home/components/Carousel/index.js
```diff
@@ -1,76 +1,142 @@
...
-    }
-  }
+  const handleClickSlide = useCallback(
+    () => {
+      if (carouselContent[imageGallery.getCurrentIndex()].ActionLink !== '') {
+        window.location = carouselContent[imageGallery.getCurrentIndex()].ActionLink;
+      }
+    },
+    [carouselContent, imageGallery],
+  );
```
EjPlatzer, line 33: Was there a reason you made this function with `useCallback`? It seems like it should be fine as a normal function.

EjPlatzer (changes requested):
Left a couple comments about ways to improve the newly-refactored components.

silasw2:
I believe I have fixed all of the issues you mentioned. Thank you for your input, especially on the redundant code. Let me know if it looks good now!

src/views/Home/components/Carousel/index.js
silasw2: I thought I had to in the moment as I was researching the differences between functional and class components, but it works fine as a normal function.

src/views/EventsAttended/index.js
```diff
@@ -7,110 +7,104 @@ import { gordonColors } from 'theme';
...
+const EventsAttended = (props) => {
+  const [events, setEvents] = useState([]);
+  const [loading, setLoading] = useState(true);
 
-  render() {
-    let content;
+  useEffect(() => {
+    const loadEvents = async () => {
+      if (props.authentication) {
+        setLoading(true);
```
cpabbot, line 17: This is unnecessary because we initialized loading to true and `loadEvents` is only called on initial page load
```suggestion
```

src/views/EventsAttended/index.js
```diff
@@ -7,110 +7,104 @@ import { gordonColors } from 'theme';
...
-                    window.location.pathname = '';
-                  }}
-                >
-                  Login
-                </Button>
-              </CardContent>
-            </Card>
-          </Grid>
+            </CardContent>
+          </Card>
```
cpabbot, line 57: You can replace all of this with the `<GordonUnauthorized feature='your attended events'/>`

src/views/Home/components/CLWCreditsDaysLeft/index.js
```diff
@@ -11,194 +11,180 @@ import './CLWChart.css';
...
-    try {
-      const daysLeftPromise = session.getDaysLeft();
-      const chapelCreditsPromise = user.getChapelCredits();
-      const currSessionPromise = session.getCurrent();
-      const daysLeft = await daysLeftPromise;
-      const chapelCredits = await chapelCreditsPromise;
-      const currSession = await currSessionPromise;
+  useEffect(() => {
+    const loadData = async () => {
+      setLoading(true);
```
cpabbot, line 23: Same here, loading is already initialized to true. Maybe it is arguably good practice to keep, but I would remove the redundancy.

cpabbot (changes requested):
Looking good! Just a couple tiny things but this looks like a major improvement, great work!

cpabbot (dismissed):
LGTM, wait for @EjPlatzer's approval before merging

src/components/Error/index.js
```diff
@@ -1,41 +1,41 @@
...
 
 GordonError.propTypes = {
   error: PropTypes.instanceOf(Error).isRequired,
   errorInfo: PropTypes.shape({
     componentStack: PropTypes.string.isRequired,
   }).isRequired,
 };
+
+
+export default GordonError;
```
cpabbot, line 41: Just so we're aware, @josh-peters-2357 pointed out that apparently this file is currently not used. Maybe it should be in the future? Or else it is pointless

src/views/EventsAttended/index.js
```diff
@@ -8,89 +8,81 @@ import { gordonColors } from 'theme';
...
-    let content;
-
-    const style = {
-      button: {
-        background: gordonColors.primary.cyan,
-        color: 'white',
-      },
+  useEffect(() => {
+    const loadEvents = async () => {
+      if (props.authentication) {
```
josh-peters-2357, line 17:
```suggestion
      setLoading(true);
      if (props.authentication) {
```
In other places where we need to load things, we typically set loading to true even if the initial value is true just to make it more clear. It's up to you whether to stick with that though; this should work fine.

silasw2: This would make sense for many of the other files that use useEffect, however, as Cam pointed out to me, events attended will only be ever loaded once, when the page is rendered. So the useState function declared at the beginning is the only time we have to set loading to true since we never update the page based on other information.

src/views/EventsAttended/index.js
```diff
@@ -8,89 +8,81 @@ import { gordonColors } from 'theme';
...
-      events: [],
-      loading: true,
-    };
-  }
-  componentDidMount() {
-    if (this.props.authentication) {
-      this.loadEvents();
-    }
-  }
+const EventsAttended = (props) => {
```
josh-peters-2357, line 11: The only thing you use from props is authentication; perhaps it would be better to only pass in authentication.

EjPlatzer (changes requested):
This is looking great! Just a few small notes but this is nearly there!

src/views/EventsAttended/index.js
josh-peters-2357: Yes, that's correct. It's more for consistency's sake; in the carousel file, for example, you DO set a loading property to true before loading, even though you initialize it to true and it is only loaded once. It is up to you what you decide on, though.

src/views/EventsAttended/index.js
silasw2: Good catch!

src/components/Error/index.js
silasw2: Its possible it could be used in the future, but honestly, if it hasn't already I wouldn't see future pages using it either.

EjPlatzer (approved):
LGTM!

PR #1186, opened 2021-06-12
(S21) Add Image to Student News (UI)

### LINKED ISSUE #731: Add an image field/container to Student News
Add an image field to Student News, so that a small visual "sign/poster/image" could be attached and listed with the item or event being posted in student news. Integrate the photo uploader. 
File size and image should scale to a normal 8.5x11 size poster (used by events), or smaller for things like lost keys or other images.

josh-peters-2357:
Adds the feature to add an image to a student news post. Closes #731 (in combination with the API pull request)

In the new post window, there is now a dropzone to add a picture (.png, .jpg or .jpeg). It can be cropped, removed, and replaced. Upon submit, the image is sent as base64 data inside a JSON object containing the rest of the news post. From there, the API handles the storage and retrieval of the image.

An author can edit their unapproved post image as well. The use cases are extremely similar since the edit and new post windows are identical in structure.

Although I worked on this project alone, much credit goes to @EjPlatzer for answering my many questions; truly, this was a joint effort after the way I incessantly flooded you with Teams messages 😄 .

**Important notes**:

This pull request CANNOT be merged alone; the API pull request with the same name must also be approved for this to work.

Only one image can be submitted to a news post, and this is a hard limit due to the structure of both the UI and API code. My thought process is that there could only be a very few cases where a student would even have the desire to add multiple images, and so this lack of flexibility is okay.

This pull request, and the API pull request, can't be moved to production before we also give Student Life a way to approve a post after seeing its image, so we probably shouldn't merge into develop just yet.

/views/news.js is still built as a class component, and in the future it needs to change.

**Known non-critical issues**:

`maxCropperWidth()` never utilizes its other cases- previous use of this same method in other files has the same problem.

Miscellaneous things are not aligned as well as they could be, and some of the layout is confusing. A good example is that on /news, for a single news entry, there is no clear separation of the description, optional image and edit/delete buttons from the category/subject/author/date section. This needs to change, but styling is not critical.

`componentWillMount` Warning (not sure if this was present before the project or not, or if it belongs to my project):

GET css Forbidden issue (not sure if this was present before the project or not, or if it belongs to my project):

PRESENT BEFORE THE PROJECT- On  /news, a news item with a long, description with strange spacing just runs off the edge- this was here before I came to this project and I wish it could've been gone by the time I leave.

PRESENT BEFORE THE PROJECT- On home page, the same no-wrap problem exists and makes viewing long news items with a picture difficult.

SOLVED (included just so you know I looked into it)-you can click edit on a post, cancel, new post, and the new post will fill in whatever the picture was from the one you clicked edit for.

SOLVED (included just so you know I looked into it)-when you click edit, the picture is view only, no cropper, but the message still says to crop

SOLVED (included just so you know I looked into it)-Editing a previously submitted post and inserting a string into the body long enough that it overflows will somehow remove the picture from the post after submission. Super strangely, if you submit the photo again it stays with the long text, and stays after reload. This behavior is super strange and not my fault :).

HYPOTHETICAL- someone could write a custom file with the proper metadata ("data:image/...") but not containing photo information. Since I don't know how to do this, I didn't test it, and I think Cropper would reject it anyways, or at least behave incorrectly and prevent anyone from posting to the API.

**Use Cases Tested**:

On new post window- add an image, remove an image, add a new image, cancel a post, submit a post, add an invalid photo type, add .jpeg, .png and.jpg, add an image with insufficient size, submit a post without an image, add an image then remove and submit

On news page- expand a news item, delete a news item, open an edit window

On edit window- view the submitted image, remove and add new image, submit an image edit, submit an edit with a changed picture and changed other components, cancel and reopen an edit window, add an invalid photo type, add .jpeg, .png and.jpg, add an image with insufficient size

On home page- display posts with images

**Ideas for the future**:

No change picture button exists in the edit window after the cropper has some picture loaded, because I couldn't figure out how to make it look good.

Move base64 data verifier to a helper service so related files can include it.

Move news loading to the render method- I was SOOOO close to getting it, but I got frustrated at part of it and gave up. It's doable though.

**Images**:
Cropper
Dropzone
Expanded news item
Home news
New edit window
Small pic


josh-peters-2357:
To answer the obvious question of whether or not the new /tmp folder with two images needs to be merged, it does not- my cleanup is not as thorough as I thought, and those changes can be discarded 😄 .

EjPlatzer:
> Move news loading to the render method- I was SOOOO close to getting it, but I got frustrated at part of it and gave up. It's doable though.

I'm very curious what you mean by this. Generally, loading data in the render method is bad practice.

josh-peters-2357:
> 
> 
> > Move news loading to the render method- I was SOOOO close to getting it, but I got frustrated at part of it and gave up. It's doable though.
> 
> I'm very curious what you mean by this. Generally, loading data in the render method is bad practice.

Oh yeah, I should clarify. Basically, because the news items are fetched but not tied to any state property, there's no way to reload them by just changing a state value. I had an idea in mind of loading the news as a state property that stored the JSX making up the news, and making a function to update that state property when a change to news is made. Unfortunately, that was harder to do than it sounded.

josh-peters-2357:
> 
> 
> > > Move news loading to the render method- I was SOOOO close to getting it, but I got frustrated at part of it and gave up. It's doable though.
> > 
> > 
> > I'm very curious what you mean by this. Generally, loading data in the render method is bad practice.
> 
> Oh yeah, I should clarify. Basically, because the news items are fetched but not tied to any state property, there's no way to reload them by just changing a state value. I had an idea in mind of loading the news as a state property that stored the JSX making up the news, and making a function to update that state property when a change to news is made. Unfortunately, that was harder to do than it sounded.

The "issue" with this is that it requires a window reload, which is annoying. And for that small an inconvenience, I moved on.

EjPlatzer:
> Oh yeah, I should clarify. Basically, because the news items are fetched but not tied to any state property, there's no way to reload them by just changing a state value. I had an idea in mind of loading the news as a state property that stored the JSX making up the news, and making a function to update that state property when a change to news is made. Unfortunately, that was harder to do than it sounded.

I'm not sure what you mean by "not tied to any state property" - when you fetch news, you update `this.state.news` and it's subcategories appropriately.
Screen Shot 2021-06-12 at 9 09 08 PM

 So the simplest fix to for getting rid of `window.location.reload` would be to refetch news and update state. You could start by simply calling `this.loadNews()` again, although you could probably do a more focused update of state in many cases.

josh-peters-2357:
> > Oh yeah, I should clarify. Basically, because the news items are fetched but not tied to any state property, there's no way to reload them by just changing a state value. I had an idea in mind of loading the news as a state property that stored the JSX making up the news, and making a function to update that state property when a change to news is made. Unfortunately, that was harder to do than it sounded.
> 
> 
> 
> I'm not sure what you mean by "not tied to any state property" - when you fetch news, you update `this.state.news` and it's subcategories appropriately.
> 
> Screen Shot 2021-06-12 at 9 09 08 PM
> 
> 
> 
>  So the simplest fix to for getting rid of `window.location.reload` would be to refetch news and update state. You could start by simply calling `this.loadNews()` again, although you could probably do a more focused update of state in many cases.
> 
> 
> 
> 

Is that something worth doing in this pull? Or should we have a separate project to fix some of the stylistic news things along with this idea?

src/views/News/index.js
```diff
@@ -146,31 +256,205 @@ export default class StudentNews extends Component {
...
+
+    /*
+    Error checking. Theoretically, this code is designed so that
+    When the get method in the API service is called from the frontend,
+    it will return the image data, even though that's not the value
+    of the image column in news entries. But in the impossible event that
+    it somhow DID return the path of the image instead of the image data,
+    not only would that produce garbage and make the cropper have trouble,
+    but it also is a potential security concern; it sends data back to the
+    client that shouldn't be sent anywhere.
```
josh-peters-2357, line 383: I should add that that would also be a security issue; if the path were returned it would include the SNID of the poster.

src/setupProxy.js
```diff
@@ -12,16 +12,19 @@ module.exports = (app) => {
   // );
 
   // @TRAIN
-  app.use(
-    '/api',
-    createProxyMiddleware({ target: 'https://360ApiTrain.gordon.edu/', changeOrigin: true }),
```
jsenning, line 17: The locally updated setupProxy.js file should probably not be in the pull-request.

josh-peters-2357: > 
> 
> The locally updated setupProxy.js file should probably not be in the pull-request.

Yeah...I had assumed this file would be ignored by Github 🤓

EjPlatzer: Yes, changes to `setupProxy.js`, and to a lesser degree `.env.development` should not be pushed without good reason. `setupProxy` is important because it affects our deployments. `.env.development` only affects developers running it locally, but we should still not change the default to an API that they aren't running, that would cause myriad issues.

src/views/News/index.js
EjPlatzer: `SNID` is an arbitrary number that uniquely identifies student news postings, so there's no issue with exposing it to the front end whatsoever. I think maybe you're confusing it with `ID_NUM`, the user's ID number?

src/views/News/index.js
```diff
@@ -188,86 +472,106 @@ export default class StudentNews extends Component {
...
-    };
+    let newImage;
 
-    // submit the news item and give feedback
-    let result = await newsService.submitStudentNews(newsItem);
-    if (result === undefined) {
-      this.updateSnackbar('News Posting Failed to Submit');
-    } else {
-      this.updateSnackbar('News Posting Submitted Successfully');
+    if (this.state.showCropper != null) {
```
EjPlatzer, line 477: You should use the Strict Inequality operator `!==` as a general principle in JavaScript. The non-strict version `!=` can behave in unexpected ways and cause frustrating bugs.

src/views/News/index.js
```diff
@@ -146,31 +256,205 @@ export default class StudentNews extends Component {
...
+    if (this.cropperRef.current === null && this.state.showCropper === null) {
+      imageData = null;
+    } else if (this.cropperRef.current === null && this.state.showCropper !== null) {
+      imageData = this.state.showCropper.replace(/data:image\/[A-Za-z]{3,4};base64,/, '');
+    } else if (this.cropperRef.current !== null) {
+      imageData = this.cropperRef.current.cropper
+        .getCroppedCanvas({ width: this.CROP_DIM })
+        .toDataURL()
+        .replace(/data:image\/[A-Za-z]{3,4};base64,/, '');
+    }
```
EjPlatzer, line 450: You could simplify the logic here a decent bit:
```suggestion
    let imageData = null;

    if (this.cropperRef.current !== null) {
       imageData = this.cropperRef.current.cropper
        .getCroppedCanvas({ width: this.CROP_DIM })
        .toDataURL()
        .replace(/data:image\/[A-Za-z]{3,4};base64,/, '');
    } else if (this.state.showCropper !== null) {
      imageData = this.state.showCropper.replace(/data:image\/[A-Za-z]{3,4};base64,/, '');
    }
```

EjPlatzer (changes requested):
Thanks for your hard work on this, this is a great start! I left a few comments with small changes that I noticed would be helpful.

That said, I really think this component needs to be refactored into a functional component before it can be merged. Its current state is very bloated and messy, which is mostly not your fault, it's because the code for handling the cropper interacts with the React Class Component in poor ways. Refactoring into a functional component will reduce the bloat quite a bit and help reveal ways the code could be further simplified and improved, especially when it comes to inconsistent or non-idiomatic ways of handling state.

The `tmp/*.png` files should not be included in this merge. I assume they were from testing locally, but we don't want them cluttering the project files. See also my other comments about `setupProxy.js` and `.env.development`

src/views/News/index.js
josh-peters-2357: Oh- yes, I am.

src/views/News/index.js
```diff
@@ -60,17 +45,110 @@ export default class StudentNews extends Component {
...
+      fab: {
+        margin: 0,
+        top: 'auto',
+        right: 40,
+        bottom: 40,
+        left: 'auto',
+        position: 'fixed',
+        zIndex: 1,
+      },
+    };
```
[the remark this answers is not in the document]
josh-peters-2357: I'm not sure why I moved it either- I'll move it.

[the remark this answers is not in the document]
josh-peters-2357: The cancel button is tied to the cropper, so I can't move it out of the class without changing it. Would it be a good idea to include it in the class, but not in the constructor as a `this` property?

src/views/News/index.js
```diff
@@ -146,31 +256,205 @@ export default class StudentNews extends Component {
     }
   }
 
-  // Function called when 'edit' clicked for a news item
+  // This should be the only time we pull from the database
```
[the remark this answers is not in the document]
josh-peters-2357: Ah yes, this comment was present before I started, and I didn't like it either- is it worth removing?

src/views/News/index.js
```diff
@@ -146,31 +256,205 @@ export default class StudentNews extends Component {
...
+    }
+  };
+
+  // TODO: Currently disabled and unused
+  search() {
+    return async (event) => {
+      // await ensures state has been updated before continuing
+      await this.setState({
+        search: event.target.value,
+      });
```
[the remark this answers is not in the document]
josh-peters-2357: I didn't author it. That all makes sense- the service methods called in the `await` statements are async methods, though. Does that make this trickier to clean up?

src/views/News/index.js
```diff
@@ -146,31 +256,205 @@ export default class StudentNews extends Component {
...
+  /*Following methods are solely related to photo submission*
+  /**********************************************************/
+
+  async clearPhotoDialogErrorTimeout() {
+    return new Promise((resolve, reject) => {
+      clearTimeout(this.photoDialogErrorTimeout);
+      this.setState({ photoDialogErrorTimeout: null, photoDialogError: null });
+      resolve(true);
+    });
+  }
```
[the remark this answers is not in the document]
josh-peters-2357: Ah, this was copied from Identification, and VS Code has been noting all this time that reject will never be read. I will change it in the next commit.

src/views/News/index.js
josh-peters-2357: Oh yes, this is much better, thanks!

src/services/news.js
```diff
@@ -5,6 +5,7 @@
  */
 // Written by Jessica Guan
 // Modified by Cameron Abbot
+//Modified by Josh Peters (for CS Practicum 2021)
```
cpabbot, line 8: We can probably remove all of these credits; Jessica was originally noted just because it wasn't tracked in Github for some reason. I think by now Student News is mature enough that we can remove all three of these lines
```suggestion

```

cpabbot (changes requested):
Quick note and also you'll need to re-checkout package-lock.json from develop
`git checkout origin/develop -- package-lock.json`

src/setupProxy.js
```diff
@@ -21,7 +21,10 @@ module.exports = (app) => {
     createProxyMiddleware({ target: 'https://360ApiTrain.gordon.edu/', changeOrigin: true }),
   );
 
-  // @LOCALHOST
-  // app.use('/api', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));
-  // app.use('/token', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));
+  //@LOCALHOST
```
josh-peters-2357, line 24:
```suggestion
  

<!--StartFragment-->  // @LOCALHOST




    
    
      
      // app.use('/api', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));




    
    
      
    <span class="blob-code-inner blob-code-marker" data-code-marker="-">  <span class="pl-c">// app.use('/token', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));</span></span><!--EndFragment-->


```

src/views/News/components/NewsList/index.js
```diff
@@ -1,164 +1,120 @@
...
-          </Grid>
-          <Grid item xs={3}>
-            <Typography variant="body1" style={headerStyle}>
-              POSTED BY
-            </Typography>
-          </Grid>
-          <Grid item xs={2}>
-            <Typography variant="body1" style={headerStyle}>
-              POSTED
+  if (window.innerWidth < BREAKPOINT_WIDTH) {
```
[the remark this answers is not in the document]
EjPlatzer: Also, the way this is currently, this will not automatically resize if the window changes size. It will only check every time it renders, and since you don't have anything watching for a resize event to cause it to rerender, it may not behave as expected.

src/views/News/index.js
```diff
@@ -134,415 +204,640 @@ export default class StudentNews extends Component {
...
-          disabled={submitButtonDisabled}
-        >
-          Submit
-        </Button>
+  });
+
+  /* Gets status of current network connection for online/offline rendering
+   *  Defaults to online in case of PWA not being possible
+   */
+  const networkStatus = JSON.parse(localStorage.getItem('network-status')) || 'online';
```
EjPlatzer, line 493: You don't need any of this. In a functional component, you can do:
```javascript
```

src/views/News/index.js
```diff
@@ -134,415 +204,640 @@ export default class StudentNews extends Component {
...
+    let editButton = (
+      <Button
+        variant="contained"
+        color="primary"
+        onClick={handleUpdate}
+        disabled={submitButtonDisabled}
+      >
+        Update
+      </Button>
+    );
```
EjPlatzer, line 536: We can consolidate these:
```suggestion
    let submitButton = (
      <Button
        variant="contained"
        color="primary"
        onClick={currentlyEditing ? handleUpdate : handleSubmit}
        disabled={submitButtonDisabled}
      >
        {currentlyEditing ? 'Update' : 'Submit'}
      </Button>
    );
```

src/views/News/index.js
```diff
@@ -134,415 +204,640 @@ export default class StudentNews extends Component {
...
+                      onChange={(event) => {
+                        setNewPostSubject(event.target.value);
+                      }}
+                      // helperText="Please enter a subject."
+                    />
+                  </Grid>
+
+                  {/* BODY ENTRY */}
+                  <Grid item xs={12}>
                     <TextField
```
EjPlatzer, line 620: While you're working in this file, we oughta update the TextField's to use the `filled` variant:
```suggestion
                    <TextField
                      variant="filled"
```

EjPlatzer (changes requested):
This is looking great! There's more polishing to be done, but it's getting to a very good state.

src/views/News/index.js
```diff
@@ -18,113 +23,177 @@ import {
...
+   *add a new picture, we want to display the cropper with the new data. When edit is
+   *clicked for the first time and when a new picture has been added, showCropper will
+   *have data, so we can't use showCropper to determine whether or not to show an img
+   *tag or a cropper. A new property was necessary. justShowPicture begins false and is
+   *made true when the edit button is clicked. When either the remove picture or cancel
+   *button is clicked from the edit window, it is set back to false. In this way there is
+   *the ability to update the view to the user. As a side note, this also creates the
+   *ability to change the message above the image in a way that makes sense in either the
+   *just-opened-an-edit or the just-submitted-new-image context.
+   */
```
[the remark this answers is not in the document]
josh-peters-2357: The main reason was that I couldn't really think of a use case where someone would realistically be editing an existing post to crop their picture more, at least not a common one. It's true, I could've just left the cropper there regardless, but the img tag looked a lot cleaner (I originally had the cropper just reappear but it did not look good at all).

src/views/News/index.js
```diff
@@ -18,113 +23,177 @@ import {
...
+  const [loading, setLoading] = useState(true);
+  const [categories, setCategories] = useState([]);
+  const [news, setNews] = useState([]);
+  const [personalUnapprovedNews, setPersonalUnapprovedNews] = useState([]);
+  const [filteredNews, setFilteredNews] = useState([]);
+  const [network, setNetwork] = useState('online');
+  const [newPostCategory, setNewPostCategory] = useState('');
+  const [newPostSubject, setNewPostSubject] = useState('');
+  const [newPostBody, setNewPostBody] = useState('');
+  const [showCropper, setShowCropper] = useState(null); //null if no picture chosen, else contains picture
```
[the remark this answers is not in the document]
josh-peters-2357: I've been thinking during this whole project that `showCropper` is pretty bad. I never changed it because hopefully in the future, some of these image-related functions and JSX bodies will be moved to a separate file; additionally, `showCropper` _does_ determine whether or not the cropper is shown. But what do you think of `cropperImageData`? I use phrases like "image data" often around the project because I feel like it's more accurate.

src/views/News/components/NewsList/index.js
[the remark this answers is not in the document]
josh-peters-2357: > 
> 
> The control flow here makes very little sense. You have if it's a small screen, show the small version, else if there's news to show, show the desktop version.
> 
> I think it should be
> 
> ```js
> if (news || personalUnapprovedNews) {
>     if (window.innerWidth < BREAKPOINT_WIDTH) {
>         // Show the mobile version
>     } else {
>         // Show the desktop version
>     }
> } else {
>     // Show some message about "no news to show"
> }
> ```

I agree- it makes no sense. When I came to the project it was designed this way, but although it behaves correctly, it doesn't read correctly. I will change it.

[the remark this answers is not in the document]
josh-peters-2357: > 
> 
> Also, the way this is currently, this will not automatically resize if the window changes size. It will only check every time it renders, and since you don't have anything watching for a resize event to cause it to rerender, it may not behave as expected.

Correct. The old class component used forceUpdate before, but when I refactored it, I wasn't sure how to convert this functionality, so I dropped that.

src/views/News/index.js
```diff
@@ -19,530 +23,764 @@ import {
...
+  const [photoDialogErrorTimeout, setPhotoDialogErrorTimeout] = useState(null);
+  const [photoDialogError, setPhotoDialogError] = useState(null);
+  const [cropBoxDim, setCropBoxDim] = useState(null);
+  const [aspectRatio, setAspectRatio] = useState(null);
+  const [snackbarOpen, setSnackbarOpen] = useState(false);
+  const [snackbarMessage, setSnackbarMessage] = useState('Something went wrong');
```
EjPlatzer, line 54: The snackbar logic in this can be drastically simplified and improved, using our custom `GordonSnackbar` component and clever use of `setState`. Checkout `src/views/InvolvementProfile/components/Membership/index.js` lines 19, 50-52, and 149-153 for an example.

src/views/News/index.js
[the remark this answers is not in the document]
EjPlatzer: cropperImageData seems like a much better name to me.

src/views/News/index.js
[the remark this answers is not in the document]
EjPlatzer: I think the cleanest way to do this is to show the cropper, and have `Change picture` and `Remove picture` buttons beneath the cropper. I think preventing the user from re-cropping their photo is not worth the tradeoff of showing a slightly prettier UI.

src/views/News/index.js
```diff
@@ -19,530 +23,764 @@ import {
...
+    fab: {
+      margin: 0,
+      top: 'auto',
+      right: 40,
+      bottom: 40,
+      left: 'auto',
+      position: 'fixed',
+      zIndex: 1,
+    },
+  };
```
EjPlatzer, line 114: I think the best way to do this is to have styles defined outside the component, and then here you can set `styles.button.cancelButton.width = showCropper ? '38%' : '86%'`.

src/views/News/index.js
```diff
@@ -19,530 +23,764 @@ import {
...
-    } else {
-      // TODO: test authentication handling and neaten code (ex. below)
-      // alert("Please sign in to access student news");
-    }
-  }
+  function imageOnLoadHelper(reader) {
+    var dataURL = reader.result.toString();
+    var i = new Image();
+    i.onload = async () => {
+      if (i.width < CROP_DIM || i.height < CROP_DIM) {
```
EjPlatzer, line 420: I don't think we actually need to enforce minimum dimensions for News. That code was leftover from uploading ID Photos where they had to be a minimum size for printing ID cards, but since news images can be whatever the user wants (as long as they get approved), it doesn't matter what size/resolution they are.

src/views/News/index.js
```diff
@@ -19,530 +23,764 @@ import {
...
+        disabled={submitButtonDisabled}
+      >
+        {currentlyEditing ? 'Update' : 'Submit'}
+      </Button>
+    );
+
+    let newsJSX;
+
+    // If the user is online
+    if (isOnline || (!isOnline && props.authentication)) {
```
EjPlatzer, line 484: The second case is never reached here because this is already inside an `if (props.authentication)` clause.

src/views/News/index.js
```diff
@@ -19,530 +23,764 @@ import {
...
-        >
-          Update
-        </Button>
-      );
+            {/* NOTE: leaving helper text for now in case
+            that is better than disabling submit button */}
+            {/* Create Posting */}
+            <Dialog open={openPostActivity} fullWidth>
+              <DialogTitle> Post on Student News </DialogTitle>
+              <DialogContent>
```
EjPlatzer, line 523: You could use the `GordonDialogBox` component to save on a lot of boiler plate and repeated code here. See `src/views/InvolvementProfile/index.js` lines 259-402 for an in-depth example.

src/views/News/index.js
```diff
@@ -19,530 +23,764 @@ import {
...
+                <Grid container>
+                  {/* CATEGORY ENTRY */}
+                  <Grid item>
+                    <TextField
```
EjPlatzer, line 527: You should go through all the TextFields and make sure they have the `variant="filled"` prop.

EjPlatzer (changes requested):
This is looking good! I left comments on some more ways this can be polished.

src/views/News/components/NewsItem/index.js
```diff
@@ -1,266 +1,211 @@
...
-        <Link className="news-authorProfileLink" to={`/profile/${posting.ADUN}`}>
-          <Typography variant="h6" className="news-column" style={{ textTransform: 'capitalize' }}>
-            {posting.author}
-          </Typography>
-        </Link>
-      );
-    }
+      </Link>
+    );
+  }
```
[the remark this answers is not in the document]
russtuck: This looks like a good change, but isn't done yet.

src/views/News/index.js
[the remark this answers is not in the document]
russtuck: This looks like it still needs to be done.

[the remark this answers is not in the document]
josh-peters-2357: Done.

src/views/News/components/NewsItem/index.js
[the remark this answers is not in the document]
josh-peters-2357: Done now, I think.

src/views/News/components/NewsItem/index.js
```diff
@@ -100,167 +72,167 @@ export default class NewsItem extends Component {
...
+        startIcon={<DeleteIcon />}
+        onClick={handleDelete}
+        className="btn deleteButton"
+      >
+        Delete
+      </Button>
+    );
+  } else {
+    deleteButton = <div></div>;
+  }
```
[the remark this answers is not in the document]
EjPlatzer: This still needs to be done. Same with the edit button below.

src/views/News/index.js
josh-peters-2357: All right, I think I have refactored to use `GordonDialogBox` correctly.

josh-peters-2357:
News now reloads upon change without refreshing the page!

src/views/News/index.js
```diff
@@ -80,6 +158,37 @@ export default class StudentNews extends Component {
...
   }
 
+  maxCropPreviewWidth() {
+    /*
+    Known non-critical issue:
+    maxCropperWidth() never utilizes its other cases.
+    Previous use of this same method in other files has the same problem.
+
+    It always uses the default case below.
+    */
```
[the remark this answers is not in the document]
josh-peters-2357: This will be fixed in the next commit!

src/views/News/components/NewsList/index.js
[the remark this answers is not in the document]
josh-peters-2357: This will be fixed in the next commit!

src/views/News/index.js
russtuck: Looks done.

russtuck (dismissed):
It appears that all comments have been resolved, and the code looks fine to me.

russtuck:
Of course the merge conflicts must be resolved.

josh-peters-2357:
The merge conflicts have been resolved. Waiting for @cpabbot and @EjPlatzer .

src/views/News/index.js
```diff
@@ -43,465 +70,556 @@ const styles = {
...
-    this.callFunction = this.callFunction.bind(this);
-  }
 
-  componentDidMount() {
-    this.setState({ loading: false });
-    this.loadNews();
-    this.loadUsername();
-    window.addEventListener('resize', this.resize);
+    loadUsername();
+  });
```
EjPlatzer, line 123: This `loadUsername()` effect is being run on every render. Can we add an empty dependency list so it only depends on the first render?

src/views/News/index.js
```diff
@@ -43,465 +70,556 @@ const styles = {
...
+    setOpenPhotoDialog(false);
+    setShowCropper(null);
+    clearPhotoDialogErrorTimeout();
+  }
+  */
+
+  function onCropperZoom(event) {
+    if (event.detail.ratio > 1) {
+      event.preventDefault();
+      React.cropperRef.current.cropper.zoomTo(1);
```
EjPlatzer, line 246: `cropperRef` is a local variable, not a property on React. Shouldn't this be:
```suggestion
     cropperRef.current.cropper.zoomTo(1);
```

src/views/News/index.js
```diff
@@ -43,465 +70,556 @@ const styles = {
...
+    if (cropperImageData !== null) {
+      let croppedImage = cropperRef.current.cropper
+        .getCroppedCanvas({ width: CROP_DIM })
+        .toDataURL();
+      newImage = croppedImage.replace(/data:image\/[A-Za-z]{3,4};base64,/, '');
+    }
+
+    //There is no else here because if there is no image, the JSON object SHOULD
+    //contain null, since the API has been designed to look for a null case and handle
+    //it, treating it appropriately to mean there is no image for the post.
```
EjPlatzer, line 320: I'm not sure this comment is necessary. I find it more confusing than helpful when making sense of this code.

src/views/News/index.js
```diff
@@ -43,465 +70,556 @@ const styles = {
...
-              <Grid item xs={12} md={12} lg={8}>
-                <Grid
-                  container
-                  alignItems="baseline"
-                  justifyContent="center"
-                  style={styles.searchBar}
-                  spacing={5}
-                >
-                  <Grid item xs={10} sm={8} md={8} lg={6}>
+              <DialogContent>
```
EjPlatzer, line 443: The `GordonDialogBox` already has a `DialogContent` wrapper tag, so I don't think this one is accomplishing much anymore.

src/views/News/index.js
```diff
@@ -43,465 +70,556 @@ const styles = {
...
-                        <a href="https://gordonedu.sharepoint.com/:b:/g/StudentLife/admin/EY22_o3g6vFEsfT2nYY-8JwB34OlYmA1oaE1f4FTGD2gew">
-                          More Details
-                        </a>
-                      </Typography>
-                    </Grid>
+                  {/* IMAGE ENTRY */}
+                  <Grid item xs={12}>
+                    <div className="gc360-photo-dialog-box">
+                      <DialogContent className="gc360-photo-dialog-box_content">
+                        <DialogContentText className="gc360-photo-dialog-box_content_text">
```
EjPlatzer, line 505: It's confusing to have DialogContent and DialogContentText wrapping the image uploader when it isn't a dialog (but is inside the DialogContent of the Post News dialog).

src/views/News/index.js
```diff
@@ -43,465 +70,556 @@ const styles = {
...
+                <IconButton
+                  key="close"
+                  aria-label="Close"
+                  color="inherit"
+                  onClick={handleSnackbarClose}
+                >
+                  <CloseIcon />
+                </IconButton>,
+              ]}
+            ></Snackbar> */}
```
EjPlatzer, line 600:
```suggestion
```

EjPlatzer (changes requested):
This is great! I have a couple questions about the code, but I did notice one major issue.

This is probably the result of merging in the restyled `GordonDialogBox`, but when I run this locally I cannot scroll down within the dialog, even though it overflows the page. That makes it impossible (at least on a screen the size of my laptop) to submit my news posting:
Screen Shot 2021-07-13 at 2 57 01 PM

I'm not totally sure what's causing that but it's probably some styling somewhere. You could mess with the css in the browser's DevTools and toggle things until you find the culprit style.

src/views/News/index.js
josh-peters-2357: I tried removing all those tags. The format becomes ugly when they're removed.

josh-peters-2357:
The issue with not being able to scroll had to do with `GordonDialogBox`, whose styling I updated.

src/views/News/index.js
EjPlatzer: That probably has more to do with the classes applied to them, but I'm not sure it's worth the time to sort that out. This doesn't matter all that much anyways.

EjPlatzer (approved):
LGTM!

PR #1187, opened 2021-06-14
(s21) Fixed Involvements Roster Group Hierarchy

roddy3:
Fixed the Hierarchy in the group page by alphabetical order. Looks much cleaner then before 

Before 
Screen Shot 2021-06-14 at 9 57 27 AM
------------------------------------------------------------------------
After 

Screen Shot 2021-06-14 at 9 59 50 AM Screen Shot 2021-06-14 at 10 00 03 AM

### DIFF
--- a/src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
+++ b/src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
@@ -1,5 +1,7 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import IMG from 'react-graceful-image';
+import PropTypes from 'prop-types';
 import {
   Accordion,
   AccordionDetails,
@@ -40,6 +42,7 @@ const PARTICIPATION_LEVELS = {
 };
 
 const MemberListItem = ({
+  props,
   member,
   isAdmin,
   isSuperAdmin,
@@ -59,6 +62,29 @@ const MemberListItem = ({
   );
   const [participation, setParticipation] = useState(member.Participation);
   const [titleComment, setTitleComment] = useState(member.Description);
+  const [avatar, setAvatar] = useState(null);
+
+  useEffect(() => {
+    const loadAvatar = async () => {
+      setAvatar(null);
+      const [{ def: defaultImage, pref: preferredImage }] = await Promise.all([
+        await user.getImage(member.AD_Username),
+      ]);
+      let tempAvatar;
+      console.log(member.AD_Username);
+      if (member.AD_Username) {
+        tempAvatar = preferredImage || defaultImage;
+      } else {
+        tempAvatar = (
+          <svg width="50" height="50" viewBox="0 0 50 50">
+            <rect width="50" height="50" rx="10" ry="10" fill="#CCC" />
+          </svg>
+        );
+      }
+      setAvatar(tempAvatar);
+    };
+    loadAvatar();
+  }, []);
 
   const handleToggleGroupAdmin = async () => {
     if (isAdmin && !isSuperAdmin && member.IDNumber === Number(user.getLocalInfo().id)) {
@@ -236,6 +262,13 @@ const MemberListItem = ({
             <Typography>
               {member.FirstName} {member.LastName}
             </Typography>
+            <IMG
+              className="people-search-avatar-mobile"
+              src={`data:image/jpg;base64,${avatar}`}
+              alt=""
+              noLazyLoad="true"
+              placeholderColor="#eeeeee"
+            />
           </Grid>
           <Grid item xs={2} style={rowStyle}>
             <Typography>{participationDescription}</Typography>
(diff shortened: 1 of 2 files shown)


Davideo37 (approved):
LGTM! Nice work on this one!

src/views/InvolvementProfile/components/Membership/components/MemberList/index.js
```diff
@@ -93,12 +93,38 @@ const MemberList = ({
       style={headerStyle}
     />
   );
+  /// function that re-arrange names alpabetically and also by Advisor, Leader, members and Guest
+  // by Hierarchy.
+  function sortByParticipation(memberArr) {
```
EjPlatzer, line 98: You have kind of imitated sorting behavior by splitting the data into arrays and the rejoining it in a new order. That certainly works, but it's highly inefficient and also more complicated than necessary. We already have a sorting function, `compareByLastThenFirst` (lines 36-50) that you could extend to sort by participation first, and then name. That would be more efficient and more understandable I think.

EjPlatzer (changes requested):
This looks like a good change! However, I think there's a better way to implement it.

src/views/InvolvementProfile/components/Membership/components/MemberList/index.js
cpabbot: That's what I suggested as well, but wasn't sure how to priority order things like "Leader" > "Member". If you know how to do this simply that might be helpful.

EjPlatzer: Here's what I came up with after a couple minutes of thinking about it:
```JavaScript
const compareByParticipationThenLastThenFirst = (a, b) => {
    if (a.ParticipationDescription !== b.ParticipationDescription) {
      if (a.ParticipationDescription === 'Advisor') return 1;
      if (b.ParticipationDescription === 'Advisor') return -1;
      if (a.ParticipationDescription === 'Leader') return 1;
      // Handle other cases
    } else {
      if (a.LastName.toUpperCase() < b.LastName.toUpperCase()) return -1;

      if (a.LastName.toUpperCase() > b.LastName.toUpperCase()) return 1;

      if (a.FirstName.toUpperCase() < b.FirstName.toUpperCase()) return -1;

      if (a.FirstName.toUpperCase() > b.FirstName.toUpperCase()) return 1;
    }

    return 0;
  };
```

We know that if two memberships have the same ParticipationDetail, then we compare them by the other properties. If they don't, then we just need to find which one has the highest priority ParticipationDetail. e.g. if `a` is an `Advisor` and we know `b` is NOT and `Advisor`, than `a` is higher sort order.

EjPlatzer: Alternatively, you could define an Enum that encodes the relative hierarchy:
```javascript
const ParticipatioinLevels = {
  Guest: 0,
  Member: 1,
  Leader:2,
  Advisor: 3
}
```
 and use that for comparison:
```javascript
    if (ParticipationLevels[a.ParticipationDescription] > ParticipationLevels[b.ParticipationDescription]) return 1;
    if (ParticipationLevels[a.ParticipationDescription] < ParticipationLevels[b.ParticipationDescription]) return -1;
```

roddy3: I tried the first one that you suggestion which work fine and it looks a lot cleaner than what i wrote before, and makes more sense and easy to read.

src/views/InvolvementProfile/components/Membership/components/MemberList/index.js
```diff
@@ -33,7 +34,21 @@ const MemberList = ({
...
   const compareByLastThenFirst = (a, b) => {
+    if (a.ParticipationDescription !== b.ParticipationDescription) {
+      if (a.ParticipationDescription == 'Advisor') return -1;
+      if (b.ParticipationDescription == 'Advisor') return 1;
+      if (a.ParticipationDescription == 'Leader') return -1;
+      if (b.ParticipationDescription == 'Leader') return 1;
+      if (a.ParticipationDescription == 'Member') return -1;
+      if (b.ParticipationDescription == 'Member') return 1;
+      if (a.ParticipationDescription == 'Guest') return -1;
+      if (b.ParticipationDescription == 'Guest') return 1;
```
EjPlatzer, line 49: If we made it to this point, we've already checked all the cases where `a !== 'Guest'` and `b !== 'Guest'`, so technically I think this code will never be reached.

roddy3: Did I So technically it is unnecessary to add Guest into the list? because all the case has been handled?

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -59,6 +62,29 @@ const MemberListItem = ({
   );
   const [participation, setParticipation] = useState(member.Participation);
   const [titleComment, setTitleComment] = useState(member.Description);
+  const [avatar, setAvatar] = useState(null);
```
cpabbot, line 65: null is default so you can just say `useState()`

cpabbot (changes requested):
Getting there! Haven't actually run the code yet but suggested some changes

PR #1200, opened 2021-06-18
S21-roster picture and Group contacts/Advisor

roddy3:
Fixed the Hierarchy in the group page by alphabetical order. Looks much cleaner then before 

Before 
Screen Shot 2021-06-14 at 9 57 27 AM
------------------------------------------------------------------------
After 

Screen Shot 2021-06-14 at 9 59 50 AM Screen Shot 2021-06-14 at 10 00 03 AM

I added pictures into the roster within the group page, for the group contact and advisor. 
before adding pictures
-----------------------
Screen Shot 2021-06-30 at 4 44 28 PM
-----------------------------------------------------------------------------------------------
Screen Shot 2021-06-30 at 4 44 39 PM

after adding pictures

Screen Shot 2021-06-30 at 4 50 44 PM

-------------------------------

Screen Shot 2021-06-30 at 4 51 15 PM


roddy3:
Hey  @EjPlatzer , I have been working on the group page, I finished the roster and looks good. I think I am almost done but I am stuck at the point of adding getting the pictures for the group contacts and Advisor. Do you have any advice on how to get the pictures working?

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -58,7 +62,28 @@ const MemberListItem = ({
...
+        MemAvatar = (
+          <svg width="50" height="50" viewBox="0 0 50 50">
+            <rect width="50" height="50" rx="10" ry="10" fill="#CCC" />
+          </svg>
+        );
+      }
+      setAvatar(MemAvatar);
+    };
+    loadAvatar();
+  }, [member.AD_Username]);
```
EjPlatzer, line 86: A couple of notes on this useEffect:

1. There's only one promise here, so you don't need `Promise.all` to await it. `Promise.all` is used to await many different promises at once, only resolving once every promise inside it has resolved.
```javascript
      const { def: defaultImage, pref: preferredImage } = await user.getImage(member.AD_Username);
```
2. Currently, you try to fetch the user's image and then check if the username is valid. You should only fetch the image if the username is valid.
3. The value of `avatar` is inconsistent here. The placeholder avatar you have is a JSX component, but the image you fetch from the API is a base64 string. The way you currently use `avatar`, only the base64 string is valid. 

```suggestion
  useEffect(() => {
    const loadAvatar = async () => {
      if (member.AD_Username) {
        const { def: defaultImage, pref: preferredImage } = await user.getImage(AD_Username);
        setAvatar(preferredImage || defaultImage);
      }
    };
    loadAvatar();
  }, [member.AD_Username]);
```

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -124,6 +150,8 @@ const MemberListItem = ({
     }
   };
 
+  console.log(member.AD_Username);
```
EjPlatzer, line 153: Make sure to remove all your console logs before you merge this.
```suggestion
```

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -277,24 +311,37 @@ const MemberListItem = ({
...
+              <Grid item xs={9} sm={10}>
+                <Grid container spacing={3} wrap="nowrap" alignItems="center">
+                  <Grid item>
+                    <IMG
+                      className="people-search-avatar-mobile"
+                      src={`data:image/jpg;base64,${avatar}`}
+                      alt=""
+                      noLazyLoad="true"
+                      placeholderColor="#eeeeee"
+                    />
```
EjPlatzer, line 324: The `IMG` component from react-graceful-image is useful when you're loading an image from a url or filepath. However, since we're requesting the image data and storing is as a state variable, it doesn't make much sense to use `react-graceful-image` because it isn't doing what it's designed to do. You might as well use the Material-UI `Avatar` component instead (see my changes to ContactListItem for an example).

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -58,7 +62,28 @@ const MemberListItem = ({
     member.ParticipationDescription,
   );
   const [participation, setParticipation] = useState(member.Participation);
-  const [titleComment, setTitleComment] = useState(member.Description);
+  const [title, setTitle] = useState(member.Description);
```
EjPlatzer, line 65: I'm guessing you renamed this for clarity (since it's now being used just for a person's title in a group)?

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -49,6 +52,7 @@ const MemberListItem = ({
   onToggleIsAdmin,
 }) => {
   const [isEditDialogOpen, setIsEditDialogOpen] = useState(false);
+  const [titleDialog, setTitleDialog] = useState(member.Description);
```
EjPlatzer, line 55: You added this new state variable but also kept the original `title` state below. Is there a reason to have both? You seem to mix them in some places, would could cause some bugs, and results in repetitive code.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -277,24 +311,37 @@ const MemberListItem = ({
...
               </Grid>
-              <Grid item xs={4} sm={3} md={2}>
+              <Grid item xs={3} sm={2}>
                 <Typography>{member.ParticipationDescription} </Typography>
-              </Grid>
-              <Grid item xs={2}>
                 <Typography>{mailLoc}</Typography>
               </Grid>
+              {/* <Grid item xs={2}>
+              </Grid> */}
```
EjPlatzer, line 338: You should remove this commented out code if it's not going to be used.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -311,19 +358,28 @@ const MemberListItem = ({
 
     content = (
       <>
-        <Grid container style={rowStyle} direction="row" alignItems="center">
-          <Grid item xs={4} style={rowStyle}>
+        <Grid container alignItems="center" spacing={2} wrap="nowrap">
+          <Grid item md={1} style={rowStyle}>
+            <IMG
```
EjPlatzer, line 363: As mentioned above, using `Avatar` makes more sense than using the `react-graceful-image` component here.

EjPlatzer (changes requested):
This is looking good! I left some comments on ways that `MemberListItem` could be further improved.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -356,4 +406,41 @@ const MemberListItem = ({
   );
 };
 
+MemberListItem.propTypes = {
+  member: PropTypes.shape({
+    MembershipID: PropTypes.number.isRequired,
+    ActivityCode: PropTypes.string.isRequired,
+    //ActivityImage: Not sure what the prop type should be, but it will be required.
```
josh-peters-2357, line 413: Do we use this? If so, we can find out what the required type is. If not, we can remove it.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -356,4 +406,41 @@ const MemberListItem = ({
...
+    //Could be removed and everything would be fine.
+    ActivityDescription: PropTypes.string,
+    ActivityImagePath: PropTypes.string,
+    SessionDescription: PropTypes.string,
+    StartDate: PropTypes.string,
+    EndDate: PropTypes.string,
+    ActivityType: PropTypes.string,
+    ActivityTypeDescription: PropTypes.string,
+    Privacy: PropTypes.string,
+    AccountPrivate: PropTypes.number,
```
josh-peters-2357, line 435: Why are these passed down then?

src/views/InvolvementProfile/components/Membership/components/MemberList/index.js
```diff
@@ -33,7 +33,19 @@ const MemberList = ({
...
+  //"member" > "Guest" then alphabetically.
+
   const compareByLastThenFirst = (a, b) => {
+    if (a.ParticipationDescription !== b.ParticipationDescription) {
+      if (a.ParticipationDescription === 'Advisor') return -1;
+      if (b.ParticipationDescription === 'Advisor') return 1;
+      if (a.ParticipationDescription === 'Leader') return -1;
+      if (b.ParticipationDescription === 'Leader') return 1;
+      if (a.ParticipationDescription === 'Member') return -1;
+      if (b.ParticipationDescription === 'Member') return 1;
```
josh-peters-2357, line 46: I have never seen a function like this, where every line is a return statement- is there a reason you can't use 
```suggestion
      if (
        (a.ParticipationDescription === 'Advisor') ||
        (a.ParticipationDescription === 'Leader') ||
        (a.ParticipationDescription === 'Member'))
          return -1;
      if (
        (b.ParticipationDescription === 'Advisor') ||
        (b.ParticipationDescription === 'Leader') ||
        (b.ParticipationDescription === 'Member'))
          return 1;
    }
```
?

roddy3: This was the better way of improving the hierarchy of the group page by still using the the `compareByLastTheFirst` function. @EjPlatzer inspired me on this one.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
roddy3: just deleted them they were not useful great catch @josh-peters-2357

josh-peters-2357: Woah woah woah no the ones you just commented out are used all over your code. Activity Image was the one that was commented out originally because it looks like it isn't used. But we needed to find out why it wasn't, and why the bottom three of the props list aren't as well.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -421,9 +421,6 @@ MemberListItem.propTypes = {
     ParticipationDescription: PropTypes.string.isRequired,
     GroupAdmin: PropTypes.bool.isRequired,
     Description: PropTypes.string.isRequired,
-
-    //Following in object are never used in this file and never passed down to another component
-    //Could be removed and everything would be fine.
```
josh-peters-2357: We didn't find out why they're not being used, instead of deleting the comment let's figure it out

src/setupProxy.js
```diff
@@ -12,19 +12,19 @@ module.exports = (app) => {
...
+  //   createProxyMiddleware({ target: 'https://360ApiTrain.gordon.edu/', changeOrigin: true }),
+  // );
+
+  // @LOCALHOST
+  app.use('/api', createProxyMiddleware({ target: 'http://localhost:1995/', changeOrigin: true }));
   app.use(
     '/token',
-    createProxyMiddleware({ target: 'https://360ApiTrain.gordon.edu/', changeOrigin: true }),
+    createProxyMiddleware({ target: 'http://localhost:1995/', changeOrigin: true }),
   );
```
josh-peters-2357, line 29: Let's undo the changes made to this file.

josh-peters-2357 (changes requested):

src/setupProxy.js
roddy3: done

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
roddy3: @josh-peters-2357 I was not sure why they were here or being used. But I think there is a reason why they are there.

josh-peters-2357: We'll look at this in the future

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
josh-peters-2357: Better now 😎🎂

josh-peters-2357 (changes requested):
Put the setup proxy back

src/setupProxy.js
```diff
@@ -22,6 +22,9 @@ module.exports = (app) => {
...
 
   // @LOCALHOST
-  // app.use('/api', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));
-  // app.use('/token', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));
+  // app.use('/api', createProxyMiddleware({ target: 'http://localhost:1995/', changeOrigin: true }));
+  // app.use(
+  //   '/token',
+  //   createProxyMiddleware({ target: 'http://localhost:1995/', changeOrigin: true }),
+  // );
```
iamphg97, line 29: Please undo this like this.

// app.use('/api', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));
// app.use('/token', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));

iamphg97 (commented):
LGTM!

src/setupProxy.js
josh-peters-2357: Yes, no changes to setupProxy

roddy3: Done!!! @iamphg97

josh-peters-2357 (approved):

roddy3 (commented):
@EjPlatzer would you mind reviewing the changes i made that you did requested?

src/components/Error/index.js
```diff
@@ -13,7 +13,7 @@ const GordonError = ({ error, errorInfo }) => {
         </Typography>
         <Typography variant="h5" gutterBottom>
           Please email the below error to
-          <a href={"mailto:cts@gordon.edu?subject=Gordon 360 Error"}> cts@gordon.edu</a>.
+          <a href={'mailto:cts@gordon.edu?subject=Gordon 360 Error'}> cts@gordon.edu</a>.
```
EjPlatzer, line 16: When a prop is just a string, you can use this simpler syntax:
```suggestion
          <a href="mailto:cts@gordon.edu?subject=Gordon 360 Error"> cts@gordon.edu</a>.
```

.env.production
```diff
@@ -1 +1 @@
-REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+REACT_APP_API_URL=http://360ApiTrain.gordon.edu/
```
EjPlatzer: This needs to be https. We shouldn't be pushing changes to this file unless there was a specific, intentional reason to do so.
```suggestion
REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
```

In the future, when adding files to your commit, don't add this file or `setupProxy.js`. That should prevent the hassle of having to revert changes.

src/views/InvolvementProfile/involvement-profile.scss
```diff
@@ -7,3 +7,15 @@
...
+  border-radius: 0.5rem;
+  height: 70px;
+  width: 70px;
+}
+
+.contact-list-image {
+  border-radius: 0.5rem;
+  height: 70px;
+  width: 70px;
+}
```
EjPlatzer, line 21: These two classes are never used, so they should be removed.

src/views/InvolvementProfile/components/Membership/components/MemberList/index.js
EjPlatzer: To clarify, we couldn't use the way you recommended @josh-peters-2357 because the order is important. Your way would return `-1` is a is a Member and b is an advisor, which is not at all the desired behavior. The way this is written, it will return the appropriate value based on which one has the "higher" role.

src/views/InvolvementProfile/components/Membership/components/MemberList/index.js
```diff
@@ -33,7 +33,19 @@ const MemberList = ({
     return () => window.removeEventListener('resize', resize);
   });
 
+  // compare names by participation Hierarchy "Advisor" > "Leader" >
+  //"member" > "Guest" then alphabetically.
+
   const compareByLastThenFirst = (a, b) => {
```
EjPlatzer, line 39: We should rename this function, since it no longer compares just by name. Also, there's a better way to do function comments in JavaScript which will give hints and documentation in VSCode.
```suggestion
    /**
   * Compares two members by their role and then their names - last then first
   *
   * Roles follow this hierarchy: Advisor > Leader > Member > Guest
   * @param {Object} a the first member to compare
   * @param {Object} b the second member to compare
   * @returns {-1 | 0 | 1} the sort order of the members: -1 if a before b, 1 if b before a, or 0 if equal
   */
  const compareByRoleThenLastThenFirst = (a, b) => {
```

This gives the following doc hint in VSCode:
Screen Shot 2021-06-30 at 4 37 25 PM

EjPlatzer (changes requested):
This is looking good! A few more things that need to be addressed or could be improved upon, but this is nearly ready to merge.

src/views/InvolvementProfile/involvement-profile.scss
roddy3: Done!!!

src/views/InvolvementProfile/components/Membership/components/MemberList/index.js
roddy3: Done!!! I really liked this way of writing comment, Thanks @EjPlatzer  for this feedback.

src/setupProxy.js
```diff
@@ -22,6 +22,9 @@ module.exports = (app) => {
   );
 
   // @LOCALHOST
-  // app.use('/api', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));
-  // app.use('/token', createProxyMiddleware({ target: 'http://localhost:2477/', changeOrigin: true }));
+  // app.use('/api', createProxyMiddleware({ target: 'http://localhost:1995/', changeOrigin: true }));
+  // app.use(
+  //   '/token',
+  //   createProxyMiddleware({ target: 'http://localhost:1995/', changeOrigin: true }),
```
cpabbot, line 28: Would probably be best to re-checkout this file and not have changes to it

cpabbot (changes requested):

src/setupProxy.js
roddy3: Done !!!

josh-peters-2357 (dismissed):
lgtm

Kevin-Chu58 (commented):
LGTM!

src/components/Error/index.js
```diff
@@ -13,7 +13,7 @@ const GordonError = ({ error, errorInfo }) => {
         </Typography>
         <Typography variant="h5" gutterBottom>
           Please email the below error to
-          <a href={"mailto:cts@gordon.edu?subject=Gordon 360 Error"}> cts@gordon.edu</a>.
+          <a href="mailto:cts@gordon.edu?subject=Gordon 360 Error"> cts@gordon.edu</a>.
```
Davideo37, line 16: I cannot approve this. This is outrageous.

Davideo37 (dismissed):
LGTM! I would recommend having someone else approve as well before merging however.

EjPlatzer (approved):
LGTM!

cpabbot (approved):
LGTM

PR #1201, opened 2021-06-21
Added Emergency Contact Information for GoPo

Kevin-Chu58:
This PR creates a new list of emergency contact information on both private and public profiles but only viewable by Gordon police accounts.

Example as follow: (My Emergency Contact Information)
_This panel is just located below the Personal Information panel if you want to know where this locates_
image
_Not sure if the work phone is really necessary because there are only 4 samples in the database have work phone filled in._

Related to this PR in API: https://github.com/gordon-cs/gordon-360-api/pull/601
Also related to #1178, but since I am not doing modification on personalInfoList, so I just created a new PR with a much better branch name.

_PS: although stafftest is set to a gordon police account, since Cameron and Roddy changed that to a super admin, its school_role is not 'gordon police' anymore, it's now 'god', so stafftest does not work on testing._


iamphg97 (approved):
LGTM! But I suggest receiving one more feedback from someone else who is more expert than me.

src/components/Profile/components/EmergencyInfoList/index.js
```diff
@@ -0,0 +1,100 @@
+import { useState } from 'react';
+import './index.css';
+import ProfileInfoListItem from '../ProfileInfoListItem';
+import user from 'services/user';
+
+import { Typography, Grid, Card, CardHeader, CardContent, List } from '@material-ui/core';
+import { useEffect } from 'react';
```
josh-peters-2357, line 7: This can be merged onto line 1 with: `import { useState, useEffect } from 'react';`

src/components/Profile/components/EmergencyInfoList/index.js
```diff
@@ -0,0 +1,100 @@
...
+
+const formatPhone = (phone) => {
+  let tele = String(phone);
+  if (tele.length === 10) {
+    return '(' + tele.slice(0, 3) + ') ' + tele.slice(3, 6) + '-' + tele.slice(6);
+  }
+  if (tele !== 'undefined') {
```
josh-peters-2357, line 14:
```suggestion
  else if (tele !== 'undefined') {
```

src/components/Profile/components/EmergencyInfoList/index.js
```diff
@@ -0,0 +1,100 @@
...
+      setEmrg(await user.getEmergencyInfo(username));
+    };
+    loadEmrg();
+  }, [username]);
+
+  const disclaimer = (
+    <Typography align="left" className="disclaimer">
+      Private: visible only to Gordon Police
+    </Typography>
+  );
```
josh-peters-2357, line 33: This could be outside the functional component, which I would suggest.

josh-peters-2357 (changes requested):
Looks pretty good! There are a couple small changes I would recommend but other than that it seems fine.

src/components/Profile/components/EmergencyInfoList/index.js
Kevin-Chu58: Yeah, this method is somewhat confusing. Because I copied the whole method from somewhere else in UI, I suppose I should also update the other one.

src/components/Profile/components/EmergencyInfoList/index.js
```diff
@@ -0,0 +1,95 @@
...
+  let tele = String(phone);
+  if (tele.length === 10) {
+    return '(' + tele.slice(0, 3) + ') ' + tele.slice(3, 6) + '-' + tele.slice(6);
+  } else if (tele !== 'undefined') {
+    return tele;
+  }
+};
+
+const EmergencyInfoList = ({ username }) => {
+  const [emrg, setEmrg] = useState([]);
```
EjPlatzer, line 18: I think you could come up with a clearer name for this variable. Maybe `contacts`?
```suggestion
  const [contacts, setContacts] = useState([]);
```

Or, since long variable names are usually better than unclear variable names, you could make it `emergencyContacts`.

src/components/Profile/components/EmergencyInfoList/index.js
```diff
@@ -0,0 +1,94 @@
...
+  useEffect(() => {
+    const loadEmrg = async () => {
+      setEmrg(await user.getEmergencyInfo(username));
+    };
+    loadEmrg();
+  }, [username]);
+
+  return (
+    <Grid item xs={12}>
+      <Card className={`emrg-info-list`}>
```
EjPlatzer, line 29:
```suggestion
      <Card className="emrg-info-list">
```

josh-peters-2357 (approved):
LGTM now!

src/components/Profile/components/EmergencyInfoList/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+import { Typography, Grid, Card, CardHeader, CardContent, List } from '@material-ui/core';
+
+const formatPhone = (phone) => {
+  let tele = String(phone);
+  if (tele.length === 10) {
+    return '(' + tele.slice(0, 3) + ') ' + tele.slice(3, 6) + '-' + tele.slice(6);
+  } else if (tele !== 'undefined') {
+    return tele;
+  }
```
EjPlatzer, line 14: The call to `String(phone)` was to make sure that calls to `tele.length` didn't return an error, I believe. However, new(ish) JavaScript syntax makes that much easier to handle:
```suggestion
  if (phone?.length === 10) {
    return `(${phone?.slice(0, 3)}) ${phone?.slice(3, 6)}-${phone?.slice(6)}`;
  } else {
    return phone
  } 
```

EjPlatzer (commented):
This looks good! I haven't previewed the UI yet so I may have stylistic thoughts, but I left a few code comments for small ways this could improve.

src/components/Profile/components/EmergencyInfoList/index.scss
```diff
@@ -0,0 +1,46 @@
+.emrg-info-list-header {
+  background-color: #014983;
+  color: #ffffff;
```
EjPlatzer, line 3: We generally try to use colors that are defined in `src/vars.scss` so that changing them in that one file will change them across the whole site. See `src/components/Profile/components/PersonalInfoList/index.scss` for an example of using color variables.

EjPlatzer (changes requested):

EjPlatzer (approved):
LGTM!

PR #1208, opened 2021-06-22
S21 Final Update: Advanced People Search Checkboxes

iamphg97:
I added "Only Alumni Option" on Advanced People Search.
This is the UI part for https://github.com/gordon-cs/gordon-360-api/pull/608.

BEFORE:
Screen Shot 2021-06-22 at 3 20 59 PM
There was only one option, "Include Alumni".

Screen Shot 2021-06-22 at 3 21 11 PM
If the user does not select "Include Alumni" option, current students/faculties/staffs show up (exclude alumni).

Screen Shot 2021-06-22 at 3 21 25 PM
If the user select "Include Alumni" option, current students/faculties/staffs and alumni show up.

NOW:
Screen Shot 2021-06-22 at 3 25 20 PM
There are two options, "Include Alumni" and "Only Alumni".

Screen Shot 2021-06-22 at 3 25 30 PM
It shows the same result as the second image on BEFORE.

Screen Shot 2021-06-22 at 3 25 44 PM

Screen Shot 2021-06-22 at 3 25 56 PM

If the user selects one of two options, the other option turns into disabled.

### DIFF
--- a/src/views/PeopleSearch/index.js
+++ b/src/views/PeopleSearch/index.js
@@ -178,6 +178,8 @@ class PeopleSearch extends Component {
         building: '',
       },
 
+      loading: true,
+
       // For April Fools:
       relationshipStatusValue: '',
 
@@ -200,6 +202,7 @@ class PeopleSearch extends Component {
   }
 
   async componentDidMount() {
+    // this.setState({ loading: true });
     if (this.props.authentication) {
       try {
         const profile = await user.getProfileInfo();
@@ -243,6 +246,8 @@ class PeopleSearch extends Component {
         this.loadSearchParamsFromURL();
       }
     }
+
+    this.setState({ loading: false });
   }
 
   async loadSearchParamsFromURL() {
@@ -515,7 +520,7 @@ class PeopleSearch extends Component {
     const networkStatus = JSON.parse(localStorage.getItem('network-status')) || 'online';
 
     if (this.props.authentication) {
-      PeopleSearchCheckbox = (
+      PeopleSearchCheckbox = !this.state.loading ? (
         <Grid item xs={12} align="center">
           <FormLabel component="legend">Type of People:</FormLabel>
           {this.state.personType && !this.state.personType.includes('alum') ? (
@@ -555,6 +560,11 @@ class PeopleSearch extends Component {
               label="Alumni"
             />
           ) : null}
+        </Grid>
+      ) : (
+        <Grid item xs={12} align="center">
+          <FormLabel component="legend">Type of People:</FormLabel>
+          <GordonLoader size={38} />
         </Grid>
       );
 
@@ -1063,7 +1073,7 @@ class PeopleSearch extends Component {
                           );
                         }}
                       >
-                        Clear All
+                        RESET
                       </Button>
                     </Grid>
                     {/* Search Button */}
--- a/src/views/PeopleSearch/index.js
+++ b/src/views/PeopleSearch/index.js
@@ -162,6 +162,7 @@ class PeopleSearch extends Component {
       // These values *must* be in the same order as services/goStalk.js search function
       searchValues: {
         includeAlumni: false,
+        onlyAlumni: false,
         firstName: '',
         lastName: '',
         major: '',
@@ -237,6 +238,7 @@ class PeopleSearch extends Component {
     this.setState({
       searchValues: {
         includeAlumni: urlParams.get('includeAlumni') || false,
+        onlyAlumni: urlParams.get('onlyAlumni') || false,
         firstName: urlParams.get('firstName')?.trim() || '',
         lastName: urlParams.get('lastName')?.trim() || '',
         major: urlParams.get('major')?.trim() || '',
@@ -286,6 +288,14 @@ class PeopleSearch extends Component {
       searchValues: {
         ...this.state.searchValues,
         includeAlumni: !this.state.searchValues.includeAlumni,
+      },
+    });
+  }
+  handleChangeOnlyAlumni() {
+    this.setState({
+      searchValues: {
+        ...this.state.searchValues,
+        onlyAlumni: !this.state.searchValues.onlyAlumni,
       },
     });
   }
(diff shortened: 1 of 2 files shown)


Kevin-Chu58 (commented):
LGTM!

EjPlatzer:
Can we come up with a better UI layout than two mutually exclusive checkboxes? It feels clunky and unintuitive to me, but I'm not sure what the best solution would be.

iamphg97:
Do you have any good ideas? How about this?
For example,
()All () Student ()Staff ()Faculty ()Alumni

EjPlatzer:
> []All []Student []Staff []Faculty []Alumni
That is definitely better (although it would mean a good bit more work in the API I'm guessing). I like that that gives the user a lot of control.

iamphg97:
How about this?? Since these checkboxes are only visible to fac/staff, I am not sure this is a priority thing to work on.
Change to this might lesser works to do and it looks better than before, I guess.

Alumni Option:
[O]Exclude []Include []Only

(Make Exclude as a default option.)

If this seems better to you, I think it is better to use Radio Button instead of a checkbox.

iamphg97:
Screen Shot 2021-06-22 at 5 14 18 PM
#

josh-peters-2357:
I definitely like the idea you mentioned earlier of having a checkbox for each role as opposed to checkboxes like "include" and "exclude"; I think this:
image
looks really intuitive.

iamphg97:
Added Student, Staff/Faculty, and alumni checkboxes.


So now...

Student Account:
Screen Shot 2021-06-24 at 11 07 43 AM
Screen Shot 2021-06-24 at 11 09 45 AM
Screen Shot 2021-06-24 at 11 09 53 AM
Screen Shot 2021-06-24 at 11 10 01 AM


Fac/Staff Account:
Screen Shot 2021-06-24 at 11 21 52 AM
Screen Shot 2021-06-24 at 11 24 18 AM
Screen Shot 2021-06-24 at 11 24 27 AM
Screen Shot 2021-06-24 at 11 24 35 AM

Works Well!

Kevin-Chu58 (commented):
LGTM, nice work!

iamphg97:
> LGTM, nice work!

Thanks for your help!

EjPlatzer:
This is a really nice improvement! I do have a couple of thoughts:

1. We should keep the default behavior of searching students and faculty/staff (e.g. set includeStudents and includeFacStaff to true initially). That will be less confusing for our users who already expect People Search to work a certain way.
2. I think this is a cleaner UI, but there might still be room for improvement. It might not be immediately obvious what the checkboxes do, since they aren't labelled. Just a simple label like "Types of People" would clarify the purpose of the checkboxes a lot. Especially since students haven't had the option to filter results that way before (although you could probably come up with better wording than that).

Kevin-Chu58:
@EjPlatzer
For the first thought, I think there may be some problems if we check the student and faculty/staff checkbox in default. What we did to the display is, check whether the account is a student, faculty/staff, or alumni. But if the student checkbox is checked in default but not being displayed, they may see the current student info which they shouldn't see.

We just did a little test, in which we set the alumni checkbox being checked in default, then log in as a student to see if alumni come up in the advanced people search. The result is yes, although we are not showing the alumni checkbox, we still can find them.

I am doing the disabling the search button in certain conditions, after Hyungyu pushed and merged this PR, I will make some modifications to his code to disable the button if none of them is checked.

iamphg97:
> @EjPlatzer
> For the first thought, I think there may be some problems if we check the student and faculty/staff checkbox in default. What we did to the display is, check whether the account is a student, faculty/staff, or alumni. But if the student checkbox is checked in default but not being displayed, they may see the current student info which they shouldn't see.
> 
> We just did a little test, in which we set the alumni checkbox being checked in default, then log in as a student to see if alumni come up in the advanced people search. The result is yes, although we are not showing the alumni checkbox, we still can find them.
> 
> I am doing the disabling the search button in certain conditions, after Hyungyu pushed and merged this PR, I will make some modifications to his code to disable the button if none of them is checked.

Here's the test result that we did.

Screen Shot 2021-06-24 at 11 56 32 AM

EjPlatzer:
> For the first thought, I think there may be some problems if we check the student and faculty/staff checkbox in default. What we did to the display is, check whether the account is a student, faculty/staff, or alumni. But if the student checkbox is checked in default but not being displayed, they may see the current student info which they shouldn't see.
> 
> We just did a little test, in which we set the alumni checkbox being checked in default, then log in as a student to see if alumni come up in the advanced people search. The result is yes, although we are not showing the alumni checkbox, we still can find them.
> 
> I am doing the disabling the search button in certain conditions, after Hyungyu pushed and merged this PR, I will make some modifications to his code to disable the button if none of them is checked.
The simplest solution to that would be to set them in `componentDidMount` based on PersonType retrieved from the database. That way the page will default to the "normal" settings that user's expect without breaking any of the privacy.

src/views/PeopleSearch/index.js
```diff
@@ -1013,8 +1027,8 @@ class PeopleSearch extends Component {
                           this.setState(
                             {
                               searchValues: {
-                                includeStudent: false,
-                                includeStaff: false,
+                                includeStudent: true,
+                                includeStaff: true,
                                 includeAlumni: false,
```
EjPlatzer, line 1032: Setting these in the `Clear All` button might have weird consequences for Alumni who are searching. Maybe use the personType state to determine what the "reset" values are?
```suggestion
                                includeStudent: this.state.personType.includes('alum') ? false : true,
                                includeStaff: true,
                                includeAlumni: this.state.personType.includes('alum') ? true : false,
```

iamphg97: That makes sense. Thanks!

iamphg97:
All the problems are fixed now! Please re-review it!

Kevin-Chu58 (commented):
LGTM! Should we restore src/views/Help/index.js?

iamphg97:
> LGTM! Should we restore src/views/Help/index.js?

Just restored it :)

Kevin-Chu58 (commented):
LGTM! Nice work!

silasw2 (approved):
LGTM! Did a really nice job with this!

silasw2 (approved):
No more checkbox appearing from thin air! LGTM! :D

PR #1209, opened 2021-06-24
Add Mailbox combinations to MyProfile view

### LINKED ISSUE #1145: Mailboxes & Combinations in 360
Jordan Tymann has requested that students be able to see their mailboxes and combinations somewhere on the MyProfile page. This would prevent a lot of the work that the mailroom has to do in managing the combinations and handing them out to the users who need to use them.

I have already put the mailboxes in our SQL Server [synergy] database at [dbo].[Mailroom_Mailboxes], and they will need to be brought into the [CCT] database.

josh-peters-2357:
Adds Mailbox combinations to MyProfile view. Now, next to your mailbox number, there is a toggle to show or hide the lock combination (by default it is hidden).

Here is a view of what it looks like on mobile when shown (random data so yall don't steal my mail):
Shown on Mobile

Here is it hidden on mobile:
Hidden on Mobile

Here is it on desktop:
Hidden on Desktop

And here is another student so you see that their combination is not visible (though their mailbox number is):
Other student profile

This frontend PR uses a newly created route in the API, so it cannot be merged on its own- the corresponding API PR needs to be merged as well.

Closes #1145 .


src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -90,15 +93,14 @@ const PersonalInfoList = ({
   const isSpousePrivate = isFacStaff && keepPrivate && SpouseName !== PRIVATE_INFO;
 
   useEffect(() => {
-    async function loadMailInfo() {
+    async function loadMailboxCombination() {
       if (myProf && isStudent) {
-        //needs to be a student? ^\._./^
```
roddy3: why did you delete this comment here, I suppose it's only for student because staff/faculty do not have it

roddy3 (commented):
looks good to me!!!

src/components/Profile/components/PersonalInfoList/index.js
josh-peters-2357: I deleted it because I initially wasn't sure if we were going to include anyone beyond students. We are not.

cpabbot (changes requested):
Can we perhaps consider a slightly different UI element? Here's one example I found, but we should think about it some
https://alvaromontoro.com/blog/67975/spoiler-sensitive-sections-in-html-and-css
image

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -55,6 +57,8 @@ const PersonalInfoList = ({
   const [isMobilePhonePrivate, setIsMobilePhonePrivate] = useState(
     Boolean(IsMobilePhonePrivate && MobilePhone !== PRIVATE_INFO),
   );
+  const [mailCombo, setMailCombo] = useState(null);
```
cpabbot, line 60: Unnecessary, sets to null by default
```suggestion
  const [mailCombo, setMailCombo] = useState();
```

.env.development
```diff
@@ -1,2 +1,2 @@
 # refers to the development environment using localhost
-REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+REACT_APP_API_URL=http://localhost:3000/
```
cpabbot, line 2:
```suggestion
REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
```

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -100,9 +104,20 @@ const PersonalInfoList = ({
   // FacStaff spouses are private for private users
   const isSpousePrivate = isFacStaff && keepPrivate && SpouseName !== PRIVATE_INFO;
 
+  useEffect(() => {
+    async function loadMailboxCombination() {
+      if (myProf && isStudent) {
+        const info = await userService.getMailboxCombination();
+        setMailCombo(info.Combination);
+        console.log(info);
```
cpabbot, line 112:
```suggestion
```

src/setupProxy.js
```diff
@@ -12,19 +12,19 @@ module.exports = (app) => {
...
+  //   createProxyMiddleware({ target: 'https://360ApiTrain.gordon.edu/', changeOrigin: true }),
   // );
+
+  // @LOCALHOST
+  app.use('/api', createProxyMiddleware({ target: 'http://localhost:3000/', changeOrigin: true }));
+  app.use(
+    '/token',
+    createProxyMiddleware({ target: 'http://localhost:3000/', changeOrigin: true }),
+  );
 };
```
cpabbot, line 30: This all needs to be removed and put back to normal

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -262,9 +277,42 @@ const PersonalInfoList = ({
...
+                    label={showMailCombo ? 'Hide Lock Combo' : 'Show Lock Combo'}
+                    labelPlacement="bottom"
+                  />
+                </Grid>
+              </>
+            )}
+          </Grid>
+        </ListItem>
+        <Divider />
+      </>
```
cpabbot, line 315: I think we can do this much simpler by just putting the necessary code within the `contentText` attribute; see how I did Clifton Strengths with the tooltip if that is helpful. I can look at this if you have trouble!

cpabbot (changes requested):
Thanks for your work on this! Few simple things that just need to be put back before being ready to go into develop, and also left a comment suggesting a simpler way to do ProfileInfoListItem

EjPlatzer:
I know it's very late to suggest this, but I thought I'd at least float the idea. Since students will presumably only need to view their lock combo very occasionally, there doesn't seem to be a good reason to keep it visible - most people will just keep it hidden and then briefly show it if they ever forget. 

So perhaps a button that would show a dialog with the combo would be more intuitive. Students could click the button when needed to see their combo, but otherwise it would never be shown?

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -285,10 +285,10 @@ const PersonalInfoList = ({
             <Grid container item xs={5} alignItems="center">
               <Typography>{'Mailbox:'}</Typography>
             </Grid>
-            <Grid container item xs={myProf ? 2 : 7} alignItems="center">
+            <Grid container item xs={myProf && mailCombo ? 2 : 7} alignItems="center">
               <Typography className={null}>{`#${Mail_Location}`}</Typography>
             </Grid>
```
EjPlatzer: We want to show their Mail Location, even if they don't have a mailbox combo, right?

josh-peters-2357: Correct. This does that, it just changes the sizing based on whether mailCombo has value or not, because the list component has a certain structure.

josh-peters-2357: No worries! You scared me for a second bc initially I wrote the code that would've done the wrong thing you were thinking of and I thought maybe I accidentally committed it 😅

josh-peters-2357:
> 
> 
> I know it's very late to suggest this, but I thought I'd at least float the idea. Since students will presumably only need to view their lock combo very occasionally, there doesn't seem to be a good reason to keep it visible - most people will just keep it hidden and then briefly show it if they ever forget.
> 
> So perhaps a button that would show a dialog with the combo would be more intuitive. Students could click the button when needed to see their combo, but otherwise it would never be shown?

Well, the toggle setup right now does just that, I think. The combo is not shown when the page loads, and the toggle is "off" by default. If a user wants, they can tap or click it and the combo will appear.

EjPlatzer:
> Well, the toggle setup right now does just that, I think. The combo is not shown when the page loads, and the toggle is "off" by default. If a user wants, they can tap or click it and the combo will appear.

That's true, but it's not the normal use for a toggle. Toggles would normally be for binary states that you want to stay set a certain way, not for a one-time viewing experience. Also, the form factor of a toggle is a little unfortunate - especially on smaller screens - and I think a more concise button would look better.

It's really not a big deal though, so feel free to ignore this.

src/components/Profile/components/PersonalInfoList/index.js
josh-peters-2357: Oh no, we can't. I initially thought that we could use the `PersonalInfoListItem` component; unfortunately, as we discussed before the component is not modular at all. There are a number of changes I needed to make that the component couldn't do.

roddy3 (dismissed):
Looks good to me!!!

josh-peters-2357:
image

cpabbot (approved):
LGTM, looks really nice!

Issue #1212, opened 2021-06-25
Add check box in people search for larger pictures


roddy3:
Having another check box that people can check it if they want to see a larger picture of people.

Davideo37:
Updated view of UI when box is unchecked vs checked.
![ui](https://user-images.githubusercontent.com/70544021/123447679-1fd1d780-d5a8-11eb-95d6-5b3dc3e24b4b.png)
![ui2](https://user-images.githubusercontent.com/70544021/123447688-22343180-d5a8-11eb-9d2c-89000b8d07c7.png)

PR #1213, opened 2021-06-25
People Search Larger Images Toggle

Davideo37:
Added a checkbox for people search on desktop that enables larger avatar previews for the results list. This should close issue #1212.

before 
Screen Shot 2021-06-25 at 4 36 58 PM

After 
image

### DIFF
--- a/src/views/PeopleSearch/components/PeopleSearchResult/index.js
+++ b/src/views/PeopleSearch/components/PeopleSearchResult/index.js
@@ -124,14 +124,7 @@ export default class PeopleSearchResult extends Component {
                   placeholderColor="#eeeeee"
                 />
               </Grid>
-              <Grid
-                item
-                style={{
-                  // a set width is necessary to keep profile images in line
-                  // while maintaining center alignment
-                  width: '260px',
-                }}
-              >
+              <Grid item xs={6}>
                 <Typography variant="h5">{fullName}</Typography>
                 <Typography variant="body2">{nickname}</Typography>
                 <Typography variant="body2">{personClassJobTitle}</Typography>
--- a/src/views/PeopleSearch/components/PeopleSearchResult/index.js
+++ b/src/views/PeopleSearch/components/PeopleSearchResult/index.js
@@ -203,7 +203,7 @@ export default class PeopleSearchResult extends Component {
             >
               <Grid item xs={1}>
                 <IMG
-                  className={'people-search-avatar'}
+                  className="people-search-avatar"
                   src={`data:image/jpg;base64,${this.state.avatar}`}
                   alt=""
                   noLazyLoad="true"
--- a/src/views/PeopleSearch/components/PeopleSearchResult/index.js
+++ b/src/views/PeopleSearch/components/PeopleSearchResult/index.js
@@ -158,7 +158,7 @@ export default class PeopleSearchResult extends Component {
                 padding: '1rem',
               }}
             >
-              <Grid item>
+              <Grid item xs={6} container justify="flex-end">
                 <IMG
                   className="people-search-avatar-large"
                   src={`data:image/jpg;base64,${this.state.avatar}`}
@@ -167,14 +167,7 @@ export default class PeopleSearchResult extends Component {
                   placeholderColor="#eeeeee"
                 />
               </Grid>
-              <Grid
-                item
-                style={{
-                  // a set width is necessary to keep profile images in line
-                  // while maintaining center alignment
-                  width: '260px',
-                }}
-              >
+              <Grid item xs={6}>
                 <Typography variant="h5">{fullName}</Typography>
                 <Typography variant="body2">{nickname}</Typography>
                 <Typography variant="body2">{personClassJobTitle}</Typography>
--- a/src/views/PeopleSearch/components/PeopleSearchResult/index.js
+++ b/src/views/PeopleSearch/components/PeopleSearchResult/index.js
@@ -160,7 +160,7 @@ export default class PeopleSearchResult extends Component {
             >
               <Grid item xs={1}>
                 <IMG
-                  className="people-search-avatar"
+                  className={'people-search-avatar'}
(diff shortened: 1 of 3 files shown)


roddy3:
worked closely with David on this one. someone else needs to review it.

Davideo37:
Updated UI with the Advanced Search "Include:" boxes

Before selection:
ui
Normal search results without Large Images selected:
ui2
Search Results with Large Images selected:
ui3
ui4

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -160,7 +203,7 @@ export default class PeopleSearchResult extends Component {
             >
               <Grid item xs={1}>
                 <IMG
-                  className="people-search-avatar"
+                  className={'people-search-avatar'}
```
silasw2, line 206: Any particular reason you changed this here? From what I know wouldn't the previous way have worked?

silasw2 (approved):
LGTM! Left a small comment and maybe wait til tomorrow to merge so you can get one more reviewer.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
Davideo37: Might've been a 'Prettier' modification. I can undo it

src/views/PeopleSearch/index.js
```diff
@@ -90,55 +90,51 @@ const noResultsCard = (
...
+      <Grid item>
+        <Typography variant="body2" style={styles.headerStyle}>
+          RESULTS
+        </Typography>
+      </Grid>
+    </Grid>
+  </div>
 );
 
 class PeopleSearch extends Component {
```
josh-peters-2357, line 140: I think we should refactor this to a functional component while we're touching these files.

src/views/PeopleSearch/index.js
```diff
@@ -560,11 +577,29 @@ class PeopleSearch extends Component {
...
+          <Media
+            query="(min-width: 960px)"
+            render={() => (
+              <Grid item xs={12} align="center">
+                <FormControlLabel
+                  control={
+                    <Checkbox
+                      checked={this.state.displayLargeImage}
+                      onChange={() => {
+                        this.handleChangeDisplayLargeImages();
```
josh-peters-2357, line 589:
```suggestion
                        this.setState({displayLargeImage: !this.state.displayLargeImage});
                        this.search();
```
This function is only called one time, and it happens to be here. It will only ever be called one time because of what it's doing, so we can put the two lines it contained in the onClick.

josh-peters-2357 (changes requested):
Looks pretty good! I left a couple comments, but I would also suggest that while we're here, we use the network hook we designed instead of what's currently here.

src/views/PeopleSearch/index.js
silasw2: This is a smart line of thinking, however, even Cam was hesitant to begin that task, at least now. For the future I think this is a good goal, but these files are quite complicated when it comes to the react tools they use so converting it to a functional component would be a longer process than we have time for currently.

src/views/PeopleSearch/index.js
Davideo37: I realize the function is only 2 lines, but so are half the other methods for handling changes in the search parameters. I was mirroring their structure which keeps the code in the JSX as clean as possible. I can still change this but I feel like even though it's only 2 lines, it would be better practice to stay consistent.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -143,6 +143,49 @@ export default class PeopleSearchResult extends Component {
...
+                />
+              </Grid>
+              <Grid
+                item
+                style={{
+                  // a set width is necessary to keep profile images in line
+                  // while maintaining center alignment
+                  width: '260px',
+                }}
+              >
```
EjPlatzer, line 177: There's an easier and better way to keep the images centered without setting a static width:
```suggestion
              <Grid item xs={6} container justify="flex-end">
                <IMG
                  className="people-search-avatar-large"
                  src={`data:image/jpg;base64,${this.state.avatar}`}
                  alt=""
                  noLazyLoad="true"
                  placeholderColor="#eeeeee"
                />
              </Grid>
              <Grid item xs={6}>
```

EjPlatzer (changes requested):
Given the time pressure we're under, this seems like a good short-term solution. I think we'll need to re-examine the search options and how to best lay them out on the search form in the long run. 

This is well-coded and works nicely - great job! I do have one suggestion, but otherwise it's coming along nicely.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
Davideo37: I didn't catch that part when I was copying from the mobile view, thanks for finding that!

EjPlatzer (dismissed):
LGTM!

src/views/PeopleSearch/index.js
josh-peters-2357: Well, consistency is good. Discreteness is good too. Although, I like efficiency and I don't like reusability when it won't get reused. But it is a difference of philosophy so I'm just gonna resolve it.

josh-peters-2357 (approved):

src/views/PeopleSearch/index.js
```diff
@@ -118,7 +118,7 @@ const peopleSearchHeaderDesktop = (
         <Typography variant="body2" style={styles.headerStyle}>
           @GORDON.EDU
           <br />
-          MAIL LOCATION
+          MAILBOX #
```
EjPlatzer, line 121: I'm not sure why this change was made, but it will be confusing for Fac/Staff results now because we have mailstops instead of mailboxes. For example, my mailstop in prod is CTS, so `Mailbox #CTS` doesn't make much sense.

PR #1217, opened 2021-06-28
Remove react import errors

EjPlatzer:
Prior to React 17, the JSX syntax (e.g. `<Component prop={prop} >...</Component>`) was supported by a JSX Transform from the React module. This meant that to use JSX in a javascript (.js) file, you had to `import React from 'react'`. This happened everywhere we used JSX (which is every component), even though we weren't explicitly making use of anything from the React module that we were importing. 

As of React 17, the JSX Transform has been completely re-written. In addition to performance gains and bug elimination, the React team also updated the JSX Transform to be usable without the React module.<sup id="#footnote1">[1](#f1)</sup> What this means for us is that it is no longer necessary to `import React from 'react'` when we don't use `React` anywhere in the file. Several new components written since we transitioned to React 17 have made use of this and do not import React. However, since we use eslint configured for a React App, the old eslint rules `react/react-in-jsx-scope` and `react/jsx-uses-react` were enforcing the (now outdated) requirement that we import React whenever we used JSX syntax. This meant that for people who have installed eslint (which ought to be but isn't our whole dev team), we were getting compile-time errors running the UI locally because eslint wanted us to import React.

I have turned off those old eslint rules so that they no longer give the incorrect error messages. This in turn revealed that we import React in MANY places where it is never subsequently used. These were warnings instead of errors, but the ~100 of them were hiding any actually relevant compiler warnings that might have been revealed. Therefore, I also ran the react-codemod script `update-react-imports` (which is written and maintained by the official React team). This script removed all the unnnecessary imports of React. In places that we were using something within the React module (e.g. `React.Fragment` or `React.forwardRef()`), the script updated the imports to import those objects directly. For example, `import React, { Component } from 'react'` becomes `import { Component, Fragment } from 'react'`. This gives us a much clearer idea of which items are actually being used in such components. In the majority of cases, `React` was never used at all, and so it was simply removed from the import list.

I went through every file changed by the codemod and double-checked the changes to ensure it didn't do anything unwanted or nefarious. I also ran the UI locally with all the changes and ensured the site still behaves as expected. 

<b id="#f1">1</b> Under the hood, the Babel compiler is actually importing the new JSX Transform if it encounters JSX syntax anywhere. This is still a nice upgrade though because we don't have to consciously import React in order to use the JSX syntax.  [↩](#footnote1)


bennettforkner (approved):
LGTM

bennettforkner (approved):
LGTM still! Thanks for all this refactoring.

bennettforkner (approved):
Yes again :)

PR #1219, opened 2021-06-28
(S21) Printable People Search Results

cpabbot:
Non-students can now print People Search results. Also adjusted mailbox labelling and fixed input mostly corrected input santitizing.

Notes:
- font key error (check console)
- override ctrl-p browser printing (currently seems to be impossible)

image
image


silasw2 (commented):
LGTM! Some good refactoring done here as well!

EjPlatzer (commented):
This looks really good! My one concern is that currently if you have more than one page of results, the pages can split anywhere in a result, which doesn't look great. I'm not sure how easy/hard that is to fix though, so it might be worth moving forward as is.

roddy3 (approved):
Looks Okay to me!!!

mFelgate:
Ah, that makes sense. And if that's the only use case then What you did def works better then a .csv haha.

src/views/PeopleSearch/index.js
```diff
@@ -461,14 +498,28 @@ class PeopleSearch extends Component {
...
-      this.updateURL();
     }
   };
 
   render() {
     const { classes } = this.props;
     let PeopleSearchCheckbox;
 
+    const printPeopleSearchHeader = (
+      <div class="test" align="center" style={{ display: 'none' }}>
```
josh-peters-2357, line 509: "class" wasn't a valid DOM property, but this code is in develop now.

cpabbot: Good catch, this will be fixed along with Clifton Strengths update. `class` is a valid DOM property it just isn't a valid MUI attribute.

EjPlatzer: This is actually a JSX thing, not Material UI (as can be seen by the fact that `<div>` is not a MUI component). The issue is that `class` is a reserved keyword in JavaScript (for creating object-oriented classes). Since JSX is a syntactic sugar over JavaScript, it couldn't use the `class` keyword, so it uses `className` for declaring the classes of a JSX component.

PR #1221, opened 2021-06-29
Swapped mailbox # heading with mail location - PATCH

cpabbot:
Patch to #1213, swapped heading back to what it was before


EjPlatzer (commented):
LGTM!

EjPlatzer (approved):

PR #1224, opened 2021-06-29
Fix quick search for people without nicknames (mostly staff)

russtuck:
Actually coded by Kevin and Cameron.  I just had a clean branch handy.


Kevin-Chu58 (approved):
LGTM!

PR #1225, opened 2021-06-29
RELEASE: Summer 2021 First Release - PATCH

cpabbot:
Fix quick people search for those without nicknames. Turns out many people have a nickname that is just set to their first name, so it was not an obvious problem.


Kevin-Chu58 (approved):
LGTM!

PR #1226, opened 2021-06-29
Use window width better in search results list

russtuck:
Changed breakpoints to smooth transition from full to partial window width


bennettforkner (approved):
LGTM!

PR #1230, opened 2021-07-01
Disable timesheets while switching HR systems

EjPlatzer:
We are switching HR solutions today, and so employment data will be unpredictably unavailable. It was decided that we should disable timesheets until the switch is complete to prevent any data loss or corruption during the switch.

These changes can be undone once the system is fully switched over tomorrow.

Note that my linter automatically removed the unused imports, but they can be restored quite easily by simply reverting this commit in the future.


josh-peters-2357 (approved):
LGTM

PR #1231, opened 2021-07-01
RELEASE: Disable Timesheets for HR Switch over

EjPlatzer:
See #1230 for more context.

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -1,1 +1,1 @@
-REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+REACT_APP_API_URL=https://360Api.gordon.edu/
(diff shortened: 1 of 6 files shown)


josh-peters-2357 (approved):
LGTM

cpabbot (changes requested):
Needs to be merged into master rather than develop

EjPlatzer:
> The changes look strange, guessing just because it hasn't updated fully. I would probably checkout a new release branch from master because then we avoid having to manually change environment variables, etc. When we did our release earlier it was very clean that way.

I'm not sure what you mean, they look totally normal to me.

EjPlatzer:
Recreating via new release branch from master

PR #1233, opened 2021-07-01
Update to people search results loading speed

josh-peters-2357:
I refactored the way people search results are rendered. They load one at a time now, as soon as the next image has been retrieved. In actuality, it looks pretty cool. However, it doesn't perform "lazy loading" where it only loads for things on the page. It goes through every result and loads them all. 

In some ways, this isn't even bad. This is not worse than what exists currently, and in most cases the results of someone's search are going to be small enough that loading them all is fine (they are about 4 MB total). But I could not figure out how to get lazy loading; it seemed hard, but I found it to be impossible (I think it might actually be possible, but I tried almost every refactoring I could think of and never found an algorithm that did it right). 

The advantage of this PR is that you can now see results loading; what exists currently instantly shows tons of results with blank images and waits for several seconds before displaying them. I have gotten rid of this odd behavior.

Feel free to run it on your own so you can see the difference in loading!

Also, if we do anything more with these files we really need to refactor to functional components.


cpabbot (changes requested):
Hmmm

josh-peters-2357:
Update: search results load very well now. The cards for each result are rendered immediately and the images for the first 50 results are loaded immediately. In the uncommon use case that the user scrolls down to past 50 results, the images load one at a time. 

The reasoning behind this is that we want the user to see things very quickly, and loading the images quickly while the cards are rendering is actually a much faster-looking process than loading one at a time (part of this has to do with the neat fade-in animation for images that is provided by the package we used, `react-graceful-image`). But we also don't want to always send GET requests for every result in a search- this might needlessly send hundreds or even thousands of GET requests to the server for no reason, which is a waste of several resources at once. So this in-between, where we load enough results immediately that most people will find what they were looking for with ease but also become more conservative with our loading after 50, seemed to be a good balance of speed and efficiency.

Added a new package, `react-in-viewport`, which handles all of the technical stuff behind conditionally running code or rendering based on whether a component is in the viewport. This is used to load result images one at a time after 50.

Refactored `PeopleSearchResult` to a functional component and updated the alt tags for the images.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -1,227 +1,239 @@
...
 
-export default class PeopleSearchResult extends Component {
-  constructor(props) {
-    super(props);
+/*Const string was created with https://png-pixel.com/ .
+ *It is a 1 x 1 pixel with the same color as gordonColors.neutral.lightGray (7/9/21)
+ *Although this doesn't use the gordonColors themes directly,
+ *the end result is much cleaner and faster than using the placeholderColor tag of react-graceful-image.
+ */
+const GORDONCOLORS_NEUTRAL_LIGHTGRAY_1X1 =
```
josh-peters-2357, line 16: Giving `` tags an image to serve as a placeholder was far better than using `null` and using the built-in placeholder feature of `react-graceful-image`; in fact, one reason why the search result images didn't load until they were all loaded was because of this use of `null`.

josh-peters-2357: The only thing I don't like about this is that it's not bound to the gordon colors in `theme`. Hypothetically, if `gordonColors.neutral.lightGray` ever changed, we'd manually change this. The only way I could think of to make this bound was to have a function that took in a color and made a 1x1 base64 pic of that color, but I couldn't figure that out.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -1,227 +1,239 @@
...
+  const [personMailLocation, setPersonMailLocation] = useState('');
 
-  componentDidUpdate(newProps) {
-    if (this.props.Person.AD_Username !== newProps.Person.AD_Username) {
-      this.loadAvatar();
+  useEffect(() => {
+    if (lazyImages === false) {
+      loadAvatar();
     }
-  }
```
josh-peters-2357, line 25: It was never possible for this to happen, I think this was copied boilerplate code? I don't have an equivalent in the functional component.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -1,227 +1,239 @@
...
-    } else {
-      avatar = (
-        <svg width="50" height="50" viewBox="0 0 50 50">
-          <rect width="50" height="50" rx="10" ry="10" fill="#CCC" />
-        </svg>
-      );
+  function displayAvatar() {
+    console.log(inViewport);
+    if (inViewport && !hasBeenRun) {
+      console.log('hey');
```
josh-peters-2357, line 51:
```suggestion

```
oops

Kevin-Chu58:
LGTM! We need more people to review it.

josh-peters-2357:
> 
> 
> LGTM! We need more people to review it.

I also need to handle the merge conflicts now, but

src/views/PeopleSearch/index.js
```diff
@@ -90,7 +90,7 @@ const styles = {
 };
 
 const noResultsCard = (
-  <Grid item xs={12}>
+  <Grid item xs={12} direction="row" justify="center" alignItems="center">
```
cpabbot, line 93:
```suggestion
  <Grid item xs={12} direction="row" justifyContent="center" alignItems="center">
```

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -1,110 +1,118 @@
...
+  }
 
-    /*** Single Size - One Column (Mobile View) ***/
-    if (size === 'single') {
-      return (
-        <>
-          <Divider />
+  return (
+    <VisibilitySensor onChange={handleVisibilityChange}>
+      <Fragment>
```
cpabbot, line 113: Probably just use the equivilant <></> empty tag shorthand since this is most conventionally used in our codebase

cpabbot (commented):
This is really neat! I didn't go through thoroughly and a lot of it isn't obvious to me so I'd have to take a closer look, but for now I just added a couple small comments. The second one is because justify is deprecated so we're trying to use only justifyContent MUI attribute.

josh-peters-2357:
Okay, all the conflicts related to needing to merge are finished.

src/views/PeopleSearch/index.js
```diff
@@ -409,13 +410,16 @@ class PeopleSearch extends Component {
     return result;
   };
 
+  // async loadImages()
+
```
cpabbot, line 414:
```suggestion
```

src/views/PeopleSearch/index.js
```diff
@@ -409,13 +410,16 @@ class PeopleSearch extends Component {
...
   async search() {
     if (!this.canSearch()) {
       // do not search, only search if there are some non-blank non-false values
     } else {
       this.setState({
         header: <GordonLoader />,
-        peopleSearchResults: null,
+        resultData: [],
```
cpabbot, line 421: does this have to be renamed? `peopleSearchResults` seems like maybe a more descriptive name but I could be wrong

src/views/PeopleSearch/index.js
```diff
@@ -426,39 +430,42 @@ class PeopleSearch extends Component {
...
             <Media query="(min-width: 960px)">
               {(matches) =>
-                peopleSearchResults.map((person) => (
+                matches && !this.state.displayLargeImage
+                  ? peopleSearchHeaderDesktop
+                  : peopleSearchHeaderMobile
+              }
+            </Media>
+          ),
+        });
```
cpabbot, line 450: Just curious, why did we have to set the header state here / why didn't we have to before?

cpabbot (commented):
Great work! Just wanted to ask you about a couple things, we can discuss quickly in person probably and then merge this in!

src/views/PeopleSearch/index.js
josh-peters-2357: Similar to the way we don't call everything related to CCT "CCT-[variableName]", we don't need to call everything related to people search "peopleSearch-[variableName]"; like obviously the results we're talking about are people search results, we're in the index file of people search.

Or maybe a better way to think of it is not to gauge it by whether it adds information to the variable purpose, but instead to gauge it by whether the information it adds adds clarity that the reader couldn't have very easily deduced already.

src/views/PeopleSearch/index.js
josh-peters-2357: Good question. The original writer of the code didn't seem to understand that there was a large difference between having one `setState` with two properties set in it and having two `setState`'s with one property set each. They are split up now, because the output before really wasn't desirable. It has something to do with using the `media` tag, I think, because it doesn't really behave the same way as a lot of other tags. 

The reason header is now _first_ is because, well, setting the header state is a lot faster and smaller, and it is higher up on the screen.

src/views/PeopleSearch/index.js
cpabbot: I suppose that's fair. The flaw comparing with "CCT-[variableName]" is just that while CCT isn't adding any information to the variable, peopleSearch is, and I would argue that we get many results in many files, so `resultData` is pretty ambiguous. That being said, I understand your point and it's not a huge deal so I won't request the change.

src/views/PeopleSearch/index.js
cpabbot: As far as I know, there isn't a large difference between one setState or two - they both queue up changes to take effect for the next render, but perhaps there is a small performance difference that ends up making a large difference somehow?

My main question was actually as to why the header is a state variable at all?

EjPlatzer: I think I can answer the second question. The header is what displays the loading state while loading search results (which may not be the ideal way to do it, but is the current situation nonetheless). Previous developers made a habit of using state to store JSX, rather than keeping simple variables (like a `loading` boolean) in state and handling the JSX based on state in the render method. There are certainly tradeoffs that could be discussed either way, but the general convention among react developers (and within our codebase) is keep state data-focused and handle the logic for rendering that data as JSX within the render method.

cpabbot: Ah gotcha, this makes sense, thank you for the clarification! That is rather unfortunate but understandable haha

cpabbot (approved):
LGTM, ready for testing in develop

PR #1234, opened 2021-07-01
S21 Disable health pop-up but keep health tab 

roddy3:
Disable the health check popup from the home page, but leave the health tab unchanged. This way, nobody will be required to make a report in order to see the home page, but they can still report if needed.

Screen Shot 2021-07-01 at 11 16 18 AM


silasw2 (commented):
Worked closely with this, someone else will have to approve. I think that it works fine, albeit a bit messy. But it is able to be undone if we need to in the future.

Davideo37 (changes requested):
Wellness stuff LGTM, I would maybe add a comment that states why the code is left commented out, and what it does/how to undo it. Besides that you may want to pull develop into this branch before merge as there seems to be some outdated code in the Involvement Profile stuff.

Davideo37 (approved):
LGTM!

Kevin-Chu58 (commented):
LGTM!

cpabbot (changes requested):
Waiting for other branch to merge first, Roddy branched out of his other branch rather than develop so this includes other changes.

mFelgate:
I don’t recommend it at this point unless you are pretty comfortable with git, but 37 commits is a lot for one pull request and it would make following the changes rly hard. I would do a git reset head to remove all the commits. This will leave all the changes you made. Then I would go back and cherry pick changes and make 5 or 6 commits with descriptions for this PR.

josh-peters-2357 (dismissed):
lgtm

cpabbot (changes requested):
Not to prolong this further, but I am concerned with how easy this will be to track down in the future if we want to revert (though we could probably just revert this commit).
Could we add a comment before each commented out section like this:
`// @WELLNESS-CHECK disabled (following lines of commented code)`
and maybe for large code blocks commented out add a `// END` line below so it's extra clear

roddy3:
> Not to prolong this further, but I am concerned with how easy this will be to track down in the future if we want to revert (though we could probably just revert this commit).
> Could we add a comment before each commented out section like this:
> `// @WELLNESS-CHECK disabled (following lines of commented code)`
> and maybe for large code blocks commented out add a `// END` line below so it's extra clear
-----------------

They were already instruction on how to revert it but I added more comment to it.

src/views/Home/index.js
```diff
@@ -1,83 +1,95 @@
...
+  //         (event.data === 'online' || event.data === 'offline')
+  //       ) {
+  //         return event.data;
+  //       }
+  //       return prevStatus;
+  //     });
+  //   });
+  //
+  //   return () => window.removeEventListener('message', () => {});
+  // }, []);
```
cpabbot, line 56:
```suggestion
  // }, []);
  // END Wellness Check disabled code
```

src/views/Home/index.js
```diff
@@ -1,83 +1,95 @@
...
       setIsAuthenticated(false);
       setLoading(false);
     }
   }, [authentication]);
-
+  // @WELLNESS-CHECK disabled to revert this you must uncomment this lines of code
   const loadPage = async () => {
     setLoading(true);
-    const [{ PersonType }, { IsValid }] = await Promise.all([
+    const [{ PersonType } /*, { IsValid }*/] = await Promise.all([
```
cpabbot, line 75:
```suggestion
    // @WELLNESS-CHECK disabled to revert this you must uncomment this lines of code
    const [{ PersonType } /*, { IsValid }*/] = await Promise.all([
```

src/views/Home/index.js
```diff
@@ -1,83 +1,95 @@
...
       // (i.e. user logs out) so that it isn't preserved falsely for the next user
-      setHasAnswered(null);
+      // setHasAnswered(null);
       setPersonType(null);
       setIsAuthenticated(false);
       setLoading(false);
     }
   }, [authentication]);
-
+  // @WELLNESS-CHECK disabled to revert this you must uncomment this lines of code
```
cpabbot, line 72:
```suggestion
```

cpabbot (commented):
Small adjustment

cpabbot (approved):
LGTM

PR #1236, opened 2021-07-02
Make the encrypted ID invisible

Kevin-Chu58:
This PR makes the encrypted ID invisible.

Example:
image

### DIFF
--- a/src/views/PeopleSearch/components/PeopleSearchResult/index.js
+++ b/src/views/PeopleSearch/components/PeopleSearchResult/index.js
@@ -52,7 +52,7 @@ export default class PeopleSearchResult extends Component {
     fullName = Person.FirstName + ' ' + Person.LastName;
 
     // set nicknames up
-    if (Person.FirstName !== Person.NickName) {
+    if (Person.NickName && Person.FirstName !== Person.NickName) {
       nickname = '(' + Person.NickName + ')';
     }
     // set classes up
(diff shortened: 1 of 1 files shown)


src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -52,7 +52,7 @@ export default class PeopleSearchResult extends Component {
     fullName = Person.FirstName + ' ' + Person.LastName;
 
     // set nicknames up
-    if (Person.NickName && Person.FirstName !== Person.NickName) {
+    if (Person.FirstName !== Person.NickName) {
```
cpabbot, line 55: What if NickName is null, won't this then become true since 'something' !== null right?

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -206,7 +206,9 @@ export default class PeopleSearchResult extends Component {
                 <Typography>{personClassJobTitle}</Typography>
               </Grid>
               <Grid item xs={2}>
-                <Typography>{Person.AD_Username}</Typography>
+                <Typography>
+                  {Person.AD_Username.includes('.') ? Person.AD_Username : null}
+                </Typography>
```
cpabbot, line 211: Haha I suppose this is a good way to do this!

cpabbot (commented):
Looks good but had one question I wanted to check

cpabbot (approved):
LGTM

PR #1241, opened 2021-07-03
RELEASE: Restore timesheets, add Update Phone dialog

EjPlatzer:
The HR Switch over is complete, so we can restore the student timesheets page (in time for students to submit their hours for the pay period that ends today).

Included in this release is the dialog to update phone numbers. This might not work in Production because of mismatched SQL permission issues, but I will fix it when I work in Tuesday. Since Timesheets is a critical system and the update phone is a small, new feature that no one knows exists, I think it is worthwhile to merge it in a (potentially) broken state for the time being.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20392,6 +20392,14 @@
         "prop-types": "^15.6.2"
       }
     },
+    "react-text-mask": {
+      "version": "5.4.3",
+      "resolved": "https://registry.npmjs.org/react-text-mask/-/react-text-mask-5.4.3.tgz",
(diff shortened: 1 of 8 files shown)


EjPlatzer:
@cpabbot  That's surprising because the only change to People Search was that 
```JavaScript
{Person.AD_Username}
```
became
```javascript
{Person.AD_Username.includes('.') ? Person.AD_Username : null}
```
I don't see how an API change (or lack thereof) would make any difference to that code. Unless `Person.AD_Username` is coming through as a non-string - but that's not an API issue, that's a UI issue:
```javascript
{Person.AD_Username?.includes('.') ? Person.AD_Username : null}
```

PR #1242, opened 2021-07-06
Re-organized the logic of setting encrypted ID to invisible

Kevin-Chu58:


cpabbot (approved):
LGTM, thanks for this improvement!

Issue #1235, opened 2021-07-02
Make GordonDialogBox prettier


iamphg97:
Now `GordonDialogBox` looks ok, but there are some rooms to improve its design.

Will make `GordonDialogBox` prettier and apply it to `UpdatePhone' and `LinksDialog`.

Now:
![124184890-35536f80-da88-11eb-895d-a7e719f8ef28](https://user-images.githubusercontent.com/70544403/124281398-ac841480-db17-11eb-8c16-6dbaa9a7d00d.png)

New:
![124184933-443a2200-da88-11eb-87b7-fc8c612a02ee](https://user-images.githubusercontent.com/70544403/124281432-b443b900-db17-11eb-8ac4-a29eb485a7f3.png)
<img width="536" alt="124184953-4a300300-da88-11eb-865a-dc7cd69eeb98" src="https://user-images.githubusercontent.com/70544403/124281450-b73ea980-db17-11eb-9930-e49f7c5b8f76.png">

iamphg97:
Duplicate #1244

PR #1244, opened 2021-07-06
Prettier Gordon Dialog Box

iamphg97:
Linked issues #1235

- [x] Update GordonDialogBox with a better design.
- [x] Check all places where use `GordonDialogBox` to make sure this update does not break anything already existed.

Checked all places and looks fine.

Before:
Screen Shot 2021-07-06 at 1 17 11 PM 1

After:
Screen Shot 2021-07-06 at 1 16 57 PM

- [x] Apply new GordonDialogBox to "Update Phone".

Looks identical.

Before:
Screen Shot 2021-07-06 at 1 35 19 PM

After:
Screen Shot 2021-07-06 at 1 35 35 PM

There are some other places where `GordonDialogBox` is not used.
And I worked on some of it...
I left some because they need some special features (such as three buttons instead of two)...

Here is one that I have worked on. (Useful Links)
Before:
Screen Shot 2021-07-06 at 2 21 13 PM

After:
Screen Shot 2021-07-06 at 2 21 19 PM


josh-peters-2357 (approved):
lgtm

src/components/Header/components/PeopleSearch/index.js
```diff
@@ -415,24 +404,15 @@ export default class GordonPeopleSearch extends Component {
               ),
             }}
           />
-          <Dialog
+          <GordonDialogBox
```
cpabbot, line 407: This should actually use GordonUnauthorized yes? I totally missed this one!

cpabbot (changes requested):
Awesome work on this!

src/components/Header/components/PeopleSearch/index.js
iamphg97: You are right! Just fixed it

silasw2 (commented):
Changing these to functional components would be helpful!

iamphg97:
Will work on it now.

cpabbot (approved):
LGTM

silasw2 (approved):
LGTM! Great work on this!

PR #1247, opened 2021-07-07
Lazy load home images

josh-peters-2357:


src/views/Home/components/Carousel/index.js
```diff
@@ -45,6 +45,7 @@ const GordonCarousel = () => {
           originalTitle: slide.Title,
         }))}
         onClick={handleClickSlide}
+        lazyLoad={true}
```
EjPlatzer, line 48: I'm not sure this actually has the intended affect, because this component is not responsible for requesting our images, so telling it to lazyLoad won't accomplish anything as far as I'm aware. We load all the images on this component's first render in the useEffect above, which happens regardless of what this ImageGallery component does.

josh-peters-2357: It does have the intended effect. When the page loads, the first and next image are loaded. Each time the gallery switches, the next image is requested. I checked that this was the behavior in browser.

josh-peters-2357: The useEffect gets different data about the images including their paths. That's why `ImageGallery` makes a map of `carouselContent`.

EjPlatzer: You are correct. I mistakenly thought that the useEffect loaded the images themselves, but it actually gets the image paths (along with other metadata) from the database. That explains why it behaves as intended. Good work!

EjPlatzer (approved):
LGTM1

PR #1251, opened 2021-07-07
Styling Example - Styled Components

silasw2:
This was an experimentation with using the MUI styled components. Overall, the implementing the styling inside of the health status page is a little too complicated for our version of MUI. With the update to v5 (if we decide to take that route) all of these changes should implement cleanly. So this is more or less an example of how it would look.
#1168 

**PROS:**

- Everything is centralized inside of the index.js file making it a little easier to understand the styling as a whole.
- Doing global styling changes (hence if we wanted to implement a dark mode) should be rather trivial by having a global MUI theme for the page that can be updated at the users request.
- Logic and useful javascript tools can be used in order to reduce the bulk of the styling code to be written, such as the switch case updating the style directly instead of calling a whole different className for the element.
- Has lasting support and tools provided by MUI, which we already use.

**CONS:**

- Switching over from our current styling would be a long process, not as easy as something like CSS modules
- We would have to update to MUI v5 in order to allow some of the styling choices we have made be done in styled components. There are roundabout ways of doing this styling in v4 but it is not intuitive and would require more work than updating to v5 in my opinion.

While the cons list looks short, it represents a large amount of time to fully complete the swap so it should be carefully considered whether we want to put in the effort.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@material-ui/icons": "^4.11.2",
         "@material-ui/lab": "^4.0.0-alpha.58",
         "@material-ui/pickers": "^3.3.10",
+        "@material-ui/styled-engine": "^5.0.0-alpha.11",
         "caniuse-lite": "^1.0.30001230",
         "chart.js": "^2.9.4",
         "classnames": ">=2.3.1",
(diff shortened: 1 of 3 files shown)


EjPlatzer:
Closing this because we have decided to use CSS Modules going forward - see https://github.com/gordon-cs/gordon-360-ui/issues/1168

PR #1254, opened 2021-07-07
Rearranged the style on advanced people search

Kevin-Chu58:
This PR brings the advanced people search a new style with more info.

**Info added**:
- Majors
- Graduating year for alumni

**Info rearranged**:
- Merge Photo, first name, last name, and Email into "Fullname"
- Rename "Class/Job Title" to "Title/Class"
- Drop the "Description" column (Merge it to "Title/Class" column)
  - If the accounts do not have a Job Title or a Class, replace it with Type
- Add PreferredClassYear to the "Title/Class" column only for alumni
- Also add major(s) to the "Title/Class" column
- Move "Mail Location" to a new column
- Make the image larger on the desktop default view\
- Display maiden name on the advanced people search

Old examples:
- Desktop Default View
image
image
- Desktop Large Image View
image
- Mobile View
image

New Examples:
- Desktop Default View
image
- Desktop Large Image View
image
- Mobile View
image
- Showing maiden name on the advanced people search
image


src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -119,12 +119,29 @@ export default class PeopleSearchResult extends Component {
...
+                  <Typography variant="body2">
+                    {Person.Major1Description}
+                    {Person.Major2Description === null || Person.Major2Description === ''
+                      ? null
+                      : ', '}
+                    {Person.Major2Description}
+                    {Person.Major3Description === null || Person.Major3Description === ''
+                      ? null
+                      : ', '}
+                    {Person.Major3Description}
```
EjPlatzer, line 140: Since `null` and the empty string are both falsy values, this can be streamlined quite a bit:
```suggestion
                    {Person.Major1Description}
                    {Person.Major2Description ? `, ${Person.Major2Description}` : null}
                    {Person.Major3Description ? `, ${Person.Major3Description}` : null}
```

josh-peters-2357 (changes requested):
Could you add `Person.Type` to the props list as required? Also, there are a number of other props used that aren't specified, I think because they're not required. But we could still add them (and you just don't add `.isRequired`), and I think we should. Your code, by the way, looks really good, although it will be nice to one day make this a functional component.

Kevin-Chu58:
> Could you add `Person.Type` to the props list as required?

image
Like this? @josh-peters-2357

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -119,15 +119,23 @@ export default class PeopleSearchResult extends Component {
...
+                    {Person.Type === 'Alum' ? ' ' + Person.PreferredClassYear : null}
+                  </Typography>
+                  <Typography variant="body2">
+                    {Person.Major1Description}
+                    {Person.Major2Description ? `, ${Person.Major2Description}` : null}
+                    {Person.Major3Description ? `, ${Person.Major3Description}` : null}
+                  </Typography>
+                  <Typography variant="body2">{Person.Email}</Typography>
+                  <Typography variant="body2">{personMailLocation}</Typography>
+                </div>
```
EjPlatzer, line 138: I like the effect of having grayer text! There is a standard MUI way to achieve this that is used elsewhere in the site - by setting the `color` prop of `Typography` to `textSecondary`. That has the disadvantage of having to be repeated for every `Typography` (unless you create a custom `SecondaryText` component).
```javascript
const SecondaryText = ({children, otherProps}) => <Typography variant="body2" color="textSecondary" {...otherProps}>{children}</Typography>
...
<SecondaryText>{Person.Email}</SecondaryText>
<SecondaryText>{personMailLocation</SecondaryText>
```

If you do choose to keep the enclosing `div` with a color that's inherited, then you should at least move the styling to a class in the `scss` file. That way the style is declared only once in the class, and each div uses that className. Then, if we ever need to change the style, we only have to change the class, not every individually styled div.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -188,34 +204,37 @@ export default class PeopleSearchResult extends Component {
...
+                  />
+                  <div>
+                    <Typography>
+                      {Person.FirstName} {nickname} {Person.LastName}
+                    </Typography>
+                    <Typography style={{ fontSize: '85%' }}>
+                      {Person.Email?.includes('.') ? Person.Email : null}
+                    </Typography>
+                  </div>
+                </div>
```
EjPlatzer, line 224: You can use the pre-existing Grid to align these, instead of adding an extra `div`.
```suggestion
              <Grid item xs={5} container alignItems="center">
                  <IMG
                    className="people-search-avatar"
                    src={`data:image/jpg;base64,${this.state.avatar}`}
                    alt=""
                    noLazyLoad="true"
                    placeholderColor="#eeeeee"
                  />
                  <div>
                    <Typography>
                      {Person.FirstName} {nickname} {Person.LastName}
                    </Typography>
                    <Typography style={{ fontSize: '85%' }}>
                      {Person.Email?.includes('.') ? Person.Email : null}
                    </Typography>
                  </div>
```

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -188,34 +204,37 @@ export default class PeopleSearchResult extends Component {
...
+                    src={`data:image/jpg;base64,${this.state.avatar}`}
+                    alt=""
+                    noLazyLoad="true"
+                    placeholderColor="#eeeeee"
+                  />
+                  <div>
+                    <Typography>
+                      {Person.FirstName} {nickname} {Person.LastName}
+                    </Typography>
+                    <Typography style={{ fontSize: '85%' }}>
```
EjPlatzer, line 220:
```suggestion
                    <Typography variant="subtitle2">
```

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -188,34 +204,37 @@ export default class PeopleSearchResult extends Component {
...
+                </div>
               </Grid>
-              <Grid item xs={2}>
+              <Grid item xs={5}>
                 <Typography>
-                  {Person.FirstName} {nickname}
+                  {personClassJobTitle ?? Person.Type}
+                  {Person.Type === 'Alum' ? ' ' + Person.PreferredClassYear : null}
+                </Typography>
+                <Typography style={{ fontSize: '85%', fontStyle: 'italic', color: '#666666' }}>
```
EjPlatzer, line 231:
```suggestion
                <Typography style={{ fontStyle: 'italic' }} color="textSecondary" variant="subtitle2">
```

EjPlatzer (changes requested):
This is an amazing visual improvement!

One note: Prefer to avoid inline styling (i.e. having a `style={{...}}` prop) whenever possible, especially if the style appears more than once. Creating reusable styles (either through scss classes or through reusable components) makes our code overall cleaner and more maintainable - and usually easier to understand as well! 

Also, using the pre-configured props for MUI elements is usually better than custom `style={{...}}` props. Mainly, the difference is that the MUI props will be consistent across the whole site. For example, specifying `color="textSecondary"` will result in the same gray text everywhere on the site, whereas using `style={{ color: ... }}` might result in different developers using different color values, which can subtly clutter the look and feel of the site.

You can visit material-ui.com for MUI documentation. For example, here is a page describing the `Typography` props: https://material-ui.com/api/typography/

src/views/PeopleSearch/components/PeopleSearchResult/index.js
Kevin-Chu58: Done.

src/views/PeopleSearch/components/PeopleSearchResult/index.js
```diff
@@ -49,6 +49,8 @@ export default class PeopleSearchResult extends Component {
   render() {
     const { Person, size } = this.props;
 
+    console.log(Person);
+
```
josh-peters-2357:
```suggestion
```

PR #1266, opened 2021-07-12
A couple changes

josh-peters-2357:
This commit just removes a couple errors from the console. There was code here and there that was generating errors but was still merged into development somehow.

- in `PersonalInfoList`, Typography uses a `<p>` tag, and `<ul>` cannot be a descendant of `<p>`.
- in `DiningBalance`, there were several div tags using "class" which is an invalid DOM property (though the classname they were looking to use does exist).
- in `InvolvementProfile`, maps were being created without unique keys, which is not good practice.
- in `PeopleSearch`, there was a `setState` that was writing to state properties that would later be used as Booleans. They were usually set to Booleans, but the very first time they are loaded (when the component is first rendered) they were being set to strings instead of Booleans. That was worrying React because if you pass a nonempty string to the `checked` property of a checkbox, it will still show. So, I changed that `setState` so it sets to true or false based on whether the value is "true" or not.


roddy3 (dismissed):
left a comment

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -264,10 +264,17 @@ const PersonalInfoList = ({
...
     myProf &&
     (isFacStaff ? (
-      <Typography align="left" className="note">
-        NOTE: To update your data, please contact{' '}
-        <a href="mailto: hr@gordon.edu">Human Resources</a> (x4828).
-      </Typography>
+      <div align="left" className="note">
+        <Typography>NOTE:</Typography>
+        <ul>
+          <li>
```
roddy3: why this change? I think the old one looks good to me. would you mind explain it to me please?

roddy3 (changes requested):
why this change?

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -264,10 +264,17 @@ const PersonalInfoList = ({
...
+            <Typography>
+              To update your data, please contact{' '}
+              <a href="mailto: hr@gordon.edu">Human Resources</a> (x4828).
+            </Typography>
+          </li>
+        </ul>
+      </div>
```
josh-peters-2357:
```suggestion
<Typography align="left" className="note">
        NOTE: To update your data, please contact{' '}
        <a href="mailto: hr@gordon.edu">Human Resources</a> (x4828).
        </Typography>
```

src/components/Profile/components/PersonalInfoList/index.js
josh-peters-2357: I think I fixed it in this newest commit.

josh-peters-2357:
> 
> 
> why this change?

We discussed this in person right? @roddy3

roddy3 (approved):

PR #1268, opened 2021-07-12
Fix Quick People Search Loading

Kevin-Chu58:
This PR fixes the display error on the quick people search.

Before, the quick people search sends requests every time when the input param is changed, which is very problematic because some requests take longer to search which overwrites the search result of the latest request that took much less time.

Before
===
(This happens if you type fast)
---
image

After
===
image

### DIFF
--- a/src/components/Header/components/PeopleSearch/index.js
+++ b/src/components/Header/components/PeopleSearch/index.js
@@ -78,7 +78,7 @@ export default class GordonPeopleSearch extends Component {
 
     let results = await peopleSearch.search(query);
     let suggestions = results.result;
-    console.log(results.result);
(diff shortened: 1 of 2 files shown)


josh-peters-2357 (approved):
It looks good to me. It's kind of sad, but I have a PR just sitting for this site with no reviews or comments where I refactored the quick people search to a functional component. Once this gets approved and merged, that PR will have merge conflicts.

Kevin-Chu58:
@josh-peters-2357  It's not too bad. Luckily my modification to your code has almost no intersections.

roddy3 (approved):
Looks Good to me!!

PR #1273, opened 2021-07-14
(S21) PATCH - Point Train to Train DB (fixed .env.production)

cpabbot:
.env.production apparently actually refers to any production environment, including both train and prod.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -92,13 +92,14 @@
     <!-- Different font key for development and production -->
     <!-- Font will not work in development and will cause a 403, because the only domains allowed are *.gordon.edu -->
     <!-- Prod font key @PROD -->
-    <link
+    <!-- <link
       rel="stylesheet"
       type="text/css"
       href="https://cloud.typography.com/7763712/7294392/css/fonts.css"
-    />
+    /> -->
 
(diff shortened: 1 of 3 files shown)


josh-peters-2357 (approved):

public/index.html
```diff
@@ -92,13 +92,14 @@
...
-    <link
+    <!-- <link
       rel="stylesheet"
       type="text/css"
       href="https://cloud.typography.com/7763712/7294392/css/fonts.css"
-    />
+    /> -->
 
     <!-- Dev font key @TRAIN -->
+    <!-- Does not currently work, returns 403 -->
```
EjPlatzer, line 102: This is actually expected to fail in Train, since we only have the rights to use the font in one website (and prod and train are technically different sites).

cpabbot: Not expected to fail in Train, this is supposed to be the alternative font. See line 93.

cpabbot: Not expected to fail in Train, this is supposed to be the alternative font. See line 93.
Edit: I am now realizing that this font key may have actually worked on train and just not on localhost since those are different sites. Does not appear to make any difference though so might not have worked.

PR #1283, opened 2021-07-16
(S21) PATCH - Update Clifton Strength Color Accessibility

cpabbot:
Increase color contrast for Clifton Strengths to improve accessibility. Thanks @iamphg97 for pointing this out to me!

Before:
image

After:
image

Simon would be so proud


iamphg97 (approved):
LGTM! 감사합니다.

PR #1286, opened 2021-07-16
Add UTC label to API build time in About page

jsenning:
The API build time is included in the API executable and displayed in the About page.  The recorded time is UTC rather than one of the USA time zones so it's helpful to explicitly state this.  The simplest way to do this is to just add "UTC" to the string displayed in the About page.


Davideo37 (approved):
LGTM!

josh-peters-2357 (approved):
LGTM

PR #1288, opened 2021-07-19
Reorganized quick people search suggestion display

Kevin-Chu58:
The old code displays nickname if there is a nickname, if not, then check if there is maiden name, if not, then only show firstname + lastname.

The new code displays nickname if there is a nickname and maiden name if there is a maiden name.

### DIFF
--- a/src/components/Header/components/PeopleSearch/index.js
+++ b/src/components/Header/components/PeopleSearch/index.js
@@ -173,16 +173,6 @@ export default class GordonPeopleSearch extends Component {
     );
   }
 
-  checkMaidenName(maidenName, lastName, userName1) {
-    if (maidenName &&
-      maidenName !== lastName &&
-      maidenName !== userName1) {
-        return ' (' + maidenName + ')';
-      }
-      else
-      return "";
-  }
-
   renderSuggestion(params) {
     const { suggestion, itemProps } = params;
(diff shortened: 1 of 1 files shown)


src/components/Header/components/PeopleSearch/index.js
```diff
@@ -215,20 +215,22 @@ export default class GordonPeopleSearch extends Component {
...
+                suggestion.LastName +
+                // If having maiden name that is unique, display that maiden name
+                (suggestion.MaidenName &&
+                suggestion.MaidenName !== suggestion.LastName &&
+                suggestion.MaidenName !== suggestion.UserName.split(/ |\./)[1]
+                ? ' (' +
+                suggestion.MaidenName +
+                ')'
+                : ''),
                 this.state.highlightQuery,
```
josh-peters-2357, line 234: This is a large number of expressions in the return of the component; I wonder if all of this should be in a function somewhere else? You could call `getHighlightedText(getSuggestedName(suggestion))` or something like that. But it's just an idea to make it neater, I don't know that it's entirely necessary. What do you think?

josh-peters-2357: (We discussed this in person; not worth doing for now.)

josh-peters-2357 (approved):
LGTM

PR #1289, opened 2021-07-19
RELEASE: Student News Images, Clifton 2.0, Disabled Wellness Check Home

cpabbot:
This release includes:
- Student News Images
- Clifton Strengths 2.0
- Wellness Check Disabled on Home Page
- Misc console warning cleanup
- UTC time to API build version

Prettier has changed parenthesis in a file, removing them while they do exist in develop. We should check on this and perhaps find a way to commit without running prettier on a release branch (we should avoid differences like that between develop and release).


iamphg97 (approved):
LGTM!

PR #1291, opened 2021-07-19
S21 Academic Check-In Final

Davideo37:
This pull request contains all the functionality for Academic Check-In, a feature of the Go site that is now implemented into 360.

Contained in this feature:

- Ability to fetch student academic holds 
- Ability to fetch and set student emergency contact information
- Ability to fetch and set student personal phone information
- Ability to set student demographic data (race/ethnicity)
- Ability to confirm a student's attendance at Gordon for Registrar's office (functionality is all there, still needs database permissions)

This pull request is connected to the API in PR https://github.com/gordon-cs/gordon-360-api/pull/648 and should only be merged alongside that one.


.env.production
```diff
@@ -1 +1 @@
-REACT_APP_API_URL=https://360Api.gordon.edu/
+REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
```
cpabbot, line 1:
```suggestion
REACT_APP_API_URL=https://360Api.gordon.edu/
```

cpabbot (changes requested):

src/views/AcademicCheckIn/components/ConfirmCheckIn/index.js
```diff
@@ -0,0 +1,127 @@
...
+  const raceNames = ['Native American', 'Asian', 'Black', 'Hawaiian', 'White', 'Prefer not to say'];
+
+  const displayRace = () => {
+    var i;
+    var names = '';
+    for (i = 0; i < raceValues.length; i++) {
+      raceValues[i] === true ? (names += raceNames[i] + ' ') : (names += '');
+    }
+    return names;
+  };
```
EjPlatzer, line 31: You could condense this quite a bit:
```suggestion
  const { ethinicity, ...raceValues } = demographic;

  const races = {
    nativeAmerican: 'Native American',
    asian: 'Asian',
    black: 'Black',
    hawaiian: 'Hawaiian',
    white: 'White',
    none: 'Prefer not to say',
  };

  const displayRace = () => races.filter((race) => raceValues[race]).join(' ');
```

src/views/AcademicCheckIn/components/PrivacyAgreement/privacy.json
```diff
@@ -0,0 +1,36 @@
+{
+  "FERPA":{
+      "header": "Notification of Rights Under FERPA",
```
EjPlatzer, line 3: What drove the decision to put this content in a json file instead of inline in the JSX? Usually, we store in josn files data that has a repetitive structure (and so can be looped over with minimal JSX code compared to explicitly repeating it all), and data that changes frequently based on site-external factors. This doesn't seem like it fits either case, so I would think it is simpler and more understandable to put it in the JSX directly.

src/views/AcademicCheckIn/components/CompletedCheckIn/index.js
```diff
@@ -0,0 +1,35 @@
...
+          Congratulations {basicInfo.studentFirstName} on completing Academic Checkin!
+        </Typography>
+        <br />
+        <Typography variant="body1" align="center" gutterBottom>
+          We are excited to welcome you to your new semester at Gordon College! Please click the
+          button below to return to the homepage.
+        </Typography>
+        <br />
+        <Grid item align="center">
+          <Button variant="contained" onClick={() => window.location.replace('')}>
```
EjPlatzer, line 26: Using `window.location` for navigation is not ideal in a Single-Page Application like 360, because it forces the entire site to reload from scratch. Instead, we use the `Link` component from `react-router` for navigation, e.g.:
```suggestion
<Button variant="contained" component={Link} to="/">
```

src/views/AcademicCheckIn/components/RaceEthnicity/info.json
```diff
@@ -0,0 +1,9 @@
+{
+    "header": "Federal Reporting Status",
```
EjPlatzer, line 2: As above, this doesn't seem like the ideal circumstances to store data as a json file.

src/views/Home/index.js
```diff
@@ -90,7 +93,9 @@ const Home = ({ authentication, onLogIn }) => {
   } // @WELLNESS-CHECK disabled to revert this you must uncomment this lines of code
   //else if (networkStatus === 'online' && !hasAnswered) {
   //return <WellnessQuestion setStatus={() => setHasAnswered(true)} />;}
-  else {
+  else if (!checkedIn && personType.includes('stu')) {
+    window.location.replace("AcademicCheckIn")
```
EjPlatzer, line 97: Please use `react-router` hooks to navigate, rather than `window.location.replace`.
```javascript
let history = useHistory()
...
history.replace("AcademicCheckIn")
```

EjPlatzer (changes requested):
This is really fantastic! I didn't get to look at the UI yet since the API hasn't been merged all the way, but overall it's looking great.

I left a few stylistic comments and questions, but the main change I'm requesting is to use `react-router` for navigation, rather than the browser built-in `window` object. The latter will reload the entire SPA whenever someone navigates, which results in long loading times and poor user experience.

src/services/wellness.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Wellness Check API call functions
+ * Wellness Check API call functions 🧙‍♂️
```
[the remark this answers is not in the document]
Davideo37: Ah I thought we missed one of those xD

src/views/AcademicCheckIn/components/AcademicCheckInWelcome/index.js
```diff
@@ -0,0 +1,205 @@
+import React from 'react';
+import { Typography, Grid } from '@material-ui/core';
+import { gordonColors } from 'theme';
+// import './index.css';
```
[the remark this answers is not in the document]
Davideo37: Good catch, I forgot we added that

src/views/AcademicCheckIn/components/PrivacyAgreement/privacy.json
Davideo37: It was primarily to save lines in the main file to improve readability as having long paragraphs that took up dozens of lines made it exhaustive to find specific code. Also this way if someone wants to update the information displayed it's much easier to find the paragraph they want to modify.

src/views/AcademicCheckIn/components/RaceEthnicity/info.json
Davideo37: Yeah, I understand this one. This was I think more because we wanted to be consistent, but in hindsight it was probably not necessary.

src/views/AcademicCheckIn/components/CompletedCheckIn/index.js
Davideo37: Seems like Link is not exported from react-router for some reason.

src/services/wellness.js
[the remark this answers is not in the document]
silasw2: Oops lol

src/views/AcademicCheckIn/components/ConfirmCheckIn/index.js
```diff
@@ -11,16 +11,9 @@ const ConfirmCheckIn = ({
...
 
-  const races = {
-    nativeAmerican: 'Native American',
-    asian: 'Asian',
-    black: 'Black',
-    hawaiian: 'Hawaiian',
-    white: 'White',
-    none: 'Prefer not to say',
-  };
+  const races = ['Native American','Asian','Black','Hawaiian','White','Prefer not to say'];
```
EjPlatzer, line 16: Does this work as intended? `raceValues` is an object of the form 
```javascript
{
  asian: boolean,
  black: boolean,
  hawaiian: boolean,
  nativeAmerican: boolean,
  none: boolean,
  white: boolean
}
```
So calling, for example, `raceValues["Native American"]` will not return `raceValues.NativeAmerican`, and the same for `None` versus `Prefer not to say`. You either need to change the names of the properties in `demographic`, or you need `races` to be an object mapping the race name to the race description, as it was before.

silasw2: Unfortunately we did not realize these were in the commit, we will fix them ASAP.

Davideo37: > 
> 
> Does this work as intended? `raceValues` is an object of the form
> 
> ```js
> {
>   Asian: boolean,
>   Black: boolean,
>   Hawaiian: boolean,
>   NativeAmerican: boolean,
>   None: boolean,
>   white: boolean
> }
> ```
> 
> So calling, for example, `raceValues["Native American"]` will not return `raceValues.NativeAmerican`, and the same for `None` versus `Prefer not to say`. You either need to change the names of the properties in `demographic`, or you need `races` to be an object mapping the race name to the race description, as it was before.

The problem I ran into with your suggested change is that ".filter" only works for Arrays, not JSON objects. So I changed it to this which works for the single word objects ('White', 'Black', etc.) but not for the longer ones. I'm not sure what other method of JSON objects would work for your suggested change, but me and Silas will look into it.

EjPlatzer: Oh, you're right, I suggested running `filter()` on an object, sorry. One way to make my original suggestion work is with `Object.keys()`, which takes an object and returns an array of the keys of the object. For example:
```javascript
const displayRace = Object.keys(races) // array of the form ['asian', 'black', 'nativeAmerican', ...]
    .filter((race) => raceValues[race]) // Above array but with only the races that apply to this user
    .map((race) => races[race]) // Map race keys to race descriptions, resulting in e.g. ['Native American', 'Black']
    .join(' '); // Concatenate into space-delimited string 'Native American Black'
```

It's a little bit messier, but still potentially clearer for people who are familiar with array methods. Without the comments (which just explain the standard behavior of well-documented JavaScript array methods), this is just one line:
```javascript
const displayRace = Object.keys(races).filter((race) => raceValues[race]).map((race) => races[race]).join(' ');
```

src/views/AcademicCheckIn/components/EmergencyContactUpdate/index.js
```diff
@@ -0,0 +1,220 @@
...
+      )}
+      <Grid item>
+        <br />
+      </Grid>
+      <Typography variant="body2" gutterBottom>
+        If, during the semester your emergency contact information must be updated, please contact
+        Student Life at 978.867.4263.
+      </Typography>
+      <Typography variant="body2" gutterBottom>
+        For international phone numbers, check the "Non-US Number" checkbox to enter information.
```
EjPlatzer, line 75: These instructions need to be updated. The checkbox is labelled `International number` now.
```suggestion
        For international phone numbers, check the "International #" checkbox.
```

EjPlatzer (changes requested):
This looks great! there are some small ways we might be able to simplify the code and make the UI sleeker in the future, but this is already in a good state!

One minor instruction text issue I noticed, but otherwise this LGTM!

EjPlatzer (dismissed):
LGTM!

russtuck (dismissed):

russtuck (dismissed):

russtuck (approved):

PR #1292, opened 2021-07-20
Update index.js

josh-peters-2357:
To account for this beautiful artifact:
image
Which is a difference that appeared in production that we didn't have when we ran this in development, we are increasing the minimum field size in the hopes that it will look better.


roddy3 (approved):
Looks Good to me!!!

PR #1294, opened 2021-07-20
Fixed special char return all results

Kevin-Chu58:
This PR prevents using special characters in the text fields of the advanced people search returning everyone.


josh-peters-2357 (approved):
LGTM

PR #1302, opened 2021-07-23
Academic Check In Quick Fixes

silasw2:
Realized that window.history is bad and does not work with all browsers (sorry @EjPlatzer for not believing you). Replaced this to use the history already defined in app.js and it seems to be fully functional.

Also implemented a small change to the phone page at request of Dr. Tuck.

### DIFF
--- a/src/views/AcademicCheckIn/index.js
+++ b/src/views/AcademicCheckIn/index.js
@@ -105,7 +105,6 @@ const AcademicCheckIn = (props) => {
     studentLastName: '',
   });
 
-
   useEffect(() => {
     const loadData = async () => {
       let profile = await user.getProfileInfo();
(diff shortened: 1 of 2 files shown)


roddy3 (approved):
looks Good to me!!!

josh-peters-2357 (approved):
LGTM

src/views/AcademicCheckIn/index.js
```diff
@@ -169,24 +168,27 @@ const AcademicCheckIn = (props) => {
...
 
+  useEffect(() => {
+    props.history.replace('/AcademicCheckIn', { step: activeStep });
+  }, [activeStep, props.history]);
+
   const handleNext = () => {
+    props.history.push('/AcademicCheckIn', { step: activeStep });
     setActiveStep((nextStep) => nextStep + 1);
-    window.history.pushState(activeStep, null, '');
+    props.history.replace('/AcademicCheckIn', { step: activeStep });
```
EjPlatzer, line 180: This is a little bit confusing, but I suspect it may not behave as you expect. It looks like you're trying to push the current active step, and then replace that entry with the next active step (i.e. what is set in `setActiveStep`).

However, `set...` does not run immediately, but rather queues a state change to be executed on React's next render. So if you're depending on `activeStep` to have updated based on the call to `setActiveStep` here, then you are creating a race condition.

PR #1303, opened 2021-07-23
Updates for the history in Academic Check In

silasw2:
Made another blunder and missed Evan's comment on the last PR. Checking if this looks better.

### DIFF
--- a/src/views/AcademicCheckIn/index.js
+++ b/src/views/AcademicCheckIn/index.js
@@ -177,13 +177,11 @@ const AcademicCheckIn = (props) => {
   const handleNext = () => {
     props.history.push('/AcademicCheckIn', { step: activeStep });
     setActiveStep((nextStep) => nextStep + 1);
-    props.history.replace('/AcademicCheckIn', { step: activeStep });
   };
 
   const handlePrev = () => {
     props.history.push('/AcademicCheckIn', { step: activeStep });
     setActiveStep((previousStep) => previousStep - 1);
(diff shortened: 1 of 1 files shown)


src/views/AcademicCheckIn/index.js
```diff
@@ -177,13 +177,11 @@ const AcademicCheckIn = (props) => {
   const handleNext = () => {
     props.history.push('/AcademicCheckIn', { step: activeStep });
     setActiveStep((nextStep) => nextStep + 1);
-    props.history.replace('/AcademicCheckIn', { step: activeStep });
   };
 
   const handlePrev = () => {
     props.history.push('/AcademicCheckIn', { step: activeStep });
     setActiveStep((previousStep) => previousStep - 1);
-    props.history.replace('/AcademicCheckIn', { step: activeStep });
```
silasw2, line 186: The original comment,



This is a little bit confusing, but I suspect it may not behave as you expect. It looks like you're trying to push the current active step, and then replace that entry with the next active step (i.e. what is set in setActiveStep).

However, set... does not run immediately, but rather queues a state change to be executed on React's next render. So if you're depending on activeStep to have updated based on the call to setActiveStep here, then you are creating a race condition.

EjPlatzer (approved):
This looks great to me! Much more sensible and less-error prone!

silasw2:
> This looks great to me! Much more sensible and less-error prone!

Sorry about that! Should work as intended now!

PR #1307, opened 2021-07-23
Removed Academic Check In from Homepage

silasw2:
Edited by @cpabbot 
This was removed since some db difference is improperly redirecting on prod


cpabbot (approved):
LGTM

PR #1316, opened 2021-07-26
CSS Modules Migration *PATCH* Quick search animation and width

### LINKED ISSUE #1312: Quick Search Results on Mobile
Styling is set such that the width of the dropdown results is flexible, but there needs to be some kind of limit as when there is a search that contains a very long name (including nickname) like typing `'ab'` the results list can overflow out of the mobile screen width. If you scroll down about 3/4 of the way for this search you will see the name that causes this particular overflow. The width should probably instead just be set to the same width as the quick search bar (or maybe full width on xs screens) and then extra content should line break.

image

cpabbot:
Patch to #1306 CSS Modules Migration
- Adds quick search animation on focus like in prod site (refactored width transition into the Header parent)
- Fixed width of People Search Results dropdown (wraps long names)

Resolves #1312

### DIFF
--- a/src/components/Header/Header.module.scss
+++ b/src/components/Header/Header.module.scss
@@ -46,9 +46,16 @@ $drawer-width: 240px;
     display: none;
   }
 
-  .people_search_container {
-    width: 9rem;
+  .people_search_container_container {
+    width: 15rem;
     text-align: right;
+  }
+
+  .people_search_container {
+    text-align: right;
+    width: 9rem;
+    transition: width 0.3s;
+    display: inline-block;
   }
 
   @media (min-width: $break-xs) {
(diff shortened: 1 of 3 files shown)


src/components/Header/Header.module.scss
```diff
@@ -46,9 +46,16 @@ $drawer-width: 240px;
     display: none;
   }
 
+  .people_search_container_container {
```
cpabbot, line 49: Oof don't mind this *whistles nonchalantly*

src/components/Header/index.js
```diff
@@ -229,8 +229,15 @@ const GordonHeader = ({ authentication, onDrawerToggle, onSignOut }) => {
             </Tabs>
           </div>
 
-          <div className={styles.people_search_container}>
-            {authentication ? <GordonPeopleSearch authentication={authentication} /> : loginButton}
+          <div className={styles.people_search_container_container}>
```
cpabbot, line 232: Needed another container to prevent the toolbar icons from shifting on focus. Don't mind the name lol...

EjPlatzer (approved):
LGTM! I wish we had a simpler class name though

PR #1317, opened 2021-07-27
close button for Links Dialog

Thatguy560:
Added a simple "Close ✕" button to the Links dialog panel, which saves you clicking outside the dialog box in order to close it.

Screenshot 2021-07-27 at 14 31 07


EjPlatzer:
Thank you for your contribution @Thatguy560! Unfortunately, it does not meet our design standards - we use the Material UI component library for standard UI components like buttons, and we also have special internal components that are used in this particular section of the code. For that reason, I'm going to have to reject this PR, but we appreciate your effort and willingness to help!

PR #1320, opened 2021-07-29
Dark Theme Groundwork

cpabbot:
Creating this PR just so that @EjPlatzer can take a look if interested, or at least as reference for future development. There is some good in here but probably needs a great deal of rework before pushing to develop.

Experimenting with dark theme was helpful in doing some code cleanup because it revealed inconsistencies in style.

### DIFF
--- a/src/app.global.scss
+++ b/src/app.global.scss
@@ -31,32 +31,38 @@ body {
   padding: 0;
 }
 
+// a {
+//   // Disable link underline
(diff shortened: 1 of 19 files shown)


amos-cha:
https://github.com/gordon-cs/gordon-360-ui/pull/1917 
issue handled (in-progress 2023/07/10) here

PR #1321, opened 2021-08-09
Prepare wellness check for start of Fall 2021

EjPlatzer:
In preparation for students returning to campus for the fall (some as early as August 13th), we are making some small updates to the Wellness Check system.

Firstly, it is being re-enabled on the Home page. However, due to backend changes to the system, it will be invisible to most users most of the time because they will have an automatically applied baseline Green status that does not expire. If the system is ever switched back into active daily check in mode, those baseline statuses will be automatically cleared and the check in will seamlessly return to the daily check in functionality used throughout the previous academic year.

Secondly, as students return to campus they are required to submit two forms before they can be marked well:
1. The Fall 2021 Return to Campus form that includes proof of testing before returning to campus.
2. The Covid-19 Vaccination Status report, which is simply a record of all students' current vaccination status, which is needed by the Health Center.
Before students return to campus, they will all be marked RED pending the submission of these two forms. Once both are submitted, their status will be cleared and they will be able to complete the check in (or have their baseline set overnight). To help students be aware of the cause for their red status and the necessary steps to clear it, the message on the Wellness page has been updated to reflect this requirement and direct students to the two forms they need to complete.


bennettforkner (approved):
LGTM

PR #1322, opened 2021-08-10
Release: 2021 Fall Start - Wellness updates, CSS Modules Migration

EjPlatzer:
This release includes the first phase of migration to CSS Modules - parity with previous mostly via scripted changes, and manually fixing errors. It also includes minor updates to the Wellness system in preparation for the start of the 2021 Fall semester. The home page wellness check in has been re-enabled, and the Health Status component has been updated to link students to the correct forms when their status is RED.


EjPlatzer:
I'm going to merge this as an administrator because we don't have reviewers and the Wellness check in is a time sensitive feature. We have tested and checked the UI after CSS Modules migration extensively. The Wellness check in has not been changed, only re-enabled and the link updates have been tested locally.

Issue #1323, opened 2021-08-10
Events Page Responsiveness broken


EjPlatzer:
The events page currently only updates layout by checking window width at render time. This means it doesn't re-render and change layout until something else updates the component. It should instead track screen width in an effect, so that it responds to layout immediately.

Nuno111:
Hello EjPlatzer I would like to tackle this issue if possible.

Many thanks

PR #1324, opened 2021-08-11
Fix event page width

Nuno111:
Created a custom hook to track width and height whenever the page is resized and adjusted events page to respond to width changes. 

This relates to issue #1323


EjPlatzer (approved):
This is excellent, thank you for the contribution @Nuno111!

PR #1325, opened 2021-08-12
Removing Academic Checkin from 360

bennettforkner:
We are sticking with the GO Gordon site for this year (2021 Fall) and will continue to test this process until then.

**Message from Evan Platzer:**
"Given that we aren't able to do a thorough test of the system soon though, I'm beginning to think that sticking with the GO site for this semester is the best call. I'll have the GoCo Labs team test it further this fall and hopefully we will be confident in it for the Spring."

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -94,11 +94,12 @@ const routes = [
     path: '/myprofile',
     component: MyProfile,
   },
-  {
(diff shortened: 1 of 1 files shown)


EjPlatzer (approved):
LGTM!

PR #1332, opened 2021-09-09
Disable the Return to Campus form links

EjPlatzer:
Per the Covid-19 Team, the semester is in full-swing and students have completed the return-to-campus process. Therefore, the return to campus form links on the wellness page can be removed.


bennettforkner (approved):
LGTM!

PR #1333, opened 2021-09-10
Fall 2021 Release 1: Remove Wellness Return to Campus Forms, remove unnecessary React imports

EjPlatzer:
Per instructions from the Covid-19 Team, the Return to Campus forms for the beginning of the Fall 2021 semester are no longer relevant. Instructions to access them have been removed from the wellness page.

Also, since we have upgraded to React 17, importing React as a full namespace in every component is no longer needed - the Babel compiler handles JSX Syntax, so we only need to import hooks and other functions from the React package that are actually used in a component. Consequently, the unused React imports have been removed from our files.


bennettforkner (approved):
LGTM! Thanks for cleaning up these unused imports.

PR #1334, opened 2021-09-16
Fixed: quick people search racing issue

Kevin-Chu58:
Just deleted two lines of extra codes that caused the result racing issue.

### DIFF
--- a/src/components/Header/components/PeopleSearch/index.js
+++ b/src/components/Header/components/PeopleSearch/index.js
@@ -101,8 +101,6 @@ const GordonPeopleSearch = ({
     //but really it's just that its capitalized what the heck
     query = query.toLowerCase();
 
-    let suggestions = await peopleSearch.search(query);
-    setSuggestions(suggestions);
     let results = await peopleSearch.renderResults(query);
     if (time < results.now) {
       setTime(results.now);


EjPlatzer (approved):
LGTM! This is working well now! I was wondering why the quick search was still giving bad results.

russtuck:
Thanks for fixing this!

Kevin-Chu58:
@EjPlatzer @russtuck Though the latest code did fix the racing issue most of the time, but it still occurs seldomly somehow. I don't really know why. Because the logic is correct, the result that isn't the latest still managed to be updated onto the display.

Code: 
image

Console.log:
image

EjPlatzer:
I'm willing to bet this is one of the many subtle issues that arise when you pair asynchronous logic with React's hidden rendering cycle. The logic is definitely correct if `setTime()` and `setSuggestions()` updated the state immediately when invoked. However, that is not what really happens. Instead, every time those functions are invoked, React will queue an update to the state to take effect during the next render cycle (which is arbitrarily determined by React). I suspect that the queries return quickly enough that several state updates get queued at once, but the time state doesn't actually get updated immediately, then only the last one queued chronologically is actually applied. Since we know that the last one queued chronologically is usually the one with the largest result set (because that took the API longer), that could easily result in an earlier query becoming the last state.

It would be interesting to compare the values being passed to those`set` functions against the actual state values on every render. You could do that by console logging the time and suggestion values inside `if (time < results.now) { ... }` and also having console logs in the component itself, like just below where the state variables are declared, to log out the actual state on every render. You'll probably see a lot of extraneous renders, but that should at least reveal if any of the updates queued by the `set` functions never actually become state.

russtuck:
Just brainstorming... 2 ideas...

Since the need is for atomic updating of the info to be displayed, is there
a way to append (time, data) tuples to a queue, and process those tuples at
render time?

Another approach would be to put the logic in the server, so it doesn't
return a result from an older request after returning one from a newer
request.  It wouldn't be perfect, since network problems could still
happen, but I think network problems are rare between our servers and
users, so it might work well.

PR #1339, opened 2021-09-20
Add Mail_Description Tooltip to Profile

bennettforkner:
Mailstops do not make it clear as to where mail will show up. The description shows a more precise location of where mailstops are located.

This pull request adds a '?' next to the mailstop listing on faculty and staff profiles that will show the description of the mailstop when hovered over or clicked (mobile)


EjPlatzer (dismissed):
LGTM!

EjPlatzer (approved):

PR #1342, opened 2021-09-23
Fix Student News Search

bennettforkner:
The search bar on the student news page ([/news](https://360.gordon.edu/news)) does not do anything as of yet.

This PR sets up logic to filter the news items whenever a search string is entered into the search box. It was originally paired with a new API endpoint, but has since been changed to rely only on the UI since it will likely always be a small results set.


src/services/news.js
```diff
@@ -83,6 +83,37 @@ const getNotExpiredFormatted = async () => {
...
+ */
+const getFilteredNews = (unexpiredNews, query) => {
+  const filteredNews = [];
+  unexpiredNews.map((newsitem) => {
+    let queryparts = query.split(' ');
+    for (let i = 0; i < queryparts.length; i++) {
+      let querypart = queryparts[i];
+      if (querypart === '') {
+        continue;
+      }
```
EjPlatzer, line 100: 3 notes:

1. You forgot to lowercase `query`.
2. A `for of` loop is much more readable here.
3. `String.split()` doesn't return empty elements, so you don't need the empty string check here.

```suggestion
    for (let queryPart of query.toLowerCase().split(' ')) {
```

src/services/news.js
```diff
@@ -83,6 +83,37 @@ const getNotExpiredFormatted = async () => {
...
 
+/**
+ * Gets unexpired student news matching a search string
+ * @param {NewsItem[]} unexpiredNews the list of unfiltered news items
+ * @param {String} query the search query
+ * @return {Promise<any>} Student news
+ */
+const getFilteredNews = (unexpiredNews, query) => {
+  const filteredNews = [];
+  unexpiredNews.map((newsitem) => {
```
EjPlatzer, line 94: Rather than creating and pushing to an intermediate array, you can use `Array.filter()` and return an array containing elements from `unexpiredNews` that return true in the code below.
```suggestion
  return unexpiredNews.map((newsitem) => {
```

src/views/News/index.js
```diff
@@ -76,6 +76,7 @@ const StudentNews = (props) => {
   const [loading, setLoading] = useState(true);
   const [categories, setCategories] = useState([]);
   const [news, setNews] = useState([]);
+  const [allNews, setAllNews] = useState([]);
```
EjPlatzer, line 79: I think a ref is more appropriate for this use case, since allNews gets set once when the component is first created, and then never gets changed. Using a state variable implies that this value will change overtime and React should re-render when it does. A ref, on the other hand, doesn't cause a re-render when it changes and it's value persists throughout all renders. This will clarify that allNews is not dynamic state, but rather is being set on load and only read thereafter.

You can switch this to a ref bu updating this line:
 ```suggestion
  const allNewsRef = useRef([]);
```

Then, when you call setAllNews, instead do `allNewsref.current = unexpiredNews`, and when you reference `allNews`, change it to `allNewsRef.current`.

EjPlatzer (changes requested):
This is great! Nice to finally have some search functionality. Still a good number of improvements that could be made, e.g. including older news entries (either an includeExpired toggle like for events, or the ability to set a minimum date?).

I left some comments for ways to clean it up a tiny bit, but overall this is great!

src/services/news.js
bennettforkner: So what about if the querypart = ''

EjPlatzer: That doesn't happen because `split()` doesn't return empty strings.

src/services/news.js
bennettforkner: I'm not adding every element to the new array and I don't want to add NULLs, so can I really use map?

EjPlatzer: Sorry, meant to change that to `unexpiredNews.filter`

src/services/news.js
bennettforkner: So it was returning an empty string earlier and that if case fixed it.

EjPlatzer: It turns out that you're absolutely right. If the final character of `query` is a space, then split returns an empty string, which then becomes a query part and matches anything. This also happens if two spaces in a row appear in `query`.

src/views/News/index.js
bennettforkner: Ah, I was not aware of the distinction. Thank you!

EjPlatzer (approved):
LGTM! A simple but very meaningful upgrade!

PR #1344, opened 2021-09-24
Update wellness contact info on question card

EjPlatzer:
The contact info is duplicated in two places, but I forgot and only changed it on the Health Status card but not the Wellness Question card.

Also, I updated the link styling to retain underline on hover.


bennettforkner (approved):
LGTM. Thanks for catching the second instance.

PR #1345, opened 2021-09-24
Release: Fix quick search, update wellness contact, add mail descriptions, fix news search

EjPlatzer:
This release is to update the wellness contact info. It also contains:
- a fix to quick people search to minimize the errors where earlier search results overwrite later search results. 
- Mail descriptions for FacStaff mailstops
- A working implementation of search on the news page
- Displaying Gordon ID on FacStaff's MyProfile view

Also inlcuded in this release is the migration from `node-sass` to `sass` for rendering scss into css. The `node-sass` package has been deprecated for awhile. We had put off migration because I mistakenly assumed that migrating to the replacement `sass` package would be an involved process. As it turns out, `sass` is a drop-in replacement for our use cases, since we just use the command-line interface to compile our scss files.


bennettforkner (approved):
LGTM -- checked this last night :)

PR #1349, opened 2021-09-27
Removing index.js.orig

bennettforkner:
Removing old file


cpabbot (approved):
LGTM, thanks for cleaning this up! At some point in the future it may be good for us to check how Student News works while offline/pwa.

PR #1350, opened 2021-09-27
Readable Session Codes on Involvements Page

### LINKED ISSUE #1179: Involvements URL encoding of seasons '201909' -> '2019fall'
Also consider if using `/2019/fall` style query parameters are something to pursue. I thought I read something about this style but cannot find anything now.

cpabbot:
E.g. `session=202109` -> `session=2021fall`
Converts back and forth straight from/to URL to avoid complications with how the session code is used elsewhere. This is intended to be a non-invasive change.

Resolves #1179


bennettforkner (approved):
LGTM!

PR #1352, opened 2021-09-29
Remove Unused/Unneeded dependencies

EjPlatzer:
We have a couple of old dependencies that never got removed after we finished using them, as well as one poorly-maintained one that can now be replaced with a trivial custom hook. I have removed these low-hanging fruit with the long-term goal of reducing and modernizing our dependencies further.

Removed dependencies:
- classnames: This had been useful in the past to handle passing complex classname props to components. It was no longer being used and is largely irrelevant with the adoption of scss modules. Once we uograde to MUIv5, it will be even less useful to us. Since it wasn't being used anywhere and it's utility to us is only decreasing, I removed it.
- react-responsive-carousel: This was previously used for the image carousel on the home page. However, at some point before I became a maintainer of the project, we switched to using react-image-gallery. The two have slightly different feature sets but react-image-gallery suits our needs and is slightly more well-maintained. Since react-responsive-carousel isn't being used (and we have no plans to return to using it), I have removed it.
- react-document-title: This small package added a custom component that would update the document title (i.e. the browser tab/page title) when it was mounted. The package never updated to using functional components and was holding back other dependencies - plus, this behavior is now trivial to achieve using hooks. I have replaced it with a custom hook `useDocumentTitle` in the one place we were using it, and removed the package.

I hope to continue slimming down our dependencies in the future - especially reducing the number of date libraries we use. Currently, we depend on Moment.js, Luxon.js, and date-fns.js, all of which achieve effectively the same thing. These are some of our heaviest dependencies, and we need at most one. However, refactoring the other two out is beyond the scope of this pull request.


README.md
```diff
@@ -113,19 +113,15 @@ VS Code users will also see a ruler at the 100 character mark, helping the devel
 
 Links to the homepages of libraries used in this project, listed here for easy reference.
 
-- [Material-UI](https://next.material-ui.com/)
+- [MUI](https://mui.com/)
 
-  Material-UI is a widely-used React implementation of Google's Material Design. It provides a comprehensive toolkit of interface components, along with a set of usability guidelines and best practices for using them.
+  MUI is a widely-used React component library originally based on Google's Material Design standards. It provides a comprehensive toolkit of interface components, along with a set of usability guidelines and best practices for using them.
```
bennettforkner, line 118: Shouldn't we say "Material-UI" somewhere in here so that people know what MUI stands for?

bennettforkner (approved):
LGTM. See minor comment

README.md
bennettforkner: What's the reason for removing it entirely?

EjPlatzer: Actually, they rebranded as of v5, and it no longer stands for `Material-UI`, the package/company is just `MUI` now. They say it stands for `**M**aterials to build **UI**s` now, but the npm package, the website, and the official name are just `MUI` now. Check out [their blog post about it](https://mui.com/blog/material-ui-is-now-mui/) for more details

bennettforkner: I figured that, but I still think that we should put "Materials to Build UIs" in the documentation

EjPlatzer: I disagree just for discoverability reasons. There is no content anywhere under the name 'Materials to Build UIs', they don't even use that name themselves. I think the only occurrence of those words might be in that one blog post. MUI is the name of the package/company.

EjPlatzer: image

image

image

image

image

Their GitHub profile, Twitter profile, and website all use MUI ubiquitously as their name, and don't reference that tag line even once. Notice how they're telling people to replace `Material-UI` with `MUI` when referring to them.

bennettforkner: You are telling me nothing, you have said this already. I clearly do not care what MUI has branded themselves as. I am saying that it would be more clear for our users if we gave the acronym a name because "MUI" means absolutely nothing to anyone who is first joining the team and is exploring our documentation. I don't think they should have to go to the website and explore it just to see what the actual name stands for.

EjPlatzer: I don't see how `Material-UI` was any more descriptive than `MUI`. And they don't have to click through to the website because the next words after `MUI` are a much more informative definition than `Materials to Build UIs`. The point is that `MUI` is no longer an acronym, it's just a name.

bennettforkner: I recognize that. I still think we should put the old name or the more descriptive name in there. It references what it used to be called if nothing else.

PR #1353, opened 2021-09-29
Refactor Availability Message into GordonLimitedAvailability Component

bennettforkner:
Repeated code refactored into a component.

This pull request completes the following:
- Refactor the aforementioned shared message into a component called `<GordonLimitedAvailability/>` and made views reference it (Timesheets, ApartApp)

### DIFF
--- a/src/components/GordonLimitedAvailability/index.js
+++ b/src/components/GordonLimitedAvailability/index.js
@@ -34,11 +34,4 @@ const GordonLimitedAvailability = ({
   );
 };
 
-GordonLimitedAvailability.defaultProps = {
-  backToLink: '',
-  backToLocation: 'Home',
-  availableTo: 'students',
-  pageName: '',
-};
-
 export default GordonLimitedAvailability;
--- a/src/components/GordonLimitedAvailability/index.js
+++ b/src/components/GordonLimitedAvailability/index.js
@@ -1,7 +1,7 @@
 import { Grid, Card, CardContent, Button } from '@material-ui/core';
 import { Link } from 'react-router-dom';
 
-const GordonLimitedAvailability = ({ pageName, backToLocation, backToLink, availableTo }) => {
+const GordonLimitedAvailability = ({ pageName = '', backToLocation = 'Home', backToLink = '', availableTo = 'students' }) => {
   let pageNameFormatted = pageName ? `The ${pageName} page` : 'This page';
   return (
     <Grid container justifyContent="center" spacing="16">
--- a/src/components/GordonLimitedAvailability/index.js
+++ b/src/components/GordonLimitedAvailability/index.js
@@ -20,7 +20,7 @@ const GordonLimitedAvailability = ({
           >
             <br />
             <h1>{pageName + ' Unavailable'}</h1>
-            <h4>{'The ' + pageName + ' page is available for ' + availableTo + ' only.'}</h4>
+            <h4>`The ${pageName} page is available for ${availableTo} only.`</h4>
             <br />
             <br />
             <Button variant="contained" component={Link} to={backToLink}>
--- a/src/components/GordonLimitedAvailability/index.js
+++ b/src/components/GordonLimitedAvailability/index.js
@@ -16,7 +16,7 @@ const GordonLimitedAvailability = ({ pageName = '', backToLocation = 'Home', bac
             <br />
             <h1>{pageName + ' Unavailable'}</h1>
             <h4>
-              {pageNameFormatted} is available for {availableTo} only.
+              {pageName ? `The ${pageName} page` : 'This page'} is available for {availableTo} only.
             </h4>
             <br />
             <br />
--- a/src/components/GordonLimitedAvailability/index.js
+++ b/src/components/GordonLimitedAvailability/index.js
@@ -1,8 +1,12 @@
 import { Grid, Card, CardContent, Button } from '@material-ui/core';
 import { Link } from 'react-router-dom';
 
-const GordonLimitedAvailability = ({ pageName = '', backToLocation = 'Home', backToLink = '', availableTo = 'students' }) => {
-  let pageNameFormatted = pageName ? `The ${pageName} page` : 'This page';
+const GordonLimitedAvailability = ({
+  pageName = '',
+  backToLocation = 'Home',
+  backToLink = '',
+  availableTo = 'students',
+}) => {
   return (
     <Grid container justifyContent="center" spacing="16">
       <Grid item xs={12} md={8}>
--- a/src/components/GordonLimitedAvailability/index.js
+++ b/src/components/GordonLimitedAvailability/index.js
@@ -1,4 +1,5 @@
 import { Grid, Card, CardContent, Button } from '@material-ui/core';
+import { Link } from 'react-router-dom';
 
 const GordonLimitedAvailability = ({
   pageName,
@@ -22,15 +23,8 @@ const GordonLimitedAvailability = ({
             <h4>{'The ' + pageName + ' page is available for ' + availableTo + ' only.'}</h4>
             <br />
             <br />
-            <Button
-              color="primary"
-              backgroundColor="white"
-              variant="outlined"
-              onClick={() => {
-                window.location.pathname = backToLink;
-              }}
-            >
-              Back To {backToLocation}
+            <Button variant="contained" component={Link} to={backToLink}>
+              Back to {backToLocation}
             </Button>
           </CardContent>
         </Card>
--- a/src/components/GordonLimitedAvailability/index.js
+++ b/src/components/GordonLimitedAvailability/index.js
@@ -1,0 +1,42 @@
+import { Grid, Card, CardContent, Button } from '@material-ui/core';
+
+const GordonLimitedAvailability = ({
+  pageName,
+  backToLocation,
+  backToLink,
+  availableTo,
+  ...OtherProps
+}) => {
+  return (
+    <Grid container justifyContent="center" spacing="16">
+      <Grid item xs={12} md={8}>
+        <Card>
+          <CardContent
+            style={{
+              margin: 'auto',
+              textAlign: 'center',
+            }}
+          >
+            <br />
+            <h1>{pageName + ' Unavailable'}</h1>
+            <h4>{'The ' + pageName + ' page is available for ' + availableTo + ' only.'}</h4>
+            <br />
+            <br />
+            <Button
+              color="primary"
+              backgroundColor="white"
+              variant="outlined"
+              onClick={() => {
+                window.location.pathname = backToLink;
+              }}
+            >
+              Back To {backToLocation}
+            </Button>
+          </CardContent>
+        </Card>
+      </Grid>
+    </Grid>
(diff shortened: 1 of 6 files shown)


src/components/GordonLimitedAvailability/index.js
```diff
@@ -0,0 +1,42 @@
...
+            <Button
+              color="primary"
+              backgroundColor="white"
+              variant="outlined"
+              onClick={() => {
+                window.location.pathname = backToLink;
+              }}
+            >
+              Back To {backToLocation}
+            </Button>
```
EjPlatzer, line 34: Setting `window.location.pathname` causes the entire site to reload and circumvents React Router. Use the `Link` component instead. See the [EventsAttended Component](https://github.com/gordon-cs/gordon-360-ui/blob/344eb36083b0b15db1fe5fdecdf7e5c78cab576f/src/views/EventsAttended/index.js) lines 43-50 for an example.

src/components/Profile/components/MembershipsList/index.js
```diff
@@ -82,6 +82,18 @@ const MembershipsList = ({ user, myProf, createSnackbar }) => {
...
+    ? myProf && (
+        <Grid container justifyContent="center">
+          <Link className="gc360_link" to="/transcript">
+            <Button variant="contained" className={styles.memberships_card_content_button}>
+              Experience Transcript
+            </Button>
+          </Link>
+        </Grid>
+      )
+    : null;
```
EjPlatzer, line 95: Mixing the ternary operator with `&&` is syntactically odd and can produce unexpected results. Include `myProf` in the boolean for the ternary instead:
```suggestion
  const transcriptButton = myProf && !PersonType?.includes('fac')
    ?  (
        <Grid container justifyContent="center">
          <Link className="gc360_link" to="/transcript">
            <Button variant="contained" className={styles.memberships_card_content_button}>
              Experience Transcript
            </Button>
          </Link>
        </Grid>
      )
    : null;
```

src/views/CoCurricularTranscript/index.js
```diff
@@ -41,6 +42,8 @@ export default class Transcript extends Component {
       /* Retrieve data from server */
       const profile = await user.getProfileInfo();
 
+      this.state.profile = profile;
```
EjPlatzer, line 45: This is bad practice, and will break if CoCurricularTranscript ever re-renders. You need to instead use setState:
```suggestion
      this.setState({ profile });
```

src/views/CoCurricularTranscript/index.js
```diff
@@ -242,7 +245,8 @@ export default class Transcript extends Component {
   };
 
   render() {
-    if (this.props.authentication) {
+    console.log(this.state.profile.PersonType);
```
EjPlatzer, line 248: Please remove this console log before merging.

src/views/CoCurricularTranscript/index.js
```diff
@@ -348,6 +354,15 @@ export default class Transcript extends Component {
...
       );
+    } else if (this.state.profile.PersonType?.includes('fac')) {
+      return (
+        <GordonLimitedAvailability
+          pageName="Experience Transcript"
+          backToLocation="My Profile"
+          backToLink="/myprofile"
+          availableTo="students and alumni"
+        />
+      );
```
EjPlatzer, line 365: I would love to see this cleaned up so we only check `PersonType` once, checking this above, before the main content.

EjPlatzer (changes requested):
This looks good! I left some notes about things that need to be fixed or could be improved.

src/components/Profile/components/MembershipsList/index.js
bennettforkner: done

src/views/CoCurricularTranscript/index.js
bennettforkner: done

src/views/CoCurricularTranscript/index.js
bennettforkner: thaanks!

src/views/CoCurricularTranscript/index.js
bennettforkner: done.

src/components/GordonLimitedAvailability/index.js
bennettforkner: commit incoming

src/components/GordonLimitedAvailability/index.js
```diff
@@ -0,0 +1,36 @@
+import { Grid, Card, CardContent, Button } from '@material-ui/core';
+import { Link } from 'react-router-dom';
+
+const GordonLimitedAvailability = ({
+  pageName,
+  backToLocation,
+  backToLink,
+  availableTo,
+  ...OtherProps
```
EjPlatzer, line 9: It looks like there's no purpose for the `OtherProps` here. It would only be useful if you were going to pass unspecified props onto a child component, but since you aren't this will not do anything.

src/views/CoCurricularTranscript/index.js
```diff
@@ -348,6 +357,15 @@ export default class Transcript extends Component {
...
       );
+    } else if (this.userIsFacStaff()) {
+      return (
+        <GordonLimitedAvailability
+          pageName="Experience Transcript"
+          backToLocation="My Profile"
+          backToLink="/myprofile"
+          availableTo="students and alumni"
+        />
+      );
```
EjPlatzer, line 368: Technically, you're still checking this twice, since you invoke the function twice. What I meant was move this return path above the other one, so you check if the user is facstaff and, if so, show them the Unavailable page, otherwise show them the Experience Transcript.

src/views/Timesheets/index.js
```diff
@@ -595,38 +596,12 @@ const Timesheets = (props) => {
...
-                    Back To Home
-                  </Button>
-                </CardContent>
-              </Card>
-            </Grid>
-          </Grid>
+          <GordonLimitedAvailability
+            pageName="TimeSheets"
+            backToLocation="Home"
+            backToLink=""
```
EjPlatzer, line 602: These would make sensible default values, so that if we didn't want to redirect the user to a specific location, we could just not pass these props and the component would send them to the Home page.

src/components/GordonLimitedAvailability/index.js
```diff
@@ -0,0 +1,36 @@
...
+        <Card>
+          <CardContent
+            style={{
+              margin: 'auto',
+              textAlign: 'center',
+            }}
+          >
+            <br />
+            <h1>{pageName + ' Unavailable'}</h1>
+            <h4>{'The ' + pageName + ' page is available for ' + availableTo + ' only.'}</h4>
```
EjPlatzer, line 23: This would be nicer as a template literal:
```suggestion
            <h4>`The ${pageName} page is available for ${availableTo} only.`</h4>
```

EjPlatzer (changes requested):
A couple more suggestions

src/views/CoCurricularTranscript/index.js
bennettforkner: Ah, yes. That makes sense!

src/views/Timesheets/index.js
bennettforkner: Yes, okay. I was not sure how to do that initially but I will figure it out :)

src/components/GordonLimitedAvailability/index.js
bennettforkner: Removed

src/views/CoCurricularTranscript/index.js
bennettforkner: Done.

src/views/Timesheets/index.js
bennettforkner: Done!

bennettforkner (commented):
Made changes as requested.

src/components/GordonLimitedAvailability/index.js
```diff
@@ -0,0 +1,40 @@
...
+          <CardContent
+            style={{
+              margin: 'auto',
+              textAlign: 'center',
+            }}
+          >
+            <br />
+            <h1>{pageName + ' Unavailable'}</h1>
+            <h4>
+              {pageNameFormatted} is available for {availableTo} only.
```
EjPlatzer, line 19: Since this is JSX, you can just put the ternary operator in this expression directly, rather than needing a variable for it.
```suggestion
  let pageNameFormatted = pageName ? `The ${pageName} page` : 'This page';
  return (
    <Grid container justifyContent="center" spacing="16">
      <Grid item xs={12} md={8}>
        <Card>
          <CardContent
            style={{
              margin: 'auto',
              textAlign: 'center',
            }}
          >
            <br />
            <h1>{pageName + ' Unavailable'}</h1>
            <h4>
              {pageName ? `The ${pageName} page` : 'This page'} is available for {availableTo} only.
```

src/components/Profile/components/MembershipsList/index.js
EjPlatzer: Did this change get reverted somehow?

src/views/CoCurricularTranscript/index.js
```diff
@@ -241,8 +244,21 @@ export default class Transcript extends Component {
     return minorsString.substr(0, minorsString.length - 2);
   };
 
+  userIsFacStaff = () => {
+    return this.state.profile.PersonType?.includes('fac');
+  };
+
   render() {
-    if (this.props.authentication) {
+    if (this.userIsFacStaff()) {
```
EjPlatzer, line 252: Since it's only called once now, you could remove the function and just put the boolean here directly.

src/components/GordonLimitedAvailability/index.js
```diff
@@ -0,0 +1,40 @@
...
+    </Grid>
+  );
+};
+
+GordonLimitedAvailability.defaultProps = {
+  backToLink: '',
+  backToLocation: 'Home',
+  availableTo: 'students',
+  pageName: '',
+};
```
EjPlatzer, line 38: The simpler (and imo more readable) way to do defaults in JS is to assign them in the function signature.

src/components/GordonLimitedAvailability/index.js
```diff
@@ -0,0 +1,40 @@
+import { Grid, Card, CardContent, Button } from '@material-ui/core';
+import { Link } from 'react-router-dom';
+
+const GordonLimitedAvailability = ({ pageName, backToLocation, backToLink, availableTo }) => {
```
EjPlatzer, line 4: You can add defaults this way:
```suggestion
const GordonLimitedAvailability = ({ pageName = '', backToLocation = 'Home', backToLink = '', availableTo = 'students' }) => {
```

EjPlatzer (changes requested):
Just a few minor comments/recommendations.

src/components/GordonLimitedAvailability/index.js
bennettforkner: Eh, I liked it more taken out. I will make the changes though

PR #1356, opened 2021-10-01
Removed On/Off Indicator if Dorm Exists

### LINKED ISSUE #1260: Remove On/Off Campus Indicator if dorm exists
Very redundant to the user to show 'on campus' indicator if the dorm is indicated right below. Consider if we can just show the dorm, or 'off campus' if the dorm does not exist. We could even change the language of "Dormitory" to "Campus Location" if we think that is a more clear indicator.

image

EDIT:
A use case I did not originally think of is students that are studying abroad. 360 shows them with their dorm but also "Away" as follows
image

I suggest perhaps just putting Dorm (On/Off Campus Status) e.g. "Nyland (Away)". Away would also be better described as "abroad" or otherwise if this is possible to improve.

silasw2:
Pretty self explanatory. If there is a value for the dorm or hall of the student, it will not show that they are on campus as it is redundant to have both. 

Here is what it looked like before:
image

Here is what it looks like now:
image

Off Campus indicator still works:
image

Closes #1260

### DIFF
--- a/src/components/Profile/components/OfficeInfoList/index.js
+++ b/src/components/Profile/components/OfficeInfoList/index.js
@@ -63,7 +63,7 @@ const OfficeInfoList = ({
     />
   ) : null;
 
-  const updateProfile =
+  const updateOfficeInfo =
     myProf && PersonType?.includes('fac') ? (
       <Typography align="left" className={styles.note}>
         NOTE: Update your office info
@@ -92,7 +92,7 @@ const OfficeInfoList = ({
             {mailstop}
             {officePhone}
             {officeHours}
-            {updateProfile}
+            {updateOfficeInfo}
           </List>
         </CardContent>
       </Card>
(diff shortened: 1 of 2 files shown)


bennettforkner (approved):
LGTM. Good job putting your time in :)

EjPlatzer:
This is a small thing, but it now seems odd to have a field labelled `On/Off Campus` if it will only ever display the value `Off Campus`. Perhaps we could combine the fields so that the `Dormitory` label shows the student's dorm if one exists, or `Off Campus` otherwise?

silasw2:
> This is a small thing, but it now seems odd to have a field labelled `On/Off Campus` if it will only ever display the value `Off Campus`. Perhaps we could combine the fields so that the `Dormitory` label shows the student's dorm if one exists, or `Off Campus` otherwise?

Yes, I'll get on that now!

bennettforkner:
> This is a small thing, but it now seems odd to have a field labelled `On/Off Campus` if it will only ever display the value `Off Campus`. Perhaps we could combine the fields so that the `Dormitory` label shows the student's dorm if one exists, or `Off Campus` otherwise?

Good call @EjPlatzer

PR #1357, opened 2021-10-01
Added: if statement to check whether clifton strength is null

Kevin-Chu58:
This PR added:

- If statement to check whether Clifton strength is null

if the first element in `cliftonStrengths.categories` is  not `null`, then set `profile.cliftonStrengths` to the `cliftonStrengths` from database.

### DIFF
--- a/src/services/user.js
+++ b/src/services/user.js
@@ -557,7 +557,14 @@ async function setAdvisors(profile) {
 
 async function setCliftonStrengths(profile) {
   const cliftonStrengths = await getCliftonStrengths(profile.AD_Username);
-  profile.CliftonStrengths = cliftonStrengths;
+  try {
+    if (cliftonStrengths.Categories[0] !== null) {
+      profile.CliftonStrengths = cliftonStrengths;
+    }
+  }
+  catch (e) {
+    console.log("Clifton strength error:", e);
+  }
 }
 
 async function setMobilePhoneNumber(value) {


EjPlatzer (changes requested):
Thanks for working on this Kevin! It's nice to see that we're no longer presenting an empty list of Clifton Strengths anymore.

However, I think there's another part to this issue, which is cleaning up the code for getting Clifton Strengths. The current state is that whenever we try to get strengths for someone who doesn't have any, we ignore all errors and console log them. This clutters up the developer tools a LOT for people who don't have strengths in the database:
Screen Shot 2021-10-03 at 7 28 00 PM
(That's just a small sampling).

The right way to fix this is to edit the code in `getCliftonStrengths` so that it doesn't try to process a cliftonStrengths object that is null. That is, it would try to fetch cliftonStrengths, and check the returned object to see whether it's null. If so, then just return null. If not, then process and return the strengths object.

src/services/user.js
```diff
@@ -517,9 +517,9 @@ const getAdvisors = async (username) => {
 };
 
 const getCliftonStrengths = async (username) => {
-  try {
-    let cliftonStrengths = await http.get(`profiles/clifton/${username}/`);
+  let cliftonStrengths = await http.get(`profiles/clifton/${username}/`);
 
+  if (cliftonStrengths.Strengths !== null && cliftonStrengths.Strengths[0] !== null) {
```
EjPlatzer, line 522: You could simplify this check by using [`optional chaining`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and checking for a ['truthy' value](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) instead of just `null`:
```suggestion
  if (cliftonStrengths?.Strengths?.[0]) {
```

PR #1359, opened 2021-10-02
Combined On/Off Campus and Dormitory into one ProfileInfoListItem

### LINKED ISSUE #1260: Remove On/Off Campus Indicator if dorm exists

silasw2:
Pretty self explanatory, this is a follow up to pull request #1356 to implement Evan's suggestions. Code works in the same way so review that pull request to see the changes. This PR is just for code clarity.

Truly closes #1260

### DIFF
--- a/src/components/Profile/components/PersonalInfoList/index.js
+++ b/src/components/Profile/components/PersonalInfoList/index.js
@@ -281,16 +281,6 @@ const PersonalInfoList = ({
       />
     ) : null;
 
-  const onOffCampus =
-    isStudent && OnOffCampus && !(BuildingDescription || Hall) ? (
-      <ProfileInfoListItem
-        title="On/Off Campus:"
-        contentText={OnOffCampus}
-        private={isCampusLocationPrivate}
-        myProf={myProf}
-      />
-    ) : null;
-
   const mail =
     isStudent && Mail_Location ? (
       <>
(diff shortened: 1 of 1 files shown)


src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -336,9 +326,16 @@ const PersonalInfoList = ({
       </>
     ) : null;
 
-  const dormInfo =
-    isStudent && (BuildingDescription || Hall) ? (
+    const campusDormInfo =
+    isStudent && OnOffCampus && !(BuildingDescription || Hall) ? (
       <ProfileInfoListItem
+        title="On/Off Campus:"
```
EjPlatzer, line 332: Can we change this to `Dormitory:` so that it's consistent with the hall value?

PR #1360, opened 2021-10-03
Added: description of group contacts and group advisors in involvement profile

Kevin-Chu58:
This PR shows:

- descriptions for every group contact and advisors in involvement profile
- If `members[i].Description === ''`, use `members[i].ParticipationDescription`

Examples:
_I used the staff-test account in 360train to add myself as an advisor of activity to check whether the code works correctly

- When the description (title) is not empty:
image
- When the description (title) is empty:
image
image

**However, I think there may be a better way to write these codes, but I cannot find other methods** 
**This demo data will be removed after anyone approves this PR**


EjPlatzer (commented):
This is a great first attempt! However, the fact that we have to fetch members and check them all against the advisors/contacts is really not ideal. There is a slightly cleaner way to do the iteration, but it still isn't ideal.

That indicates to me that we should rethink things in the database/API instead. It would be trivial in the backend to add the `Description` attribute to the advisors/contacts when they're fetched. That would be a much better and simpler way to implement this.

EjPlatzer:
> @EjPlatzer I checked out the API and stored procedure `EMAILS_PER_ACT_CDE` in 360train.
> 
> I modified that procedure but haven't updated it yet. The only code I added was `CONCAT(mem.ACT_CDE,mem.COMMENT_TXT) as Title`. Do you think this is the best solution when I pass the parameters to the front-end and use split to get the right data?

Thanks for looking into this! This is a good approach, but there are a couple of issues with it. The first is that you would need to change all of the stored procedures that are used to access emails per involvement code (if you check https://github.com/gordon-cs/gordon-360-api/blob/develop/Gordon360/Services/EmailService.cs you can see there are actually 4 different ones). That's because in order for it to work, you have to change the `EmailViewModel` in the API at the same time as you change the stored procedures, otherwise the ViewModel will break. The other issue is that, because the ViewModel and the Stored Procedures are then tightly coupled, you'd need to also change the stored procedures in the Prod database at the same time as you push the release with these changes, which is a really fragile deployment process.

The cleaner way to do this is to make use of Entity Framework, so instead of using stored procedures, you query the memberships table directly from the C# code (and Entity framework takes care of turning the C# into SQL). There's a decent example of using the membershipRepository for querying the memberships table in the [MembershipRequestService.ApproveMembership](https://github.com/gordon-cs/gordon-360-api/blob/a4aa719a4f851ba9aae594effe9d5d2cb837d875/Gordon360/Services/MembershipRequestService.cs#L72) method.

The idea would be to refactor all the methods in EmailService to use the Membership Repository, but you could start with just he group admin and advisor ones.

Kevin-Chu58:
This the new display in the past 360 involvement. The frontend now checks whether `Description` is null or empty, if not, the UI shows that person's description in `dbo.MEMBERSHIP`.
demo1
demo2

src/views/InvolvementProfile/components/ContactListItem/index.js
```diff
@@ -43,7 +43,11 @@ const ContactListItem = ({ contact }) => {
         </Avatar>
       </ListItemAvatar>
       <Link href={`/profile/${contact.Email.split('@')[0]}`}>
-        <ListItemText primary={`${contact.FirstName} ${contact.LastName}`} />
+        <ListItemText
+          primary={`${contact.FirstName} ${contact.LastName} ${
+            contact.Description?.trim() ? '- ' + contact.Description : ''
+          }`}
```
EjPlatzer, line 49: Do you think it would look nicer as secondary list item text?
```suggestion
          primary={`${contact.FirstName} ${contact.LastName}`}
          secondary={contact?.Description}
```

EjPlatzer (commented):
One question about the visual aspect

EjPlatzer (approved):
LGTM! Thanks for seeing this through several iterations!

PR #1365, opened 2021-10-05
Refactor Clifton Data into Separate File

bennettforkner:
This PR is just for cleaning up the code so that the Clifton Strengths objects are not cluttering the user service code.


EjPlatzer (approved):
Beautiful! Thank you!

PR #1367, opened 2021-10-05
Add User context to store info of currently logged in user

EjPlatzer:
One of the most common operations in our components is to get info about the currently logged-in user. This includes person type, name, email, and many more things that are relevant in many (maybe even most) of our components.

Tied to this data is the authentication state - whether the user is currently logged-in. If they are authenticated, then we know some things about, and if they aren't logged-in then we know next-to-nothing.

Managing authentication state has previously been done by passing it as a prop all the way down from our top-level App.js component. This is an incredibly messy solution because it means that every component that needs auth state - as well as all of their parent components, needs to pass through the authentication prop. This is especially unfortunate because it results in dozens of components that accept and pass on the authentication prop without ever needing it themselves, effectively bloating that component with complexity unrelated to it's behavior/responsibilities.

Furthermore, the situation with user info is even worse, because currently every component that needs user info fetches it independently from the API, resulting in tens or possibly even hundreds more API calls than ought to be necessary.

The solution to both of these problems is to store auth/user state in one React Context object, and make it available via React's Context API to every component across the site - without the need for passing props or redundant fetching.

Consequently, I have created the UserContext, and it's useUser and useUpdateUser hooks. UserContext - like our other pre-existing context objects, is responsible for one piece of site-wide data. In this case, that data is info about the currently logged-in user. If the user is authenticated, UserContext stores the user's profile info. If the user is not authenticated, UserContext is null. In this way, we can replace the multitudinous calls to `userService.getProfileInfo()` and the redundant `authentication` prop with a simple `useUser()` invocation in any functional component.

To handle updates to the user/auth state, UserContext also provides the `useUpdateUser` hook. This hook returns a `updateUser` function that, when called, forces the UserContext to update. This hook is only needed when the user signs in/out at the moment. In the future, it might be useful when changing the user's data. It replaces the `onAuthChange` function from App.js that was passed as a prop to every component that signed the user in/out, as well as every intervening component. Now, any component that updates the user's state can call `useUpdateUser` to get the update function and update the user across every component when needed.

This pull request introduces the new Context/hooks, and gives an example implementation of `useUser` in the GordonNavAvatar. Updating every relevant component to use the new hook is beyond this scope, and can be done incrementally. The simplest way to achieve this would be to find every component that calls `userService.getProfileInfo()` (with no parameters, which is different from passing the username of a user other than the current one), and every component that depends on the `authentication` prop, and refactoring them to use the new hook. Once all components use the new hook, the authentication prop can be cleaned out of App.js and the component tree


bennettforkner (approved):
LGTM. Thanks for taking the time to refactor this so we can avoid having duplicated code and multitudinous API requests.

PR #1368, opened 2021-10-06
Disable schedule on public Profile

EjPlatzer:
We have been informed that showing a student's schedule to FacStaff without the student's consent is a potential privacy violation. We will revisit the privacy surrounding schedules when it is rewritten/fixed, but in the meantime the simplest solution is to disable the schedule when showing a student's PublicProfile.


bennettforkner:
> LGTM. I'm not sure I understand how it is a privacy issue to show schedule to facstaff but not an issue to show to other students. I assume that is the case since you included it in your logic though.

Ah, I did not see the `!` or the fact that it was from the profile. All set then!

PR #1370, opened 2021-10-06
Minor News UI Adjustments

cpabbot:
- made font sizes consistent with events and people search
- added underline for SNID link
Note: this #fff was hard coded because I could not figure out a simple way to set it to inherit the white color (probably would need to rearrange the elements), which doesn't matter right now but will have to be corrected in a future themed world.

On hover:
image

### DIFF
--- a/src/views/PeopleSearch/index.js
+++ b/src/views/PeopleSearch/index.js
@@ -1,50 +1,49 @@
-import { Component } from 'react';
-import GordonUnauthorized from 'components/GordonUnauthorized';
-import GordonOffline from 'components/GordonOffline';
 import {
   Button,
   Card,
+  CardActions,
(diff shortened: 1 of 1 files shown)


src/views/News/components/NewsItem/NewsItem.module.scss
```diff
@@ -9,6 +9,10 @@
   .news_authorProfileLink {
     color: inherit;
   }
+  .news_authorProfileLink:hover {
+    text-decoration: underline;
+    text-decoration-color: #fff;
```
EjPlatzer, line 14: At the very least, this oughta use the white variables from `_vars.scss`. Inheriting would be nice but we can at least use scss variables that can get overridden.

Issue #1406, opened 2021-11-13
Migrate to TypeScript


EjPlatzer:
# What and Why?

This issue proposes and tracks a migration of the 360 UI from JavaScript to TypeScript. TypeScript is a superset of JavaScript that adds static typing syntax. TypeScript compiles to JavaScript at build time, after running type-correctness checks that catch many insidious JavaScript bugs. This provides type-safety, more helpful Intellisense, and better self-documenting code. For more information, visit the [TypeScript website](https://www.typescriptlang.org/) and read about [Why TypeScript](https://www.typescriptlang.org/why-create-typescript).

The main goals of this migration are to:
- prevent, discover, and eliminate bugs by:
  - revealing TypeErrors at build time (e.g. calling `.length` on an object with no `length` property)
  - documenting data types in accordance with the API types, catching inconsistencies
  - ensuring that changes in one place result in the correct changes in all dependent code
- increase the capabilities of Intellisense, type hinting, and other Developer Experience tools
- reduce the boilerplate of JSDoc typing, allowing JSDoc comments to specify only important context/additional info

# How?
Incrementally. TypeScript can be enabled on a per-file (or even more granular) basis, which enables us to migrate only the most important parts of the app first, and immediately reap the benefits of doing so. As such, I have divided the migration effort into roughly 5 phases. The phases are split by conceptual piece of the app, and are not equal in breadth or complexity.

## Phases

I propose to migrate to TypeScript in these 5 general phases, ordered according to what will give the greatest benefit for the least effort:
- [ ] Phase 1: Migrate all services, along with necessary dependencies
- [ ] Phase 2: Migrate Contexts and hooks
- [ ] Phase 3: Migrate `src/components` shared components
- [ ] Phase 4: Migrate PWA code in `public` folder
- [ ] Phase 5: Migrate `src/views` single-use components

### Phase 1

The services cover 90% of the data types in use throughout our app, since they are solely responsible for fetching data from the API and other external data sources. By getting these data types statically determined and clearly documented, we increase the typed-ness of the rest of the app markedly for very little effort. TypeScript will be able to infer the types of most state, props, etc. based on the data passed in. 

### Phase 2

Contexts and hooks also affect data flow and they are relatively more complex than the rest of the app. However, they contain very little code, so correctly typing them will be not much work but can improve Developer Experience drastically, especially for less-experienced developers.

### Phase 3

[...]

PR #1407, opened 2021-11-13
TypeScript Migration Phase 1: Services

EjPlatzer:
This pull request migrates our services in `src/services` from JavaScript to TypeScript, completing Phase 1 of the TypeScript migration. For the context and motivation of the migration, see issue #1406.

Migrating our services to TypeScript enabled me to discover and fix several bugs, reduce the size of the code by nearly 2,000 lines without reducing functionality, and make the services easier to use, maintain, and extend in the future.

## JSDoc Changes

In so far as they add type info to JavaScript code, TypeScript and JSDoc fulfill the same purpose. However, they are not exactly the same. JSDoc also lends the ability to add extra information beyond types that will be shown in Intellisense and be visible to future maintainers. Conversely, TypeScript checks the correctness of type annotations, rather than parroting whatever type information it is given as JSDoc does. Therefore, they are actually complimentary tools in some circumstances.

Where a JSDoc comment existed purely to add type info, I have removed the comment in favor of letting the TypeScript types speak for themselves. This was, I found, the vast majority of JSDoc comments in our codebase. However, there are some comments that add helpful/crucial information beyond the types. These latter comments I preserved, and I think that by pruning the boilerplate comments I have helped draw attention to the important info in the remaining comments.

Many, many of the types given in JSDoc comments were either not defined or simply incorrect. As mentioned above, JSDoc has no issue with this, and will show these types in Intellisense and make suggestions based on them. This has the potential to actively increase the incidence of bugs in our codebase. When converting these incorrect/undefined types to TypeScript, the TypeScript compiler forced me to fix/define them. This revealed several bugs, increased the clarity and transparency of dataflow, and corrected an abundance of misleading type hinting.


src/services/http.ts
```diff
@@ -47,34 +46,17 @@ const makeHeaders = (headerOptions) => {
...
-  });
-
 /**
  * Parse an HTTP response
  *
- * @param {Response} res HTTP response
- * @returns {Promise.<Any>} Resolves with response body; rejects on non-2xx
- * response code
+ * @param res The response to parse
+ * @returns The parsed response
```
russtuck: Is it clear from broader context, or perhaps not even true, that the promise "rejects on a non-2xx response code"?

russtuck:
I'm making this required reading for CPS 353.  (All the conversation, and at least a good sample of the code changes.  Reading every line of code is NOT required for the class.)

src/services/http.ts
EjPlatzer: No, it's not. I'm not sure why I changed that comment, since it's helpful context for this function. I'll restore it.

cpabbot (commented):
Didn't go through all changes and don't have a lot of experience with TypeScript, but so far everything looks pretty clean and straightforward. Nice work refactoring to simplify a lot of complexity and clarify naming. Won't hit approve yet until code can be fully examined and understood, whether by myself or someone else.

PR #1410, opened 2021-11-15
Temporarily hard code Slider images while WWW is down

EjPlatzer:
Our home page slider normally fetches images for display from `www.gordon.edu`. The WWW site is currently in the process of being restored, so this shims in hard-coded images hosted in Sharepoint for the time being.

**When reviewing, please run the UI locally and ensure you can see the images on the homepage. There may be some permissions quirks with Sharepoint so I want to make sure the images are visible to anyone.**


cpabbot (changes requested):
Access to sharepoint images appears to be denied for both tested users (my personal account and student test). Can the permissions be adjusted so anyone can view? Or does this create permissions complications with CORS or something

EjPlatzer:
> Access to sharepoint images appears to be denied for both tested users (my personal account and student test). Can the permissions be adjusted so anyone can view? Or does this create permissions complications with CORS or something

It is possible to give broad access to view the files within Sharepoint, but that's a different link with different permissions. The links I ended up using are to the 'original file' for the images which is just the raw image data rather than a whole sharepoint item. I was hoping the permissions would transfer to the originals but apparently it does not.

Sadly, I think this means we'll need a different hosting solution. I'll look into an alternative tomorrow.

bennettforkner (dismissed):
LGTM, but why arent we hosting these on our site locally?

EjPlatzer:
> LGTM, but why arent we hosting these on our site locally?

Do you mean hosting them within IIS? I honestly hadn't thought of that. Once RDP is working, that'll be trivially easy.

bennettforkner (approved):
LGTM, thanks!

PR #1418, opened 2021-11-29
Fetch Slider images from DB again

EjPlatzer:
We have restructured how slider images are stored and retrieved. The image metadata is now stored in the CCT database (rather than the now obsolete WWW database). The images themselves are stored in the API filesystem. Currently, the image files and metadata have to be manually managed by ISG, but it should be simple enough to create a page for site admins to manage them.


cpabbot (approved):
LGTM, tested on my personal device and appears to pull images correctly.

PR #1421, opened 2021-12-02
Realign develop master

EjPlatzer:
This pull request merges master back into develop, realigning the history of the branches. I also took the liberty of centralizing the settings that differed between train and prod so that they are both environment variables (API URL and Font URL). Every reference to these settings now references the environment variables in .env.[ENVIRONMENT]. This will make it easy to notice which environment is pointed to, and also to keep environments consistent.

The next release will need to be extra careful that the variables in `.env.production` are set to `@PROD`, and not `@TRAIN`.


bennettforkner (approved):
LGTM

PR #1422, opened 2021-12-02
Upgrade dependencies

EjPlatzer:
This pull request applies non-breaking upgrades to our dependencies. We had 17 dependencies in need of minor and patch versions. These upgrades had no breaking changes at all (I checked the release notes to ensure they were following semver and weren't releasing minor version upgrades with breaking changes).

I also upgraded the following dependencies to a new major version, where the new major version didn't include any breaking changes that affected us:
- react-device-detect
- eslint-plugin-jsdoc
- husky
- lint-staged
- stylelint
- stylelint-config-recommended-scss

The Stylelint upgrade didn't include breaking changes, but did add/fix rules that detected issues/bad practice in our scss files. I fixed the small number of warnings.


bennettforkner (approved):
LGTM!

PR #1423, opened 2021-12-02
Show Deciding when no majors, and hide for alum without majors

EjPlatzer:
The official term for students who haven't picked a major yet is 'Deciding', not 'Undecided'. Also, alumni who have no majors are not 'Deciding', they just don't have majors listed in our system. Hence, we only show the majors list item if the profile belongs to a student, or to an alum who has majors.


bennettforkner (approved):
LGTM!

PR #1428, opened 2022-01-06
Wellness: Add StatusDescription

EjPlatzer:
This PR is the counterpart to gordon-cs/gordon-360-api#674. It shows the StatusDescription on the user's HealthStatus, if there is a StatusDescription to show.

It also refactors the HealthStatus scss module to reduce code duplication and simplify it greatly. By using CSS custom properties, we can declare the styles for the component only once, and override the dynamic properties (in this case, just the background and text color) depending on which color of status is showing. This is a much cleaner alternative to the previous approach of redeclaring the same properties for each color, only changing the background and text color (and changing it in several places). This also helps ensure consistency across the styles.

### DIFF
--- a/src/services/wellness.js
+++ b/src/services/wellness.js
@@ -33,6 +33,7 @@ export const StatusColors = {
  * @property {StatusColor} Status The user's status
  * @property {Date} Created when the status was created
  * @property {boolean} IsValid whether the status has expired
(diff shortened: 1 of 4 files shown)


josh-peters-2357 (approved):
LGTM

PR #1437, opened 2022-01-10
Fix Service Worker loading env vars

EjPlatzer:
Our service worker - which is responsible for caching data for the Progressive Web App - previously had hard-coded paths for the api url and the font source, which differ between Train and Production. Consequently, these paths had to be manually updated and corrected when doing releases, which created lots of room for error. Since the errors dealt with the cache, it made them even harder to track down.

In #1421, I centralized our environment settings to all be based on the values in `.env*`. However, this turned out to be an issue for the service worker scripts. Normally, Create React App takes care of providing environment variables at build time (or runtime in development). Essentially, Create React App has configured Webpack to extract the variables from the relevant `.env*` file and replace all instances of `process.env.[VAR]` with the correct variable. However, the service worker scripts are in the `public` folder, which is not touched by our build process because these files are intended to be served as-is from the web server. This means that the environment variables don't get processed in the files.

The solution is to move `pwa.js` from `public/` to `src/` where it is part of the normal build process. Then, rather than being loaded at runtime via `public/index.html`, it is called from `src/index.js` to register the service-worker while the application is running. This way, `pwa.js` is processed during the regular build and has access to the environment variables. It can then pass those variables as url params to the service-worker script when it registers them.

The end result is that our service-worker scripts don't have to undergo major changes, but they can still make use of the centralized environment variables like the rest of the application.


russtuck:
Thank you for tracking this down, fixing it, and explaining it so well!

bennettforkner (approved):
LGTM.

PR #1432, opened 2022-01-06
Update Enrollment Check-In for 2021SP Semester

EjPlatzer:
Addresses the following issues with the Enrollment Check-in process:
- Rename from academic checkin to Enrollment Check-In
- Reword instructions for registering for classes given current conditions
- Don't display instructions for clearing holds when there are no holds.

Fixes #1430 #1431

PR #1440, opened 2022-01-20
Hotfix: Enrollment Check in - Fix major holds

EjPlatzer:
It turns out that #1432 broke the holds text for major holds in the enrollment check-in process. I had intended to simplify the `displayMajorHolds` method using an array instead of multiple variables to store the needed hold text. However, displaying the resulting list of holds text using `Array.protoype.join()` did not work as expected because the elements of the array were not string but rather `react.element`s. This meant that when displaying holds, they would be displayed as `[object Object]` to the end user. #1439 fixed this for single holds (which I thought at the time was the only issue), but not for multiple holds.

This hotfix has re-structured the way hold text is displayed. Rather than calling a function that returns the hold text, we now have a React component which is displayed only when needed. The component determines which hold texts to show internally, and so doesn't have to concatenate several hold text elements together.

The hotfix has been tested thoroughly in Train.


bennettforkner (approved):
LGTM!

PR #1447, opened 2022-03-03
Add space before selection date

EjPlatzer:


bennettforkner (approved):
Thanks for all your hard work to go through this file and really gut it for clearer messaging. This change is going to be so helpful in users' understanding of the details of the Apartment Application!

PR #1455, opened 2022-03-22
Fix Apart App (and user) services

EjPlatzer:
This fixes in issue in housing.ts and user.ts where optional username params were still appending a trailing slash when the optional param was empty, resulting in paths that ended in `//`. The simple fix ought to be just removing the trailing slash, but apparently that breaks ASP.NET's routing. We should try removing trailing slashes again once the API is using ASP.NET Core.


bennettforkner (approved):
LGTM

PR #1456, opened 2022-03-22
Fix People Search class filter

EjPlatzer:
People Search was broken since the switch to TypeScript because the class filter defaults to '', rather than null or a meaningful variable.


bennettforkner (approved):
SFFB!

PR #1457, opened 2022-03-23
Remove Static Height for Health Box

bennettforkner:
The wellness status module does not show the "Status: ..." message if your screen is too wide (too wide being most computer monitor sizes). This is due to a static height being set for the colored box which is not tall enough to show the message.

My proposal is that we simply remove the css for the height and let the block size itself. I have tested with yellow and green status with many different page sizes and that is the most responsive option.

Current:
image
image

Proposed:
image
image


EjPlatzer (approved):
Thanks for making this better!

PR #1458, opened 2022-03-28
Fix image post

EjPlatzer:
This pull request fixes image posting, which was broken by #1407 because it relied on an unclear and undocumented shortcut to manage HTTP Content-Type headers.

The browser Fetch API that handles HTTP requests automatically creates the Content-Type header based on the type of the request body. However, when the request body is a string, Fetch sets the content-type to 'text/plain'. In almost every request that 360 sends with a body, the body is a string but the content-type is actually application/json. Thus, the default behavior for HTTP HTTP headers has historically been to add the authorization token (is the user is authenticated), and set the content-type to application/json. 

This works in almost every case. For requests with no body (such as GET and DELETE), the content-type is ignored. For requests with a body (namely POST and PUT, application/json is almost always the correct content-type. The one exception (currently) is when we POST an image to the backend. Images are sent as base64 encoded strings, but that string is part of a FormData object, so that the data can be streamed as a multipart request. Thus, the content-type when posting an image should be multipart/form-data. The Fetch API knows this and will give requests with a FormData body the appropriate Content-type (and other relevant headers). However, if we manually overwrite the content-type to be application/json, that breaks these requests.

In the past, we handled this by having an optional `headerOptions` argument for the `http.post` function. When specifying headers, we checked whether the `headerOptions` argument was defined:
 - if not, then the content-type is specified as application/json
 - if so, the content-type is left unspecified so the Fetch API can specify it properly.

Note that we didn't actually use the value of the `headerOptions` argument at all, we just checked whether or not it was passed at all.

Thus, the way that images were posted in the past was by passing a dummy object as the `headerOptions` argument to `http.post`, thereby preventing the content-type from being manually specified, allowing the Fetch API to set it automatically. This has several drawbacks. The first and most notable is that the value of the `headerOptions` argument is meaningless. Thus, in the old code, it was often misleading or confusing. For example, `postIDImage` used to pass as `headerOptions` this object: `{ key: 'this is a post'}`. That had no clear meaning and looked like useless code. Furthermore, if we wanted to pass legitimate header options to be used in the request, there was no way to do so. Putting any real headers on the `headerOptions` object wouldn't set those as headers in the request, it would just prevent the content-type from being set to application/json.

The above work-around broke when I refactored our services to TypeScript, because I saw the inane `headerOptions` argument that didn't have any real headers and removed it, thinking that it was dead code. Instead of restoring the old, hacky and confusing method for managing content-type, I have now refactored the http service to handle content-type appropriately depending on the nature of the request. The default `post` function will now always post it's body as JSON with the content-type application/json. For posting images, I created the new function postImage, which will post an image as FormData with the correct content-type and other headers. Seperating postImage as it's own function has the added advantage of centralizing the code to transform a base64 encoded string into the proper FormData format, which was previously handled independently in every service that needed to post an image. This reduces code duplication and better encapsulates the details of the HTTP protocol inside the http service.

In the process, I also fixed the http service so that we can now pass any custom headers as part of a request, as the original code seemed to intend. This is not currently being used but may become useful in the future.

### DIFF
--- a/src/services/activity.ts
+++ b/src/services/activity.ts
@@ -33,32 +33,8 @@ type ActivityEdit = {
 const editActivity = (activityCode: string, data: ActivityEdit): Promise<ActivityEdit> =>
   http.put(`activities/${activityCode}`, data);
 
-const setActivityImage = (activityCode: string, dataURI: string): Promise<any> => {
-  const imageData = new FormData();
-  const blob = dataURItoBlob(dataURI);
-  const type = blob.type.replace('image/', '');
-  imageData.append('canvasImage', blob, 'canvasImage.' + type);
-  return http.post(`activities/${activityCode}/image/`, imageData);
-};
-
-// convert data to blob
-function dataURItoBlob(dataURI: string) {
-  // convert base64/URLEncoded data component to raw binary data held in a string
-  var byteString;
-  if (dataURI.split(',')[0].indexOf('base64') >= 0) byteString = atob(dataURI.split(',')[1]);
-  else byteString = unescape(dataURI.split(',')[1]);
-
-  // separate out the mime component
-  var mimeString = dataURI.split(',')[0].split(':')[1].split(';')[0];
-
-  // write the bytes of the string to a typed array
-  var ia = new Uint8Array(byteString.length);
-  for (var i = 0; i < byteString.length; i++) {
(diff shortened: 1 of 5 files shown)


bennettforkner (approved):
LGTM!

PR #1469, opened 2022-05-04
Display login error messages from API

EjPlatzer:
The TokenIssuer returns errors in a different format then our CustomException format from the API. In order to properly inform user's of an error signing them in, we need to display the `error_description` property from the API.


bennettforkner (approved):
Clear and concise. Good fix. Thanks!

PR #1472, opened 2022-05-26
Alumni Update Form

amos-cha:
DONE:
- Converted repeated code into components
- Created components subfolder
- Reduced number of useStates
- Adjusted code styling/readability 
- Simplified Snackbar

TODO:
- Functionality is still under investigation

### DIFF
--- a/src/views/AlumniUpdateForm/components/UpdateForm/index.js
+++ b/src/views/AlumniUpdateForm/components/UpdateForm/index.js
@@ -249,7 +249,7 @@ const UpdateForm = ({ profile }) => {
 
   if (!isOnline) return <GordonOffline feature="Update Profile" />;
 
-  if (isUserAlumni)
+  if (!isUserAlumni)
     return (
       <GordonLimitedAvailability
         pageName="Alumni Update Form"
--- a/.env.development
+++ b/.env.development
@@ -5,9 +5,9 @@
 # REACT_APP_FONT_URL=https://cloud.typography.com/7763712/6754392/css/fonts.css
 
 # @TRAIN
-# REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
-# REACT_APP_FONT_URL=https://cloud.typography.com/7763712/7294392/css/fonts.css
+ REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+ REACT_APP_FONT_URL=https://cloud.typography.com/7763712/7294392/css/fonts.css
 
 # @LOCALHOST
- REACT_APP_API_URL=http://localhost:51626/
- REACT_APP_FONT_URL=https://cloud.typography.com/7763712/7294392/css/fonts.css
+# REACT_APP_API_URL=http://localhost:51626/
+# REACT_APP_FONT_URL=https://cloud.typography.com/7763712/7294392/css/fonts.css


src/views/Update/UpdatePage/index.js
```diff
@@ -0,0 +1,257 @@
...
+                info={info}
+                onChange={handleChange}
+              />
+            : <UpdateGrid
+                info={info}
+                onChange={handleChange}
+              />
+        return content;
+      })
+    }
```
EjPlatzer, line 192: You can simplify this a fair bit by changing the format of `data` slightly.

Firstly, notice that you're passing exactly the same props to `UpdateForm`, `UpdateSelect`, and `UpdateGrid`, which indicates that you can extract the logic for deciding which type of input this is to a custom component, e.g. `ProfileUpdateField`. Then, the content of `infoMap` would be just `data.map((info) => <ProfileUpdateField info={info} onChange={handleChange} />

Secondly, we can extract the `personalInfo`, `emailInfo`, etc. arrays out of the component so that they aren't recreated on every render. The only reason they're inside the component currently is because of the reference to `userInfo` for the `value` field. But notice that `value` is ALWAYS `userInfo[name]`. So we can just remove value and add it in when we map the info to components.

Thirdly, if `data` is an array of `{label: string, name: string, type: 'checkbox' | 'textField' | 'select', menuItems?: string[]}` objects, then we can greatly simplify the logic for deciding which input this is.

Putting together all three of those suggestions, you end up with for example:

```jsx
const infoMap = (fields) => {
  return fields.map((field) => <ProfileUpdateField label={field.label} name={field.name} value={userInfo[field.name]} type={label.type} menuItems={field.menuItems} onChange={handleChange} />
}

// In `src/views/Update/components/ProfileUpdateField/index.js`
const ProfileUpdateField = ({label, name, value, type, menuItems, onChange}) => {
  let field;
  switch (type) {
    case 'textField':
      field = (<TextField
        className="disable_select"
        style={{ width: 252, }}
        label={label}
        name={name}
        value={value}
        onChange={onChange}
      />);
      break;
    case 'checkbox':
      field = (<FormControlLabel
        control={<Checkbox checked={value}
        onChange={onChange} />}
        label={label}
        name={name}
      />);
      break;
    case 'select':
      field = (<FormControl style={{ width: 252 }}>
        <InputLabel>{label}</InputLabel>
        <Select
          label={label}
          name={name}
          value={value}
          onChange={onChange}
        >
          {menuItems.map((info) => (
            <MenuItem value={info.value}>{info.value}</MenuItem>
          ))}
        </Select>
      </FormControl>);
      break;
  }
  return (
    <Grid item xs={9} md={3} lg={3}>
      {content}
    </Grid>
  );
}
```

Since `infoMap` is now just a call to `map`, all it does is conceal what's happening without actually encapsulating anything. So I think you could just replace calls to `infoMap` with e.g. `personalInfoFields.map(...)` etc.

src/views/Update/UpdatePage/index.js
```diff
@@ -0,0 +1,257 @@
...
+                    {saveButton}
+                  </Grid>
+                </CardContent>
+              </Card>
+              <Typography variant="subtitle1">
+                Found a bug?
+                <a href="mailto:cts@gordon.edu?Subject=Gordon 360 Bug">
+                  <Button
+                    style={{ color: gordonColors.primary.cyan }}>Report to CTS</Button>
+                </a>
```
EjPlatzer, line 236: Rather than wrapping the button in an anchor tag, you can just set the `href` prop of the button and it will behave like a link.

Note though that you should only do this for non-navigational links like `mailto`. `http[s]` links should use the `Link` component from react router for proper navigation).

Also, you can just use `color="primary"` to set the color.

```suggestion
                <Button
                    href="mailto:cts@gordon.edu?Subject=Gordon 360 Bug" color="primary">Report to CTS</Button>
```

src/views/Update/index.js
```diff
@@ -0,0 +1,19 @@
...
+ * Sends an update form to the development office
+ */
+
+const Update = (props) => {
+  const { profile, loading } = useUser();
+  return loading
+    ? <GordonLoader />
+    : profile
+      ? <UpdatePage profile={profile} authentication={props.authentication} />
+       :<GordonUnauthorized feature={'the Alumni Update Page'} />
```
EjPlatzer, line 16: This is one case where `if` statements are probably much more readable than nested ternary operators.

EjPlatzer (commented):
This is coming along nicely!

I have many comments. Several of them are clean up from the previous owner of this project. I also had a couple about convention and best practice in React. Plus, I left a few suggestions that together are a decent refactor of the way you're mapping profile info fields to components.

src/services/update.ts
```diff
@@ -0,0 +1,13 @@
+import http from './http';
+
+type UpdatedFields = {
+  field: string;
+  value: string;
+}
+
+const requestInfoUpdate = async ( email_content: Array<UpdatedFields> ) => {
```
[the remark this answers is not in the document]
amos-cha: I'm not sure if I can do that, typescript wants me to put a type after the colon. 
the alternative would be:
`const requestInfoUpdate = async ( updatedFields ) => {`
which could be more readable? I'm not too sure.

public/sw_guest_cache.js
```diff
@@ -40,6 +40,7 @@ const static360Cache = [
   '/admin',
   '/news',
   '/myprofile',
+  '/update',
```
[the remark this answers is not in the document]
amos-cha: got it

src/views/Update/UpdatePage/index.js
```diff
@@ -0,0 +1,257 @@
...
+              />
+            : <UpdateGrid
+                info={info}
+                onChange={handleChange}
+              />
+        return content;
+      })
+    }
+
+    if (isOnline && isUserStudent) {
```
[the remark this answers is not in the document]
amos-cha: ahh I see, will implement

amos-cha:
@EjPlatzer, I have a really whacky way of re-formatting the country names from the Jenzabar table. I'm pretty sure there has to be a better way of implementing ln 102 - 109 on UpdatePage, but I can't really think of one at the moment. If you have time for a full code review at some point that would be really helpful too ^^

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+    return label;
+  };
+
+  /**
+   * @param updatedInfo updated information fields object
+   * @param currentInfo old/saved information fields object
+   * @returns {Array<Object>} Array of updated fields in a subobject
+   */
+  function getUpdatedFields(updatedInfo, currentInfo) {
+    var updatedFields = [];
```
EjPlatzer, line 185: Do not use the `var` keyword. It is bad practice in modern JavaScript.
```suggestion
    const updatedFields = [];
```

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+          field: field,
+          value: currentInfo[field],
+          label: getAssociatedLabel(field),
+        });
+      }
+    }
+    return updatedFields;
+  }
+
+  //handles confirmation window's final submit
```
EjPlatzer, line 198: As above, comments on functions should use JSDoc when they are needed. This comment is not currently adding anything. Just rename the function if `handleConfirm` is not clear enough.

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+      createSnackbar(
+        'A request to update your information has been sent. Please check back later.',
+        'info',
+      );
+      setSaving(false);
+      handleWindowClose();
+    });
+  };
+
+  //executes on confirmation window cancel & submit
```
EjPlatzer, line 218: This comment doesn't add anything. Please remove it.

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+import useNetworkStatus from 'hooks/useNetworkStatus';
+import { ProfileUpdateField, NotAlumni, ContentCard } from '..';
+import GordonDialogBox from 'components/GordonDialogBox';
+import GordonUnauthorized from 'components/GordonUnauthorized';
+import { ConfirmationRow } from '../ConfirmationRow';
+import { ConfirmationWindowHeader } from '../ConfirmationHeader';
+
+/**
+ * Sends an update form to the development office
+ */
```
EjPlatzer, line 25: This comment is both unnecessary and untrue. This comment is for the entire component. Firstly, you should probably rename the component for clarity. Secondly, if you do feel that a comment explaining the purpose of the component is necessary, please make it correct:
```suggestion
/**
 * A form for alumni to request an update to their profile information.
 */
```

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+    >
+      Update
+    </Button>
+  );
+
+  /**
+   * @param {Array<Object>} fields static fields that define label, name, value, type
+   * @returns JSX elements of all elements in array with correct type
+   */
+  const infoMap = (fields) => {
```
EjPlatzer, line 251: This comment can be much clearer. Also, please rename this function as its current name is too vague to be helpful.
```suggestion
  /**
   * @param {Array<{name: string, label: string, type: string, menuItems: string[]}>} fields array of objects defining the properties of the input field
   * @returns JSX input for each field
   */
  const mapFieldsToInputComponents = (fields) => {
```

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+        name={field.name}
+        value={updatedInfo[field.name]}
+        type={field.type}
+        menuItems={field.menuItems}
+        onChange={handleChange}
+      />
+    ));
+  };
+
+  // following 3 consts determine confirmation window and its contents
```
EjPlatzer, line 264: This comment is unnecessary.

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+  const [confirmRows, setConfirmRows] = useState('');
+  const [changeReason, setChangeReason] = useState('');
```
EjPlatzer, line 266: By convention, please put all hook calls at the top of the component.

src/views/Update/components/UpdatePage/index.js
```diff
@@ -0,0 +1,365 @@
...
+        style={{
+          width: '100%',
+          minWidth: 504,
+        }}
+      >
+        {getUpdatedFields(currentInfo, updatedInfo).map((field) => (
+          <ConfirmationRow field={field} prevValue={currentInfo[field.field]} />
+        ))}
+      </Grid>,
+    );
```
EjPlatzer, line 281: There is no reason to have `confirmRows` as a state variable. Just put this JSX inline where `{confirmRows}` is currently used. Making it a state variable just complicates the code and introduces more room for bugs (especially state update race conditions).

EjPlatzer (changes requested):
This is looking good! I have several suggestion for cleanup and code quality, but the functionality is getting pretty polished.

src/views/Update/components/UpdatePage/index.js
amos-cha: Yep I did this in the next commit (haven't pushed)

src/views/Update/components/ConfirmationRow/index.js
```diff
@@ -0,0 +1,66 @@
...
+          <Grid item>
+            <Typography
+              variant="caption"
+              style={{
+                paddingLeft: '10px',
+                paddingRight: '10px',
+                color: `${gordonColors.neutral.grayShades[900]}`,
+              }}
+            >
+              {prevValue === '' || false ? 'No previous value' : prevValue}
```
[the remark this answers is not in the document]
amos-cha: Thought in retrospect there are other issues with this, I'll have to think of a better solution

amos-cha:
fixed alignment bug on confirmation window, fixed countries showing wrong capitalization upon select, implemented a clearer truth value confirmation"

src/views/AlumniUpdateForm/components/ConfirmationRow/index.js
```diff
@@ -1,8 +1,32 @@
...
+  console.log(prevValue);
+  if (typeof prevValue === 'boolean') {
+    console.log('hi');
+  }
+  const currentValue =
+    typeof prevValue === 'boolean' ? (
+      <Check className={styles.icon_current} />
+    ) : (
+      <Typography variant="subtitle2" className={styles.text_current}>
+        {`${field.value}`}
```
EjPlatzer, line 17: This syntax is wonderful 😃 

This is a JSX expression (outermost brackets), with a template string (backticks ``) containing an expression that evaluates to a string.

The simpler way to do this is to put the expression `field.value` as the value of the JSX expression:
```suggestion
        {field.value}
```

src/views/Update/components/ConfirmationRow/index.js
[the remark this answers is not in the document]
amos-cha: Fixed it with a check if boolean type then resolved the True False with an mui-icon

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -237,6 +237,27 @@ const PersonalInfoList = ({
...
+  const updateInfoButton = myProf ? (
+    <Grid container justifyContent="center">
+      <Link className="gc360_link" to="/myprofile/update">
+        <Button variant="contained" className={styles.update_info_button}>
+          Update Information
+        </Button>
+      </Link>
+    </Grid>
+  ) : null;
+
```
EjPlatzer, line 260: Once everything else is done and tested, please clean this up so there's only one (uncommented) updateInfoButton that's only visible to alumni.

EjPlatzer (changes requested):
This is looking really good! Just a couple of minor clean-up requests

src/views/AlumniUpdateForm/components/UpdateForm/index.js
```diff
@@ -0,0 +1,349 @@
...
+  { label: 'Do Not Mail', name: 'doNotMail', type: 'checkbox' },
+];
+
+/**
+ * A form for alumni to request an update to their profile information.
+ */
+
+const UpdateForm = (props) => {
+  const history = useHistory();
+  const returnToMyProfile = useCallback(() => history.push('/myprofile'), [history]);
```
EjPlatzer, line 73: This doesn't need to be a callback, it's not expensive to re-compute.

src/views/AlumniUpdateForm/components/UpdateForm/index.js
```diff
@@ -0,0 +1,349 @@
...
+
+  const isOnline = useNetworkStatus();
+  const profile = props.profile;
```
EjPlatzer, line 76: You can destructure props instead
```suggestion
const UpdateForm = ({ profile }) => {
  const history = useHistory();
  const returnToMyProfile = useCallback(() => history.push('/myprofile'), [history]);

  const isOnline = useNetworkStatus();
```

src/views/AlumniUpdateForm/components/UpdateForm/index.js
```diff
@@ -0,0 +1,349 @@
...
+  const isUserAlumni = profile.PersonType.includes('alu');
```
EjPlatzer, line 77: This needs to be if the user is ONLY an alumni. We don't want this accessible to people who are alumni and also a student or facstaff, since Development would not be in charge of their data.
```suggestion
  const isAlumni = profile.PersonType === 'alu';
```

src/views/AlumniUpdateForm/index.js
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * A form for alumni to request an update to their profile information.
+ */
+
+const AlumniUpdateForm = (props) => {
+  const { profile, loading } = useUser();
+  return loading ? (
+    <GordonLoader />
+  ) : profile ? (
+    <UpdateForm profile={profile} authentication={props.authentication} />
```
EjPlatzer, line 15: This authentication prop is both deprecated and unused
```suggestion
    <UpdateForm profile={profile} />
```

src/views/AlumniUpdateForm/index.js
```diff
@@ -0,0 +1,21 @@
+import GordonLoader from 'components/Loader';
+import GordonUnauthorized from 'components/GordonUnauthorized';
+import { useUser } from 'hooks';
+import UpdateForm from './components/UpdateForm';
+
+/**
+ * A form for alumni to request an update to their profile information.
+ */
+
+const AlumniUpdateForm = (props) => {
```
EjPlatzer, line 10: Props aren't used in this component
```suggestion
const AlumniUpdateForm = () => {
```

src/views/AlumniUpdateForm/index.js
```diff
@@ -0,0 +1,21 @@
...
+  const { profile, loading } = useUser();
+  return loading ? (
+    <GordonLoader />
+  ) : profile ? (
+    <UpdateForm profile={profile} authentication={props.authentication} />
+  ) : (
+    <GordonUnauthorized feature={'the Alumni Update Page'} />
+  );
```
EjPlatzer, line 18: I think this is more readable as `if` statements than nested ternary expressions.
```suggestion
  if (loading) return <GordonLoader />
  
  if (!profile) return <GordonUnauthorized feature={'the Alumni Update Page'} />
  
  return <UpdateForm profile={profile} />
```

EjPlatzer (changes requested):
A couple more tweaks 😄

bennettforkner (changes requested):
LGTM

src/services/update.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const requestInfoUpdate = async (updatedFields: Array<ProfileFieldUpdate>) => {
+  await http.post('profiles/update/', updatedFields);
+};
+
+const getAllStates = async (): Promise<State[]> => await http.get('addresses/states/');
+
+const getAllCountries = async (): Promise<Country[]> => await http.get('addresses/countries/');
+
+export { requestInfoUpdate, getAllStates, getAllCountries };
```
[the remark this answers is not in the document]
EjPlatzer: Yes, I meant to ask for that but forgot, thank you for catching it. 

I think `profileInfoUpdate.ts` might be clearest.

PR #1477, opened 2022-06-22
Redirect for auth, fixing Safari login

EjPlatzer:
The `loginPopup` authentication flow of `@azure/msal-browser` is not working on Safari (on both Macos and iOS). The login redirect works well on all browsers (and actually simplifies the code a little bit), so I am switching to that.


amos-cha:
I see Apple is truly built different. LGTM

amos-cha (approved):
I see Apple is truly built different. LGTM

PR #1479, opened 2022-06-22
Remove hard-coded number left in from testing

EjPlatzer:
I accidentally left a hard-coded debugging value in the PersonalInfoList. This has been removed.


amos-cha (approved):
Never seen this number in my life

PR #1486, opened 2022-08-22
Fix updating involvements

EjPlatzer:
Depends on gordon-cs/gordon-360-api#698

This PR fixes involvement updates. Since upgrading the API, involvement updates have been broken because of type mismatches between the API and the UI. I have resolved those mismatches, but it required the introduction of a new type, InvolvementUpdate, which is introduced to the API in gordon-cs/gordon-360-api#698. Therefore, this PR will not work without that API PR. This is not a breaking change because the functionality targeted is already broken 😅.


bennettforkner (approved):
LGTM!

PR #1491, opened 2022-08-25
Only show page after profile and data is loaded

EjPlatzer:
We were setting `loading` to false while the profile was still loading, before data loaded. That meant that as soon as the profile loaded (and before the data was all loaded), we were showing the page, which is not supposed to happen. Now, we only show the page after the profile and data has loaded (even if the loaded value is null).


bennettforkner (approved):
LGTM, thanks!

PR #1493, opened 2022-08-25
Beautify Guest welcome

EjPlatzer:
The Guest Welcome page changed shape once the old login form was removed. I have updated it to be slightly more appealing by putting a max width on the card and adding a (technically redundant but still helpful) login button.
# Before
image

# After
image

### DIFF
--- a/src/views/Home/components/GuestWelcome/GuestWelcome.module.scss
+++ b/src/views/Home/components/GuestWelcome/GuestWelcome.module.scss
@@ -4,6 +4,8 @@ div.gw_card {
   background-color: $primary-blue;
   color: $neutral-white;
   opacity: 0.92;
+  max-width: 25rem;
+  padding: 1rem;
 }
 
 .gw_container {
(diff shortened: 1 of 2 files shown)


bennettforkner:
Hey Evan, thanks for getting to this. Looks a lot better with the sizing on desktop. 2 things:

How does this look on mobile with the block sizing? Any change?

Can we add a little bit more margin between the guest message and the login button? It looks a bit cluttered in the picture you sent.

EjPlatzer:
> How does this look on mobile with the block sizing? Any change?

I'm not sure what you mean by `block sizing`, do you mean the `max-width`? It doesn't have any effect on mobile because it's a max width.

> Can we add a little bit more margin between the guest message and the login button? It looks a bit cluttered in the picture you sent.

Done:
image

bennettforkner (approved):
Thanks a bunch looks great. And thanks for confirming no effect on mobile.

PR #1494, opened 2022-08-26
Display meal plan when it exists

EjPlatzer:
Some Faculty and Staff have a valid meal plan, so we should show that, rather than assuming all FacStaff only have a dining dollars balance.


bennettforkner (dismissed):
LGTM. Thanks for coming up with a fix so fast.

bennettforkner (approved):
LGTM

PR #1497, opened 2022-08-31
Show ID Number on public profile as Sensitive Info

EjPlatzer:
This Pull Request shows ID Number on students' public profile to members of the `SensitiveInfoView` authorization group. The group is carefully curated to only include the subset of Student Life and Campus Safety staff that need access to that data to perform their jobs.


bennettforkner (approved):
LGTM

PR #1498, opened 2022-08-31
Rename prop `isSuperAdmin` -> `isSiteAdmin`

EjPlatzer:
I somehow missed renaming the props for MemberList and AdminCard even though I renamed the props for the other components (including their child components). This fixes that mistake so that the correct prop is passed all the way through the component tree.


bennettforkner (approved):
LGTM

PR #1506, opened 2022-09-12
Clifton strengths csv upload

bennettforkner:
Clifton Strengths data currently must be manually inserted into the database periodically by developers. This PR adds a feature to the admin page to allow an admin end-user to upload Clifton Strengths data. It also opens the doors for other features to use the `SpreadsheetUploader` component for importing large data files. The user workflow for this process is as follows.

1) The user is presented with a button to start the upload on the `/admin` page.
image

2) The user can either drag and drop a file or click the drop zone to choose a file from their computer. A template can also be downloaded beforehand to confirm columns are correct.
image

3) The user is presented with the data that they uploaded to give opportunity to double-check.
image

4) Upload results are displayed on the screen for each uploaded row.
image


src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+        DateCompleted: cs.DateCompleted,
+        Themes: [cs.Theme1, cs.Theme2, cs.Theme3, cs.Theme4, cs.Theme5],
+        Private: false,
+      };
+    });
+
+    CliftonStrengthsService.postCliftonStrengths(formattedData)
+      .then((data) => {
+        setUploadResults(
+          data.map((r, index) => (
```
EjPlatzer, line 55: The variable name `r` is not very descriptive. Perhaps `uploadResult` would be better?

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+              <TableCell style={{ fontSize: '1rem' }}>{r.AccessCode}</TableCell>
+              <TableCell
+                style={{
+                  backgroundColor: successResults.some((s) => s === r.UploadResult)
+                    ? gordonColors.secondary.greenShades.secondary
+                    : gordonColors.secondary.red,
+                  color: 'white',
+                  fontSize: '1rem',
+                }}
+              >
```
EjPlatzer, line 67: Please use css instead of inline styles. Also, the `sx` prop doesn't work in our current version of MUI.

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+          'Theme1',
+          'Theme2',
+          'Theme3',
+          'Theme4',
+          'Theme5',
+          'AccessCode',
+        ]}
+        buttonName="Upload Strengths"
+        template={CliftonStrengthsUploadTemplate}
+      ></SpreadsheetUploader>
```
EjPlatzer, line 136: This can be a self-closing tag
```suggestion
       />
```

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+            onClick={() => setIsSpreadsheetUploaderOpen(true)}
+            style={{ margin: '50px auto' }}
+          >
+            Upload Clifton Strengths
+          </Button>
+        </Grid>
+        {loading ? (
+          <>
+            <GordonLoader />
+            <div style={{ height: '10px' }}></div>
```
EjPlatzer, line 98: Again, I think css would be a better method to space these apart (if that is needed).

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+import { gordonColors } from 'theme';
+import GordonLoader from '../../../../components/Loader';
+import SpreadsheetUploader from '../../../../components/SpreadsheetUploader';
+import CliftonStrengthsService from '../../../../services/cliftonStrengths';
+import CliftonStrengthsUploadTemplate from './cliftonStrengthsUploadTemplate.csv';
+
+const headerStyle = {
+  backgroundColor: gordonColors.primary.blue,
+  color: gordonColors.neutral.grayShades[100],
+};
```
EjPlatzer, line 23: Please use CSS classes instead of javascript style classes.

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+      .catch((error) => {
+        setUploadResults(<h1>An Error has occurred while uploading.</h1>);
+        setLoading(false);
+      });
+  };
+
+  return (
+    <>
+      <Card>
+        <CardHeader style={headerStyle} align="center" title={`Clifton Strengths`} />
```
EjPlatzer, line 84: When passing a string prop, you can pass the string in double quotes, instead of as a string in brackets:
```suggestion
        <CardHeader style={headerStyle} align="center" title="Clifton Strengths" />
```

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+        <Grid container justify="center">
+          <Button
+            variant="contained"
+            color="primary"
+            onClick={() => setIsSpreadsheetUploaderOpen(true)}
+            style={{ margin: '50px auto' }}
```
EjPlatzer, line 90: Please use css instead of inline styles.

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+        ) : (
+          uploadResults && (
+            <Table>
+              <TableHead>
+                <TableRow>
+                  <TableCell style={{ fontWeight: 'bold', fontSize: '1rem' }}>Email</TableCell>
+                  <TableCell style={{ fontWeight: 'bold', fontSize: '1rem' }}>
+                    Access Code
+                  </TableCell>
+                  <TableCell style={{ fontWeight: 'bold', fontSize: '1rem' }}>
```
EjPlatzer, line 109: Please use CSS instead of inline styles.

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+        setUploadResults(
+          data.map((r, index) => (
+            <TableRow key={index} sx={{ '&:last-child td, &:last-child th': { border: 0 } }}>
+              <TableCell style={{ fontSize: '1rem' }}>{r.Email}</TableCell>
+              <TableCell style={{ fontSize: '1rem' }}>{r.AccessCode}</TableCell>
+              <TableCell
+                style={{
+                  backgroundColor: successResults.some((s) => s === r.UploadResult)
+                    ? gordonColors.secondary.greenShades.secondary
+                    : gordonColors.secondary.red,
```
EjPlatzer, line 63: I think a colored icon would be a more aesthetically pleasing way to distinguish the result than coloring the entire table cell. This is also more consistent with the rest of the site, and more accessible to people with colorblindness.

package.json
```diff
@@ -11,6 +11,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@material-ui/pickers": "^3.3.10",
+    "@mui/x-data-grid": "^4.0.1",
```
EjPlatzer, line 14: It looks like you added this package but don't use it anywhere. If you don't plan on using it, please remove it.

EjPlatzer (changes requested):
This is looking really good! I have a lot of small notes, but overall I think it's really good.

I have only one high-level thought. It seems odd that we display the data as a pillar of cards before the user uploads, but then display the upload results as a table. I would prefer that both were tables, or at least standard columnar layouts. A pillar of cards will be really unfriendly when the user uploads potentially hundreds of rows at once. won't it?

src/components/SpreadsheetUploader/index.js
```diff
@@ -0,0 +1,189 @@
...
+            keys.length > maxColumns ||
+            keys.length < requiredColumns.length ||
+            requiredColumns.some((column) => !keys.includes(column)) ||
+            keys.some((key) => !requiredColumns.includes(key) && !otherColumns.includes(key))
+          );
+        })
+      ) {
+        setError('Please make sure your file matches the template file.');
+        return;
+      }
```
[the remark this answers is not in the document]
EjPlatzer: It took me a minute to understand that `keys` here represents the names of the columns of the given row. Could you use a more descriptive name, possibly `columns` or `columnNames`?

src/components/SpreadsheetUploader/SpreadsheetUploader.module.scss
```diff
@@ -0,0 +1,19 @@
...
+
+.dropzone {
+  transition: border-color 0.3s, background-color 0.3s, box-shadow 0.3s;
+  min-width: 18rem;
+  min-height: 8rem;
+  display: flex;
+  border-style: dashed;
+  border-width: 0.2rem;
+  border-radius: 1rem;
+  border-color: darkgray;
```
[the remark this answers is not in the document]
bennettforkner: I definitely agree with you that there are benefits to each. This form declares each property individually if you are not as familiar with the structure of the combined form, but sacrifices conciseness. Nevertheless, I think the combined form is clear enough...making changes.

src/components/SpreadsheetUploader/index.js
```diff
@@ -0,0 +1,189 @@
+import { Button, Card, Typography } from '@material-ui/core';
+import GordonDialogBox from 'components/GordonDialogBox';
+import GordonSnackbar from 'components/Snackbar';
+import { useEffect, useState } from 'react';
+import Dropzone from 'react-dropzone';
+import { Link } from 'react-router-dom';
+import { gordonColors } from 'theme';
+import XLSX from 'xlsx';
+import SpreadsheetSVG from './spreadsheet.svg';
```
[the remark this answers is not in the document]
bennettforkner: Good catch, I have updated it to an MIT open licensed image:

image
Source: https://iconscout.com/icon/spreadsheet-12

src/components/SpreadsheetUploader/index.js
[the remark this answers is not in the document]
bennettforkner: Implementing changes:
```suggestion
      uploadedData.forEach((row) => {

        let columnNames = Object.keys(row);

        if (error) return;

        if (columnNames.length > maxColumns)
          setError("File has too many columns");
        if (columnNames.length < requiredColumns.length)
          setError("File has too few columns");
        if (requiredColumns.some((column) => !columnNames.includes(column)))
          setError("File is missing required columns");
        if (columnNames.some((columnName) => !requiredColumns.includes(columnName) && !otherColumns.includes(columnName)))
          setError("File contains extra columns");
      });
      
      if (error) return;
```

src/components/SpreadsheetUploader/index.js
```diff
@@ -0,0 +1,189 @@
...
+        ) : null}
+        {data
+          ? data.map((row, index) => {
+              const fields = requiredColumns.map((c) => {
+                return row[c] ? (
+                  <Typography variant="body2">
+                    <b>{c}:</b> {row[c]}
+                  </Typography>
+                ) : null;
+              });
```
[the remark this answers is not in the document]
bennettforkner: I feel like sometimes this hurts code readability when it is a multi-line expression with JSX in the return. That is why I added brackets, not because I don't know about expression syntax.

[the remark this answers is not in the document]
bennettforkner: Nevertheless, I am making changes as you requested

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -0,0 +1,141 @@
...
+  const [isSpreadsheetUploaderOpen, setIsSpreadsheetUploaderOpen] = useState(false);
+  const [uploadResults, setUploadResults] = useState(null);
+  const [loading, setLoading] = useState(false);
+
+  const handleSubmitStrengths = (uploadedData) => {
+    setLoading(true);
+
+    let formattedData = uploadedData.map((cs) => {
+      // The date completed comes out of the
+      // spreadsheet uploader as a number of days since 1/1/1900.
```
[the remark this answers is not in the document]
bennettforkner: Wild guess

src/components/SpreadsheetUploader/index.js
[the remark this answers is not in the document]
EjPlatzer: That looks good! My only concern is that `error` is not updated synchronously by `setError`, so you may get some unexpected behavior here. There are two simple ways to resolve that:
1. Introduce a local `error` variable and then call `setError` outside the loop before returning if the local `error` is not null
2. Still use `Array.some` instead of `Array.forEach`, but still `setError` inside the loop. That way, you don't have to evaluate the whole dataset when you're only looking for the first error that exists.

Personally, I think that option #2 results in cleaner flow overall:
```suggestion
      const hasError = uploadedData.some((row) => {
            let columnNames = Object.keys(row);
	
	        if (columnNames.length > maxColumns) {
	          setError("File has too many columns");
	        } else if (columnNames.length < requiredColumns.length) {
	          setError("File has too few columns");
	        } else if (requiredColumns.some((column) => !columnNames.includes(column))) {
	          setError("File is missing required columns");
	        } else if (columnNames.some((columnName) => !requiredColumns.includes(columnName) && !otherColumns.includes(columnName))) {
	          setError("File contains extra columns");
	        } else {
	          return false;
	        }
	        
	        return true;
	      });
	      
	      if (hasError) return;
```

src/views/Admin/components/CliftonStrengthsUpload/index.js
bennettforkner: I think this looks good:
image

src/views/Admin/components/CliftonStrengthsUpload/index.js
bennettforkner: I have no idea where that came from. I don't believe I copied it and I definitely didn't add the sx myself. Hmmm...

Removed.

src/components/GordonDialogBox/index.js
```diff
@@ -38,9 +38,16 @@ const GordonDialogBox = ({
   children,
   ...otherProps
 }) => {
+  const handleClose = (event, reason) => {
+    if (typeof cancelButtonClicked === 'function') {
+      cancelButtonClicked();
+    }
+  };
```
bennettforkner, line 45: This is the recommended way to resolve the deprecation of backdrop-related Dialog hooks. See https://robertmarshall.dev/blog/disablebackdropclick-of-forwardrefdialog-is-deprecated-solution-and-example/

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -9,21 +9,20 @@ import {
...
-const headerStyle = {
-  backgroundColor: gordonColors.primary.blue,
-  color: gordonColors.neutral.grayShades[100],
-};
+import UploadFailedSVG from './uploadFailed.svg';
+import UploadSuccededSVG from './uploadSucceded.svg';
 
 const successResults = ['Success', 'Added', 'Modified'];
 
+const initDate = parseISO('1900-01-01T00:00:00');
```
EjPlatzer, line 24: What was the reasoning for using `date-fns/parseISO` instead of `new Date()` here?

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -52,20 +49,34 @@ const CliftonStrengthsUpload = () => {
...
               <TableCell
-                style={{
-                  backgroundColor: successResults.some((s) => s === r.UploadResult)
-                    ? gordonColors.secondary.greenShades.secondary
-                    : gordonColors.secondary.red,
-                  color: 'white',
-                  fontSize: '1rem',
-                }}
+                className={
+                  successResults.some((s) => s === uploadResult.UploadResult)
```
EjPlatzer, line 60: I'm noticing that we compute this value three times for each uploadResult. Would it make sense to assign to a variable instead and reuse the variable?

src/components/GordonDialogBox/index.js
EjPlatzer: I don't think this accomplishes anything besides preventing `cancelButtonClicked` from being called if it's not a function (which is not an awful thing, but that would only apply when someone passes something that isn't a function and is not `undefined` as the value of `cancelButtonClicked`).

Are you just trying to get the dialog to close when the backdrop is clicked? If so, the simplest solution is to pass `onClose` to the `GordonDialogBox`, rather than changing the default behavior of the `GordonDialogBox`.

EjPlatzer (changes requested):
Looking good! Just a couple more things.

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -9,21 +9,20 @@ import {
...
+import GordonLoader from 'components/Loader';
+import SpreadsheetUploader from 'components/SpreadsheetUploader';
+import { addDays, parseISO } from 'date-fns';
 import { useState } from 'react';
-import { gordonColors } from 'theme';
-import GordonLoader from '../../../../components/Loader';
-import SpreadsheetUploader from '../../../../components/SpreadsheetUploader';
-import CliftonStrengthsService from '../../../../services/cliftonStrengths';
+import CliftonStrengthsService from 'services/cliftonStrengths';
+import styles from './CliftonStrengthsUpload.module.scss';
```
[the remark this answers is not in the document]
EjPlatzer: Looks like you already fixed this, nevermind.

src/components/GordonDialogBox/index.js
bennettforkner: Yea, that's essentially what it was doing except also checking the type. I will remove that type check.

src/views/Admin/components/CliftonStrengthsUpload/index.js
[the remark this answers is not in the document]
bennettforkner: You must have reviewed before I made the commit that added it. It is here now

src/views/Admin/components/CliftonStrengthsUpload/index.js
bennettforkner: Nothing particular, just a decision.

src/components/GordonDialogBox/index.js
```diff
@@ -41,6 +41,7 @@ const GordonDialogBox = ({
   return (
     <Dialog
       className={styles.gc360_gordondialogbox}
+      onClose={cancelButtonClicked}
```
EjPlatzer, line 44: This changes the default behavior for all GordonDialogBoxes to close whenever someone clicks outside the dialog, hits esc, or presses a close button. While I don't necessarily think that's a bad thing, I do think that it needs to be a considered and tested change, especially since some of our dialogs are multistep forms where closing unexpectedly could be quite frustrating to the user. Perhaps we could create an issue about this, but for this current pull request, I would think it's better to just pass an `onClose` function to the `GordonDialogBox` in `SpreadSheetUploader`, which will have the same effect for that dialog but won't change every other use of our dialog component site-wide.

bennettforkner:
> FYI, we already have good ❌ and ✅ icons provided by `@material-ui/icons`, I don't think we need to add two new SVGs for that. See the [`Snackbar`](https://github.com/gordon-cs/gordon-360-ui/blob/develop/src/components/Snackbar/index.js) for example usage.

All set. I removed the SVGs and used the check and X from MUI.

src/views/Admin/components/CliftonStrengthsUpload/index.js
```diff
@@ -51,23 +50,21 @@ const CliftonStrengthsUpload = () => {
         setUploadResults(
           data.map((uploadResult, index) => {
             const uploadResultSuccess = successResults.some((s) => s === uploadResult.UploadResult);
-
+            console.log(uploadResultSuccess);
```
EjPlatzer, line 53: Is this intended to go into production, or was it just for debugging purposes?

EjPlatzer (approved):
LGTM!

PR #1518, opened 2022-09-22
Add Rec-IM Initial Route and Folder Organization

amos-cha:
+ Route /recim
+ Routed currently through Header (can be switched up), requires log-in, currently uses Event icon
+ Primary view found in views/recim/index.js
+ Service file found in service/recim.ts


Kevin-Chu58:
LGTM!

Kevin-Chu58 (dismissed):

src/components/Header/index.js
```diff
@@ -1,6 +1,7 @@
 import { useIsAuthenticated } from '@azure/msal-react';
 import { AppBar, Button, IconButton, Tab, Tabs, Toolbar, Typography } from '@material-ui/core';
 import EventIcon from '@material-ui/icons/Event';
+import SportsFootballIcon from '@material-ui/icons/SportsFootball';
```
cpabbot, line 4: I think this is probably fine for now, we'll have to think about Rec-IM icon at some point down the line. It'd be cool to have an icon that encapsulates all of rec-im, but this matches well

cpabbot (approved):
Looks good, thanks for getting that initial framework down. Everything here makes sense to me, only thing will be figuring out that icon which we can change later on.

src/components/Header/index.js
amos-cha: I've uploaded a set of icons that MUI provides for sports, perhaps it would be a good idea to bring it up to our customer and ask if they would like a completely new icon or if any of the sport icons would suit their taste.

PR #1520, opened 2022-09-25
Add GET Leagues & POST League services

amos-cha:


src/services/recim.ts
```diff
@@ -1,3 +1,31 @@
+import { DateTime } from 'luxon';
```
silasw2, line 1: Queried reason for luxon import, conversed in person. Just for typing.

silasw2 (approved):
LGTM!

src/services/recim.ts
```diff
@@ -1,3 +1,31 @@
...
 import http from './http';
 
 // all Gordon Rec-IM http requests
+type League = {
+  ID: number;
+  Name: string;
+  RegistrationStart: DateTime;
```
EjPlatzer, line 8: I don't think this does what you think it does. This 'DateTime' is completely unrelated to C#'s DateTime type. Also, because you're serializing/deserializing JSON with this type, it will just be a string in transit. That means that you would have to manually convert the string date into a Luxon DateTime, which is not what you are doing.

amos-cha:
Correct me if I'm wrong @EjPlatzer, C#'s DateTime comes back as a string which can be parsed using JS' in built Date.parse() function. So I've adjusted them to just a string to be handled in UI.

src/services/recim.ts
amos-cha: ahhh, i see okay. I made an assumption that it worked because it did not throw me errors when running

PR #1525, opened 2022-09-28
Refactor people search

EjPlatzer:
This is a refactoring of the People Search page.

The main focus of the refactoring was to move from Class components to function components. In the process, I also refactored the components to TypeScript, cleaned up the styling and markup quite a bit, and streamlined the design of the search results to one unified, responsive design that looks good on all screen sizes.


src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -0,0 +1,493 @@
...
+              name="residence_hall"
+              value={searchParams.residence_hall}
+              updateValue={handleUpdate}
+              options={halls}
+              Icon={FaBuilding}
+              select
+            />
+          </Grid>
+
+          {isTodayAprilFools() ? (
```
bennettforkner, line 334:
```suggestion
          {isTodayAprilFools ? (
```

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -0,0 +1,493 @@
...
+  'Jesus Is My Significant Other',
+  'Waiting For Her Boaz',
+  'Waiting For His Proverbs 31 Woman',
+  "It's Complicated",
+];
+
+const searchPageTitle = (
+  <>
+    Search the
+    <b style={{ color: gordonColors.primary.cyan }}> Gordon </b>
```
bennettforkner, line 72:
```suggestion
    <b style={{ color: gordonColors.primary.blue }}> Gordon </b>
```

bennettforkner (changes requested):
Just a few changes, otherwise looks good.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -0,0 +1,493 @@
...
+  building: '',
+};
+
+const { serializeSearchParams, deserializeSearchParams } =
+  searchParamSerializerFactory(defaultSearchParams);
+
+const isTodayAprilFools = () => {
+  const todaysDate = new Date();
+  return todaysDate.getMonth() === 3 && todaysDate.getDate() === 1;
+};
```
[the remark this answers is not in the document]
EjPlatzer: Technically, this introduces a bug where we create a new `Date` twice, and the second one will possibly have a later date than the first. In reality, that shouldn't matter because the expected difference in time between the two dates is at most milliseconds.

My other concern would be that, if we statically calculate this outside the component, it might be evaluated at build time (or some other point), rather than being recalculated on render, meaning the value would become stale and inaccurate.

bennettforkner (approved):

PR #1528, opened 2022-09-30
Check Empty/null GraduationYear for Display

### LINKED ISSUE #1378: Graduating Seniors show an empty Graduation Year
The logic for showing Graduation Year on PublicProfile does not factor in the case where a student is a graduating senior because they are already considered alumni in the system. We should add a check not display their graduation year if it is null or empty.

image

Kevin-Chu58:
Reformed logic to `isAlumni && !profile.PreferredGraduationYear.trim().isEmpty()` to check whether Graduation Year is null or empty.

- If not null/empty, show the graduation year in display
- else, no display at all

Closes #1378.


Kevin-Chu58:
I wasn't able to test this on the frontend, do you know if there's a test account for alumni?

EjPlatzer:
> I wasn't able to test this on the frontend, do you know if there's a test account for alumni?

We unfortunately don't have a test account that has the relevant data. We will be able to test this again in the Spring when seniors are marked as alum prior to their graduation, but in the meantime we can at least verify that it isn't a regression (i.e. that we can still see graduation years on true alumni).

bennettforkner (approved):
Looks good to me. Thanks Kevin for the quick action on this!

src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -257,7 +257,7 @@ const PersonalInfoList = ({ myProf, profile, createSnackbar }) => {
     createSnackbar(status.message, status.type);
   };
 
-  const graduationYear = isAlumni ? (
+  const graduationYear = isAlumni && !profile.PreferredClassYear.trim().isEmpty()? (
```
EjPlatzer, line 260: Please make sure you have the `Prettier` VS Code extension enabled to automatically format files when you save them.

In this case, the question mark should be separate from the function call because it is ternary conditional operator and not part of the syntax of the function call.

```suggestion
  const graduationYear = isAlumni && !profile.PreferredClassYear.trim().isEmpty() ? (
```

EjPlatzer (changes requested):

EjPlatzer (approved):
LGTM! Thanks for taking care of this!

Kevin-Chu58:
> LGTM! Thanks for taking care of this!

No problem! BTW, I think I need the permission to merge this pull request.

PR #1531, opened 2022-10-04
Rec-IM Home Initial Layout

cpabbot:
## Rec-IM Home Initial Layout
Initial layout creation with two cards for "upcoming events" and "my teams" with some starter listings within them. These listings have not yet been fleshed out; they are merely concept starters.

**Notes:**
- On mobile, "my teams" should probably be at the top before "upcoming events"
- We'll need to replace the Rec-IM image obviously with a high quality PNG
- More of a general 360 thing, but I think the card headers should be smaller

image

Resolves https://github.com/gordon-cs/RecIM-Docs/issues/30


src/views/RecIM/RecIM.module.scss
```diff
@@ -1 +1,14 @@
 @import '../../vars';
+
+.card {
+  background-color: #014983;
+  color: #ffffff;
```
bennettforkner, line 5: Make sure to avoid hard-coded color codes in the long term. 
```suggestion
  background-color: $primary-blue;
  color: $neutral-white;
```

bennettforkner (approved):
Looks great -- just saw this come through so I thought I would take a look. Suggesting one change now, but I'm sure we'll have chances later to walk through the code and approve.

src/views/RecIM/RecIM.module.scss
cpabbot: I'll change that - thanks for catching that! I copied that from another place styling card headers, so there are other places that need to be updated at some point, and I think it would be good to have a generic GordonCard styling. More global styling is something to think about on the 360 side.

amos-cha (commented):
Looks good! However, I'm wondering if it would better in the long run to have a single "Listing" component that could take in types like League, Team, Series, Match, Users and so on or if each type is specialized enough that we should have separate components for each.

Kevin-Chu58 (approved):
LGTM!

PR #1532, opened 2022-10-06
Replacing isEmpty() to str.length === 0

### LINKED ISSUE #1378: Graduating Seniors show an empty Graduation Year

Kevin-Chu58:
Replaced isEmpty() with checking the string length logic.

Closes #1378.

### DIFF
--- a/src/components/Profile/components/PersonalInfoList/index.js
+++ b/src/components/Profile/components/PersonalInfoList/index.js
@@ -258,7 +258,7 @@ const PersonalInfoList = ({ myProf, profile, createSnackbar }) => {
   };
 
   const graduationYear =
-    isAlumni && !profile.PreferredClassYear.trim().length === 0 ? (
+    isAlumni && profile.PreferredClassYear?.trim() ? (
       <ProfileInfoListItem title={'Graduation Year:'} contentText={profile.PreferredClassYear} />
     ) : null;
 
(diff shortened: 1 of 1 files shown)


src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -257,9 +257,10 @@ const PersonalInfoList = ({ myProf, profile, createSnackbar }) => {
     createSnackbar(status.message, status.type);
   };
 
-  const graduationYear = isAlumni && !profile.PreferredClassYear.trim().isEmpty() ? (
-    <ProfileInfoListItem title={'Graduation Year:'} contentText={profile.PreferredClassYear} />
-  ) : null;
+  const graduationYear =
+    isAlumni && !profile.PreferredClassYear.trim().length === 0 ? (
```
EjPlatzer, line 261: Since you're just trying to check whether this string has any content, we can take advantage of JavaScript's truthiness rules to simplify this. Also, we should be careful of the case where `PreferredClassYear` is null (even though that ought to be impossible, the integrity of our Profile data is unfortunately suspect).
```suggestion
    isAlumni && profile.PreferredClassYear?.trim() ? (
```

This way, we check that the person is an alumni, and that their preferred class year is not null or undefined and contains content other than whitespace.

EjPlatzer (changes requested):

EjPlatzer (approved):
LGTM!

PR #1533, opened 2022-10-06
Remove ID Numbers from Membership, Request, and Admin Routes

bennettforkner:
The Membership, Request, and Admin services have been using Gordon ID numbers in the post body when they pass information to the API. Gordon IDs should be protected from view so that other users cannot see them unless absolutely necessary.

This PR along with API [#720](https://github.com/gordon-cs/gordon-360-api/pull/720) remove ID numbers from objects used in the UI and clean up some of the code surrounding the two resources.


src/services/request.ts
```diff
@@ -8,52 +8,51 @@ export enum RequestStatus {
...
-  ActivityDescription: string;
-  /** Comment or text */
+export type RequestUpload = {
+  ACTCode: string;
+  SessCode: string;
+  Username: number;
+  PartCode: Participation;
+  DateSent: string;
   CommentText: string;
+  Status: RequestStatus;
```
EjPlatzer, line 18: The naming of these fields seems like a regression. Can we use the full word, instead of abbreviating the `Code` fields?

EjPlatzer (commented):
This is looking really good so far! Just a couple of very minor things. Looking forward to once this is ready.

.env.development
```diff
@@ -1,10 +1,10 @@
...
+# Refers to the development environment using localhost
+
+# @PROD
+# REACT_APP_API_URL=https://360Api.gordon.edu/
+
+# @TRAIN
+# REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+
+# @LOCALHOST
+ REACT_APP_API_URL=http://localhost:51626/
```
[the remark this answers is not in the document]
bennettforkner: Yea, didnt mean to commit that -- although I usually do it about once per PR.

src/views/InvolvementsAll/index.js
```diff
@@ -83,12 +83,10 @@ const InvolvementsAll = ({ location, history }) => {
       setAllInvolvements(await involvementService.getAll(selectedSession));
       setTypes(await involvementService.getTypes(selectedSession));
       if (profile) {
-        setMyInvolvements(
-          await membershipService.getSessionMembershipsWithoutGuests(
-            profile.AD_Username,
-            selectedSession,
-          ),
-        );
+        console.log('"' + selectedSession + '"');
```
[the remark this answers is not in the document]
bennettforkner: ofc :)

src/services/request.ts
bennettforkner: Done

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -231,7 +232,7 @@ const MemberListItem = ({
           cancelButtonClicked={() => setIsRemoveAlertOpen(false)}
         >
           Are you sure you want to remove {member.FirstName} {member.LastName} (
-          {member.ParticipationDescription}) from this involvement?
+          {member.ParticipationDescription.trim()}) from this involvement?
```
EjPlatzer, line 235: Should we be trimming this on the backend?

src/views/InvolvementsAll/index.js
```diff
@@ -83,12 +83,9 @@ const InvolvementsAll = ({ location, history }) => {
...
       if (profile) {
-        setMyInvolvements(
-          await membershipService.getSessionMembershipsWithoutGuests(
-            profile.AD_Username,
-            selectedSession,
-          ),
-        );
+        membershipService
+          .getSessionMembershipsWithoutGuests(profile.AD_Username, selectedSession)
+          .then(setMyInvolvements);
```
EjPlatzer, line 88: Couple of thoughts:
1. It's weird to change this syntax in this file without changing anything else in this file.
2. It's weird to change only this instance of this syntax in this file, when the exact same syntax is used directly above it
3. This actually breaks the loading behavior of this page. By using `.then` instead of `await`, we allow execution of the rest of this function before the promise is fulfilled, which means we set loading to false before `MyInvolvements` has loaded. That makes this change a bug.

EjPlatzer (changes requested):
I had a few comments. Some of them are incidental to your changes, so feel free to disregard if you don't think they belong in the scope of this PR.

src/views/Admin/components/AdminList/index.js
```diff
@@ -24,27 +23,29 @@ const AdminList = () => {
...
-    let email = newAdminEmail;
+    let email;
+    let username;
     if (!newAdminEmail.toLowerCase().includes('@gordon.edu')) {
       email = newAdminEmail + '@gordon.edu';
+      username = newAdminEmail;
+    } else {
+      email = newAdminEmail;
+      username = newAdminEmail.replace('@gordon.edu', '');
     }
```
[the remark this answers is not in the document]
bennettforkner: Yes, that makes sense. Probably `newAdmin`

src/views/Admin/components/AdminList/index.js
```diff
@@ -24,27 +23,29 @@ const AdminList = () => {
...
+
     let data = {
-      ID_NUM: addID,
-      EMAIL: email,
-      USER_NAME: email.split('@')[0],
-      SUPER_ADMIN: true, //Used to be distinction between superadmin (godmode), admin, and groupadmin
-      //now just superadmin and groupadmin
+      Email: email,
+      Username: username,
+      IsSuperAdmin: true,
```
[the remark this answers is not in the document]
bennettforkner: Nope, should be trivial to remove.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -129,22 +122,30 @@ const MemberListItem = ({
   const confirmLeave = async () => {
     await membership.remove(member.MembershipID);
     let inInvolvement = await membership.search(
-      member.IDNumber,
+      member.Username,
       member.SessionCode,
       member.ActivityCode,
     )[0];
```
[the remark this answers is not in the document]
bennettforkner: Sure

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
bennettforkner: I don't know. I just noticed a problem and fixed it.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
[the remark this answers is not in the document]
bennettforkner: Done

src/views/Admin/components/AdminList/index.js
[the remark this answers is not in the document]
bennettforkner: On second thought I imagine that we are going to do major refactoring of the admin activity area with our new Azure authentication in place. I think the change would be better suited there.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.js
```diff
@@ -129,22 +122,30 @@ const MemberListItem = ({
...
+    } else {
+      leaveText = `in removing ${member.Username}`;
+    }
+
     if (inInvolvement) {
-      createSnackbar('Leaving involvement failed', 'error');
+      createSnackbar(`Failed ${leaveText}`, 'error');
     } else {
-      createSnackbar('Leaving involvement succeeded', 'success');
+      createSnackbar(`Succeeded ${leaveText}`, 'success');
```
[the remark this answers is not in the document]
bennettforkner: Changes made

src/views/InvolvementsAll/index.js
bennettforkner: Yea, I think I reverted other changes here but somehow missed this one. Good catch with the bug.

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.js
```diff
@@ -52,20 +52,19 @@ const AdminCard = ({ createSnackbar, isSiteAdmin, involvementDescription, onAddM
...
-        ACT_CDE: involvementCode,
-        SESS_CDE: sessionCode,
-        // TODO: Fix API to accept username instead of ID and then remove Group Admin privilege to access ID.
-        ID_NUM: (await membershipService.getEmailAccount(memberEmail)).GordonID,
-        PART_CDE: participationCode,
-        COMMENT_TXT: titleComment,
-        GRP_ADMIN: false,
+        Activity: involvementCode,
+        Session: sessionCode,
+        Username: username,
```
EjPlatzer, line 63: So you're calling `setUsername`, but then referencing `username` in the same function. There's no guarantee that `username` will have updated by the time you reference it here, which could cause serious and unpredictable issues.

EjPlatzer (changes requested):
This looks good! Just one bug I noticed this time around.

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.js
```diff
@@ -52,20 +52,20 @@ const AdminCard = ({ createSnackbar, isSiteAdmin, involvementDescription, onAddM
   };
 
   const handleAddMember = async () => {
-    let memberEmail = username;
-    if (!memberEmail.toLowerCase().includes('@gordon.edu')) {
-      memberEmail = memberEmail + '@gordon.edu';
+    let formattedUsername = username.toLowerCase();
+    if (!formattedUsername.includes('@gordon.edu')) {
+      formattedUsername = formattedUsername.replace('@gordon.edu', '');
     }
```
EjPlatzer, line 58: I think you inverted this process without inverting the `if`. Shouldn't this be 'if the username **includes** `'@gordon.edu'`'?

```javascript
    let formattedUsername = username.toLowerCase();
    if (formattedUsername.includes('@gordon.edu')) {
      formattedUsername = formattedUsername.replace('@gordon.edu', '');
    }
```

Also, I think this could be cleaned up in a couple of ways, now that I'm thinking about it:
1. Would it not be more efficient to always remove the ending `'@gordon.edu'`, rather then checking if it exists before removing it? Calling `replace()` with an unmatched pattern doesn't affect the output at all.
So you could rewrite it as:
```javascript
    const formattedUsername = username.toLowerCase().replace('@gordon.edu', '');
```
2. Technically, we want to replace `'@gordon.edu'` at the END of the string, not wherever we find it (although it should only exist at the end). We could do this with a RegExp pattern input to `replace`, but actually I think it would be clearer to use the `substring` and `lastIndexOf` functions on JS strings. The key thing to note is that `substring` returns the portion of the string from the `start` index (first param), up to but not including the `end` index. Also, I did some quick perf-testing and the latter is approximately 50-60% more performant than the former:
```javascript
    let domainIndex = username.toLowerCase().lastIndexOf('@gordon.edu');
    if (domainIndex === -1) domainIndex = username.length;
    const formattedUsername = username.substring(0, domainIndex);
```

Now at this point, this almost feels like it deserves to be a separate utility function:
```javascript
const stripDomain = (input, domain = '@gordon.edu') => {
    const domainIndex = input.toLowerCase().lastIndexOf(domain);
    
    // if domain was not found, return entire input
    if (domainIndex === -1) return input;
    
    // return input until domain
    return input.substring(0, domainIndex);
}
```

I think that at the very least we should adapt the 'always replace' style, but having this as it's own utility function might be really useful.

bennettforkner: Love it

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.js
```diff
@@ -27,6 +27,16 @@ const headerStyle = {
   padding: '10px',
 };
 
+const stripDomain = (input, domain = '@gordon.edu') => {
```
EjPlatzer, line 30: Do we want to move this to `utils.ts` so that it can be re-used in other places more easily?

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.js
```diff
@@ -52,10 +62,7 @@ const AdminCard = ({ createSnackbar, isSiteAdmin, involvementDescription, onAddM
   };
 
   const handleAddMember = async () => {
-    let formattedUsername = username.toLowerCase();
-    if (!formattedUsername.includes('@gordon.edu')) {
-      formattedUsername = formattedUsername.replace('@gordon.edu', '');
-    }
+    let formattedUsername = stripDomain(username.toLowerCase());
```
EjPlatzer, line 65: I don't think the `.toLowerCase()` is necessary here - it was needed to match `'@gordon.edu'`, but since `stripDomain` handles that, I think you can leave the casing as-is without any issues, right?

bennettforkner:
I was just thinking for consistent uploads to the API

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.js
bennettforkner: I was just thinking for consistent uploads to the API

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.js
bennettforkner: Perhaps that makes sense. I often forget that we have a utils service.

src/services/utils.ts
```diff
@@ -152,5 +152,24 @@ export const searchParamSerializerFactory = <
...
+    const domainIndex = input.toLowerCase().lastIndexOf(domain);
+
+    // if domain was not found, return entire input
+    if (domainIndex === -1) return input;
+
+    // return input until domain
+    return input.substring(0, domainIndex);
+  };
+
+  return { serializeSearchParams, deserializeSearchParams, stripDomain };
```
EjPlatzer, line 174: I think you embedded this inside `searchParamSerializerFactory`.

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.js
```diff
@@ -18,6 +18,7 @@ import GordonDialogBox from 'components/GordonDialogBox';
 import { useParams } from 'react-router';
 import involvementService from 'services/activity';
 import membershipService from 'services/membership';
+import utils from 'services/utils';
```
EjPlatzer, line 21: Utils does not have a default export, you'll want to `import { stripDomain } from 'services/utils'`

EjPlatzer (changes requested):

EjPlatzer (approved):
LGTM!

PR #1534, opened 2022-10-08
Initial page routing logic

silasw2:
Contained in this PR,
- Default league and team pages to test if the routing has worked. These will be propagated with actual styling later
- React router dom page logic. New routes have been added linking to league and team page with the parameters for the page in the URL.
- Included `views` folder inside of `RecIM` to contain league, team, and match views within the RecIM section.


EjPlatzer:
It might make more sense to use a nested router in this case, since all the Rec IM passed will be nested under the Rec IM route. That way you can spare the tiring logic and passes for the Rec IM App out from the broader 360 routing.

silasw2:
Abstracted home page out and put routing at root level in RecIM component.

cpabbot (approved):
LGTM. This looks great! We mentioned considering using TeamName in URL later on if we keep it unique per league, but maybe not and either way will be post-MVP

PR #1538, opened 2022-10-12
Adjusted Routes to match DB/API, Hotfix on the league form not closing on confirmation submission

amos-cha:


cpabbot (commented):
This is good, but now things would be partially league and partially activity. Is this PR just to update the endpoint fetch stuff, and then plan to create a different PR to do all the other changes? That would be okay. If so, I can make a PR for that this evening if no one is working on it.

amos-cha:
> This is good, but now things would be partially league and partially activity. Is this PR just to update the endpoint fetch stuff, and then plan to create a different PR to do all the other changes? That would be okay. If so, I can make a PR for that this evening if no one is working on it.

that is perfectly fine, at some point someone just needs to implement the changes in this PR.

PR #1539, opened 2022-10-13
[Rec-IM] Refactored Listings

cpabbot:
### Simple PR:
Refactored TeamListing and LeagueListing into a RecIM component Listing. We could modify this to be a single default returned with a parameter specifying which type of listing it is, but this will depend on how similar they end up being. This simplification is super helpful though because now the different Listings can share a single stylesheet and some logic. They will also be reused in various parts of the RecIM platform, so it is helpful to have these scoped to RecIM, rather than to specific views.

Note: League will be renamed to Activity, but that will be separate PR later on


src/views/RecIM/components/Listing/index.js
```diff
@@ -0,0 +1,47 @@
...
+      </Link>
+    </>
+  );
+};
+
+// We could also use ParticipantID (not student ID) if we have that and prefer it to AD_Username
+const ParticipantListing = ({ username }) => {
+  return (
+    <>
+      <Link to={`/recim/profile/${username}`} className="gc360_link">
```
EjPlatzer, line 38: Question for your team: would it make sense to incorporate this RecIM user info into the regular user profile? I'm imagining a new card on the profile page that has RecIM info attached to, similar perhaps to the current Involvements list?

I don't know your requirements or what information is included in the RecIM user's profile. I do, though, think it would be nice to have one unified profile rather than dispersing a user's info across the site.

amos-cha: I'm not sure exactly why we have a participant profile page? I thought that we weren't going to have profiles for each person, one of the reasons against it was that we would most likely need to implement some sort of search feature for RecIM participants which would conflict/look weird with people search already existing in the 360 Header. 

Unless this is a typo and you meant to send directly to the 360 user profile (line 38). Though in that case, I would argue that it might be a hassle to need to get back to where you saw the people listing. 
i.e.
Press RecIM -> Activity -> Team -> Person

To answer @EjPlatzer's question, I don't think that is too hard to do. But it might be excess information that would further clutter the Profile view as it is. Moreover, we would have to handle cases where PERSON has not signed up for RecIM in which we might just have an empty card which might look awkward. I'm also pretty sure that our customer is not putting individual statistics high on their priority list, so the card would be relatively empty (Though this might be subject to change.)

Otherwise, @cpabbot code looks good.

amos-cha (approved):
Looks good! I mentioned my only issue in @EjPlatzer's comment.

src/views/RecIM/components/Listing/index.js
cpabbot: Whoops haha sorry to create confusion! @amos-cha is exactly right and this was just a typo 🙃. I will merge this and just change that in some other PR.

> "I would argue that it might be a hassle to need to get back to where you saw the people listing."

Good point too; you should just be able to click back to return but I get what you mean. Maybe in the future especially if we had some Rec-IM specific stuff we could have some kind of profile tooltip/popup card of some kind with Rec-IM stats and a link to 360 profile - and then this card component we could simply add to the 360 profile page as well if users wanted it. I'm going to put this idea in the fridge; for now, I think 360 profile linking will be MVP sufficient.

PR #1543, opened 2022-10-15
Finished team page layout and photo fetching for participants

silasw2:
This is what the current team page looks like (and subsequently what the participant listing looks like).

image

Somethings to consider in this PR:
- Maybe switch from the grid div system we have been using for listings and switch to MUI List as it would be more synonymous with the rest of 360
- Pass user id probably better than AD username, but this requires us have dummy data and a complete participant fetching protocol as we need more data than just user ID

### DIFF
--- a/src/views/RecIM/components/Listing/index.js
+++ b/src/views/RecIM/components/Listing/index.js
@@ -37,10 +37,10 @@ const TeamListing = ({ activityID, teamID }) => {
 const ParticipantListing = ({ username }) => {
   const [avatar, setAvatar] = useState('');
 
-  const [name, setName] = useState({
-    firstname: '',
-    lastname: '',
-  });
+  // const [name, setName] = useState({
+  //   firstname: '',
+  //   lastname: '',
+  // });
 
   useEffect(() => {
     const loadAvatar = async () => {
--- a/src/views/RecIM/components/Listing/index.js
+++ b/src/views/RecIM/components/Listing/index.js
@@ -1,6 +1,8 @@
-import { Grid } from '@material-ui/core/';
+import { Grid, Avatar, ListItemAvatar, ListItem, ListItemText } from '@material-ui/core/';
 import styles from './Listing.module.css';
 import { Link } from 'react-router-dom';
+import { useState, useEffect } from 'react';
+import user from 'services/user';
 
 const ActivityListing = ({ activityID }) => {
   return (
@@ -33,14 +35,37 @@ const TeamListing = ({ activityID, teamID }) => {
 
 // We could also use ParticipantID (not student ID) if we have that and prefer it to AD_Username
 const ParticipantListing = ({ username }) => {
+  const [avatar, setAvatar] = useState('');
+
+  const [name, setName] = useState({
+    firstname: '',
+    lastname: '',
+  });
+
+  useEffect(() => {
+    const loadAvatar = async () => {
+      if (username) {
+        const { def: defaultImage, pref: preferredImage } = await user.getImage(username);
+        setAvatar(preferredImage || defaultImage);
+      }
+    };
+    loadAvatar();
+  }, [username]);
   return (
-    <>
-      <Link to={`/profile/${username}`} className="gc360_link">
-        <Grid container className={styles.listing}>
-          <Grid item>Participant Listing</Grid>
-        </Grid>
-      </Link>
-    </>
+    <ListItem key={username} disableGutters={true}>
+      <Grid container alignItems="center" className={styles.listing}>
+        <ListItemAvatar>
+          <Avatar
+            src={`data:image/jpg;base64,${avatar}`}
+            className={styles.avatar}
+            variant="rounded"
+          ></Avatar>
+        </ListItemAvatar>
+        <Link to={`/profile/${username}`} className="gc360_link">
+          <ListItemText primary={username} />
+        </Link>
+      </Grid>
+    </ListItem>
   );
 };
 
--- a/src/views/RecIM/views/Team/index.js
+++ b/src/views/RecIM/views/Team/index.js
@@ -69,6 +69,9 @@ const Team = () => {
             {rosterCard}
           </Grid>
         </Grid>
+        <p>
+          Activity ID: {activityID} Team ID: {teamID} (for testing purposes only)
+        </p>
       </>
     );
   }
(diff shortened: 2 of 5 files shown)


src/views/RecIM/components/Listing/index.js
```diff
@@ -33,14 +35,37 @@ const TeamListing = ({ activityID, teamID }) => {
 
 // We could also use ParticipantID (not student ID) if we have that and prefer it to AD_Username
 const ParticipantListing = ({ username }) => {
+  const [avatar, setAvatar] = useState('');
+
+  // const [name, setName] = useState({
+  //   firstname: '',
+  //   lastname: '',
+  // });
```
amos-cha, line 43: I think the user hook has a fullName field that we could use for a clean first/last name. I did a little bit of digging and we can't really get the gordon id very easily from the hook so we might need to add an AD_Username to a Participant viewmodel.

src/views/RecIM/views/Team/index.js
```diff
@@ -1,12 +1,80 @@
...
+    // The user is not logged in
+    return <GordonUnauthorized feature={'the Rec-IM page'} />;
+  } else {
+    return (
+      <>
+        <Grid container alignItems="center" className={styles.activityHeader}>
+          <Grid item>
+            <img src={''} alt="Team Icon" width="85em"></img>
+          </Grid>
+          &nbsp;&nbsp;&nbsp;&nbsp;
```
amos-cha, line 59: Not sure if this is a temporary or permanent fix to the spacing issue? I'll assume it's temporary.

amos-cha (approved):
Looks good! I left a couple notes.

src/views/RecIM/views/Team/index.js
silasw2: This is a Cam question, as it was in his code for the activity/home page layout that I tried to mimic.

src/views/RecIM/components/Listing/index.js
silasw2: Good note for the future!

silasw2:
I am going to leave branch here for cam to review.

bennettforkner:
"Pass user id probably better than AD username, but this requires us have dummy data and a complete participant fetching protocol as we need more data than just user ID"

Hey @silasw2, I don't know if this is what you are suggesting, but we avoid using people's Gordon IDs anywhere we can in 360. Our end goal is that the only places we use IDs are where explicitly necessary such as showing the user their ID # on their profile. Since we have not followed the practice of creating a separate user ID just for 360, we use the account AD_Username as their identifier.

amos-cha:
> "Pass user id probably better than AD username, but this requires us have dummy data and a complete participant fetching protocol as we need more data than just user ID"
> 
> Hey @silasw2, I don't know if this is what you are suggesting, but we avoid using people's Gordon IDs anywhere we can in 360. Our end goal is that the only places we use IDs are where explicitly necessary such as showing the user their ID # on their profile. Since we have not followed the practice of creating a separate user ID just for 360, we use the account AD_Username as their identifier.

@bennettforkner if this is the case, would it be better to start identifying people in our Participant Table (back end) by AD_Username? or in the backend we should still be using Jenzabar ID's

bennettforkner:
> 
> @bennettforkner if this is the case, would it be better to start identifying people in our Participant Table (back end) by AD_Username? or in the backend we should still be using Jenzabar IDs


The answer is yes. If we were completely redesigning 360 or if it was worth it to do a complete overhaul of the backend, we would really just create our own unique keys (rather than using a string username) that we would use as identifiers everywhere. Unfortunately, that will likely never get the attention it would need as we do not have the bandwidth to address it. Therefore, we settle for what we can fix now, which is primarily using ad_username (or just "username") as the identifier between the UI and API. See [API #578](https://github.com/gordon-cs/gordon-360-api/issues/578) for details.

amos-cha:
> > @bennettforkner if this is the case, would it be better to start identifying people in our Participant Table (back end) by AD_Username? or in the backend we should still be using Jenzabar IDs
> 
> The answer is yes. If we were completely redesigning 360 or if it was worth it to do a complete overhaul of the backend, we would really just create our own unique keys (rather than using a string username) that we would use as identifiers everywhere.
> 
> Unfortunately, that will likely never get the attention it would need as we do not have the bandwidth to address it. Therefore, we settle for what we can fix now, which is primarily using ad_username (or just "username") as the identifier between the UI and API. See [API #578](https://github.com/gordon-cs/gordon-360-api/issues/578) for the raised issue about it.

Got it, I'll start re-factoring our schema and API accordingly then.

bennettforkner:
And I'm realizing that you were just talking about your participation table for RecIM. I would guess we want to keep it consistent, but I wonder what @EjPlatzer thinks.

bennettforkner:
> Got it, I'll start re-factoring our schema and API accordingly then.
I would wait for Evan to confirm. I'm not 100% sure on your schema specifically. @EjPlatzer, should they use GordonIDs, AD_Username, or a new key as their unique identifier for users in the recim tables?

silasw2:
> "Pass user id probably better than AD username, but this requires us have dummy data and a complete participant fetching protocol as we need more data than just user ID"
> 
> Hey @silasw2, I don't know if this is what you are suggesting, but we avoid using people's Gordon IDs anywhere we can in 360. Our end goal is that the only places we use IDs are where explicitly necessary such as showing the user their ID # on their profile. Since we have not followed the practice of creating a separate user ID just for 360, we use the account AD_Username as their identifier.

This is what I thought, but Amos had mentioned using ID as an identifier, but he might have made participant ID so maybe I misunderstood.

src/views/RecIM/views/Team/index.js
cpabbot: Haha yeah temporary, we can change it at some point.

EjPlatzer:
I'll add my two cents to the conversation even though this has been closed. I think Bennett was trying to clarify that we want to keep ID Numbers out of the Front End as much as possible - which means not using them to communicate with the API, in addition to stripping them from database objects before sending them to the UI, and not displaying them anywhere they aren't strictly necessary in the UI.

However, Gordon ID is still the preferred user identifier in the database. Luckily, the fix is pretty easy - use AD Username as an identifier in the UI and when communicating with the API. In the API, it is fairly trivial to accept AD Usernames and then get the ID from the username so that you can use the ID to communicate with the database.

TLDR: Avoid using/having Gordon ID in the UI. Use AD Username to identify users. In the API, accept AD Username but use Gordon ID.

PR #1546, opened 2022-10-17
update the profile PersonalInfoList props

Kevin-Chu58:
Issue:

> in `src/components/Profile/index.js`, `PersonalInfoList` has a **network** input, which is imported as `useNetworkStatus` from `hooks/useNetworkStatus`, but then inside `PersonalInfoList/index.js`, there's no props for **network**, and `useNetworkStatus` from `hooks/useNetworkStatus` was imported again.


src/components/Profile/components/PersonalInfoList/index.js
```diff
@@ -38,7 +37,7 @@ const formatPhone = (phone) => {
   }
 };
 
-const PersonalInfoList = ({ myProf, profile, createSnackbar }) => {
+const PersonalInfoList = ({ myProf, profile, network, createSnackbar }) => {
```
bennettforkner, line 40:
```suggestion
const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }) => {
```
It might be clearer to name this property in a boolean-ish way rather than just "network". Could you change this in the 3+ places where it appears?

bennettforkner: Also @EjPlatzer , is it better to use the hook everywhere or is it alright to pass the value from higher level components down.

Kevin-Chu58: This Evan's response to my concern with the duplicated imports, then he told me the imports are find, but not calling the hooks multiple times. I also checked out other components inside Profile, the Identification also uses network prop.
image

bennettforkner: Great, that answers that then. Just the change to `isOnline` for my recommendation then.

Kevin-Chu58: image
I don't think that works, because isOnline has already declared as a const.

bennettforkner: If you delete the line for defining the const after updating the property name, it should work.

bennettforkner:
It seems that network is not actually what you think it is here. If you look at line 21 of [src/components/Profile/index.js](https://github.com/gordon-cs/gordon-360-ui/pull/1546/files#diff-8822edc1d3d5441a27c2284d3e26fc1a0eaf07110853a57776f5c95ffa2c92e7), you can see that network is actually a string that is either set to `"online"` or `"offline"`. This means that if you treat it as a boolean value (line 190 of [src/components/Profile/components/PersonalInfoList/index.js](https://github.com/gordon-cs/gordon-360-ui/pull/1546/files#diff-5e1568f32be0ab5bc3733113538070917e4d1b8d357aee2b6827b6a30d209d1f)), it will always be `true`.

I would recommend going up the chain and modifying [src/components/Profile/index.js](https://github.com/gordon-cs/gordon-360-ui/pull/1546/files#diff-8822edc1d3d5441a27c2284d3e26fc1a0eaf07110853a57776f5c95ffa2c92e7) to pass `isOnline` to the `<PersonalInfoList>` component rather than passing `network`.

bennettforkner:
I hate to make this a bigger issue and add more work to a small PR, but there are 3 other cases where a `network` is being used unnecessarily when it could just be `isOnline` passed to the child components:
image

The children are all handling network as a string and checking if it is equal to `online` like you can see below. It would be much simpler to use the boolean `isOnline` value and eliminate `network` completely as it is not much of a descriptive term.
image

Your choice whether or not to address those ones. It would be some nice code cleanup if we were able to do it.

bennettforkner (changes requested):
See my earlier comment concerning the value of `network`. The code will currently not work as intended.

src/components/Profile/components/Identification/index.js
```diff
@@ -648,7 +648,7 @@ const Identification = ({ profile, myProf, network, createSnackbar }) => {
                       alt="Profile"
                     />
 
-                    {network === 'online' && myProf && (
+                    {isOnline === 'online' && myProf && (
```
bennettforkner, line 651: Since isOnline is a boolean, you no longer have to compare it to a string value `"online"`
```suggestion
                    {isOnline && myProf && (
```

src/components/Profile/components/Identification/index.js
```diff
@@ -659,7 +659,7 @@ const Identification = ({ profile, myProf, network, createSnackbar }) => {
                       </Typography>
                     )}
                   </div>
-                  {network === 'online' && myProf && (
+                  {isOnline === 'online' && myProf && (
```
bennettforkner, line 662:
```suggestion
                  {isOnline && myProf && (
```

src/components/Profile/components/Identification/index.js
```diff
@@ -753,15 +753,15 @@ const Identification = ({ profile, myProf, network, createSnackbar }) => {
                   </a>
                 </Grid>
 
-                {network === 'online' && createPhotoDialogBox()}
+                {isOnline === 'online' && createPhotoDialogBox()}
```
bennettforkner, line 756:
```suggestion
                {isOnline && createPhotoDialogBox()}
```

src/components/Profile/components/Identification/index.js
```diff
@@ -753,15 +753,15 @@ const Identification = ({ profile, myProf, network, createSnackbar }) => {
...
               </Grid>
             </Grid>
           </Grid>
         ) : (
           <GordonLoader />
         )}
 
-        {userProfile && network === 'online' && myProf && (
+        {userProfile && isOnline === 'online' && myProf && (
```
bennettforkner, line 764:
```suggestion
        {userProfile && isOnline && myProf && (
```

bennettforkner (changes requested):
This is looking great. Just one more update in a few places to finish cleaning up!

src/components/Profile/components/Identification/index.js
Kevin-Chu58: You are right, I missed this one.

bennettforkner (approved):
Looks good to me, assuming the GitHub actions checks pass.

PR #1549, opened 2022-10-28
Update lists to MUI and abstract List Components

silasw2:
This PR abstracts the listing we were doing with divs, to using MUI lists to link our pages together. It also moves these lists to a component file so they can be reused in different sections of the RecIM pages. Here is a picture of what it looks like with the new lists,

image

Ignore the PR I accidentally put into the merge into develop category!

### DIFF
--- a/src/views/RecIM/components/List/Listing/index.js
+++ b/src/views/RecIM/components/List/Listing/index.js
@@ -4,37 +4,38 @@ import { Link } from 'react-router-dom';
 import { useState, useEffect } from 'react';
 import user from 'services/user';
 
-const ActivityListing = ({ activityID }) => {
+const ActivityListing = ({ activity }) => {
   return (
-    <>
-      <Link to={`/recim/activity/${activityID}`} className="gc360_link">
-        <Grid container className={styles.listing}>
-          <Grid item>Activity Listing</Grid>
-          {/* include: 
+    <ListItem button component={Link} to={`/recim/activity/${activity.ID}`} className="gc360_link">
+      <Grid container className={styles.listing}>
+        <Grid item>Activity Listing</Grid>
+        {/* include:
           - activity type (activity, tournament, one-off)
           - registration deadline IF there is one (start date as well for admin only)
           - date(s) of activity (ex. season date range or tournament date)
           */}
-        </Grid>
-      </Link>
-    </>
+      </Grid>
+    </ListItem>
   );
 };
 
-const TeamListing = ({ activityID, teamID }) => {
+const TeamListing = ({ team }) => {
   return (
-    <>
-      <Link to={`/recim/activity/${activityID}/team/${teamID}`} className="gc360_link">
-        <Grid container className={styles.listing}>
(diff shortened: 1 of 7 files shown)


src/views/RecIM/components/List/Listing/index.js
```diff
@@ -61,23 +64,26 @@ const ParticipantListing = ({ username }) => {
...
+    <ListItem
+      button
+      component={Link}
+      to={`/recim/activity/${match.activityID}/match/${match.ID}`}
+      className="gc360_link"
+    >
+      <Grid container className={styles.listing}>
+        <Grid item>Team A vs Team B</Grid>
+      </Grid>
+    </ListItem>
```
amos-cha, line 86: Next PR should have some of these details fleshed out.
i.e.
Team A vs Team B can already be detailed to Match.Team[0].Name vs Match.Team[1].Name for any match instance that is not Ladder styled. You could probably even handle this with a foreach with a catch the last element in the array.

src/views/RecIM/components/List/index.js
```diff
@@ -0,0 +1,24 @@
...
+  let content = matches.map((match) => <MatchListing match={match} />);
+  return <List>{content}</List>;
+};
+
+const TeamList = ({ teams }) => {
+  let content = teams.map((team) => <TeamListing team={team} />);
+  return <List>{content}</List>;
+};
+
+export { ActivityList, ParticipantList, MatchList, TeamList };
```
amos-cha, line 24: I want to say you could make this cleaner by making a generalized List component for us or maybe make our Listing component more generalized/simplified. 

How you could do this would be via:

_in the Listing component_
Changing the ______Listing to Listing = ({data, type}) 
^data would be the full array of objects
using a switch case on type to determine if it is a Participant,Match,Activity...so on and so forth

Then you could probably just do the mapping within either an overhead function in Listing or somewhere else.

src/views/RecIM/views/Activity/index.js
```diff
@@ -32,8 +27,12 @@ let teamsCard = (
...
     <CardContent>
       {/* if I am apart of any active teams, map them here */}
-      <div className={styles.listing}>{/* <TeamListing activityID={123456} teamID={789} /> */}</div>
-      <div className={styles.listing}>{/* <TeamListing activityID={12345} teamID={987} /> */}</div>
+      <TeamList
+        teams={[
+          { activityID: '123456', ID: '789' },
+          { activityID: '12345', ID: '987' },
+        ]}
+      />
```
amos-cha, line 35: same comment on match

amos-cha (approved):
Looks good! but There are a few things I prefer to see for the next PR (mainly the generalized listing) you could either modify it and I'll re-review it on this PR or it can come in the next one.

src/views/RecIM/components/List/index.js
silasw2: I think this would almost be the same amount of code with less clarity. Abstracting like this also gives us flexibility if a component needs react state or other hooks.

PR #1550, opened 2022-10-31
[Rec-IM] Merge in develop to upgrade to MUI5

cpabbot:
## Merge in develop
Upgrades MUI to v5, which will be helpful for our use of the new DateTime picker! Plus other features/changes.

**_NOTES:_**
- **Important:** Had to replace the proxy stuff that @EjPlatzer removed in commit https://github.com/gordon-cs/gordon-360-ui/pull/1536/commits/10d8f74ff05a12f3786f4aadd168fe9f94557e61 for unknown reason that broke stuff with CORS. (I'm sure there's in reality a long conversation here but for now this works and @EjPlatzer can discuss this with us later)
- There was surprisingly only one merge conflict, which was due to the fact that Material UI imports are now `mui/...` in place of `material-ui/...`


silasw2 (approved):
It's good to merge develop every so often, thanks for doing this! LGTM!

silasw2 (changes requested):
People tab is not working currently but that is because train is not working either.

silasw2 (approved):
LGTM!

PR #1554, opened 2022-11-02
[Rec-IM] Basic instructions started for SSH tunneling

cpabbot:
## Documentation update: local front end, VM backend

*****THIS IS A DRAFT*****

By no means finished nor complete. Creating draft PR so it will be easily viewable. @silasw2 should add instructions for SSH Tunneling at some point and we might also want to consider putting this in a separate doc.


silasw2:
@cpabbot What do think is left to add to this branch?

bennettforkner:
This could probably be merged directly into develop when approved. Since it doesn't have anything specifically to do with the RecIM project, do you want to update the PR destination to develop?

PR #1556, opened 2022-11-05
[Rec-IM] MUI5 DateTime Picker Implementation

cpabbot:
## Rec-IM MUI5 DateTime Picker

Implemented the MUI5 DateTime Picker as an InformationField for use in the CreateActivity form. Note that MUI does not return an event onChange but rather its value, so I passed an additional optional parameter to specify the target source in these cases.


amos-cha (approved):
I like it! Excited to see what the new Activity creation looks like :)

PR #1558, opened 2022-11-06
Gordon RecIM 22-23 (alpha release)

amos-cha:
# Gordon RecIM Documentation :bouncing_ball_person:	

### 22-23 Gordon College CS Senior Project [UI] <br>
@cpabbot , @amos-cha , @Kevin-Chu58 , @silasw2   <br>

## Development and Documentation Links
[Project Overview](https://docs.google.com/document/d/1nPTZvDK5QJR_4jwQ1q15j0-IzSj0WXo407QA76Hncyw/edit?usp=sharing)<br>
[Project Problem Statement](https://docs.google.com/document/d/1U6wZTeWqoG7LTugwx-uXYL68V5OYWVNPbZAH16Tqfsw/edit?usp=sharing) <br>
[Project Design Document](https://docs.google.com/document/d/1TRP1-OYy-Bnh1vrc2JHyZX3OFJ61amXItTHDSanf3ok/edit?usp=sharing) <br>
[End-to-End Document ](https://docs.google.com/document/d/13W2WR0zIey6RegaFtyP7E2-btKdQihZqECBgGDiShQ8/edit?usp=sharing) <br>


.env.development
```diff
@@ -4,7 +4,10 @@
 # VITE_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
- VITE_API_URL=https://360ApiTrain.gordon.edu/
+# VITE_API_URL=https://360ApiTrain.gordon.edu/
+
+# @RECIM
+VITE_API_URL=https://360ApiRecIM.gordon.edu/
```
amos-cha, line 10: This will have to be reverted before merge

src/views/RecIM/views/Admin/index.jsx
```diff
@@ -0,0 +1,102 @@
...
+      setParticipants(await getParticipants());
+    };
+    loadActivities();
+    loadTeams();
+    // if you don't do this, you can see every participant via looking at the network tab
+    // on console, feel free to add the loadActivities/Teams to this 'if' statement if you desire
+    // but at the very least, participants need to be hidden
+    if (user?.IsAdmin) {
+      loadParticipants();
+    }
```
EjPlatzer, line 62: 1. It would be more efficient to load these in parallel using `Promise.all`.
2. If `loadActivities` and `loadTeams` don't need to re-run when `user` changes, they should be in a separate `useEffect`.

src/views/RecIM/views/Home/index.jsx
```diff
@@ -0,0 +1,276 @@
...
+      setActivities(await getActivities());
+      setInvites(await getTeamInvites());
+      if (profile) {
+        setParticipant(await getParticipantByUsername(profile.AD_Username));
+        setMyTeams(await getParticipantTeams(profile.AD_Username));
+      }
+      setLoading(false);
+    };
+    loadData();
+  }, [profile, openActivityForm, openWaiver, openCreateSeriesForm]);
```
EjPlatzer, line 78: Do you need to reload all this data every time any of these changes?

src/views/RecIM/components/Forms/WaiverForm/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+          incomplete or incorrect instruction, and of impacts with other persons or equipment. This
+          agreement is made in sole consideration of Gordon College permitting my use of the Bennett
+          Center or other Gordon facilities, equipment, or services. I agree to comply with any
+          Gordon College and Bennett Center rules and regulations. Hours, services and policies are
+          subject to change, without prior notice, in the sole discretion of Gordon College. I agree
+          to accept such reasonable changes as a condition of membership. I certify that I have read
+          this application in its entirety and that by signing below I agree to be bound by all of
+          its terms and conditions. My signature below provides permission and waives/release claims
+          for any minor participating because of my membership.
+          {''}
```
EjPlatzer, line 184:
```suggestion
```

src/views/RecIM/views/Team/index.jsx
```diff
@@ -0,0 +1,194 @@
...
+    const loadTeamData = async () => {
+      setLoading(true);
+      setTeam(await getTeamByID(teamID));
+      if (profile) {
+        setUser(await getParticipantByUsername(profile.AD_Username));
+      }
+      setLoading(false);
+    };
+    loadTeamData();
+  }, [profile, teamID, openTeamForm, openInviteParticipantForm]);
```
EjPlatzer, line 40: This looks like it should be two separate `useEffect`s, one for participant and one for team.

src/views/RecIM/views/Team/index.jsx
```diff
@@ -0,0 +1,194 @@
...
+  // @TODO modify above dependency to only refresh upon form submit (not cancel)
+
+  //checks if the team is modifiable by the current user
+  useEffect(() => {
+    let hasCaptainPermissions = false;
+    if (user) {
+      if (team) {
```
EjPlatzer, line 47:
```suggestion
    if (user && team) {
```

src/views/RecIM/views/Team/index.jsx
```diff
@@ -0,0 +1,194 @@
...
+        let role =
+          team.Participant.find((teamParticipant) => teamParticipant.Username === user.Username)
+            ?.Role ?? 'Invalid';
```
EjPlatzer, line 50: Since `role` is only referenced here, there's no need to provide a default value, just let it be undefined.
```suggestion
        const role =
          team.Participant.find((teamParticipant) => teamParticipant.Username === user.Username)
            ?.Role;
```

EjPlatzer (changes requested):
This is looking pretty good. I am eagerly awaiting the form refactoring you mentioned. I might have a little bit more feedback after we do a full demo, but for now this looks pretty good!

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -0,0 +1,358 @@
...
+
+const ActivityListing = ({ activity }) => {
+  const [activityType, setActivityType] = useState();
+  const [currentCapacity, setCurrentCapacity] = useState(<GordonLoader size={15} inline />);
+  useEffect(() => {
+    const loadActivityType = async () => {
+      let activityTypes = await getActivityTypes();
+      setActivityType(
+        activityTypes.find((activityType) => activityType.ID === activity.TypeID).Description,
+      );
```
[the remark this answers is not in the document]
amos-cha: This was a mistake on the API's part which is a part of the new API refactor

.env.production
```diff
@@ -5,4 +5,7 @@
 # VITE_ANALYTICS_ID=UA-3559968-3
 
 # @TRAIN
-VITE_API_URL=https://360ApiTrain.gordon.edu/
+# REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+
+# @RECIM
+REACT_APP_API_URL=https://360ApiRecIM.gordon.edu/
```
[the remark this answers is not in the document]
amos-cha: made a note up top

src/views/RecIM/views/Activity/components/ScheduleList/index.jsx
```diff
@@ -0,0 +1,230 @@
...
+        </Typography>
+      </Typography>,
+    );
+    setOpenAutoSchedulerDisclaimer(true);
+    handleClose();
+  };
+
+  const handleConfirmAutoSchedule = () => {
+    scheduleSeriesMatches(series.ID).then((res) => {
+      console.log(res);
```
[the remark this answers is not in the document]
amos-cha: the console.logs are actually just placeholders for the return with snackbar instances (they're in progress)

src/views/RecIM/components/Forms/components/ConfirmationHeader/ConfirmationHeader.module.scss
```diff
@@ -0,0 +1,16 @@
...
+  padding-left: 5px;
+  padding-right: 5px;
+}
+
+.header_style {
+  color: $primary-blue;
+  @extend %header_style;
+
+  &_text {
+    color: #9f9f9d;
```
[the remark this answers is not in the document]
amos-cha: noted. I think this was a remanent of the AlumniUpdateForm which has the same .scss file. A change might have to be made there.

src/services/recim/activity.ts
```diff
@@ -32,19 +34,26 @@ type CreatedActivity = BaseActivity & {
...
+  SportID: number;
+  TypeID: number;
   StatusID: number;
   Logo: string;
   Completed: boolean;
 };
 
 //Activity Routes
-const createActivity = (newActivity: UploadActivity): Promise<CreatedActivity> =>
+const createActivity = async (newActivity: UploadActivity): Promise<CreatedActivity> =>
```
EjPlatzer, line 51: You don't need the `async` here, since you aren't using `await` at all. Since `http.get` already returns a Promise, it is already asynchronous by default. In fact, adding `async` here is purely a performance reduction because it adds another, unnecessary Promise wrapper around the `http.get` Promse.

src/services/recim/match.ts
```diff
@@ -37,35 +37,72 @@ type CreatedMatchStats = {
...
+const updateMatchAttendance = (
+  attendance: UploadMatchAttendance,
+  matchID: number,
+): Promise<Attendance[]> => http.put(`recim/matches/${matchID}/attendance`, attendance);
+
+const updateAttendance = (matchID: number, attendance: Attendance): Promise<Attendance> =>
+  http.post(`recim/matches/${matchID}/attendance`, attendance);
+
+const removeAttendance = (matchID: number, attendance: Attendance): Promise<Attendance> =>
+  http.del(`recim/matches/${matchID}/attendance`, attendance);
```
EjPlatzer, line 102: Does this route actually need the entire `Attendance` object in the body? Typically, you just need the ID of the attendance to be deleted.

src/views/RecIM/components/Forms/ActivityForm/index.jsx
```diff
@@ -168,9 +168,9 @@ const ActivityForm = ({
         registrationStart: activity.RegistrationStart,
         registrationEnd: activity.RegistrationEnd,
         typeID:
-          activityTypes.find((type) => type.ID === activity.TypeID) == null
+          activityTypes.find((type) => type.Description === activity.Type) == null
             ? ''
-            : activityTypes.find((type) => type.ID === activity.TypeID).Description,
+            : activityTypes.find((type) => type.Description === activity.Type).Description,
```
EjPlatzer: Firstly, you can make this more efficient:
```suggestion
          activityTypes.find((type) => type.Description === activity.Type)?.Description ?? '',
```

But secondly, do you need to search all activityTypes? Can't you just use `activity.Type` directly now?
```suggestion
	          activity.Type,
```

src/views/RecIM/components/Forms/MatchForm/index.jsx
```diff
@@ -132,15 +127,15 @@ const MatchForm = ({ closeWithSnackbar, openMatchForm, setOpenMatchForm, activit
...
         StatusID:
           matchStatus.find((type) => type.Description === match.Status) == null
             ? ''
             : matchStatus.find((type) => type.Description === match.Status).Description,
         SurfaceID:
-          surfaces.find((type) => type.Description === match.Surface) == null
+          surfaces.find((type) => type.Name === match.Surface) == null
             ? ''
-            : surfaces.find((type) => type.Description === match.Surface).Description,
+            : surfaces.find((type) => type.Name === match.Surface).Name,
```
EjPlatzer: As with activityTypes above, can't you use `match.Surface` directly now?:
```suggestion
          match.Surface,
```

EjPlatzer (changes requested):

src/components/Header/index.jsx
```diff
@@ -211,6 +213,7 @@ const GordonHeader = ({ onDrawerToggle }) => {
               />
               {requiresAuthTab('People', <PeopleIcon />)}
               {requiresAuthTab('Timesheets', <WorkIcon />)}
+              {requiresAuthTab('Rec-IM', <RecIMIcon />)}
```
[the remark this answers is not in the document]
amos-cha: Yep, we have https://github.com/gordon-cs/gordon-360-ui/pull/1726 ready to merge into this branch just before the push.

src/components/Nav/components/NavLinks/index.jsx
```diff
@@ -173,6 +186,7 @@ const GordonNavLinks = ({ onLinkClick }) => {
         {eventsButton}
         {peopleButton}
         {timesheetsButton}
+        {recimButton}
```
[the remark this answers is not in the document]
amos-cha: mentioned above, but re-linking just incase we lose these threads
https://github.com/gordon-cs/gordon-360-ui/pull/1726

src/views/RecIM/components/Forms/ActivityForm/index.jsx
```diff
@@ -0,0 +1,393 @@
...
+    ).ID;
+    if (activity) {
+      activityRequest.statusID = activityStatusTypes.find(
+        (type) => type.Description === activityRequest.statusID,
+      ).ID;
+      editActivity(activity.ID, activityRequest).then((res) => {
+        setSaving(false);
+        closeWithSnackbar({
+          type: 'success',
+          message: 'Your new activity has been created or whatever message you want here',
```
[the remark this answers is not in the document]
amos-cha: The close with snackbar "sassy" comments were a placeholder for actually returning proper responses from the API. They were put in when I made placed the skeleton of the form in the UI back in October/November. At that point we had no API response so what you see now is residual snackbar information that has not been propagated. 

This, to me, is a blocking change that we need to make before we merge develop as we don't want errors hanging the UI.

src/services/recim/team.ts
```diff
@@ -0,0 +1,133 @@
...
+const getTeamInvites = async (): Promise<Team[]> => http.get(`recim/Teams/invites`);
+
+const respondToTeamInvite = async (
+  teamID: number,
+  response: string,
+): Promise<CreatedTeamParticipant> => http.patch(`recim/Teams/${teamID}/invite/status`, response);
+
+//temporary solution, may need a cleaner route implementation or have attendance count return in team
+const getParticipantAttendanceCountForTeam = (teamID: number, username: string): Promise<number> =>
+  http.get(`recim/Teams/${teamID}/attendance?username=${username}`);
```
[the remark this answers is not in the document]
amos-cha: this is an API change that can be handled. 👍🏼

src/services/recim/team.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+const editTeamParticipant = async (
+  teamID: number,
+  editedParticipant: UploadTeamParticipant,
+): Promise<CreatedTeamParticipant> =>
+  await http.patch(`recim/Teams/${teamID}/participants`, editedParticipant);
+
+const deleteTeamParticipant = async (teamID: number, username: string) =>
+  await http.del(`recim/Teams/${teamID}/participants?username=${username}`);
```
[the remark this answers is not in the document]
amos-cha: ditto below

src/services/recim/series.ts
```diff
@@ -0,0 +1,118 @@
...
+const getAllSeries = (): Promise<Series[]> => http.get(`recim/series`);
+
+const editSeries = (seriesID: number, updatedSeries: PatchSeries): Promise<CreatedSeries> =>
+  http.patch(`recim/series/${seriesID}`, updatedSeries);
+
+const getSeriesSchedule = (seriesID: number): Promise<SeriesSchedule> =>
+  http.get(`recim/series/${seriesID}/schedule`);
+
+const putSeriesSchedule = async (schedule: UploadSeriesSchedule): Promise<SeriesSchedule> =>
+  http.put(`recim/series/schedule`, schedule);
```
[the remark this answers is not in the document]
amos-cha: I was made to believe that put and post had 1 key difference where put can be executed in succession multiple times without side effect, while post would obviously make _n_ number of objects. Is there a reason that you need to specify a location for put? I also can't find anything on that from googling.

src/services/recim/activity.ts
```diff
@@ -0,0 +1,89 @@
...
+  http.post('recim/activities', newActivity);
+
+const deleteActivity = async (ID: number): Promise<CreatedActivity> =>
+  http.del(`recim/activities/${ID}`);
+
+const getActivityByID = (ID: number): Promise<Activity> => http.get(`recim/activities/${ID}`);
+
+const getActivities = (active: boolean, time: String): Promise<Activity[]> => {
+  if (time) return http.get(`recim/activities?active=${active}&time=${time}`);
```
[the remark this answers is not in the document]
amos-cha: https://github.com/gordon-cs/RecIM-Docs/issues/171

src/views/RecIM/components/Forms/ActivityForm/index.jsx
amos-cha: @EjPlatzer This is actually a reverse null coalescence. I want it to be an empty string **if** it's null

EjPlatzer: > @EjPlatzer This is actually a reverse null coalescence. I want it to be an empty string **if** it's null

```javascript
activity.Type ?? ''
```

amos-cha: Actually no you're right, I think this was the workaround for the janky ID logic I had before

src/views/RecIM/components/List/Listing/Listing.module.scss
```diff
@@ -24,22 +25,31 @@
...
 .acceptButton {
-  background-color: #007000;
+  background-color: $primary-cyan;
   color: $neutral-white;
 }
 
 .acceptButton:hover {
-  background-color: #009000;
+  background-color: $primary-cyan-shade;
 }
```
EjPlatzer: You shouldn't need all these button styles, you can just the built-in MUI color props. `reject` and `accept` can definitely be replaced with `color="primary"` and `color="secondary"` respectively. I can't remember if we have the red color setup as a button color yet, but if not you can easily add it.

src/_vars.scss
```diff
@@ -5,9 +5,11 @@
 // Ensure that the relative path (i.e. `../../`, etc.) is correct when importing from a nested
 // component.
 
-// For use in a .js file, see `src/theme.js`.
+// For use in a .ts file, see `src/theme.ts`.
 $primary-blue: #014983;
+$primary-blue-tint: #016bc0;
 $primary-cyan: #00aeef;
+$primary-cyan-shade: #0098d0;
```
EjPlatzer, line 12: It looks like you're only using these colors for the button hover shades, but MUI will take care of that for you if you use the built-in color props as mentioned in my other comment.

EjPlatzer: So please remove these essentially redundant colors.

src/views/RecIM/components/List/Listing/Listing.module.scss
```diff
@@ -1,3 +1,4 @@
+@use 'sass:color';
```
EjPlatzer, line 1: What is this used for?

EjPlatzer (changes requested):

src/views/RecIM/components/List/Listing/Listing.module.scss
amos-cha: Not used anymore, it was originally used to pass a weird linting check where we couldn't use `mix`. I wanted to make a specific hover color which is why you see the residual changes on blue-tint and shades. I'll remove it.

src/views/RecIM/components/List/Listing/Listing.module.scss
cpabbot: I see what you're saying here; it probably would be advantageous to use MUI styling as you suggest because that will automatically give us the correct hover tints and shades. The reason we did not do that originally was because MUI IconButton color attribute refers to the color of the icon itself and not the background color. We can change this by swapping to MUI Button, which will still require custom styling (to make the button circular) but will remove the need for those extra colors.

amos-cha: Is this true? The MUI IconButton takes `color` as a parameter (true enough) but this changes the `color` property as expected which is in this case would have been the little `✔️` or `✖️` inside of the circular icon. Whereas I want to modify the outside circular button and maintain the white   `✔️` or `✖️` .

(@cpabbot commented, I forgot to submit this comment)

EjPlatzer: Thanks @cpabbot and @amos-cha for clarifying - I didn't realize this was being used only for Icon Buttons. It's true that MUI doesn't currently support outlined Icon Buttons - that is a limitation of Material Design v2. It's been fixed by Material Design v3, but MUI is not being quick to adopt it (for a number of reasons that we don't need to get into here).

For now, I think what you have is a fine way to achieve a contained icon button. Maybe you could consider renaming the classes to clarify that they're for icon buttons only?

src/_vars.scss
EjPlatzer: As mentioned above, these are needed to tint contained Icon Buttons on hover.

However, there is still some inconsistency to clean up here. Why is the blue a `tint` but the cyan is  `shade`?

Did you have a specific reason for darkening the blue on hover but brightening the cyan?

If we are going to add these as global colors, they at least need to be consistent and clearly named.

amos-cha: Answering the last question first:
It was more of a design choice where the `cyan` was too bright to make brighter on hover (just hurt my eyes) and `blue` was too dark to darken on hover (just not enough contrast, and the problem is exacerbated by its proximity to other `blue` items), thus I thought a tint would give it the necessary _pop_ for visual clarity. 

Regarding why it's a tint and cyan is a shade: the blue tint is literally a lighter version (higher up on theme.tx) and the cyan shade is a darker version. Thus, it just made more sense to me to name the blue a tint since it's not a shade of `primary-blue`

src/services/recim/activity.ts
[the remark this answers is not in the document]
bennettforkner: @amos-cha I cannot view this issue. It gives me a 404.

src/_vars.scss
EjPlatzer: > Answering the last question first: It was more of a design choice where the `cyan` was too bright to make brighter on hover (just hurt my eyes) and `blue` was too dark to darken on hover (just not enough contrast, and the problem is exacerbated by its proximity to other `blue` items), thus I thought a tint would give it the necessary _pop_ for visual clarity.

That's what I figured, thanks.

> Regarding why it's a tint and cyan is a shade: the blue tint is literally a lighter version (higher up on theme.tx) and the cyan shade is a darker version. Thus, it just made more sense to me to name the blue a tint since it's not a shade of `primary-blue`

Firstly, it's not obvious that "tint" and "shade" mean something different here - they seem like arbitrary synonyms rather than a meaningful technical distinction.

Secondly, both of these colors are shades of the respective base color in the sense that they appear in the `blueShades` and `cyanShades` list in our theme. The `$primary-blue-tint` color is [`gordonColors.primary.blueShades.300`](https://github.com/gordon-cs/gordon-360-ui/blob/068ca7ae435a0be1c24d7f37ac8b8a2848cc6f38/src/theme.ts#L12), and the `$primary-cyan-shade` is [`gordonColors.secondary.cyanShades.600`](https://github.com/gordon-cs/gordon-360-ui/blob/068ca7ae435a0be1c24d7f37ac8b8a2848cc6f38/src/theme.ts#L33). In my opinion, these should be called `$primary-blue-300` and `$primary-cyan-600` for full clarity.

amos-cha: In respect to the current `theme.ts` I agree that naming it `$primary-blue-___` & `primary-cyan-____` would make the most sense. I'll make that change. 👍🏼 

However, just as a personal note, looking back at `theme.ts`, the term `shade` is not accurate for half those values. Shades are meant to be explicitly darker than a specified color as in art they're used as mixtures of black + color. Whereas, tints are meant to be the opposite, being white + color. So for pure terminology accuracy [gordonColors.primary.blue](https://github.com/gordon-cs/gordon-360-ui/blob/068ca7ae435a0be1c24d7f37ac8b8a2848cc6f38/src/theme.ts#L7) should be [gordonColors.primary.blueShades.50](https://github.com/gordon-cs/gordon-360-ui/blob/068ca7ae435a0be1c24d7f37ac8b8a2848cc6f38/src/theme.ts#L9). Though, obviously it's the center hue since it would save time and characters to not have to specify `primary.blueShades.500` so I get that.

src/views/RecIM/components/List/Listing/Listing.module.scss
amos-cha: renamed them to specify icon: i.e `acceptIcon`

src/_vars.scss
EjPlatzer: > However, just as a personal note, looking back at `theme.ts`, the term `shade` is not accurate for half those values. Shades are meant to be explicitly darker than a specified color as in art they're used as mixtures of black + color. Whereas, tints are meant to be the opposite, being white + color.

This is true when using "shade" and "tint" as technical terms in the fields of art and color theory. However, it is not true in general English usage. More importantly for this context, it is common practice in UI Design to refer to the "shades" (sometimes also variants) of a theme color, where "shade" has the meaning "a color, especially with regard to how light or dark it is or as distinguished from one nearly like it." (cf. https://www.google.com/search?q=shade+meaning).

> So for pure terminology accuracy [gordonColors.primary.blue](https://github.com/gordon-cs/gordon-360-ui/blob/068ca7ae435a0be1c24d7f37ac8b8a2848cc6f38/src/theme.ts#L7) should be [gordonColors.primary.blueShades.50](https://github.com/gordon-cs/gordon-360-ui/blob/068ca7ae435a0be1c24d7f37ac8b8a2848cc6f38/src/theme.ts#L9). Though, obviously it's the center hue since it would save time and characters to not have to specify `primary.blueShades.500` so I get that.

I'm not exactly sure what your point here is: `gordonColors.primary.blue` and `gordonColors.primary.blueShades.500` are the same, with the former being the short-hand for the latter. Although in a more important sense the former is more accurate because it is actually the base color, and the other shades are simply variations on that theme color.

src/views/RecIM/components/List/Listing/Listing.module.scss
cpabbot: Ideal would be acceptIconButton (distinct from an icon) but regardless I don't think the exact name here is super critical. These aren't global styles and there is no need to be overly detailed naming selectors beyond what it intends to target (the specific accept button on this page).

src/_vars.scss
amos-cha: 1. fair enough
2. was more so referring to the event that `blueShade.50` became the new `primary.blue` which would satisfy the art definition of `shade` 

but neither here nor there, I made the adjustment to refer directly to the name in `theme.ts` which should setup a framework for future colors to be added 😎. Should show up on the next PR

src/app.module.scss
```diff
@@ -15,9 +15,6 @@
   padding: 1rem 0.5rem;
   width: 100%;
   background-color: $neutral-light-gray;
-  &_container {
-    height: 100%;
-  }
```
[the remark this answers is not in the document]
cpabbot: I remember this line created an overflow issue. I tried to test this but now can't find where this selector is used at all and got very confused, and then found that 360 development team actually removed the selector in develop several months back. See commit https://github.com/gordon-cs/gordon-360-ui/commit/ccdb54904c180eb3c91fe0cd645d1db069f28f46, not sure which PR this was apart of. So this problematic line actually now has no significance at all as there is no longer any `app_main_container` in place.

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -0,0 +1,358 @@
...
+          <Grid item>
+            {match.Team[0]?.Name ?? <i>TBD</i>} vs. {match.Team[1]?.Name ?? <i>TBD</i>}
+          </Grid>
+        </Grid>
+      </ListItemButton>
+    </ListItem>
+  );
+};
+
+export { ActivityListing, TeamListing, ParticipantListing, MatchListing };
```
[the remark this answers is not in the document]
cpabbot: The reason we did this initially was so that they could share styling and common features, but now that they have become more complex, I think we may be able to factor these out and use `Listing` as a shell component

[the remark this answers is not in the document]
amos-cha: I feel like refactoring the Listing as a shell component might be difficult due to the intricacies of each listing component with each listing having a completely different grid layout to another one. The only one that is remotely similar is Participant and Team, but even within Team there are all the record types on the right side. Perhaps we could create a new file for each but it would make it look incredibly bloated in terms of file numbers. @EjPlatzer is this more of a stylistic change or is there a specific reason you want them in separate files?

src/components/Profile/components/VictoryPromiseDisplay/index.js
```diff
@@ -0,0 +1,357 @@
...
+import { ReactComponent as OffIM } from './images/Off-IM.svg';
+import { ReactComponent as OffLS } from './images/Off-LS.svg';
+import { ReactComponent as OffLW } from './images/Off-LW.svg';
+import { ReactComponent as OnCC } from './images/On-CC.svg';
+import { ReactComponent as OnIM } from './images/On-IM.svg';
+import { ReactComponent as OnLS } from './images/On-LS.svg';
+import { ReactComponent as OnLW } from './images/On-LW.svg';
+import styles from './VictoryPromiseDisplay.module.css';
+
+export default class VictoryPromiseDisplay extends Component {
```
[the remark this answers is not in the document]
amos-cha: I think it might have been a merge error and I haven't been keeping up to date with exact changes in develop.

src/views/RecIM/components/Forms/SurfaceForm/index.jsx
```diff
@@ -0,0 +1,93 @@
...
+      type: 'text',
+      error: errorStatus.Name,
+      helperText: '*Required',
+      required: true,
+    },
+    {
+      label: 'Description',
+      name: 'Description',
+      type: 'text',
+      error: errorStatus.Name,
```
EjPlatzer, line 26: This looks like a typo, shouldn't it be errorStatus.Description?

src/views/RecIM/components/List/Listing/index.jsx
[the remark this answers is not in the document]
EjPlatzer: > The reason we did this initially was so that they could share styling and common features, but now that they have become more complex, I think we may be able to factor these out and use `Listing` as a shell component

They aren't actually sharing any code in the current iteration. If you plan to factor out repeated code in the future, that should be done via a re-usable component.

Currently, these are just a bunch of completely independent components sharing the file. The only thing they have somewhat in common is imports (but those don't exist in production builds so it doesn't make any sense to group components based on shared imports, otherwise we'd just write everything in the same file).

> I feel like refactoring the Listing as a shell component might be difficult due to the intricacies of each listing component with each listing having a completely different grid layout to another one. The only one that is remotely similar is Participant and Team, but even within Team there are all the record types on the right side.

This is, in my mind, proof that they should not be in the same file. IF they don't have anything in common beyond "being a list", why are they in the same file? And if they do have something in common, that thing in common should be extracted out to a re-usable component that also has it's own file.

> Perhaps we could create a new file for each but it would make it look incredibly bloated in terms of file numbers. @EjPlatzer is this more of a stylistic change or is there a specific reason you want them in separate files?

This is a somewhat fundamental code quality concern, so I wouldn't describe it as stylistic necessarily. You're thinking of this backward. Files exist so that we can keep them each small and self-contained. What you currently have is an extraordinarily bloated file (ideally, every file would have <100 lines of code ignoring imports). We use files to breakup a codebase into small, self-contained chunks that are easy to read and fully understand. So saying that breaking a bloated file into multiple files would cause "bloat" is exactly opposite the whole point of organizing source code into files.

Also, it's a react convention that each component should have it's own file. This is not a hard and fast rule, but it is a well-established convention and the exceptions (which mostly include extremely simple "helper" components) don't apply to this case.

Sorry if I'm coming across as harsh, that is not my intent, I just want to make sure I'm being clear.

src/components/Profile/components/VictoryPromiseDisplay/index.js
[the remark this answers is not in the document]
EjPlatzer: Yes, it must've been a merge error.

src/views/RecIM/components/List/Listing/index.jsx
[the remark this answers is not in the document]
amos-cha: No I get it, makes sense to me! 👍🏼

src/views/RecIM/components/Forms/Form/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+    return fields.map((field) => (
+      <InformationField
+        key={field.name}
+        error={field.error}
+        label={field.label}
+        name={field.name}
+        helperText={field.helperText}
+        value={newInfo[field.name]}
+        type={field.type}
+        menuItems={field.menuItems}
```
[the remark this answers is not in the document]
EjPlatzer: Given that we can condense this function so much, and that it's only ever called within `DialogContent`, and that it's just a wrapper around `fields.map` without actually adding any behavior, I think it would be cleaner to move the map to inline inside of `DialogContent` and get rid of this function entirely.

src/views/RecIM/components/Forms/Form/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+            !isSaving && handleConfirm(newInfo, handleWindowClose);
+          }}
+          buttonName="Confirm"
+          // in case you want to authenticate something change isButtonDisabled
+          isButtonDisabled={false}
+          cancelButtonClicked={() => {
+            if (!isSaving) setOpenConfirmWindow(false);
+          }}
+          cancelButtonName="Cancel"
+        >
```
EjPlatzer, line 200: This is more of a big picture thing, but I think the nested dialog here is really unfortunate. I would prefer a two-step form within a single dialog (i.e., you fill out the form and then hit a `Next` button, which shows you the confirmation step and the Submit button.

Also, for most of these I don't think you need the `Confirm` step at all. My rule of thumb would be, if the user can easily correct mistakes (e.g. fixing a team name) then the confirm is more cumbersome than useful.

EjPlatzer (commented):
I have some suggestions on ways to simplify and improve the `Form` component.

src/services/recim/series.ts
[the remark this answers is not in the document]
bennettforkner: @amos-cha it's not specific to PUT requests, it's a REST API convention to use a unique URI for referring to a specific resource. In this case, you are overwriting a specific resource (it may be null before the call, but you're still overwriting that location technically), so you should refer to a unique location.

https://restfulapi.net/resource-naming/ is very helpful for understanding the structure.

package.json
```diff
@@ -27,6 +27,7 @@
     "prop-types": ">=15.7.2",
     "react": "^17.0.2",
     "react-big-calendar": "^0.38.1",
+    "react-brackets": "^0.4.7",
```
EjPlatzer, line 30: We usually require new dependencies to be approved before adding them to the project.

This particular dependency is not very healthy, and I do not think we should add it for the following reasons:
1. It is a pre-1.0 package, meaning it is open to breaking changes and has no expectation of stability.
2. It is not popular and has a limited community, meaning there are not many eyes on it to detect bugs and security issues, and not much incentive for the maintainer or the users to continue developing and supporting it.
3. The maintenance is sparse and inconsistent, meaning there isn't someone actively devoted to keeping it secure and up to date with the rest of the eco-system.

That is a recipe for bugs, security issues, and difficulties when upgrading other packages. It's package health score is only 56/100.

src/services/recim/activity.ts
```diff
@@ -59,7 +59,7 @@ const getActivityByID = (ID: number): Promise<Activity> => http.get(`recim/activ
 
 const getActivities = (active: boolean, time: String): Promise<Activity[]> => {
   if (time) return http.get(`recim/activities?active=${active}&time=${time}`);
-  if (active) return http.get(`recim/activities?active=${active}`);
+  if (typeof active !== 'undefined') return http.get(`recim/activities?active=${active}`);
```
EjPlatzer, line 62:
```suggestion
  if (active !== undefined) return http.get(`recim/activities?active=${active}`);
```

src/services/recim/match.ts
```diff
@@ -123,6 +123,13 @@ const updateMatch = (ID: number, updatedMatch: PatchMatch): Promise<CreatedMatch
 const deleteMatchCascade = async (matchID: number): Promise<CreatedMatch> =>
   http.del(`recim/matches/${matchID}`);
 
+// shortcut service to delete a list of matches
+const deleteMatchList = async (matches: Match[]) => {
+  for (let i = 0; i < matches.length; i++) {
+    await http.del(`recim/matches/${matches[i].ID}`);
+  }
+};
```
EjPlatzer, line 131: This will be MUCH more performant without the inner `await`, since you can send all the requests in parallel instead of awaiting each request before sending the next one.

Also, generally better to use a `for in` loop to iterate arrays or else `Array.map`, since plain for loops clutter the important logic and introduce extra room for errors:
```suggestion
const deleteMatchList = (matches: Match[]) => 
  Prmoise.all(matches.map(match => deleteMatchCascade(match.ID)));
```

EjPlatzer (changes requested):
A couple small suggestions.

Overall, I must say I'm super impressed by this. Y'all have created an excellent interface! I know I've been nitpicky with my reviews, but I think this is a clean, well-architected project.

src/services/recim/match.ts
amos-cha: Unfortunately we can't actually do this, I wanted to mess with cleaner implementations of javascript (I don't think I have those commits) but the main issue is that our actual `delete` route, since it also patches all associated sub resources i.e Activity -> series -> team/match, we HAVE to await each call or else under load (even just 15-20 matches) there are 500 errors thrown all over the place.

package.json
amos-cha: I see, we use it for our visual display of brackets. Perhaps we could argue that this can be a temporary implementation and have our own component to be developed in the near future?

EjPlatzer: Yeah, I see that y'all are using it quite extensively.

I don't think it's reasonable at this stage to ask you to do a full rewrite of the bracket UI, so your suggestion seems like the best path forward.

src/services/recim/match.ts
EjPlatzer: Hmm, I'm looking at the code for `MatchService.DeleteMatchCascadeAsync` that is called by this route in the backend, and all that you're doing there is setting the status of this match and all of it's Match Team records to Deleted. That shouldn't cause any Database conflicts.

EjPlatzer: This is not a blocking issue, so it doesn't need to be something you investigate deeply right now, I just am not sure how it could be causing the problems you described.

.env.development
```diff
@@ -4,7 +4,11 @@
 # VITE_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
- VITE_API_URL=https://360ApiTrain.gordon.edu/
+# VITE_API_URL=https://360ApiTrain.gordon.edu/
+
+# @RECIM
+ VITE_API_URL=https://360ApiRecIM.gordon.edu/
```
bennettforkner, line 10:
```suggestion
 VITE_API_URL=https://360ApiTrain.gordon.edu/
```

.env.production
```diff
@@ -5,4 +5,7 @@
 # VITE_ANALYTICS_ID=UA-3559968-3
 
 # @TRAIN
-VITE_API_URL=https://360ApiTrain.gordon.edu/
+# VITE_API_URL=https://360ApiTrain.gordon.edu/
+
+# @RECIM
+VITE_API_URL=https://360ApiRecIM.gordon.edu/
```
bennettforkner, line 11:
```suggestion
 VITE_API_URL=https://360ApiTrain.gordon.edu/
```

package.json
amos-cha: https://github.com/gordon-cs/RecIM-Docs/issues/203

src/views/RecIM/components/Forms/WaiverForm/index.jsx
```diff
@@ -53,8 +53,11 @@ const WaiverForm = ({ username, createSnackbar, openWaiverForm, setOpenWaiverFor
 
   let waiverContent = (
     <Typography margin={4}>
+      <Typography variant="h5">
+        Note: This form is currently not legally binding, a paper form will be provided in the
+        meantime. Please contact a Rec-IM staff to sign up
+      </Typography>
```
bennettforkner, line 59:
```suggestion
      <Typography variant="h5">
        Please note: This acknowledgement is just a formality, you will need to sign a physical
        waiver before competing. Please contact
        <a href='mailto:rec.im@gordon.edu'>Rec.IM@gordon.edu</a>
        to sign up.
      </Typography>
```
I think this is a little bit better wording. Also it's good to give them a specific way to sign the paper waiver because otherwise, it's super hard to figure out who is in charge of Rec/IM and get their email.

src/views/RecIM/views/Activity/index.jsx
```diff
@@ -159,17 +161,25 @@ const Activity = () => {
...
+    }
+    if (activity.SoloRegistration) {
+      setLoading(true);
+      const profileInfo = await userService.getProfileInfo(profile.AD_Username);
+      const request = {
+        Name: profileInfo.fullName,
+        ActivityID: activityID,
+      };
+      await createTeam(profile.AD_Username, request);
+      setReload(!reload);
```
bennettforkner, line 177: There is a "special" way that react handles states which makes this not always work as you intend it to. Instead, inversions like this should be done with a lambda function passed in like so:
```suggestion
      setReload(r => !r);
```

src/views/RecIM/views/Activity/index.jsx
```diff
@@ -444,6 +452,15 @@ const Activity = () => {
...
               activityID={activityID}
             />
+            <WaiverForm
+              username={profile.AD_Username}
+              createSnackbar={createSnackbar}
+              onClose={() => {
+                setReload((prev) => !prev);
+              }}
+              openWaiverForm={openWaiver}
+              setOpenWaiverForm={(bool) => setOpenWaiver(bool)}
```
bennettforkner, line 462:
```suggestion
              setOpenWaiverForm={(w) => setOpenWaiver(w)}
```
Naming variables after a type is not the best convention because it quickly confuses the reader. My suggestion is to use a one-character, throw-away, variable name.

src/views/RecIM/components/Header/index.jsx
```diff
@@ -0,0 +1,52 @@
+import { Grid, AppBar, Breadcrumbs, Typography } from '@mui/material';
+import { Link as LinkRouter } from 'react-router-dom';
+import styles from './Header.module.css';
+import HomeIcon from '@mui/icons-material/Home';
+
+const RecIMBreadcrumb = ({ link, children }) => {
```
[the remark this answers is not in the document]
amos-cha: I'm struggling to understand what you mean here. If I'm not mistaken, createBrowserRouter creates a `router` object which cannot be rendered under 360's Router. However, the 360 Header is integral to Rec-IM as it is right now as without the header, the connection between 360 and recim is severed with the only connection being just a url name. 

The alternatives in my mind are that I either:
1) Add the crumb property to all routes in 360, and if they're null then they're null.
2) I leave it the way it is 
3) we refactor the header out of the router (_somehow_)

As is for now, I'll make an issue for this but I don't see a really easy solution without jumping over a lot of unnecessary hurdles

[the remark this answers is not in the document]
amos-cha: https://github.com/gordon-cs/RecIM-Docs/issues/214

.env.production
```diff
@@ -5,4 +5,4 @@
 # VITE_ANALYTICS_ID=G-2FE78G0CBN
 
 # @TRAIN
-VITE_API_URL=https://360ApiTrain.gordon.edu/
+ VITE_API_URL=https://360ApiTrain.gordon.edu/
```
amos-cha, line 8:
```suggestion
VITE_API_URL=https://360ApiTrain.gordon.edu/
```

.env.development
```diff
@@ -8,3 +8,4 @@
 
 # @LOCALHOST
 # VITE_API_URL=http://localhost:51626/
+
```
amos-cha, line 11:
```suggestion
```

PR #1559, opened 2022-11-07
[Rec-IM] Create Team Form

cpabbot:
## Rec-IM - Create Team Form

**Notes:**
- This is just the UI form groundwork. API integration will come separate.
- Lot of unnecessary code right now, but not sure yet how to best abstract so leaving clutter for later.

image

Does **not** resolve https://github.com/gordon-cs/RecIM-Docs/issues/59 (will come later)


src/services/recim.ts
```diff
@@ -16,13 +17,21 @@ type Activity = {
   Completed: boolean;
 };
 
+type Team = {
```
silasw2, line 20: Is this the way the ViewModel is represented coming from the database? We shouldn't use typing unless we know for sure what it looks like.

src/services/recim.ts
```diff
@@ -33,7 +42,11 @@ const getAllActivities = async (active: boolean, time: String): Promise<Object[]
 };
 
 const getAllSports = async (): Promise<Object[]> => {
-  return await http.get(`recim/sports`)
-}
+  return await http.get(`recim/sports`);
+};
+
+const createNewTeam = async (newTeam: Team) => {
```
silasw2, line 48: This doesn't have a return value at all.

src/views/RecIM/components/CreateActivityForm/index.js
```diff
@@ -219,6 +218,10 @@ const CreateActivityForm = ({
         type={field.type}
         menuItems={field.menuItems}
         onChange={handleChange}
+        xs={12}
+        sm={6}
+        md={4}
+        lg={3}
```
silasw2, line 224: A little interesting. But OK.

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -1,11 +1,4 @@
-import {
-  Grid,
-  Avatar,
-  ListItemAvatar,
-  ListItem,
-  ListItemText,
-  Typography,
-} from '@mui/material';
+import { Grid, Avatar, ListItemAvatar, ListItem, ListItemText, Typography } from '@mui/material';
```
silasw2, line 1: Did this get un-prettier?

silasw2 (changes requested):
Couple small notes. Mostly in the services file.

src/services/recim.ts
amos-cha: Screen Shot 2022-11-14 at 5 39 15 PM
when creating a team (which is what you're doing right now) you will need to look at the ____UploadViewModels)
This is what your Team Type should look like

src/views/RecIM/components/CreateActivityForm/CreateActivityForm.module.scss
```diff
@@ -1,12 +0,0 @@
...
-.form_title {
-  background-color: $primary-blue;
-  color: $neutral-white;
-  text-align: center;
-  grid-row: 1;
-
-  & :global(.MuiTypography-h6) {
-    font-size: 2em;
-  }
-}
```
amos-cha, line 12: I do not mind you deleting the contents of this file however, I don't know if you should be deleting the scss of this index.

cpabbot: The `form_title` selector is not used anywhere (must be a remnant from when you were creating the dialog box or something) so I removed it, and then deleted the file because there was nothing else in it.

amos-cha (approved):
👍🏼

silasw2 (approved):
Cam and I worked on this together, and I requested changes on Cam's initial commits. Approving here since he changed what I needed and Amos approved.

PR #1565, opened 2022-11-11
Update Workflows for Node v16

EjPlatzer:
GitHub is bumping the minimum Node Version of their actions to Node v16, which requires updating the action versions in our workflows.


bennettforkner (approved):
LGTM!...Even though it's a draft ;)

bennettforkner (approved):

PR #1568, opened 2022-11-14
Update Contributors for 2021/2022 School Year

bennettforkner:
Looks like we missed this update last Spring. Adding developers that worked in the Fall of 2021 while they were able.


EjPlatzer (approved):
LGTM!

PR #1569, opened 2022-11-14
Patch: GordonDialogBox Content Padding

bennettforkner:
The top padding on the GordonDialogBox component is currently not being applied because MuiCard-Root is taking a greater CSS weight. This means that the dialog box is showing the text in a popup with no space between it and the header as below:
image
This PR increases the specificity of the `.gc360_gordondialogbox` CSS class by repeating the tag in the SCSS file. After these changes, the GordonDialogComponent looks like this:
image
(disregard font styling between production and development on localhost)


src/components/GordonDialogBox/GordonDialogBox.module.scss
```diff
@@ -1,6 +1,6 @@
 @import '../../vars';
 
-.gc360_gordondialogbox {
+.gc360_gordondialogbox .gc360_gordondialogbox {
```
EjPlatzer, line 3: Rather than bumping the specificity of the entire dialog (which would make it harder for anything styling the dialog to override it), wouldn't it be better to override the underlying MUI style?

If you change `&_content {` on line 12 to `&_content :global(MuiDialogTitle-root)`, it should override that specific style.

Also, I believe that `.gc360_gordondialogbox .gc360_gordondialogbox` would not work because it selects an element with the class `.gc360_gordondialogbox` **inside** an element with the class `.gc360_gordondialogbox`, right? Shouldn't it be `.gc360_gordondialogbox.gc360_gordondialogbox` instead?

EjPlatzer (approved):
LGTM!

PR #1573, opened 2022-11-17
Patch: Fix styling of memberships list

bennettforkner:
It seems that the MUIv5 update messed with some of our styling for the MembershipsList component. Images loose their left padding.

Additionally, it seems that we lost some styling with the member's name link in that it is no longer vertically aligned with the rest of the text content.

This PR fixes those two issues by reworking styles for the rows & content.

Current:
image
New:
image


EjPlatzer (approved):
LGTM! Although I would prefer to use a class rather than inline styles.

PR #1575, opened 2022-11-21
[Rec-IM] UI Service Routes

amos-cha:
- Created all PATCH & POST types
- Created all Proposed Routes (https://docs.google.com/document/d/1-inU_yD_QgUsPekB-m_QOyPHPszeC1EuDmxdcDfWbv0/)

DO NOT MERGE UNTIL THIS API PR IS MERGED:
https://github.com/gordon-cs/gordon-360-api/pull/754


src/services/recim.ts
```diff
@@ -3,28 +3,71 @@ import http from './http';
 // all Gordon Rec-IM http requests
```
EjPlatzer, line 3: Would it make sense to split this file up into a folder containing one service per API resource? e.g. instead of `services/recim.ts`, you'd have `services/recim/activity.ts`, `services/recim/team.ts`, etc.?

cpabbot: I think this is probably a good idea; could we put them all in a `services/recim` folder? Also we probably will have some services that don't fall under one of the main categories right, so perhaps we could have a `services/recim/recim.ts` as well? Can't think of exactly what routes that would be though.

src/views/RecIM/views/Home/index.js
```diff
@@ -134,13 +134,15 @@ const Home = () => {
...
-        /> : null}
+        {openCreateActivityForm ? (
+          <CreateActivityForm
+            closeWithSnackbar={(status) => {
+              handleCreateActivityForm(status);
+            }}
+            openCreateActivityForm={openCreateActivityForm}
+            setOpenCreateActivityForm={(bool) => setOpenCreateActivityForm(bool)}
+          />
+        ) : null}
```
amos-cha, line 145: somehow my prettier formatted it differently than someone else's?

src/services/recim/activity.ts
```diff
@@ -0,0 +1,88 @@
+import http from '../http';
+import { sport } from './sport';
+import { team } from './team';
+import { series } from './series';
+
+export type activity = {
```
EjPlatzer, line 6: FYI, types are capitalized in JS/TS by convention.

src/services/recim/activity.ts
```diff
@@ -0,0 +1,88 @@
...
+  MaxCapacity: number;
+  SoloRegistration: boolean;
+  Logo: string;
+  Completed: boolean;
+};
+
+//Activity Routes
+const createNewActivity = async (newActivity: uploadActivity): Promise<Object[]> => {
+  return await http.post('recim/activities', newActivity);
+};
```
EjPlatzer, line 62: You can simplify this syntax significantly if you so choose:
```suggestion
const createNewActivity = (newActivity: uploadActivity): Promise<Object[]> =>  http.post('recim/activities', newActivity);
```

This is functionally equivalent because `http.post` returns a promise, so declaring async and `await`ing it (thereby wrapping the call to `http.post` in a Promise) is the same as just returning the interior Promise.

Also, an arrow function that is just one return statement can use the expression syntax `(params) => result` instead of the traditional statement syntax `(params) => { return result; }`.

The first is potentially a minor performance improvement, but ultimately neither make a huge difference and might potentially confuse developers who are less familiar with these concepts, so this condensed form is not **necessarily** better - although I personally do find it more readable.

So this is just a notice to make sure you're aware that this is an option in JS/TS, not a "you should write it this way" kind of suggestion.

amos-cha: i see, I can make those changes.

src/services/recim/activity.ts
```diff
@@ -0,0 +1,83 @@
...
+  RegistrationStart: string;
+  RegistrationEnd: string;
+  SportID: number;
+  StatusID: number;
+  MinCapacity: number;
+  MaxCapacity: number;
+  SoloRegistration: boolean;
+  Logo: string;
+  Completed: boolean;
+};
```
bennettforkner, line 56: I don't think it's necessary to have two separate objects here. You can just make use of the optional fields to merge the two together.

src/services/recim/activity.ts
```diff
@@ -0,0 +1,83 @@
...
+): Promise<Activity[]> => {
+  if (registrationOpen) {
+    return http.get(`recim/activities?registrationOpen=${registrationOpen}`);
+  } else {
+    if (time) {
+      return http.get(`recim/activities?active=${active}&time=${time}`);
+    } else {
+      return http.get(`recim/activities?active=${active}`);
+    }
+  }
```
bennettforkner, line 77: If you just add all of the potential parameters to the URL, the nulls will just appear as null to the API. You can simplify to:
```suggestion
  return http.get(`recim/activities?active=${active}&time=${time}&registrationOpen=${registrationOpen}`);
```

src/services/recim/match.ts
```diff
@@ -0,0 +1,75 @@
...
+  MatchID: number;
+  ParticipantID: number;
+};
+
+//Match Routes
+const createMatch = (newMatch: UploadMatch): Promise<CreatedMatch> =>
+  http.post('recim/matches', newMatch);
+
+const createMatchAttendance = (username: string, matchID: number): Promise<CreatedAttendance> =>
+  http.post(`recim/matches/attendance?matchID=${matchID}&username=${username}`);
```
bennettforkner, line 63: Data in a created object should go in a post body rather than the URL GET parameters. Also, the match identifier should be a part of the URI path, not a GET/POST parameter.
```suggestion
const createMatchAttendance = (username: string, matchID: number): Promise<CreatedAttendance> =>
  http.post(`recim/matches/${matchID}/attendance`, username);
```

src/services/recim/participant.ts
```diff
@@ -0,0 +1,106 @@
...
+  ID: number;
+  ActivityID: number;
+  ParticipantID: number;
+  PrivTypeID: number;
+  isFreeAgent: boolean;
+};
+
+//Participant Routes
+const createParticipant = (ID: number): Promise<Participant> =>
+  http.post(`recim/participants/${ID}`);
```
bennettforkner, line 67: I think this should technically be a PUT operation. Since you already know the unique ID that you want to use (ID), you are 'putting' data in that location. POST operations are for when you want the API to figure out an identifier to use for the object data that you are passing to it.

src/services/recim/participant.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const getParticipants = (): Promise<Participant[]> => http.get(`recim/participants`);
```
bennettforkner, line 69: What UI functionality is this route used for? It seems like it could become a lot of overhead on the API & UI if there are no filters applied.

src/services/recim/participant.ts
```diff
@@ -0,0 +1,106 @@
...
+const sendNotification = (
+  username: string,
+  notification: UploadParticipantNotification,
+): Promise<CreatedParticipantNotification> =>
+  http.post(`participants/${username}/notifications`, notification);
+
+const editParticipantActivity = (
+  username: string,
+  updatedParticipant: PatchParticipantActivity,
+): Promise<CreatedParticipantActivity> =>
```
bennettforkner, line 89: Here as well as in the API, it is a bit strange to return a "created" object type from an "edit" function.

src/services/recim/series.ts
```diff
@@ -0,0 +1,59 @@
...
+  StatusID: number;
+};
+
+//Series Routes
+const createSeries = async (
+  referenceSeriesID: number,
+  newSeries: UploadSeries,
+): Promise<CreatedSeries> => {
+  var subQuery = referenceSeriesID === null ? '' : `?referenceSeriesID=${referenceSeriesID}`;
+  return await http.post(`recim/series${subQuery}`, newSeries);
```
bennettforkner, line 49: Any data that is part of the object to be created should be in the post body. Could you clarify exactly what referenceSeriesID is supposed to do here?

If the purpose is to create a series using another existing one as a template, I would recommend creating a different API route & UI service method for that.

Even that recommendation would not be considered truly "restful" as operations should perform one action (create, retrieve, update, or destroy). When you create a copy of another resource, I would think that the most responsible way to do so would be to first retrieve the series, remove the ID and update as necessary, then pass the updated object to the POST route.

src/services/recim/participant.ts
bennettforkner: At the very least, pagination would help here.

src/services/recim/team.ts
```diff
@@ -0,0 +1,89 @@
...
+
+type PatchTeam = {
+  Name: string;
+  StatusID: number;
+  Logo: string;
+};
+
+//Team Routes
+const createTeam = (username: string, newTeam: UploadTeam): Promise<CreatedTeam> =>
+  http.post(`recim/Teams?username=${username}`, newTeam);
```
bennettforkner, line 64: What is the reason for passing `username` as a GET parameter here? The currently authenticated username is already accessible to the API controller, so that shouldn't be needed.

src/services/recim/team.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const getTeamByID = (ID: number): Promise<Team> => http.get(`recim/Teams/${ID}`);
+
+const addParticipantToTeam = async (
+  username: string,
+  TeamID: number,
+  RoleID: number,
+): Promise<CreatedTeamParticipant> => {
+  var query = RoleID === null ? '' : `?RoleID=${RoleID}`;
+  return await http.post(`recim/Teams/Participants?username=${username}&TeamID=${TeamID}${query}`);
```
bennettforkner, line 74: RoleID and username should be part of the post body since they are object properties of a TeamParticipant. Also, you should be referencing the specific Teams resource in the URI path.
```suggestion
  return await http.post(`recim/Teams/${TeamID}/Participants`, {Username: username, RoleID: RoleID});
```

src/services/recim/team.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const editTeamParticipant = async (
+  username: string,
+  TeamID: number,
+  RoleID: number,
+): Promise<CreatedTeamParticipant> => {
+  var query = RoleID === null ? '' : `?RoleID=${RoleID}`;
+  return await http.patch(`recim/Teams/Participants?username=${username}&TeamID=${TeamID}${query}`);
```
bennettforkner, line 83: Same thing here with the patch body as above in the post method.

src/views/RecIM/components/Forms/CreateActivityForm/index.js
```diff
@@ -187,7 +188,7 @@ const CreateActivityForm = ({
       (sport) => sport.Name === activityCreationRequest.sportID,
     )[0].ID;
 
-    createNewActivity(activityCreationRequest).then(() => {
+    createActivity(activityCreationRequest).then(() => {
```
bennettforkner, line 191: This success message should only be shown if the createActivity data post operation was successful. Otherwise, you should show an error message. Please confirm that the value returned from `createActivity()` is checked to confirm creation.

src/views/RecIM/components/Forms/CreateTeamForm/index.js
```diff
@@ -122,9 +122,9 @@ const CreateTeamForm = ({
 
     let teamCreationRequest = { ...currentInfo, ...newInfo };
 
-    console.log(teamCreationRequest)
+    console.log(teamCreationRequest);
```
bennettforkner, line 125: Remove console logging here

src/views/RecIM/components/Forms/CreateTeamForm/index.js
```diff
@@ -122,9 +122,9 @@ const CreateTeamForm = ({
 
     let teamCreationRequest = { ...currentInfo, ...newInfo };
```
bennettforkner, line 123: I'm not sure "...Request" is the best terminology here. A request is a different thing than the object to upload. Perhaps `teamUpload` or `newTeam`.

bennettforkner (commented):
Good looks so far. I had a few comments.

I wanted to consider how you use your "Patch" and "Upload" types overall. I'm not sure that you need two different types in all of your services for the two operations. I would recommend using nullable properties instead.

src/services/recim/activity.ts
amos-cha: I guess this goes back to if it would be better to merge objects together in which case I got to ask the question: Why not do that in the API end as well?

src/services/recim/activity.ts
```diff
@@ -0,0 +1,83 @@
...
+  SportID: Sport;
+  MinCapacity: number;
+  MaxCapacity: number;
+  SoloRegistration: boolean;
+  Logo: string;
+  Series: Series[];
+  Team: Team[];
+};
+
+type CreatedActivity = {
```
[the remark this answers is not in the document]
amos-cha: They should be all be named correctly as I worked with both in parallel. The only differences being small naming changes as the API has relatively complex/long names that are not necessary for the UI to know about.

src/services/recim/match.ts
amos-cha: since username is a string, i think asp.net automatically puts it in the query. I could switch it to FromBody but regardless it might be a bit odd just to have a single string there.

src/services/recim/team.ts
amos-cha: This would require another viewmodel/class in the API and, for consistency sake, another type for the type {Username, RoleID} as I don't think it would be very clean to have an injected object like that. Right now I'll pass it as /Teams/teamID/username=username&roleID=roleID

src/views/RecIM/components/Forms/CreateActivityForm/index.js
amos-cha: This is not necessarily in the scope of my PR as anything to do with the forms should be handled by the PRs implementing the forms. I'll make a note of this an request a fix to someone in the team.

src/views/RecIM/components/Forms/CreateTeamForm/index.js
amos-cha: I'll make a note for the team. I only did a git pull from the current WIP branch

src/views/RecIM/components/Forms/CreateTeamForm/index.js
amos-cha: I think the Request terminology was originally used since it was the passed/requested object to the route. The naming was inherited from the AlumniUpdateForm earlier this summer

src/services/recim/match.ts
bennettforkner: ASP.NET checks all potential places where data could reside. If it does not find it in the query string, then it will check the post body.

If you are posting data, it should be in the body, even if it is just a string. "Query" parameters are called query parameters because their primary purpose is to filter a dataset, not to pass data to the endpoint.

src/services/recim/team.ts
bennettforkner: You should have a type for any data that you pass to the API, no matter how small of an object. (of course, if it is a native type like a string, bool, or int, it does not need its own type)

Data should not be sent in the query string of a post request.

src/views/RecIM/components/Forms/CreateTeamForm/index.js
bennettforkner: But AlumniUpdateForm was actually sending an information update request to development. This is actually going to make changes in the database. The issue is kind of a result of the overloaded term, 'request'. (http request vs request for info update)

src/services/recim/team.ts
amos-cha: In that case, I'll make a new viewmodel for it on the API end and make that a new type.

src/views/RecIM/components/Forms/CreateTeamForm/index.js
amos-cha: I see, I'll make a note for the team to make an update on that front.

Kevin-Chu58 (approved):
LGTM!

amos-cha:
Gone over with @cpabbot  in person

PR #1578, opened 2022-11-28
Remove admin list and point users to groups site

bennettforkner:
The admin list is no longer relevant due to the refactoring of authorization by @EjPlatzer to use Azure Active Directory groups.

The functionality of the admin list does nothing except change the rows of an admin table that is no longer used for anything.

Our new method for allowing end-users to update AD groups is through the groups.gordon.edu website from on-campus. This will allow all essential users to be able to modify the 360-SiteAdmin-SG group as necessary.


EjPlatzer (approved):
LGTM!

PR #1579, opened 2022-11-29
Dependency Bump - moment, minimist, terser, loader-utils, minimatch, recursive-readdir

bennettforkner:
I am confident in all of these except `moment` as I could not properly test `moment` in the schedule, so I will confirm with students that nothing is more broken once this is in train.


EjPlatzer:
> as I could not properly test `moment` in the schedule
I'm not sure what there is to test, since the schedule has been broken for a long time. Please fix the conflicts and then this looks good to me.

EjPlatzer (approved):
LGTM!

PR #1580, opened 2022-11-29
Disable Demographic Questions in Enrollment Check-in

EjPlatzer:
Per request of the registrar, this PR will disable the demographic (race/ethnicity) questions in the Enrollment Check-in process.

I have enabled enrollment check-in in Train for testing purposes. I have tested it with a student test account and it seems to work well, but given the number of edge cases surrounding enrollment check-in, we'll want to test more thoroughly in Train.


bennettforkner (approved):
LGTM

PR #1582, opened 2022-11-29
Fix react router Navigate

EjPlatzer:
There were a couple lingering components that broke on upgrade to react-router, mainly because of using `props.history` or `Route` instead of `Navigate`. I have fixed these last few.


bennettforkner (approved):
LGTM!

Issue #1542, opened 2022-10-14
Student news -- posting images is broken


bennettforkner:
It seems that the UI is sending the fields of a new student news post as GET (URL) parameters rather than a post body. This means that when you post an image, the base64 encoded image data is sent as a part of the URL. The URL is then too long for browsers to send, so they error out with a 431 HTTP code.

This data should be sent as a posted body rather than URL parameters.

PR #1584, opened 2022-11-30
student news upload image

Kevin-Chu58:
This pr requires the API update as well, otherwise by simply merging this to the develop, the image will set to `null` in the API.
 - move image from route to body in service request
 
image
image

Link #1542.
API PR: https://github.com/gordon-cs/gordon-360-api/pull/759.


src/services/news.ts
```diff
@@ -101,20 +101,21 @@ const getNewsByCategory = async (category: number): Promise<NewsItem[]> =>
...
   categoryID: number,
   body: string,
   image?: string,
-): Promise<NewsItem | undefined> {
+): Promise<NewsItem | undefined> => {
   try {
     return await http.post(
-      `news?subject=${subject}&categoryID=${categoryID}&body=${body}&image=${image ?? ''}`,
+      `news?subject=${subject}&categoryID=${categoryID}&body=${body}`,
+      image ?? '',
```
bennettforkner, line 113: Hey @Kevin-Chu58, I know this is still a draft, but I wanted to note that this entire query string should be in the body, not just the image. Perhaps a `StudentNewsUpload` type would help?

bennettforkner (changes requested):

bennettforkner (commented):
This looks good. I am holding off on actually approving because the auto-deployment works on the UI but not on the API. We will have to deploy the API manually before we merge this.

bennettforkner (approved):
Approved, pending lint and build actions succeed.

PR #1586, opened 2022-12-01
Allow additional format of xlsx upload

bennettforkner:
@ChrisImagineer noticed that our spreadsheet uploader was actually not working for excel ".XLSX" files even though we were stating that it should on the upload page.

I did some research and discovered that there is actually another format for Excel files that we have to allow. This change adds that format to the acceptedTypes variable.


bennettforkner:
I have confirmed that this fixes @ChrisImagineer 's issues.

EjPlatzer (approved):
LGTM

PR #1587, opened 2022-12-01
Implement API/SQL changes to vm's

amos-cha:
Adjustment PR introducing changes from the SQL Refactor conducted by @Kevin-Chu58 

API PR required:
https://github.com/gordon-cs/gordon-360-api/pull/761

### DIFF
--- a/src/services/recim/participant.ts
+++ b/src/services/recim/participant.ts
@@ -8,6 +8,7 @@ export type Participant = {
   Role: string;
   Status: string;
   Notification: ParticipantNotification[];
+  IsAdmin: boolean;
 };
 
 type PatchParticipantActivity = {
@@ -17,7 +18,7 @@ type PatchParticipantActivity = {
 };
 
 type PatchParticipantStatus = {
-  StatusDescription: string;
+  StatusID: number;
   EndDate: string;
 };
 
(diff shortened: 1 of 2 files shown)


src/services/recim/participant.ts
```diff
@@ -8,6 +8,7 @@ export type Participant = {
   Role: string;
   Status: string;
   Notification: ParticipantNotification[];
+  IsAdmin: boolean;
```
bennettforkner, line 11: What is this "isAdmin" value here for? How is it to be used? If it is for authorization of global RecIM admins, it should be received from the useAuthGroups hook rather than as a part of this object.

amos-cha: this is a local type for rec-im staff. the AuthGroup will be used for RecIM Directors

EjPlatzer: To clarify, "rec-im staff" means student workers who are working for RecIM, as opposed "RecIM Directors" being full-time staff members who oversee RecIM, right?

It might be worth having a discussion (over Teams probably) about the differences in these permission levels and the pros and cons of using Auth Groups as opposed to a flag in the database.

Kevin-Chu58 (approved):
LGTM! Thank you for your good work!

src/services/recim/participant.ts
amos-cha: Yes, staff implies student workers

amos-cha: We can have that discussion with ISG on our upcoming meeting. I think our MVP will be due when we get back as there is a hefty chunk of work yet to be done.

PR #1591, opened 2022-12-10
[RecIM] Team Page Fetch Logic and Tools

silasw2:
This PR serves as the place where all team page logic and work will take place. See (eventual) linked RecIMDocs issues for details.

### DIFF
--- a/src/views/RecIM/components/List/Listing/index.js
+++ b/src/views/RecIM/components/List/Listing/index.js
@@ -65,12 +65,12 @@ const ActivityListing = ({ activity }) => {
   );
 };
 
-const TeamListing = ({ team, activityID }) => {
+const TeamListing = ({ team }) => {
   return (
     <ListItem
       button
       component={Link}
-      to={`/recim/activity/${activityID}/team/${team.ID}`}
+      to={`/recim/activity/${team.ActivityID}/team/${team.ID}`}
       className="gc360_link"
     >
       <Grid container className={styles.listing}>
@@ -83,25 +83,24 @@ const TeamListing = ({ team, activityID }) => {
 // We could also use ParticipantID (not student ID) if we have that and prefer it to AD_Username
 const ParticipantListing = ({ participant }) => {
   const [avatar, setAvatar] = useState('');
-
-  // const [name, setName] = useState({
-  //   firstname: '',
-  //   lastname: '',
-  // });
+  const [name, setName] = useState('');
 
   useEffect(() => {
     const loadAvatar = async () => {
-      if (participant.username) {
+      if (participant.Username) {
         const { def: defaultImage, pref: preferredImage } = await user.getImage(
-          participant.username,
+          participant.Username,
         );
         setAvatar(preferredImage || defaultImage);
+        const splitName = participant.Username.split('.');
+        setName(splitName[0] + ' ' + splitName[1]);
       }
     };
     loadAvatar();
-  }, [participant.username]);
+  }, [participant.Username]);
+
   return (
-    <ListItem key={participant.username} disableGutters={true}>
+    <ListItem key={participant.Username} disableGutters={true}>
       <Grid container alignItems="center" className={styles.listing}>
         <ListItemAvatar>
           <Avatar
@@ -110,8 +109,8 @@ const ParticipantListing = ({ participant }) => {
             variant="rounded"
           ></Avatar>
         </ListItemAvatar>
-        <Link to={`/profile/${participant.username}`} className="gc360_link">
-          <ListItemText primary={participant.username} />
+        <Link to={`/profile/${participant.Username.toLowerCase()}`} className="gc360_link">
+          <ListItemText primary={name} />
         </Link>
       </Grid>
     </ListItem>
--- a/src/views/RecIM/views/Activity/index.js
+++ b/src/views/RecIM/views/Activity/index.js
@@ -12,7 +12,7 @@ import { getActivityByID } from 'services/recim/activity';
 
 const Activity = () => {
   const { activityID } = useParams();
-  const { profile, /* profileLoading */ } = useUser();
+  const { profile } = useUser();
   const [loading, setLoading] = useState(true);
   const [activity, setActivity] = useState({});
   const [openCreateTeamForm, setOpenCreateTeamForm] = useState(false);
@@ -25,14 +25,13 @@ const Activity = () => {
       setLoading(false);
     };
     loadData();
-  }, [activityID]);
+  }, [activityID, openCreateTeamForm]);
+  // ^ May be bad practice, but will refresh page on dialog close
 
   const handleCreateTeamForm = (status) => {
     //if you want to do something with the message make a snackbar function here
     setOpenCreateTeamForm(false);
   };
-
-  console.log(activity)
 
   // profile hook used for future authentication
   // Administration privs will use AuthGroups -> example can be found in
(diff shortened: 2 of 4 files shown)


.env.development
```diff
@@ -4,7 +4,7 @@
 # REACT_APP_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
-REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+# REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
 
 # @LOCALHOST
-# REACT_APP_API_URL=http://localhost:51626/
+REACT_APP_API_URL=http://localhost:51626/
```
amos-cha, line 10: revert env changes before we push into `recim`

src/views/RecIM/views/Activity/index.js
```diff
@@ -47,12 +47,13 @@ const Activity = () => {
...
         <CardContent>
-          {/* if there are games scheduled, map them here */}
-          <MatchList matches={[{ activityID: '123456', ID: '789' }]} />
-          {/* else "no schedule yet set" */}
-          <Typography variant="body1" paragraph>
-            Games have not yet been scheduled.
-          </Typography>
+          {activity.Series[0].Match ? (
+            <MatchList matches={activity.Series[0].Match} />
+          ) : (
```
amos-cha, line 52: Theoretically will work, but we do not have a way of creating series objects right now from the UI.

src/views/RecIM/views/Match/index.js
```diff
@@ -1,27 +1,68 @@
...
+  useEffect(() => {
+    const loadMatch = async () => {
+      setLoading(true);
+      setMatch(await getMatchByID(matchID));
+      setLoading(false);
+    };
+    loadMatch();
+  }, [matchID]);
+
+  console.log(match);
```
amos-cha, line 50: remember to remove this console log before push

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -110,24 +109,24 @@ const ParticipantListing = ({ participant }) => {
             variant="rounded"
           ></Avatar>
         </ListItemAvatar>
-        <Link to={`/profile/${participant.username}`} className="gc360_link">
-          <ListItemText primary={participant.username} />
+        <Link to={`/profile/${participant.Username.toLowerCase()}`} className="gc360_link">
```
bennettforkner, line 112: Why are you going tolowercase here? That has not been our practice in the past for 360 profiles. It is just extra code as far as functionality goes.

src/views/RecIM/views/Match/index.js
```diff
@@ -1,27 +1,68 @@
...
+const RosterCard = (participants, teamName) => {
   return (
+    <Card>
+      <CardHeader title={teamName} className={styles.cardHeader} />
+      <CardContent>
+        <ParticipantList participants={participants} />
+      </CardContent>
+    </Card>
+  );
+};
```
bennettforkner, line 33: Simplify this
```suggestion
const RosterCard = (participants, teamName) => (
  <Card>
    <CardHeader title={teamName} className={styles.cardHeader} />
    <CardContent>
      <ParticipantList participants={participants} />
    </CardContent>
  </Card>
);
```

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -97,25 +97,24 @@ const TeamListing = ({ team }) => {
...
     const loadAvatar = async () => {
-      if (participant.username) {
+      if (participant.Username) {
         const { def: defaultImage, pref: preferredImage } = await user.getImage(
-          participant.username,
+          participant.Username,
         );
         setAvatar(preferredImage || defaultImage);
+        const splitName = participant.Username.split('.');
+        setName(splitName[0] + ' ' + splitName[1]);
```
cpabbot, line 110: Can we fetch the actual name for this? If it's more difficult we can do it later and merge this PR for now, but this solution is not ultimately ideal since names differ from usernames.

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -124,23 +123,29 @@ const ParticipantListing = ({ participant }) => {
...
       </Grid>
     </ListItem>
   );
 };
 
-const MatchListing = ({ match }) => {
+const MatchListing = ({ match, activityID }) => {
+  if (match.Team.length === 0) {
+    return null;
+  }
```
cpabbot, line 137: Probably a cleaner way to do this. At very least we can do a `if(!match?.Team?.length) return null;` which I'll commit

cpabbot (changes requested):
Looking good! Going to commit a couple changes just to clean up a couple lines logically, but otherwise just look at my question about participant names. If you think we should worry about it later (if too complicated or something) I can approve, but I imagine this is doable.

src/views/RecIM/components/List/Listing/index.js
amos-cha: I think we can use Gordon's profile hook to fetch proper names.
However, if it ends up being taxing/slow, I can add it as a part of our Participants table

cpabbot: Agreed, I think that's best solution too. Maybe I'll try adding it.

cpabbot (approved):
LGTM. Re-check before pushing and review the modifications I committed to this branch.

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -124,23 +128,30 @@ const ParticipantListing = ({ participant }) => {
...
       component={Link}
-      to={`/recim/activity/${match.activityID}/match/${match.ID}`}
+      to={`/recim/activity/${activityID}/match/${match.ID}`}
       className="gc360_link"
     >
       <Grid container className={styles.listing}>
-        <Grid item>Team A vs Team B</Grid>
+        <Grid item>
+          {match.Team[0].Name} vs. {match.Team[1].Name}
+        </Grid>
```
amos-cha, line 154: @silasw2, I think eventually we need to find a way to separate these team "columns" or somehow handle matches with more than 2 teams. For now this is okay.

src/views/RecIM/views/Team/index.js
```diff
@@ -1,95 +1,104 @@
...
+            <MatchList matches={team.Match} activityID={team.ActivityID} />
+          ) : (
+            <Typography variant="body1" paragraph>
+              No matches scheduled at this time!
+            </Typography>
+          )}
+        </CardContent>
+      </Card>
+    );
+
```
amos-cha, line 101: Is there any added benefit of making these variables if they're only ever used once?

amos-cha (commented):
Looks good but I got a few style questions

src/views/RecIM/components/List/Listing/index.js
cpabbot: Agreed, excellent point. I almost just added a commit to list teams if more than two, but I think instead I'll just add a bug issue to the project board for later so we can check with Rec-IM how they want to handle this.

src/views/RecIM/views/Team/index.js
cpabbot: Just helps clean up return so it's easier to read, but that's about all

src/views/RecIM/components/List/Listing/index.js
silasw2: @amos-cha Yea, I am going to loop over them, will come with my next commit. Thanks for pointing it out!

silasw2: I actually think that for now, it might be best to leave it like this, and our UI design for the team page will need to change in the future to be able to handle multiple events. There just needs to be a bit more brainstorming around how it will work.

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -106,10 +106,15 @@ const ParticipantListing = ({ participant }) => {
           participant.Username,
         );
         setAvatar(preferredImage || defaultImage);
-        const splitName = participant.Username.split('.');
-        setName(splitName[0] + ' ' + splitName[1]);
       }
     };
+    const loadUserInfo = async () => {
```
silasw2, line 111: This is good, thanks for making this change Mr. Cam.

PR #1592, opened 2022-12-12
Type updates

amos-cha:


amos-cha:
handle updates from https://github.com/gordon-cs/gordon-360-api/pull/767

PR #1594, opened 2022-12-23
[Rec-IM] Fix datetime

cpabbot:
Fixes couple DateTime things:
- DateTime pickers re-added after being removed in previous merge conflicts
- Package dependency swapped from dayjs to date-fns, which is what 360 currently uses
- Confirmation window now shows DateTime as a properly formatted (with timezone) string

Noted a bug where activities are all showing time of midnight... Is this a DB or API issue that we are already aware of?


Kevin-Chu58:
> Noted a bug where activities are all showing time of midnight... Is this a DB or API issue that we are already aware of?
I think the time is set to the midnight in default.

Kevin-Chu58 (approved):
LGTM! Good work.

PR #1596, opened 2022-12-31
[Rec-IM] Roster Updates

cpabbot:
## Roster Updates
- More options menu to remove a team participant or make co-captain
- Invite participant window, where a list of participants can be invited at once

To do: redundancy check - prevent invites being sent to people that are already on the team or have already been invited.

**Note: there is no API functionality with this yet, this is simply the UI work**

image
image

### DIFF
--- a/src/views/RecIM/components/Forms/InviteParticipantForm/index.js
+++ b/src/views/RecIM/components/Forms/InviteParticipantForm/index.js
@@ -1,4 +1,4 @@
-import { Grid, List } from '@mui/material';
+import { Grid } from '@mui/material';
 import { useState, useEffect } from 'react';
 import GordonDialogBox from 'components/GordonDialogBox';
 import { ParticipantList } from './../../List';
@@ -57,12 +57,7 @@ const InviteParticipantForm = ({
     >
       <Grid container alignItems="center" direction="column" spacing={2} p={2}>
         <Grid item sx={{ width: '100%' }}>
-          <List>
-            <ParticipantList minimal participants={inviteList} callbackFunction={removeInvite} />
-            {/* {inviteList?.map((username) => (
-              <ListItem>{username}</ListItem>
-            ))} */}
-          </List>
+          <ParticipantList minimal participants={inviteList} callbackFunction={removeInvite} />
         </Grid>
         <Grid item>
           <GordonQuickSearch
--- a/src/views/RecIM/components/Forms/InviteParticipantForm/index.js
+++ b/src/views/RecIM/components/Forms/InviteParticipantForm/index.js
@@ -1,206 +1,77 @@
-import { Grid } from '@mui/material';
-import { useState, useMemo, useEffect } from 'react';
+import { Grid, List } from '@mui/material';
+import { useState, useEffect } from 'react';
 import GordonDialogBox from 'components/GordonDialogBox';
-import { InformationField } from '../components/InformationField';
-import { ConfirmationWindowHeader } from '../components/ConfirmationHeader';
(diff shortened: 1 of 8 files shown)


src/views/RecIM/components/Forms/InviteParticipantForm/index.js
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  const handleSubmit = () => {
+    inviteList.forEach((value) => {
+      let participantData = {
+        Username: value.Username,
+        RoleTypeID: 2,
+      };
+      addParticipantToTeam(teamID, participantData).catch((reason) => {
+        console.log(reason);
```
amos-cha, line 49: console log

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -108,30 +133,48 @@ const ParticipantListing = ({ participant }) => {
...
-        <Link to={`/profile/${participant.Username}`} className="gc360_link">
-          <ListItemText primary={name} />
-        </Link>
-      </Grid>
+    // first ListItem is used only for paddings/margins
+    // second ListItem (nested inside) is used to layout avatar and secondaryAction
+    <ListItem key={participant.Username}>
+      <ListItem
+        secondaryAction={
+          minimal ? (
```
amos-cha, line 145: we might want a variable with better naming, I'm not sure what minimal means

amos-cha (changes requested):
Looks good, though I'm not sure what _minimal_ means and there is a console.log to remove

src/views/RecIM/components/List/Listing/index.js
silasw2: I tend to agree with this, I think cam was the one who put minimal in. In reality, I think that this is to identify whether the participant listing is for the quick search add, or the team/match pages.

src/views/RecIM/components/Forms/InviteParticipantForm/index.js
silasw2: This is actually a placeholder, and does not necessarily need to be removed as it is catching the error the backend throws. Ultimately we should be adding snackbar notifications whenever an operation does not succeed. This is what that is in place of. I can remove it though if we would like to go back through later and add more in depth error handling.

amos-cha: i see, as long as there is a purpose then it's alright. however, errors that get thrown by the backend would appear on the console log regardless of catch.

amos-cha (approved):
Looks good, just the thing about the variable "minimal". @cpabbot I'll leave you to make that change

src/views/RecIM/components/Forms/InviteParticipantForm/index.js
cpabbot: That all sounds good, and yeah that makes sense to just use it as a placeholder until snackbar is implemented. I assume snackbar is not working at the moment? We can add later

src/views/RecIM/components/List/Listing/index.js
cpabbot: Haha I saw your question about minimal and was like yeah that's dumb, then realized I did that haha. Idea was that this is not marked for a specific purpose, it is simply a minimal version of the ParticipantListing - so just name and profile image for now. We can update naming later on as we figure out what all our use cases end up being.

cpabbot: I realized there will probably be multiple versions of ParticipantListing, so we can update that later on as that becomes more clear what the distinctions are.

PR #1608, opened 2023-01-09
HOTFIX: Fix people search navigation

### LINKED ISSUE #1607: People search results links are broken

bennettforkner:
As @russtuck points out ( closes #1607 ), people search is currently linking users to a broken page from the results objects. This is because the link is using a relative path (starting from `/people`) rather than a static path (starting from `/`). This PR sets the path as static.


EjPlatzer (approved):
Thanks for fixing this!

PR #1613, opened 2023-01-12
Clifton Import Cleanup & File Restructure

### LINKED ISSUE #1585: Match Clifton CSV import fields to source file
The clifton strengths upload module on the admin page currently takes in a list of PascalCase column names. @chrisimagineer has requested that we instead make these match the format of the file exported from Clifton for ease of use. The ideal format is as follows (note spaces between some words):

Last Name,First Name,Email,Status,Date Completed,Theme_1,Theme_2,Theme_3,Theme_4,Theme_5,Access Code

bennettforkner:
This PR re-organizes the file structure of the Clifton Strengths importer so that it matches what is initially exported from Gallup. I have also added padding to the GordonSpreadsheetUploader and have centered the download link so that it looks nicer to the end user.

Closes #1585


src/components/SpreadsheetUploader/SpreadsheetUploader.module.scss
```diff
@@ -20,8 +20,11 @@
   }
   &_templatelink {
     color: $primary-blue;
+    display: block;
     text-align: center;
+    font-size: 13px;
```
EjPlatzer, line 25: Can we use `rem` instead of `px` so that the font size will scale properly if the user sets their font-size differently? That's pretty important for accessibility.

PR #1614, opened 2023-01-12
Delete codeql-analysis.yml

bennettforkner:
Requested to be deleted by Evan


EjPlatzer (approved):
Since this is now a built-in GitHub feature, we can safely disable this action.

PR #1620, opened 2023-01-19
Refactor VictoryPromise to function component

EjPlatzer:
This PR continues the effort to migrate our remaining Class-based components to function components, focusing on the VictoryPromise component for display on MyProfile.

During the refactoring, I also refactored the component in other ways, since I was re-writing it almost from scratch:
- I separated out the Graph display and the Icon display as child components. This simplified the logic of each component greatly, so that the parent VictoryPromise component is concerned only with the overall logic of the Victory Promise module (fetching data and showing the correct display), while the children take care of the particulars of displaying the data in the desired way.
- I moved the SVG icons inline (since SVGs are valid HTML/JSX), for easier dynamic styling. This allowed me to directly specify the `fill` property of the SVGs based on the state of each Victory Promise category, which simplified the styling logic greatly.
- I migrated the components to TypeScript for more safety and simpler maintenance. This also allowed some more ergonomic patterns (such as passing string literal props that must be one of a specific set of strings).
- I streamlined the markup and styling. The end result looks functionally identical, but uses less and simpler HTML and CSS to achieve it.


src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * and background colors). The elements are placed in clockwise order on the graph, starting at
+ * the top right. For 4 elements numbered 0 through 3, we get the order:
+ *
+ * 3  0
+ *
+ * 2  1
+ *
+ * To ensure that the graph data lines up with icon order, we use the following map to determine
+ * the array index of each category.
+ */
```
bennettforkner, line 102: Is this the best place for this documentation? It feels like a lot of comment just for explaining what an external package does with our input.

src/components/Profile/components/VictoryPromise/components/VictroyPromiseIcons.tsx
```diff
@@ -0,0 +1,267 @@
+import { Typography } from '@mui/material';
```
bennettforkner, line 1: This filename is spelled incorrectly.

src/components/Profile/components/VictoryPromise/components/VictroyPromiseIcons.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+const CategoryDescription = {
+  christian_character:
+    'Opportunities encouraging faith formation and its connection to living, learning and leading with others',
+  intellectual_maturity:
+    'Opportunities to extend critical reasoning, deepen understanding, and ignite imagination',
+  lives_of_service:
+    "Opportunities to lend one's strengths and talents with our partners to our neighbors",
+  leadership_worldwide:
+    "Opportunities to develop one's understanding and influence in God's amazing, dynamic and challenging world",
```
bennettforkner, line 16:
```suggestion
    'Opportunities encouraging faith formation and its connection to living, learning and leading with others',
  intellectual_maturity:
    'Opportunities to extend critical reasoning, deepen understanding, and ignite imagination',
  lives_of_service:
    'Opportunities to lend one's strengths and talents with our partners to our neighbors',
  leadership_worldwide:
    'Opportunities to develop one's understanding and influence in God's amazing, dynamic and challenging world',
```
I don't mind double quotes if that is what you prefer, but we should be consistent about which quotes we are using in the same object.

src/components/Profile/components/VictoryPromise/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+            {displayMode === DisplayMode.Icon ? (
+              <IconDisplay scores={scores} />
+            ) : (
+              <GraphDisplay scores={scores} />
+            )}
+          </Box>
+          {isOnline && (
+            <Link href="https://www.gordon.edu/victorypromise" className="gc360_text_link">
+              Click here for more information!
+            </Link>
```
bennettforkner, line 60: This should open in a new window since it is not a 360 link, right? I don't think that's the default for the `Link` component.

bennettforkner (changes requested):
A few change requests

src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
EjPlatzer: Yes, because it's documenting the (otherwise arbitrary) reasoning for the values in the GraphOrder object. By putting it here directly above the GraphOrder object, we make it so that anyone hovering over/inspecting GraphOrder will see this documentation.

src/components/Profile/components/VictoryPromise/components/VictroyPromiseIcons.tsx
EjPlatzer: Good catch!

src/components/Profile/components/VictoryPromise/components/VictroyPromiseIcons.tsx
EjPlatzer: Prettier did that automatically. As is apparent with GitHub's syntax highlighting, the double quotes are needed on the last two because they contain single quotes. However, Prettier defaults to single quotes when double quotes aren't needed, which is why they're inconsistent. The only way to override that would be to disable Prettier formatting for these lines, which I don't think is a good option.

src/components/Profile/components/VictoryPromise/index.tsx
EjPlatzer: Good question! The confusion is because `Link` is an overloaded component name. We usually use the `Link` component from `react-router-dom`, which is used for site-internal links like you're describing.

This is the `Link` component from `@mui/material`, which is just an anchor tag styled to be consistent with our other MUI components. That's why this one has an `href` prop like a normal anchor tag while the `react-router-dom` `Link` only accepts `to`, not `href`.

bennettforkner: So my question is, shouldn't we be opening this in a new window using target _blank?

EjPlatzer: Oh right I completely misinterpreted your question. Yes, we should be opening it in a new window probably.

bennettforkner (approved):

PR #1621, opened 2023-01-19
Migrate from Create React App to Vite

EjPlatzer:
This project was originally scaffolded with [Create React App (CRA)](https://github.com/facebook/create-react-app). Although CRA was at one time the recommended way of starting a new React project, it has now become a hindrance to maintenance and development. The main issues with Create React App are:
   
1. Lagging Updates
1. Bloated Config
1. Dwindling Support

[Vite](https://vitejs.dev/) is a frontend tool that enables a significantly improved development experience with a comparable maintenance cost and production output as CRA.

# CRA Problems

Create React App is convenient because it creates an opinionated environment for running a React app on top of [webpack](https://webpack.js.org/). It does this by hard-coding the webpack configuration into the CRA package, as well as adding lots of semi-extraneous dependencies.

Although CRA is very convenient for starting a new project with React - because you don't have to know how to configure e.g. webpack, babel, ESLint, etc. to get started - it becomes a chore and a source of friction for maintaining a React app long term.

## Lagging Updates
   As it ages, CRA sees increasingly fewer and less frequent releases, which means the hard-coded dependencies and configuration show their age more and more. For example, we cannot upgrade our React version until CRA has released a new version that targets that version of React.

## Bloated Config

Since CRA was the de facto option for so many years, it has become bloated with configurations and dependencies for all the different kinds of environments one might want to conveniently spin-up for React - most of which are not relevant to us.

## Dwindling Support

The team of maintainers for CRA has dwindled, and the React community is increasingly moving on to better alternatives that have sprung up in the many years since CRA was first popular.

# CRA Alternatives

There is more than one way to build a React app - CRA has many viable alternatives.

## Ejecting

The "simplest" option to migrate away from CRA is to "eject" - which means to uninstall CRA and instead manually manage all of the dependencies and configuration that CRA was doing for us. Although this is certainly a viable solution in that it allows us to manage our own environment, it has the significant drawback of increasingly the overall complexity of maintenance.

For example, we would now have to maintain our own webpack, babel, ESLint, and TypeScript configurations. We would have to manually review these configurations to remove the cruft that isn't relevant to our project. In the longer term, we would need to do the work of keeping these tools all in-sync and updating them as appropriate.

None of that is prohibitive, but it is unnecessary when there exists an even easier and better way.

## Vite

Vite is a modern tool for building frontend applications. It is newer than some alternatives (such as webpack), but has already gained widespread and massive adoption, both within the React eco-system and across the entire frontend sphere. Vite is the perfect solution for us because it manages to solve all the problems of CRA while also giving an overall better developer experience.

One upside of Vite as compared to other frontend build tools is it's simplicity. When performing this migration, my steps were essentially "install vite, add a simple `vite.config.ts` file, and go". There were a couple of syntactic updates needed (environment variable syntax and using the `.jsx`/`.tsx` extension for files with JSX syntax), but that was essentially it.

Vite essentially offers the same ease of maintenance as CRA, but without the downsides. It won't suffer from lagging updates because it is not a meta-package that mostly aggregates other tools - it is a self-contained build tool. When I switched dependencies from CRA to Vite, our `packge-lock.json` file shrunk by >22,000 lines. Vite does not suffer from a bloated config because it gives us full control of our configuration, while having practical defaults so that we don't have to hand-craft a massive and unique configuration. And it's support is currently second-to-none and only going up. 

Vite's killer feature is using the browser-native ES Modules to serve source code during development, as opposed to the bundling approach of tools like Webpack. This results in a unbelievably faster development experience (on my machine vite took ~500ms to start up the development server, while webpack regularly took >5 minutes). This speed up doesn't only affect startup times - it is perhaps most notable during development, because serving unbundled source code allows Vite to update only the changed files, instead of updating the entire project every time a file changes as bundlers do.

For more info on Vite's advantages over other build tools, check out their [Why Vite](https://vitejs.dev/guide/why.html) page.


.env
```diff
@@ -1,4 +1,4 @@
-REACT_APP_API_URL=/
-REACT_APP_AZURE_AD_CLIENT=abc3be17-5b4c-48c7-9cd8-db28399d772b
-REACT_APP_AZURE_AD_TENANT=0251ecdb-4e2d-4077-9498-40328ac794e9
-REACT_APP_FONT_URL=https://cloud.typography.com/7763712/615144/css/fonts.css
\ No newline at end of file
+VITE_API_URL=/
+VITE_AZURE_AD_CLIENT=abc3be17-5b4c-48c7-9cd8-db28399d772b
+VITE_AZURE_AD_TENANT=0251ecdb-4e2d-4077-9498-40328ac794e9
+VITE_FONT_URL=https://cloud.typography.com/7763712/615144/css/fonts.css
```
bennettforkner, line 4: Do these three really need to start with VITE (or REACT)? Can't they just be AZURE_AD_CLIENT, AZURE_AD_TENANT, etc?

bennettforkner (changes requested):

.env
EjPlatzer: Yes, they do. CRA required that any variables available to the client start with `REACT_`, whereas Vite requires that they start with `VITE_`.

[According to Vite](https://vitejs.dev/guide/env-and-mode.html#env-fileshttps://vitejs.dev/guide/env-and-mode.html#env-files), this is a security mechanism:

> To prevent accidentally leaking env variables to the client, only variables prefixed with VITE_ are exposed to your Vite-processed code.

src/index.tsx
```diff
@@ -3,6 +3,7 @@ import { MsalProvider } from '@azure/msal-react';
 import { ThemeProvider, StyledEngineProvider } from '@mui/material/styles';
 import UserContextProvider from 'contexts/UserContext';
 import { register } from 'pwa';
+import React from "react";
```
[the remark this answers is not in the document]
EjPlatzer: Yeah I think I probably typed that and it didn't lint normally.

.env
bennettforkner: Thanks for clarifying!

bennettforkner (approved):
Look good to me! How do you want to go about merging this? Are you ready to do so now?

PR #1623, opened 2023-01-19
Remove Wellness Check from Nav Header

bennettforkner:
This PR removes the wellness check from the Nav header as approved by Chris Imming. The URL `.../wellness` will still work as before, but it will be obfuscated and un-referenced.


src/components/Header/index.js
```diff
@@ -210,7 +203,6 @@ const GordonHeader = ({ onDrawerToggle }) => {
               />
               {requiresAuthTab('People', <PeopleIcon />)}
               {/* {requiresAuthTab('Timesheets', WorkIcon)} */}
```
EjPlatzer, line 205: Should we put Timesheets back in the header now that we're removing wellness?

bennettforkner: I didn't know that was why it was removed. Yes! Let's add it back.
```suggestion
              {requiresAuthTab('Timesheets', WorkIcon)}
```

src/components/Header/index.js
```diff
@@ -41,13 +40,7 @@ const GordonHeader = ({ onDrawerToggle }) => {
...
     // Tab url regular expressions must be listed in the same order as the tabs, since the
     // indices of the elements in the array on the next line are mapped to the indices of the tabs
-    let urls = [
-      /^\/$/,
-      /^\/involvements\/?$|^\/activity/,
-      /^\/events\/?$/,
-      /^\/people$/,
-      /^\/wellness$/,
-    ];
+    let urls = [/^\/$/, /^\/involvements\/?$|^\/activity/, /^\/events\/?$/, /^\/people$/];
```
EjPlatzer, line 43: I think we need to add timesheets to the tab index.

EjPlatzer (approved):
LGTM!

PR #1624, opened 2023-01-19
[Rec-IM] Waiver Form 

amos-cha:
Screen Shot 2023-01-19 at 3 36 02 PM
Waiver Form is not finalized as this is not the official waiver which is in progress by @bennettforkner. However, this is the waiver form pulled from the official Bennett Center membership application form https://gordonedu.sharepoint.com/WebLinks/Shared%20Documents/Bennett%20Center/Membership%20Application%20(FINAL).pdf?ga=1

Usage:
- Both the checkbox and Electronic signature must be filled out
- Follows our default Form format which @silasw2 plans to factor re-used functions
- "I Agree" enters the profile's AD_Username as a participant
- Cancel returns the profile to the 360/Home to prevent seeing data without being a part of the Participant List

### DIFF
--- a/src/app.global.scss
+++ b/src/app.global.scss
@@ -179,3 +179,7 @@ strong {
     }
   }
 }
+
+.accentText {
+  color: $primary-cyan;
+}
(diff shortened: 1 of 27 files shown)


cpabbot (approved):
LGTM, can't test the form itself but looks good! Do we have an account to test this with?

amos-cha:
> LGTM, can't test the form itself but looks good! Do we have an account to test this with?

I deployed it and got Josh Peters and a few other random students to test what it would look like, then confirmed their permissions and existence by direct query on the database

PR #1628, opened 2023-01-20
Refactor EventItem to function component

EjPlatzer:
I refactored the `EventItem` component from a Class component to a function component. In the process, I removed the redundant `CollapsibleEventItem` component. Both components have a collapsible description and show the same data, but are laid out for different screen sizes. I was able to have one component cover all screen sizes by shifting layout dynamically and improving the formatting slightly. I also refactored `EventItem`'s parent component `EventList` minorly.


bennettforkner (approved):
LGTM!

PR #1629, opened 2023-01-21
Refactor last minor class components

EjPlatzer:
This PR refactors the last small, self-contained class components into function components.

I also took the time to do some minor refactoring of the affected components, and sometimes related components when it made sense (i.e. properly extracting shared styles).


src/services/session.ts
```diff
@@ -7,7 +7,7 @@ type Session = {
   SessionEndDate: string;
 };
 
-type DaysLeft = [daysRemaining: number, daysCompleted: number];
+type DaysLeft = [daysRemaining: number, daysInSemester: number];
```
bennettforkner, line 10: Has this been changed in the API? Was it named incorrectly before?

bennettforkner (changes requested):

src/services/session.ts
EjPlatzer: So this is a JS "tuple", aka an array. So the names are only for TypeScript/Intellisense reference, the elements of the array aren't actually named. Also, this is the correct name, matching the one in the API. It was incorrect before but inconsequential.

bennettforkner: Oh yes, thanks for clarifying for me. Great!

bennettforkner (approved):
LGTM!

PR #1630, opened 2023-01-21
[Rec-IM] Match Editing and Creation

silasw2:
To clarify a bit more on the purpose of this PR, this is the edit icon that can allow a referee or admin to edit the score, sportsmanship, and attendance of the teams involved in the match.

Also this will enable admins to create new matches that were not auto scheduled.


cpabbot (approved):
LGTM

PR #1634, opened 2023-01-23
[Rec-IM] Quick Series UI Update

cpabbot:
## Rec-IM Minor Series Listing UI Update

- changed icons
- fixed listing UI


amos-cha (approved):
Looks good, BUT can we merge this **AFTER** the other PR's so that it doesn't accidentally get overwritten by the incoming PR's

PR #1640, opened 2023-01-25
FIXED - MemberListItem pointing to the wrong URL on mobile.

Khalil-Ullah-Hafeez:
Changed `AD_Username` property to `Username` of the `member` object.


bennettforkner:
Thank you for proposing this fox @Khalil-Ullah-Hafeez . We are actually keeping this issue open for the moment as we intend to solve this issue as part of a demo for our new developers. I am closing this even though it is a partially correct solution.

PR #1641, opened 2023-01-26
[Rec-IM] Admin Page v1

cpabbot:
WIP, but can merge whenever and continue to be developed.

### DIFF
--- a/src/services/recim/activity.ts
+++ b/src/services/recim/activity.ts
@@ -68,7 +68,7 @@ const getActivityByID = (ID: number): Promise<Activity> => http.get(`recim/activ
 
 const getActivities = (active: boolean, time: String): Promise<Activity[]> => {
   if (time) return http.get(`recim/activities?active=${active}&time=${time}`);
-  else if (active) return http.get(`recim/activities?active=${active}`);
+  if (active) return http.get(`recim/activities?active=${active}`);
   return http.get(`recim/activities`);
 };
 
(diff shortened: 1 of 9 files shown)


src/services/recim/activity.ts
```diff
@@ -66,9 +66,10 @@ const createActivity = (newActivity: UploadActivity): Promise<CreatedActivity> =
 
 const getActivityByID = (ID: number): Promise<Activity> => http.get(`recim/activities/${ID}`);
 
-const getAllActivities = (active: boolean, time: String): Promise<Activity[]> => {
+const getActivities = (active: boolean, time: String): Promise<Activity[]> => {
   if (time) return http.get(`recim/activities?active=${active}&time=${time}`);
-  return http.get(`recim/activities?active=${active}`);
+  else if (active) return http.get(`recim/activities?active=${active}`);
```
amos-cha, line 71:
```suggestion
  if (active) return http.get(`recim/activities?active=${active}`);
```
else is redundant on return

src/services/recim/team.ts
```diff
@@ -69,11 +69,14 @@ type PatchTeam = {
 };
 
 //Team Routes
+const getTeams = (active: boolean): Promise<Team[]> => {
+  if (active) return http.get(`recim/Teams?active=${active}`);
+  return http.get(`recim/Teams`);
+};
```
amos-cha, line 75: There is a bug in the API, I assume that this route should automatically give all teams if there is no active tag, but right now it simply has active = true/false

amos-cha (changes requested):
Clean-up logic that I wanted to suggest is in this PR: https://github.com/gordon-cs/gordon-360-ui/pull/1652

src/services/recim/team.ts
amos-cha: fixed here: 
https://github.com/gordon-cs/gordon-360-api/pull/793

amos-cha (approved):

amos-cha:
Looks Good :)

PR #1643, opened 2023-01-26
Fix clifton strengths external links

### LINKED ISSUE #1590: Clifton Strength Links on Profile Broken
The Clifton strengths display on a user's profile should link each strength to the proper Gallup page for the relevant strength. It seems that currently, these links direct you to the 360 homepage.

EjPlatzer:
Clifton Strengths on the profile page is supposed to link to the Gallup introduction to each theme. However, because the link was using React Router's `Link` tag, the `href` prop (which was pointing to the correct page) was being ignored because the `Link` component requires a `to` prop instead.

Closes #1590


bennettforkner (approved):
LGTM!

PR #1649, opened 2023-01-29
[Rec-IM] Activity Layout Changes and Referee Match Stat Editing Functionality

silasw2:
This PR fixes and changes a few things,

- Removes Series listing as it is only used one place and should not be a list to begin with
- Created series headings to be looped over in the activity page so all matches and series can be shown
- Adds the ability for referees to modify match statistics
- Creates a new design for the buttons to be in within their respective cards
- Fixed SCSS global styling issues
- Migrated some functions to a helper file since they are used in multiple locations

Closes https://github.com/gordon-cs/RecIM-Docs/issues/106
Closes https://github.com/gordon-cs/RecIM-Docs/issues/67

### DIFF
--- a/src/services/recim/match.ts
+++ b/src/services/recim/match.ts
@@ -42,7 +42,7 @@ type CreatedMatchStats = {
 
 type PatchMatchStats = {
   TeamID: number;
-  Status: string;
+  StatusID: number;
   Score: number;
   Sportsmanship: number;
 };
--- a/src/services/recim/match.ts
+++ b/src/services/recim/match.ts
@@ -7,7 +7,7 @@ import { Activity } from './activity';
 export type Match = {
   ID: number;
   Scores: TeamMatchHistory[];
-  Activity: Activity; // This will only be used for activity ID
+  Activity: Activity; // This will only be used for activity ID and name
   Time: string;
   Surface: string;
   Status: string;
--- a/src/services/recim/match.ts
+++ b/src/services/recim/match.ts
@@ -42,9 +42,9 @@ type CreatedMatchStats = {
 
 type PatchMatchStats = {
   TeamID: number;
-  StatusID: number;
-  Score: number;
-  Sportsmanship: number;
+  StatusID?: number;
+  Score?: number;
+  Sportsmanship?: number;
 };
 
 type PatchMatch = {
--- a/src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
+++ b/src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
@@ -1,0 +1,254 @@
+import { Grid } from '@mui/material';
+import { useState, useMemo, useEffect } from 'react';
+import GordonDialogBox from 'components/GordonDialogBox';
+import { createTeam, editTeam, getTeamStatusTypes } from 'services/recim/team';
+import { InformationField } from '../components/InformationField';
+import { ConfirmationWindowHeader } from '../components/ConfirmationHeader';
+import { ConfirmationRow } from '../components/ConfirmationRow';
+import { ContentCard } from '../components/ContentCard';
+import GordonLoader from 'components/Loader';
+import { useUser } from 'hooks';
+
+const EditMatchStatsForm = ({
+  match,
+  closeWithSnackbar,
+  openEditMatchStatsForm,
+  setOpenEditMatchStatsForm
+}) => {
+  const [errorStatus, setErrorStatus] = useState({
+    Name: false,
+    ActivityID: false,
+    Logo: false,
+    statusID: false,
+  });
+
+  const { profile } = useUser();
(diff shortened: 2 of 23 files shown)


src/services/recim/match.ts
```diff
@@ -42,7 +42,7 @@ type CreatedMatchStats = {
 
 type PatchMatchStats = {
   TeamID: number;
-  Status: string;
+  StatusID: number;
```
amos-cha, line 45: TY for this fix btw

src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
```diff
@@ -0,0 +1,245 @@
...
+      label: 'Status',
+      name: 'StatusID',
+      type: 'select',
+      menuItems: matchStatus.map((type) => {
+        return type.Description;
+      }),
+      error: errorStatus.Name,
+      helperText: '*Required',
+    },
+  ];
```
amos-cha, line 59: Your error handling does not work? 
errorStatus.Name does not exist, you have:
`const [errorStatus, setErrorStatus] = useState({
    Score: false,
    StatusID: false,
  });`

src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
```diff
@@ -0,0 +1,245 @@
...
+
+  const allFields = [createMatchStatsField].flat();
+
+  const currentInfo = useMemo(() => {
+    return {
+      TeamID: teamMatchHistory.TeamID,
+      Score: teamMatchHistory.TeamScore,
+      Sportsmanship: '',
```
amos-cha, line 67:
```suggestion
      Sportsmanship: 0,
```
despite sportsmanship not working fully at the moment, I don't know if TypeScript will be happy casting an empty string to a number. Once I implement the sportsmanship functionality, we might also get errors if we don't make the change now.

src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  const handleConfirm = () => {
+    debugger;
+    setSaving(true);
+    let matchStatsRequest = { ...currentInfo, ...newInfo };
+    matchStatsRequest.Score = parseInt(matchStatsRequest.Score);
+    matchStatsRequest.Sportsmanship = parseInt(matchStatsRequest.Sportsmanship);
+    matchStatsRequest.TeamID = parseInt(matchStatsRequest.TeamID);
+    matchStatsRequest.StatusID = parseInt(
```
amos-cha, line 150: we don't need to parse int if we handle field validation (which was set to errorStatus.Name which did not exist)

amos-cha (changes requested):
Amazing stuff! A few questions regarding error handling but beyond that, I think you've done some great stuff.

src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
```diff
@@ -0,0 +1,244 @@
...
+          Field: key,
+          Value: value,
+          Label: getFieldLabel(key),
+        });
+    });
+    return updatedFields;
+  }
+
+  const handleConfirm = () => {
+    debugger;
```
amos-cha, line 144:
```suggestion

```

amos-cha (changes requested):

src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
silasw2: Whoops nice catch!

src/views/RecIM/components/Forms/EditMatchStatsForm/index.js
silasw2: I removed this, but am waiting for patch to be fixed before I can see if the typescript can handle the change.

silasw2: It seems that the TS is ok with having them be passed as strings, so I will resolve this for now.

amos-cha (approved):
Looks Good!

cpabbot (dismissed):
LGTM

src/views/RecIM/components/List/Listing/index.js
```diff
@@ -21,8 +21,56 @@ import EventAvailableIcon from '@mui/icons-material/EventAvailable';
...
+//             </Grid>
+//           </Grid>
+//         </Grid>
+//         <Grid item xs={12} sm={4}>
+//           {status()}
+//         </Grid>
+//       </Grid>
+//     </ListItem>
+//   );
+// };
```
amos-cha, line 72: @cpabbot why not just delete this?

amos-cha (changes requested):

amos-cha (approved):

PR #1650, opened 2023-01-30
[Rec-IM] User-Test#1 - Fix null isAdmin on reload, fix logic error

amos-cha:
bugs found in user test 1


cpabbot (approved):
LGTM, clean good work!

PR #1657, opened 2023-02-04
[Rec-IM] Don't show Inactive Team Participants for Non-Admins

silasw2:


cpabbot (changes requested):
This is great! One question - should this even be handled by the UI or should the API not even return inactives unless you are admin? For now this is fine, I honestly don't know.

Requesting one small change - can we invert the logic so the tag is instead `showInactive`? I think this is slightly more readable and then the default (without adding that tag) is correctly not including inactive (which we generally don't want).

Also if you do that, I thought of one possible logical simplification but now I'm thinking this is more complicated. You can leave as is but this is one other option: (I think how you had it is more readable though).
```
!(showInactive || participant.Role === "Inactive") || !showInactive && ... [jsx]
```

src/views/RecIM/views/Team/index.js
```diff
@@ -153,7 +153,7 @@ const Team = () => {
         <CardHeader title="Roster" className={styles.cardHeader} />
         <CardContent>
           {hasPermissions ? (
-            <ParticipantList participants={team.Participant} showParticipantOptions />
+            <ParticipantList participants={team.Participant} showParticipantOptions showInactive />
```
amos-cha, line 156: I like this change! Makes it much cleaner :)

amos-cha (approved):
Looks good!

cpabbot (approved):
LGTM! Thanks for quick change

PR #1664, opened 2023-02-07
Refactor Experience Transcript

EjPlatzer:


package.json
```diff
@@ -66,6 +65,7 @@
     "sass": "^1.44.0",
     "stylelint": "^14.9.0",
     "stylelint-config-recommended-scss": "^5.0.2",
+    "typescript": "^4.9.4",
```
bennettforkner, line 68: Feels like this isn't in the scope of this PR ;)

PR #1666, opened 2023-02-09
Use simpler Memberships endpoint

EjPlatzer:
This PR makes use of the simpler and more flexible `GET /api/memberships` endpoint introduced by gordon-cs/gordon-360-api#814 to simplify, unify, and streamline the handling of memberships in the UI.

The main gains are in `services/membership.ts`, where much of the boilerplate for fetching and filtering different data points around a user's or involvement's membership have all been unified into the simple `membershipService.get(queryParams)` function. This enables different parts of the application to query exactly the memberships data needed in an ergonomic way, without exposing any more (or any private) membership data.

In the process of this refactor, I also made sure that the `Participation` enum is used consistently instead of repeated (and error-prone) string literals when referencing the `Participation` field of membership objects.


src/services/membership.ts
```diff
@@ -48,115 +57,76 @@ export type MembershipUpload = {
   Privacy: boolean;
 };
 
+const get = async (queryParams: {
+  involvementCode?: string;
+  username?: string;
+  sessionCode?: string;
+  participationTypes?: Participation[] | Participation;
+}): Promise<MembershipView[]> => http.get(`memberships/${http.toQueryString(queryParams)}`);
```
bennettforkner, line 65: You use a `/` before the query string here whereas in emails.ts you do not (line 18). Is there a reason for this inconsistency or can we keep it the same in one pull request?

bennettforkner (changes requested):

src/services/http.ts
```diff
@@ -132,12 +132,33 @@ const dataURItoBlob = (dataURI: string) => {
...
 };
 
+const toQueryString = (
+  queryParams?: Record<string | number | symbol, string | number | boolean | Array<any>>,
+): string => {
+  if (!queryParams) return '';
+
+  const urlSearchParams = new URLSearchParams();
+
+  Object.entries(queryParams).forEach(([key, value]) => {
```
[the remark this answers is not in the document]
EjPlatzer: Yeah, I can add that.

src/services/membership.ts
EjPlatzer: Ah, it shouldn't have the slash, that's leftover from a previous url that used route params instead of query params. Thanks for catching that!

bennettforkner (approved):
LGTM!

PR #1669, opened 2023-02-11
[Rec-IM] navUpdate review

amos-cha:
@cpabbot

### DIFF
--- a/src/views/RecIM/components/Header/index.js
+++ b/src/views/RecIM/components/Header/index.js
@@ -5,8 +5,6 @@ import { DateTime } from 'luxon';
 import styles from './Header.module.css';
 import { useUser } from 'hooks';
 import { getParticipantByUsername } from 'services/recim/participant';
-import ActivityForm from 'views/RecIM/components/Forms/ActivityForm';
-import TeamForm from 'views/RecIM/components/Forms/TeamForm';
 import EditMatchStatsForm from 'views/RecIM/components/Forms/EditMatchStatsForm';
 import GordonLoader from 'components/Loader';
 import recimLogo from './../../recim_logo.png';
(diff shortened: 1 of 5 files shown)


src/views/RecIM/components/Header/index.js
```diff
@@ -24,22 +21,27 @@ const RecIMBreadcrumb = ({ link, children }) => {
...
+}) => {
   const { profile } = useUser();
   const [user, setUser] = useState();
-  const [openActivityForm, setOpenActivityForm] = useState(false);
-  const [openTeamForm, setOpenTeamForm] = useState(false);
   const [isCaptain, setIsCaptain] = useState(false);
-  const [team0Score, setTeam0Score] = useState(0);
-  const [team1Score, setTeam1Score] = useState(0);
-  const [openEditMatchStatsForm, setOpenEditMatchStatsForm] = useState(false);
-  const [selectedScores, setSelectedScores] = useState();
```
amos-cha, line 36: Having the form logic within the header makes this component **far** too bloated. Furthermore, it prevents refresh on save since the fetch is happening on the actual ____ page

src/views/RecIM/components/Header/index.js
```diff
@@ -24,22 +21,27 @@ const RecIMBreadcrumb = ({ link, children }) => {
...
-const Header = ({ page, expandable = false, match, team, activity, home, admin }) => {
+const Header = ({
+  page,
+  expandable = false,
+  match,
+  team,
+  activity,
+  home,
+  admin,
+  setOpenHeaderForm,
```
amos-cha, line 32: All forms simply are just setOpen____Form so this would just make it more simple. As opposed to a bunch of states.

src/views/RecIM/components/Header/index.js
```diff
@@ -218,6 +170,10 @@ const Header = ({ page, expandable = false, match, team, activity, home, admin }
                   {match?.Team[0]?.Name ?? 'No team yet...'}
                 </Typography>
               </LinkRouter>
+              <Typography className={styles.subtitle}>
+                {/* once this is added to the API, it will instantly work */}
+                {match?.Team[0]?.TeamRecord.Win ?? 0}W : {match?.Team[0]?.TeamRecord.Loss ?? 0}L
+              </Typography>
```
amos-cha, line 176: I viewed this as a non-admin, the header looks a bit naked without the Sportsmanship line. Added the W/L column just to fill negative space. Plus, it is a relatively nice feature. Feel free to adjust the styling. I just stole the subtitle style you had (obviously)

src/views/RecIM/views/Activity/index.js
```diff
@@ -44,7 +46,14 @@ const Activity = () => {
...
     loadData();
-  }, [profile, activityID, openTeamForm, openCreateSeriesForm, openCreateMatchForm]);
+  }, [
+    profile,
+    activityID,
+    openTeamForm,
+    openCreateSeriesForm,
+    openCreateMatchForm,
+    openActivityForm,
+  ]);
```
amos-cha, line 56: This is an example of how we force a page reload upon submit (specifically when the openActivityForm is changed from true->false)

amos-cha (commented):
In general, it was a really good change. I think we need a future PR to turn all this into a useContext so that we don't have to pass so many random things up and down and back up again. This is the cleanest "working" implementation I could think of without using Context. 

Good work @cpabbot !

PR #1672, opened 2023-02-15
Wrap ShiftDisplay in LocalizationProvider

EjPlatzer:
The LocalizationProvider for Timesheets was wrapping the whole page except for the `ShiftDisplay`. When someone tried to edit shifts within the `ShiftDisplay`, they would receive a MUI error because no LocalizationProvider was present.


bennettforkner (approved):
Thanks for a simple fix!

amos-cha:
Thank you @EjPlatzer!

PR #1674, opened 2023-02-15
[Rec-IM] Quick Add Nav Button to Rec-IM

cpabbot:
added drawer button, added dash to header button

### DIFF
--- a/src/views/RecIM/components/List/Listing/index.js
+++ b/src/views/RecIM/components/List/Listing/index.js
@@ -269,9 +269,30 @@ const MatchListing = ({ match, activityID }) => {
         className={styles.listing}
       >
         <Grid container>
-          <Grid item>
-            {match.Team[0]?.Name} vs. {match.Team[1]?.Name}
-          </Grid>
+          {match.Team.length > 2 ? (
+            match.Team.map((team) => (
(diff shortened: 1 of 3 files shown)


src/views/RecIM/components/List/Listing/index.js
```diff
@@ -269,9 +269,30 @@ const MatchListing = ({ match, activityID }) => {
...
+              <>
+                <Grid item xs={10}>
+                  {team.Name}
+                </Grid>
+                <Grid item xs={2} textAlign="center">
+                  vs.
+                </Grid>
+              </>
+            ))
+          ) : (
```
amos-cha, line 283: What happens to the last team in that list?
wouldn't it look like:

Team A vs Team B vs Team C vs
?

Also wouldn't this be an incredibly long list of names? I argue we just throw a:
<Typography style={styles.subtitle} >
Multiple Teams
</Typography>

amos-cha (changes requested):
I think it's a good change, but while you're adding the nav button, can you also add recim to the burger menu? There is a good chance our first activity will be primarily mobile users, so we should at least have rec-im on the burger menu to prevent having to manually cast the route.

PR #1677, opened 2023-02-16
Clifton Strengths to show on both Fac/Staff and Student Profiles

### LINKED ISSUE #1642: Fac/Staff Clifton Strengths do not show up on profile
I'm not sure when these disappeared, but @ejplatzer and my Clifton Strengths are no longer displaying on the profile page.

charlielee144r:
This PR fixes some logic in fetching a user profile so that Clifton Strengths show on the profiles of both Fac/Staff and Students if it is available (and/or public?). 

Fixes #1642

### DIFF
--- a/src/services/user.ts
+++ b/src/services/user.ts
@@ -262,13 +262,16 @@ const getProfileInfo = async (username: string = ''): Promise<Profile | undefine
   if (!profile) return undefined;
 
   const fullName = `${profile?.FirstName} ${profile?.LastName}`;
+  const getCliftonStrengths = await CliftonStrengthsService.getCliftonStrengths(
+    profile.AD_Username,
+  );
 
   if (isStudent(profile)) {
     return {
       ...profile,
       fullName,
       Advisors: await getAdvisors(profile.AD_Username),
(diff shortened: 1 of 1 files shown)


src/services/user.ts
```diff
@@ -262,13 +262,16 @@ const getProfileInfo = async (username: string = ''): Promise<Profile | undefine
   if (!profile) return undefined;
 
   const fullName = `${profile?.FirstName} ${profile?.LastName}`;
+  const getCliftonStrengths = await CliftonStrengthsService.getCliftonStrengths(
```
bennettforkner, line 265: I wouldn't call this `getCliftonStrengths`. That name implies that it is a function, not a variable.
```suggestion
  const cliftonStrengths = await CliftonStrengthsService.getCliftonStrengths(
```

bennettforkner (changes requested):
Looking good, just a variable name change that I am requesting.

src/services/user.ts
charlielee144r: That makes sense. Thank you!

charlielee144r (commented):
Done!

bennettforkner (approved):
Looks Good To Me!

EjPlatzer (approved):

PR #1678, opened 2023-02-16
Fix Case for Showing no Events Message

bennettforkner:
This check to see if there are any events with the current filters is checking for `length < 0` which is a case that will never occur. This PR updates the logic to check `length <= 0` instead.

### DIFF
--- a/src/components/EventList/index.jsx
+++ b/src/components/EventList/index.jsx
@@ -65,7 +65,7 @@ const EventList = ({ events }) => {
       <CardHeader title={header} className={styles.header} />
       <Grid>
         <List className="gc360_event_list" disablePadding>
-          {events?.length <= 0
+          {events?.length < 1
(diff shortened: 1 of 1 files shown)


src/components/EventList/index.jsx
```diff
@@ -65,7 +65,7 @@ const EventList = ({ events }) => {
       <CardHeader title={header} className={styles.header} />
       <Grid>
         <List className="gc360_event_list" disablePadding>
-          {events?.length < 0
+          {events?.length <= 0
```
EjPlatzer, line 68: I think `< 1` is probably clearer here

bennettforkner:
```suggestion
          {events?.length < 1
```

EjPlatzer (approved):

PR #1684, opened 2023-02-18
[Rec-IM] Activity Start and End

cpabbot:
Adds start datetime and end datetime to the activity listing and page.

**Known issue:**
- I disabled lint warning to add the const activityFields to the dependency array. I think this approach should be adjusted in the future and I did not want to mess with it too much.


src/services/recim/activity.ts
```diff
@@ -4,58 +4,39 @@ import { Team } from './team';
...
+  ID: number;
+  RegistrationOpen: boolean;
   Logo: string;
   Series: Series[];
   Team: Team[];
 };
 
-type CreatedActivity = {
+type CreatedActivity = BaseActivity & {
   ID: number;
```
amos-cha, line 29: I don't mind this logic if it works, but if you start it here make sure you handle it in the other service files eventually!

amos-cha (commented):
I'm gonna finish my review after I handle my PR.

src/views/RecIM/components/Helpers/index.js
```diff
@@ -1,11 +1,35 @@
 import { DateTime } from 'luxon';
 
 const standardDate = (date, includeTime) => {
+  if (!DateTime.isDateTime(date)) date = DateTime.fromISO(date);
```
amos-cha, line 4: this looks incredibly cursed

amos-cha: Is there no way to explicitly cast date to a datetime without checking if !DateTime.isDateTime(date)?

src/views/RecIM/components/Helpers/index.js
```diff
@@ -1,11 +1,35 @@
...
+    return `${standardDate(startDateTime, true).replace(':00', '')} - ${endTime}`;
+  }
+  // Jan 1 3:00pm - Jan 31 3:00pm
+  let isOneYear = startDateTime.year === endDateTime.year;
+  if (isOneYear) {
+    return `${standardDate(startDateTime)} - ${standardDate(endDateTime)}`;
+  }
+  // Jan 1 2023 3:00pm - Jan 1 2024 3:00pm
+  return `${startDateTime.toLocaleString(DateTime.DATE_MED)} - ${endDateTime.toLocaleString(
+    DateTime.DATE_MED,
```
amos-cha, line 31: you're going to have to explain this to me in person

amos-cha (dismissed):
Looks good, but I will need some clarification in person

src/views/RecIM/components/Helpers/index.js
cpabbot: Not that I'm aware of, since they are specific formats I don't think there's a way to cast like that but not sure.

src/views/RecIM/components/Helpers/index.js
cpabbot: I'm not sure if it's the cleanest way to do it, but this is all just to simplify formatting as you can see in the comments above each section. Only show the year if they are different years; if it's a one day tournament just show the date once and then two times; etc.

amos-cha: ah, okay but what is the usecase for over a year, we do that in testing but there should never be a rec-im activity that lasts more than a couple months at max

bennettforkner: I might call this `formatDateTimeRange(...)` to imply that it is returning a string format of the date/s passed in. I think that makes a little bit more sense to a developer without context.

bennettforkner: > ah, okay but what is the usecase for over a year? we do that in testing but there should never be a rec-im activity that lasts more than a couple months at max

I'd push back on this @amos-cha because I believe this is a generic function that promises to handle date ranges and return a proper formatted string. It should not be tied in with the logic of the rest of the application unless it has a more tailored function.

cpabbot: > I might call this `formatDateTimeRange(...)` to imply that it is returning a string format of the range of the dates passed in. I think that makes a little bit more sense to a developer without your context.

Agreed and I had that same thought as well, but we have an existing helper function called standardDate() so I adopted that format for this. I'll change this though.

src/views/RecIM/components/Forms/ActivityForm/index.jsx
```diff
@@ -195,10 +225,12 @@ const ActivityForm = ({
       if (currentInfo[field] !== newInfo[field]) {
         hasChanges = true;
       }
-      handleSetError(field, newInfo[field] === '');
-      hasError = newInfo[field] === '' || hasError;
+      let isFieldRequired = activityFields.find((n) => n.name === field).required;
+      handleSetError(field, !newInfo[field] && isFieldRequired);
+      if (!newInfo[field] && isFieldRequired) hasError = true;
```
amos-cha, line 230: Note: done for a trial run to see if there is a performance difference

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -97,11 +93,13 @@ const ActivityListing = ({ activity }) => {
             </Grid>
           </Grid>
           <Grid item container xs={12} sm={7} direction="column" spacing={1}>
-            <Grid item>
-              <Typography sx={{ color: 'gray', fontWeight: 'bold' }}>
-                ActivityStart - ActivityEnd
-              </Typography>
-            </Grid>
+            {activity.StartDate && (
```
amos-cha, line 96:
```suggestion
            {activity.StartDate && activity.EndDate && (
```
Since both are nullable, I don't know if we should be just checking for a single one

amos-cha (changes requested):
1 minor change, you can commit straight from github if you agree.

amos-cha (approved):
Dope! Curious to see how the new methodology pans out!

PR #1685, opened 2023-02-18
[Rec-IM] Cohesive Button Design

silasw2:
This PR makes the team page and home page have admin button styling identical to the activity page button styling introduced a while ago. Below are pictures with the look of the new pages.

Home Page:
image

Team Page:
image


amos-cha (approved):
Looks good, thanks for the double catch on CanCreateTeam!, I saw that and was going to make that an issue.

PR #1686, opened 2023-02-19
[Rec-IM] Allow Form Submit on Button Key Press

silasw2:
This is going to remain a draft for a while I believe, and may end up morphing into the branch that handles our form cleanup. Essentially, we could put a listener for an enter keypress on the GordonDialogBox component, but that makes it so if we press enter once, both the form submits and the confirmation is dismissed (if Cam has any advice for this you can see the commented code in GordonDialogBox file). I opted to place the listener in our activity page. I did not make it activate the handle submit, as it would need to be wrapped in a useCallback hook. Right now, the keypress will open the confirmation page, if it is able to be opened but I would like some input before I move further in this direction.

### DIFF
--- a/src/views/RecIM/components/Forms/ActivityForm/index.js
+++ b/src/views/RecIM/components/Forms/ActivityForm/index.js
@@ -278,6 +278,20 @@ const ActivityForm = ({
     setNewInfo(currentInfo);
   };
 
+  useEffect(() => {
+    const listener = (event) => {
+      if (event.code === 'Enter' || event.code === 'NumpadEnter') {
+        if (!openConfirmWindow && !disableUpdateButton) {
+          setOpenConfirmWindow(true);
+        }
+      }
+    };
+    document.addEventListener('keypress', listener);
+    return () => {
+      document.removeEventListener('keypress', listener);
+    };
+  }, [openConfirmWindow, disableUpdateButton]);
+
   /**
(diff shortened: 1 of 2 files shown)


src/views/RecIM/components/Forms/ActivityForm/index.js
```diff
@@ -278,6 +278,20 @@ const ActivityForm = ({
     setNewInfo(currentInfo);
   };
 
+  useEffect(() => {
+    const listener = (event) => {
+      if (event.code === 'Enter' || event.code === 'NumpadEnter') {
```
amos-cha, line 283: I'm going to comment on this line just so that we can see replies instead of scrolling for a year. I honestly think that it would be a great addition to the GordonDialogBox. However, I do think that we should handle some sort of _timeout_ to ensure that the confirmation window doesn't instantly get submitted if the User presses enter too quickly. This shouldn't affect other DialogBoxes as I'm pretty sure the standard usecase revolves around actually filling out a form.

bennettforkner: I would be concerned with implementing this globally since it is not necessarily a feature that we would want on all dialog boxes. Instead, I would use a `closeDialog()` function passed into the child component (in these cases, a form), then handle the key press in the child component.

PR #1688, opened 2023-02-20
[Rec-IM] activity registration condition fix + team captain immutability 

amos-cha:
If a user wasn't participating in the activity, the condition was overwritten despite the activity registration being closed.

captain status cannot be changed. listing updated to add minimal tag if admin.

### DIFF
--- a/src/views/RecIM/components/List/index.jsx
+++ b/src/views/RecIM/components/List/index.jsx
@@ -26,7 +26,7 @@ const ParticipantList = ({
       <ParticipantListing
         key={participant.username}
         participant={participant}
-        minimal={minimal}
+        minimal={minimal || participant.Role === 'Team-captain/Creator'} //temporarily disable captains from leaving or changing their status
         callbackFunction={callbackFunction}
         showParticipantOptions={showParticipantOptions}
       />
--- a/src/views/RecIM/components/List/index.jsx
+++ b/src/views/RecIM/components/List/index.jsx
@@ -26,7 +26,7 @@ const ParticipantList = ({
       <ParticipantListing
         key={participant.username}
         participant={participant}
-        minimal={minimal || participant.Role === 'Team-captain/Creator'} //temporarily disable captains from leaving or changing their status
+        minimal={minimal || participant.Role === 'Team-captain/Creator'} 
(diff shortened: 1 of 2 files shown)


src/views/RecIM/components/List/index.jsx
```diff
@@ -26,7 +26,7 @@ const ParticipantList = ({
       <ParticipantListing
         key={participant.username}
         participant={participant}
-        minimal={minimal}
+        minimal={minimal || participant.Role === 'Team-captain/Creator'} //temporarily disable captains from leaving or changing their status
```
silasw2, line 29: Interesting, good change! The naming on that variable should change though, and I think changing the comment to just remain there (so remove the temporarily) would be good. While that might be a little out of scope of the PR, it is something we have been meaning to do so it would be great if it could be now.

src/views/RecIM/views/Activity/index.jsx
```diff
@@ -60,11 +60,10 @@ const Activity = () => {
   useEffect(() => {
     if (activity && userTeams && user) {
       let participating = false;
-      setCanCreateTeam(activity.RegistrationOpen);
       userTeams.forEach((team) => {
         if (team.Activity.ID === activity.ID) participating = true;
       });
-      setCanCreateTeam(!participating || user.IsAdmin);
+      setCanCreateTeam(!participating || user.IsAdmin || activity.RegistrationOpen);
```
silasw2, line 66: Another good change!

silasw2 (changes requested):
See comments for details.

src/views/RecIM/components/List/index.jsx
amos-cha: I've made it an issue to be handled better. But for now, without those usecases, we shouldn't allow a team to accidentally have no people.

amos-cha, line 29:
```suggestion
        minimal={minimal || participant.Role === 'Team-captain/Creator'} 
```

silasw2: Ok, sounds good!

silasw2 (approved):
LGTM!

PR #1691, opened 2023-02-21
[Rec-IM] Visual Bug fix

amos-cha:
The main one is that I was unaware of what Minimal meant. I thought it implies "no options" but then did not realize we had a separate tag for it.


cpabbot (approved):
LGTM

PR #1694, opened 2023-02-22
Fix Happy Birthday Message

bennettforkner:
The "is birthday today" logic in the user service was comparing date ISO formats to see if a user's birthday was the current date. Unfortunately, the user's birthday in ISO format also includes the year that they were born so this will essentially never resolve to true.

This PR updates the logic so that it compares the month number and the day of the month (1-30ish) rather than the ISO date string.

This change will officially activate the "Happy Birthday" message feature for the first time in Production.

### DIFF
--- a/src/services/user.ts
+++ b/src/services/user.ts
@@ -1,8 +1,8 @@
 import { DateTime } from 'luxon';
 import { Platform, platforms, socialMediaInfo } from 'services/socialMedia';
 import CliftonStrengthsService, { CliftonStrengths } from './cliftonStrengths';
-import { Class } from './peopleSearch';
 import http from './http';
+import { Class } from './peopleSearch';
 import { Override } from './utils';
 
 type CLWCredits = {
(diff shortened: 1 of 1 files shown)


Kevin-Chu58 (approved):
LGTM! That's why I never got anything in my birthday :)

src/services/user.ts
```diff
@@ -249,7 +249,7 @@ const getBirthdate = async (): Promise<DateTime> =>
 
 const isBirthdayToday = async () => {
   const birthday = await getBirthdate();
-  return birthday?.toISODate() === DateTime.now().toISODate();
+  return birthday?.month === DateTime.now().month && birthday?.day === DateTime.now().day;
```
EjPlatzer, line 252: I know that this is not part of this pull request, but I would love to see the birthday logic cleaned up to not use the `Luxon` library. Everything you're doing with these dates can be done really easily with native JS dates:
```typescript
const getBirthdate = async (): Promise<Date> =>
  http.get('profiles/birthdate').then(Date.parse) // Date.parse handles ISO 8601 format natively

const isBirthdayToday = async () => {
  const now = Date.now();
  const birthdate = await getBirthdate();
  return birthdate.getMonth() === now.getMonth() && birthdate.getDate() === now.getDate();
}
```

EjPlatzer (commented):
This makes sense, but I wonder why we're using `Luxon` for this, when JS Dates can do all of this natively.

PR #1697, opened 2023-02-22
HOTFIX: Add Involvement Join Buttons Back

bennettforkner:
The recent PR, #1695, only fixed 1/2 instances of the incorrect logic. This PR resolves the second and last of these.

### DIFF
--- a/src/views/InvolvementProfile/components/Membership/index.jsx
+++ b/src/views/InvolvementProfile/components/Membership/index.jsx
@@ -22,7 +22,7 @@ const Membership = ({ isAdmin, isSiteAdmin, involvementDescription, toggleIsAdmi
   const [isMemberAndNonGuest, setIsMemberAndNonGuest] = useState(false);
 
   useEffect(() => {
-    const loadMembers = async () => {
+    const loadMembershipStats = async () => {
       setLoading(true);
 
       try {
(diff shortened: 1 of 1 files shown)


src/views/InvolvementProfile/components/Membership/index.jsx
```diff
@@ -33,7 +35,11 @@ const Membership = ({ isAdmin, isSiteAdmin, involvementDescription, toggleIsAdmi
         setFollowersNum(followersNum);
         setMembersNum(membersNum);
 
-        if ((membership && membership.Participation !== Participation.Guest) || isSiteAdmin) {
+        setIsMemberAndNonGuest(
+          (membership && membership.Participation !== Participation.Guest) || isSiteAdmin,
+        );
+
+        if (isMemberAndNonGuest) {
```
EjPlatzer, line 42: This will not work (reliably) because `setState` does not run sequentially, the state update gets queued to run in the future. So the `setIsMemberAndNonGuest` call will schedule the update to `isMemberAndNonGuest` but within this function the value of `isMemberAndNonGuest` will not immediately change so the `if` on line 42 won't be true when it ought to be.

src/views/InvolvementProfile/components/Membership/index.jsx
```diff
@@ -19,8 +19,10 @@ const Membership = ({ isAdmin, isSiteAdmin, involvementDescription, toggleIsAdmi
   const { involvementCode, sessionCode } = useParams();
   const { profile } = useUser();
 
+  const [isMemberAndNonGuest, setIsMemberAndNonGuest] = useState(false);
```
EjPlatzer, line 22: This variable name is misleading because it is also true if the user is a SiteAdmin.

Maybe `canViewAllMembers` would be better, since that's what the value is being used for?

EjPlatzer (changes requested):

src/views/InvolvementProfile/components/Membership/index.jsx
bennettforkner: Well the value is also used for the join and subscribe button display, that is why I didnt use that name

EjPlatzer: That's not entirely accurate. It's checked to decide if we should display the `MemberList` component, and the logic is that, if we don't show the `MemberList` component then we should show the join and subscribe buttons.

EjPlatzer: So I still think the name `canViewAllMembers` is accurate.

bennettforkner: I prefer shouldShowMemberships as I think it makes more sense in that it surrounds a conditional state on the component rather than a condition on the user.

bennettforkner:
I also just removed the additional logic that I put in for not showing the empty table of members. That was only necessary because the second check was broken and I did not realize it during my first PR.

EjPlatzer (approved):
LGTM!

PR #1699, opened 2023-02-22
[Rec-IM] HMR vite fix 

amos-cha:
Ideas:
- rename all files to explicitly state component name


bennettforkner:
Hey Amos, most of these changes should be proposed against develop as they have nothing to do with recim. After that, I would pull them to recim and add your recim-specific changes. Otherwise, this is going to clutter and taint your potential merge into develop, making it near impossible to review.

amos-cha:
> Hey Amos, most of these changes should be proposed against develop as they have nothing to do with recim. After that, I would pull them to recim and add your recim-specific changes. Otherwise, this is going to clutter and taint your potential merge into develop, making it near impossible to review.

True, I was leaving this branch so I could get back to it at some point if we bothered trying to handle it. I'll close the PR but keep the branch for now. I might just drop this idea as a whole since I have a lot of other tasks to handle.

PR #1700, opened 2023-02-24
[Rec-IM] Fix quick search overlay bug

cpabbot:


amos-cha (changes requested):
Ni c e

amos-cha (approved):

amos-cha:
image
bug still exists

PR #1702, opened 2023-02-25
Create new 409 Error 'Conflict' to give a 'Member Already Added' error message for involvements

### LINKED ISSUE #1701: Involvements: Snackbar not generated when adding an already existing member
When adding an already existing member to an involvement/club, there is no snackbar with an error message. 

The case to catch this error is being bypassed, resulting in no snackbar message at all.

charlielee144r:
The current mode for catching 'Member Already Added' exceptions/errors is bypassed because the error message does not return exactly what is expected by the code (replicated by trying to add an already existing member to a club/involvement). 
As a result, there is currently no visual/UI response when trying to add a member that already exists. There is only a console error.

~~This PR redirects the catching of the error to the default of a switch/case statement so that a snackbar shows (at the very least) that an error has occurred.~~

~~Bennett and I discussed this very abruptly during our last weekly team meeting so I'm not certain if this is the best way to approach this issue. It may require or benefit from a fix on the API side?~~

This PR creates a new error 'Conflict Error' which is caught when trying to add an already existing member. The caught error will product a snackbar that states that `'username' is already a member`.

Fixes #1701

### DIFF
--- a/src/views/InvolvementProfile/components/Membership/components/AdminCard/index.jsx
+++ b/src/views/InvolvementProfile/components/Membership/components/AdminCard/index.jsx
@@ -21,6 +21,7 @@ import membershipService from 'services/membership';
 import { stripDomain } from 'services/utils';
 import { gordonColors } from 'theme';
 import RequestsReceived from './components/RequestsReceived';
+import { ConflictError, NotFoundError } from 'services/error';
 
 const headerStyle = {
   backgroundColor: gordonColors.primary.blue,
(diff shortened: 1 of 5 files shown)


src/views/InvolvementProfile/components/Membership/components/AdminCard/index.jsx
```diff
@@ -71,18 +71,15 @@ const AdminCard = ({ createSnackbar, isSiteAdmin, involvementDescription, onAddM
...
+          break;
 
-          default:
-            console.log(error);
-            break;
-        }
+        default:
+          createSnackbar('This member could not be added', 'error');
+          console.log(error);
+          break;
```
bennettforkner, line 82: You don't have to break the default case since it is at the end of the switch statement. Up to you if you want to keep it or not.
```suggestion
```

bennettforkner (dismissed):
LGTM! Just one unnecessary line in the code, but that won't cause an error.

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.jsx
charlielee144r: I'll remove it. Thank you!

bennettforkner (approved):

EjPlatzer (commented):
This is a good bandaid for the problem that adding a redundant member fails silently. However, this only gives a really generic error message that doesn't inform the user **why** their action failed.

The API returns a `409 Conflict` response when you try to add a user to an involvement that they're already a member of. It would be nice if we added in some error-handling to the `http` service to detect the 409 status and return a `ConflictError` object, which we can then detect here and handle with a more informative error message.

charlielee144r:
Closed due to unwanted files in PR. Continued [here](https://github.com/gordon-cs/gordon-360-ui/pull/1710)

PR #1704, opened 2023-02-26
[Rec-IM] Edit Series

silasw2:
This PR allows editing of a specific series from its additional options drop-down in the scheduling page. It also involved renaming the form as a whole since it is not longer just a create form.

### DIFF
--- a/src/services/recim/series.ts
+++ b/src/services/recim/series.ts
@@ -89,8 +89,9 @@ const getSeriesTypes = (): Promise<Lookup[]> => http.get(`recim/series/lookup?ty
 
 const getAllSeries = (): Promise<Series[]> => http.get(`recim/series`);
 
-const editSeries = (seriesID: number, updatedSeries: PatchSeries): Promise<CreatedSeries> =>
-  http.patch(`recim/series/${seriesID}`, updatedSeries);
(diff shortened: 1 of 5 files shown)


src/services/recim/series.ts
```diff
@@ -89,8 +89,9 @@ const getSeriesTypes = (): Promise<Lookup[]> => http.get(`recim/series/lookup?ty
 
 const getAllSeries = (): Promise<Series[]> => http.get(`recim/series`);
 
-const editSeries = (seriesID: number, updatedSeries: PatchSeries): Promise<CreatedSeries> =>
-  http.patch(`recim/series/${seriesID}`, updatedSeries);
+const editSeries = (seriesID: number, updatedSeries: PatchSeries): Promise<CreatedSeries> => {
+  return http.patch(`recim/series/${seriesID}`, updatedSeries);
+};
```
amos-cha, line 94:
```suggestion
const editSeries = (seriesID: number, updatedSeries: PatchSeries): Promise<CreatedSeries> =>
  http.patch(`recim/series/${seriesID}`, updatedSeries);
```

amos-cha (changes requested):
Small change but other than that, it looks amazing! Good work! Also cool check on the reload callback. I think that should be an implementation we include throughout so we prevent the whole list of dependencies.

amos-cha (approved):

PR #1705, opened 2023-02-27
[Rec-IM] Series Schedule Form

silasw2:
This PR adds a form to allow admins to set a series schedule for the auto-scheduler to use for scheduling matches. Review away.

### DIFF
--- a/src/views/RecIM/components/Forms/SeriesScheduleForm/index.jsx
+++ b/src/views/RecIM/components/Forms/SeriesScheduleForm/index.jsx
@@ -1,0 +1,331 @@
+import { Grid } from '@mui/material';
+import { useState, useEffect, useMemo } from 'react';
+import GordonLoader from 'components/Loader';
+import GordonDialogBox from 'components/GordonDialogBox';
+import { ConfirmationRow } from '../components/ConfirmationRow';
+import { ConfirmationWindowHeader } from '../components/ConfirmationHeader';
+import { ContentCard } from '../components/ContentCard';
+import { InformationField } from '../components/InformationField';
+import { putSeriesSchedule } from 'services/recim/series';
+import { getMatchSurfaces } from 'services/recim/match';
+
+const SeriesScheduleForm = ({
+  closeWithSnackbar,
+  openSeriesScheduleForm,
+  setOpenSeriesScheduleForm,
+  seriesID,
+}) => {
+  const [errorStatus, setErrorStatus] = useState({
+    AvailableDays: false,
+    AvailableSurfaceIDs: false,
+    DailyStartTime: false,
+    DailyEndTime: false,
+    EstMatchTime: false,
+  });
+
+  // Fetch data required for form creation
+  const [loading, setLoading] = useState(true);
+  const [surfaces, setSurfaces] = useState([]);
+
+  useEffect(() => {
+    const loadData = async () => {
+      setLoading(true);
+      setSurfaces(await getMatchSurfaces());
+      setLoading(false);
+    };
+    loadData();
+  }, []);
+
+  const seriesScheduleFields = [
+    {
+      label: 'Monday',
+      name: 'Mon',
+      type: 'checkbox',
+      error: errorStatus.AvailableDays,
+      helperText: '*Required',
+    },
+    {
+      label: 'Tuesday',
+      name: 'Tue',
+      type: 'checkbox',
+      error: errorStatus.AvailableDays,
+      helperText: '*Required',
+    },
+    {
+      label: 'Wednesday',
+      name: 'Wed',
+      type: 'checkbox',
+      error: errorStatus.AvailableDays,
+      helperText: '*Required',
+    },
+    {
+      label: 'Thursday',
+      name: 'Thu',
+      type: 'checkbox',
+      error: errorStatus.AvailableDays,
+      helperText: '*Required',
+    },
+    {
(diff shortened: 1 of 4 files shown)


bennettforkner:
Without doing a lot of investigation and evaluation of your code, I'm wondering if much of the logic for this form and forms like it can be refactored out to a shared component. It would be less difficult because the fields are already split out into their own object. Does this make sense to do in your context?

src/views/RecIM/components/Forms/SeriesScheduleForm/index.jsx
```diff
@@ -0,0 +1,331 @@
...
+          Label: getFieldLabel(key),
+        });
+    });
+    return updatedFields;
+  }
+
+  const handleConfirm = () => {
+    setSaving(true);
+
+    let dayOfTheWeek = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
```
bennettforkner, line 224: It would technically be more efficient to put this as a const outside of the component itself. Otherwise, it is getting re-created every time `handleConfirm` is called (which may only be once anyway, but it's good practice to do it consistently)

bennettforkner:
> this is actually something on our todo list. I believe @silasw2 actually has this planned next

Awesome

src/views/RecIM/views/Activity/components/ScheduleList/index.jsx
```diff
@@ -29,16 +31,26 @@ const ScheduleList = ({ isAdmin, series, activityID, reload, setReload }) => {
...
     handleClose();
   };
 
+  const handleSeriesSchedule = () => {
+    setOpenSeriesScheduleForm(true);
+    handleClose();
+  };
+
   const handleEditSeriesForm = (status) => {
     setReload((prev) => !prev);
```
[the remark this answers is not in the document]
silasw2: This is a really good point. However, since we fetch the entire pages information in one call, we cannot just load a single piece (eager loading). The only way we could accomplish this is to make a new get route, but that does not make sense to have the data in the eager load and the other get.

src/views/RecIM/components/Forms/SeriesScheduleForm/index.jsx
```diff
@@ -0,0 +1,331 @@
...
+      name: 'Sun',
+      type: 'checkbox',
+      error: errorStatus.AvailableDays,
+      helperText: '*Required',
+    },
+    {
+      label: 'Surfaces',
+      name: 'AvailableSurfaceIDs',
+      type: 'multiselect',
+      menuItems: surfaces.map((surface) => surface.Description),
```
amos-cha, line 93:
```suggestion
      menuItems: surfaces.map((surface) => surface.Name),
```

src/views/RecIM/components/Forms/SeriesScheduleForm/index.jsx
```diff
@@ -0,0 +1,331 @@
+import { Grid } from '@mui/material';
+import { useState, useEffect, useMemo } from 'react';
+import GordonLoader from 'components/Loader';
+import GordonDialogBox from 'components/GordonDialogBox';
+import { ConfirmationRow } from '../components/ConfirmationRow';
+import { ConfirmationWindowHeader } from '../components/ConfirmationHeader';
+import { ContentCard } from '../components/ContentCard';
+import { InformationField } from '../components/InformationField';
+import { putSeriesSchedule } from 'services/recim/series';
+import { getMatchSurfaces } from 'services/recim/match';
```
amos-cha, line 10:
```suggestion
import { getSurfaces } from 'services/recim/match';
```

src/views/RecIM/components/Forms/SeriesScheduleForm/index.jsx
```diff
@@ -0,0 +1,331 @@
...
+  });
+
+  // Fetch data required for form creation
+  const [loading, setLoading] = useState(true);
+  const [surfaces, setSurfaces] = useState([]);
+
+  useEffect(() => {
+    const loadData = async () => {
+      setLoading(true);
+      setSurfaces(await getMatchSurfaces());
```
amos-cha, line 33:
```suggestion
      setSurfaces(await getSurfaces());
```

amos-cha (changes requested):
Looks really good! 
There were a few updates since you've been gone and I suggested the locations. (Once you merge `recim` you should see them) I'm pretty sure this is all that has changed for series. 

After that, you're cleared for merge!

src/views/RecIM/components/Forms/SeriesScheduleForm/index.jsx
```diff
@@ -0,0 +1,331 @@
...
+
+    // To get days of the week into array form for pass to API
+    dayOfTheWeek.forEach((day) => {
+      seriesRequest.AvailableDays[day] = seriesRequest[day];
+      delete seriesRequest[day];
+    });
+
+    seriesRequest.AvailableSurfaceIDs = seriesRequest.AvailableSurfaceIDs.map(
+      (surfaceDescription) => {
+        return surfaces.find((surface) => surfaceDescription === surface.Description).ID;
```
amos-cha, line 237:
```suggestion
        return surfaces.find((surface) => surfaceDescription === surface.Name).ID;
```

amos-cha (approved):
Merged my own fix. Should be good to go!

PR #1706, opened 2023-02-28
change Gordon's favicon from black to blue

### LINKED ISSUE #1156: Favicon Background on Dark Mode
Not to become too picky with the favicon (but I will)
Consider if we can put a white background around the favicon only when the browser is in dark mode. This also affects the PWA icon.
Getting started note: there is a `prefers-color-scheme` media query that may work and browser support seems to be decent https://caniuse.com/?search=prefers-color-scheme

image

ArabellaJi:
This PR is to change Gordon's favicon from black to blue. This way, people can see the favicon more clearly when using dark mode.

image

closes #1156


ArabellaJi:
Oh I see. Also, it looks like they need different kinds of icons. Is there a good way for me to find different sized icons or resize them? Also, are there any differences between "android-icon-72x72.png" and "apple-icon-72x72.png"?

EjPlatzer:
> Also, it looks like they need different kinds of icons. Is there a good way for me to find different sized icons or resize them?

Good question! There are lots of tools that will generate the different variations (both sizes and platforms) for you from one base image. The most well-known is [favicon.io](https://favicon.io). However, I think the base image that we harvested from the www site last week is too low resolution for that to work well.

The College Creative team manages this kind of branding content for our websites, so they might be able to help you by either providing the base image or possibly the entire set of favicons. You could reach out to them at design@gordon.edu and explain your goal and ask if they have any resources to assist you.

ArabellaJi:
Sounds good! Will do now!

EjPlatzer (changes requested):
I think that we should remove all the old images (the ones that don't have replacements) so that it's consistent across platforms.

Also, please be sure to update the root `index.html` file so that it references only the updated images.

EjPlatzer (approved):
Looks great! Thanks for updating this!

EjPlatzer (approved):

Issue #1571, opened 2022-11-15
Allow FacStaff to add calendar links to their profile


bennettforkner:
Some professors have asked for the ability to link students to their scheduling calendar (Calendy or other) so that students can easily set up a time to connect.

After discussion with @chrisimagineer, we determined that the best place to have this would probably be in the social links at the top of a user's profile. Since there is already infrastructure built out for that, it will be trivial to add another field.

bennettforkner:
Working on this issue will likely involve working with all three main aspects of 360: database, API, and UI.

You could start exploring the files in the following areas:
UI:
**profile component & subcomponents:** src/components/Profile/components/Identification/components/SocialMediaLinks/
**user service:** src/services/user.ts

API:
**profile controller:** Gordon360/Controllers/ProfilesController.cs (UpdateLinkAsync() method)
**profile service:** Gordon360/Services/ProfileService.cs (UpdateProfileLinkAsync() method)
**CUSTOM_PROFILE database model:** Gordon360/Models/CCT/CUSTOM_PROFILE.cs -> would need to add a field to the database, then have Entity Framework automatically update this and other database object model files.

PR #1707, opened 2023-02-28
Faculty Calendar LInk

jakedcolbert:
Added calendar link. Requires validation #1571

### DIFF
--- a/src/services/socialMedia.tsx
+++ b/src/services/socialMedia.tsx
@@ -1,4 +1,11 @@
-import { FaFacebookF, FaHandshake, FaInstagram, FaLinkedin, FaTwitter, FaCalendar } from 'react-icons/fa';
+import {
+  FaFacebookF,
+  FaHandshake,
+  FaInstagram,
+  FaLinkedin,
+  FaTwitter,
+  FaCalendar,
+} from 'react-icons/fa';
 
 export enum Platform {
   Facebook = 'Facebook',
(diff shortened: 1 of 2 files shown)


EjPlatzer:
> Hey @jakedcolbert , looks like the `Lint and Build` GitHub action failed here. Seems like there might be a typescript object that does not have the calendar link property that you added.

More specifically, it seems that the `userService` in `src/services/user.ts` encodes that profiles retrieved from the API will have each member of the `Platform` enum present as a property, but you haven't updated the `BaseProfileInfo` type to reflect the new Platform type. Therefore, when we try to `formatSocialMediaLinks` on a profile (user.ts line 189), typescript notices an error because we would try to access `profile.Calendar` which doesn't exist.

The simple way to fix that is to just add `Calender` as a property of `BaseProfileInfo` underneath `LinkedIn`.

There is another way to fix it such that the `Platform` properties on the profile are not hard-coded but are actually computed from the `Platform` enum. To do so, you would add a new type in `socialMedia.tsx` like the following:
```typescript
type PlatformLinks = {
    [p in Platform]: string
}
```
and then you can change `BaseProfileInfo` in `user.ts` to be the intersection of the other profile items and the properties of `PlatformLinks`, i.e.:
```typescript
type BaseProfileInfo = {
  ID: string,
  ...
} & PlatformLinks
```

Then, you can remove the `LinkedIn: string` members from `BaseProfileInfo` because they're already included in the `PlatformLinks` type.

EjPlatzer (approved):
LGTM!

russtuck:
I really like this feature, but it doesn't look like it's working...  I added my link in my profile (https://russtuck.youcanbook.me), and can see it there.  But when I go to my public profile, it's not visible.  (The only link is linkedin.)

Issue #1653, opened 2023-01-30
Personal Information Privacy Toggle Broken


bennettforkner:
The Public/Private toggle on the Personal Information card on `MyProfile` is not displaying correctly. When I set my PI to private and then reload the page, the toggle is still set to Public even though the change worked.

![Image](https://user-images.githubusercontent.com/43764277/215543326-5d516fbf-6f02-44cf-b9d8-5425c0eb97e2.png)

bennettforkner:
The issue is likely with something here:

https://github.com/gordon-cs/gordon-360-ui/blob/a2d7c3bf58d3f13d960c3365dc7fc86f89302154/src/components/Profile/components/PersonalInfoList/index.jsx#L486-L498

jalajk24:
@bennettforkner Can I works on this issue ? I am new to open source contribution

bennettforkner:
@jalajk24 we welcome you to open a pull request and propose code changes if you so desire. Unfortunately this application can only be run with our local database, so you will have a hard time testing.

PR #1709, opened 2023-03-03
Fac/Staff Personal Info Toggle Visual Fix

charlielee144r:
This PR fixes state logic for showing which selection (from private or public) personal information currently is. Previously, the logic for setting the state of the toggle always returned false when there was no Home phone no. added (it seems), which automatically set the toggle to 'true' (which is shown as public by the toggler). So whether or not you set your profile as private or public with the toggle, it would show that it was set to public (see #1653). 

More testing and logic handling is required before this is ready. It seems to work when there is no home phone no. associated with a profile, but have not been able to test it with a profile with a home phone no.

This might be good to discuss in a meeting while looking through the code together.

### DIFF
--- a/src/components/Profile/components/PersonalInfoList/index.jsx
+++ b/src/components/Profile/components/PersonalInfoList/index.jsx
@@ -79,8 +79,6 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }) => {
   const isCampusLocationPrivate = isStudent && keepPrivate && profile.OnOffCampus !== PRIVATE_INFO;
 
   // Students' home phone is always private. FacStaffs' home phone is private for private users
-  // Why isHomePhonePrivate always no/false?
-  // For some reason, showing the disclaimer text...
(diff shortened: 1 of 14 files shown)


src/components/Profile/components/PersonalInfoList/index.jsx
```diff
@@ -507,6 +503,7 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }) => {
                   <Switch
                     onChange={handleChangeHomePhonePrivacy}
                     color="secondary"
+                    // saying isHomePhonePrivate is always returning 'no' - which means it is always public
                     checked={!isHomePhonePrivate}
```
bennettforkner, line 507: I'm not sure I understand why this switch is based off of the `isHomePhonePrivate` variable. It pertains to more fields than just that, so I would think we would want to base it off of the whole. Would it make more sense to base this on the value of `(isStudent || keepPrivate)`?

bennettforkner: Of course, I know you didn't write this part!

charlielee144r:
Weird merge issues - continued on #1734

PR #1711, opened 2023-03-05
[Rec-IM] Match Listing Update

cpabbot:
## Rec-IM Match Listing Update

Note: 'vs.' to be replaced with the score upon completion, and TBD if the score does not exist at that point.

image

### DIFF
--- a/src/views/RecIM/components/List/Listing/index.jsx
+++ b/src/views/RecIM/components/List/Listing/index.jsx
@@ -396,6 +396,7 @@ const MatchListing = ({ match, activityID }) => {
             </Grid>
           </Grid>
           <Grid item container direction="column" xs={6}>
+            {!match.Team?.length && <Typography>No teams yet placed in this match.</Typography>}
             {match.Team.map((team) => (
               <Grid item container>
                 <Grid item xs={10}>
--- a/src/views/RecIM/components/List/Listing/index.jsx
+++ b/src/views/RecIM/components/List/Listing/index.jsx
@@ -345,9 +345,29 @@ const MatchListing = ({ match, activityID }) => {
         to={`/recim/activity/${activityID}/match/${match.ID}`}
         className={styles.listing}
       >
-        <Grid container>
+        <Grid container direction="column" alignItems="center">
+          <Grid item container>
+            <Grid item xs={5}>
+              <Typography className={styles.listingTitle}>
+                {match.Team[0]?.Name ?? <i>TBD</i>}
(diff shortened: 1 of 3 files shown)


amos-cha:
@cpabbot Personal suggestion, I think it would look better/make more sense if the listing ordering went like this:

Screen Shot 2023-03-07 at 5 28 01 PM

It makes more sense reading:
March 2nd, 7:00PM. Josh and 2 will be playing at ________

Also it's more balanced

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -338,6 +340,56 @@ const ParticipantListing = ({ participant, minimal, callbackFunction, showPartic
...
+              </Grid>
+              <Grid item xs={2} textAlign="center">
+                {/* show scores only if match is in the present/past */}
+                {DateTime.now() > DateTime.fromISO(match.Time) ? (
+                  <Typography>
+                    {match.Scores?.find((matchTeam) => matchTeam.TeamID === match.Team[0]?.ID)
+                      ?.TeamScore ?? 'TBD'}{' '}
+                    :{' '}
+                    {match.Scores?.find((matchTeam) => matchTeam.TeamID === match.Team[1]?.ID)
+                      ?.TeamScore ?? 'TBD'}
```
amos-cha, line 367:
```suggestion
                    {match.Scores?.find((matchTeam) => matchTeam.TeamID === match.Team[0]?.ID)
                      ?.TeamScore ?? 0}{' '}
                    :{' '}
                    {match.Scores?.find((matchTeam) => matchTeam.TeamID === match.Team[1]?.ID)
                      ?.TeamScore ?? 0}
```

Scores should be 0 not tbd?

amos-cha (changes requested):
Looks good otherwise!

src/views/RecIM/components/List/Listing/index.jsx
cpabbot: I actually did have 0 at first, but then thought that that would technically be incorrect right? Because the score could have been anything and just not entered yet. It could be 5:0 but not updated, so a 0:0 showing would be inaccurate. And then how do you tell the difference between a TBD score and a final score of 0:0 if the score did indeed end up being 0:0 but also it could just be that it wasn't entered yet?

amos-cha (approved):
Looks good, but we'll have to think about what we want to show for a game that hasn't started yet.

PR #1720, opened 2023-03-09
[Rec-IM] API/Develop Update

amos-cha:
Handles name changes + Activity Delete route

### DIFF
--- a/src/views/RecIM/components/Forms/ActivityForm/index.jsx
+++ b/src/views/RecIM/components/Forms/ActivityForm/index.jsx
@@ -274,9 +274,6 @@ const ActivityForm = ({
   const handleConfirm = () => {
     setSaving(true);
     let activityRequest = { ...currentInfo, ...newInfo };
-    // solution while not handling activity max capacity
-    activityRequest.maxCapacity =
-      activityRequest.maxCapacity === '' ? 0 : activityRequest.maxCapacity;
 
     activityRequest.sportID = sports.find((sport) => sport.Name === activityRequest.sportID).ID;
     activityRequest.typeID = activityTypes.find(
--- a/src/views/RecIM/components/Forms/ActivityForm/index.jsx
+++ b/src/views/RecIM/components/Forms/ActivityForm/index.jsx
@@ -274,6 +274,10 @@ const ActivityForm = ({
   const handleConfirm = () => {
     setSaving(true);
     let activityRequest = { ...currentInfo, ...newInfo };
+    // solution while not handling activity max capacity
+    activityRequest.maxCapacity =
+      activityRequest.maxCapacity === '' ? 0 : activityRequest.maxCapacity;
+
     activityRequest.sportID = sports.find((sport) => sport.Name === activityRequest.sportID).ID;
     activityRequest.typeID = activityTypes.find(
       (type) => type.Description === activityRequest.typeID,
--- a/src/views/RecIM/components/Forms/ActivityForm/index.jsx
+++ b/src/views/RecIM/components/Forms/ActivityForm/index.jsx
@@ -66,7 +66,7 @@ const ActivityForm = ({
       name: 'startDate',
       type: 'datetime',
       error: errorStatus.startDate,
-      required: true,
+      required: false,
       helperText: '*Required',
     },
     {
@@ -168,9 +168,9 @@ const ActivityForm = ({
         registrationStart: activity.RegistrationStart,
         registrationEnd: activity.RegistrationEnd,
         typeID:
-          activityTypes.find((type) => type.ID === activity.TypeID) == null
+          activityTypes.find((type) => type.Description === activity.Type) == null
             ? ''
-            : activityTypes.find((type) => type.ID === activity.TypeID).Description,
+            : activityTypes.find((type) => type.Description === activity.Type).Description,
         sportID:
           sports.find((type) => type.ID === activity.Sport.ID) == null
             ? ''
--- a/src/views/RecIM/components/Forms/ActivityForm/index.jsx
+++ b/src/views/RecIM/components/Forms/ActivityForm/index.jsx
@@ -112,7 +112,7 @@ const ActivityForm = ({
         return sport.Name;
       }),
       error: errorStatus.sportID,
-      required: false,
+      required: true,
       helperText: '*Required',
     },
     {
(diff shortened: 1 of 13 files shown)


src/views/RecIM/components/Forms/ActivityForm/index.jsx
```diff
@@ -168,9 +168,9 @@ const ActivityForm = ({
         registrationStart: activity.RegistrationStart,
         registrationEnd: activity.RegistrationEnd,
         typeID:
-          activityTypes.find((type) => type.ID === activity.TypeID) == null
+          activityTypes.find((type) => type.Description === activity.Type) == null
             ? ''
-            : activityTypes.find((type) => type.ID === activity.TypeID).Description,
+            : activityTypes.find((type) => type.Description === activity.Type).Description,
```
cpabbot, line 173: I'm confused by this, we're setting typeID to activityTypeDescription?

src/views/RecIM/components/Forms/ActivityForm/index.jsx
```diff
@@ -274,6 +274,10 @@ const ActivityForm = ({
   const handleConfirm = () => {
     setSaving(true);
     let activityRequest = { ...currentInfo, ...newInfo };
+    // solution while not handling activity max capacity
+    activityRequest.maxCapacity =
+      activityRequest.maxCapacity === '' ? 0 : activityRequest.maxCapacity;
```
cpabbot, line 279: maxCapacity should be nullable, is this the temp solution due to it not being treated as nullable by UI? Could probably easily fix

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -136,6 +133,125 @@ const ActivityListing = ({ activity }) => {
...
+            </MenuItem>
+            <MenuItem dense onClick={() => console.log('delete')} className={styles.rejectButton}>
+              Delete
+            </MenuItem>
+          </Menu>
+        )}
+      </ListItem>
+    </ListItem>
+  );
+};
```
cpabbot, line 254: I still don't know about this new listing, but I guess as long as the old one is there commented out I don't mind including this

src/views/RecIM/components/List/index.jsx
```diff
@@ -38,9 +42,9 @@ const ParticipantList = ({
 };
 
 const MatchList = ({ matches, activityID }) => {
-  if (!matches?.length) return <Typography>No matches to show.</Typography>;
+  if (!matches?.length || !matches[0]) return <Typography>No matches to show.</Typography>;
```
cpabbot, line 45: This shouldn't be necessary. If there is no 0 match then matches will have a length of zero.

src/views/RecIM/components/List/index.jsx
```diff
@@ -38,9 +42,9 @@ const ParticipantList = ({
...
   let content = matches.map((match) => (
-    <MatchListing key={match.ID} match={match} activityID={activityID} />
+    <MatchListing key={match?.ID} match={match} activityID={activityID} />
```
cpabbot, line 47: Unnecessary check because we are mapping from an existing (match) so this will never be null.

cpabbot (changes requested):
Good catches on several bugs and thanks for updating to new API updates!

src/views/RecIM/components/Forms/ActivityForm/index.jsx
amos-cha: This was a mistake when we first added type to Activity. I returned a type id which is contrary to the rest of our API where we return the description/string. This was an amendment.

src/views/RecIM/components/Forms/ActivityForm/index.jsx
amos-cha: I think I fixed this already but forgot to push the change.

src/views/RecIM/components/List/Listing/index.jsx
amos-cha: The more I look at it, the less I really like it. I am not a huge fan of putting the delete ONLY in the activity page though. Perhaps just have it on the admin page? Then we could use something similar to the minimal tag on the home page and show options for the admin page.

src/views/RecIM/components/List/index.jsx
amos-cha: Unfortunately this is not the case. I quite literally added that because it returned me an array with a single null value inside
```
[ null ]
```
hilarious right?

src/views/RecIM/components/List/index.jsx
amos-cha: a remnant of me slowly tracing bugs.

amos-cha, line 47:
```suggestion
    <MatchListing key={match.ID} match={match} activityID={activityID} />
```

src/views/RecIM/components/List/Listing/index.jsx
cpabbot: Oh ya, big fan of this! Maybe just comment out the new one and we can implement that later on for admin page?

src/views/RecIM/components/List/index.jsx
cpabbot: How is the array length determined? That smells like an API bug... We could do this though if that's how it is rn

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -16,35 +16,29 @@ import styles from './Listing.module.css';
...
       setActivityType(
         activityTypes.find((activityType) => activityType.ID === activity.TypeID).Description,
       );
     };
-    const calculateCurrentCapacity = async () => {
-      let fullActivity = await getActivityByID(activity.ID);
-      setCurrentCapacity(fullActivity.Team?.length);
-    };
     loadActivityType();
-    calculateCurrentCapacity();
```
cpabbot, line 47: The other capacity thing is commented out but this one is deleted. At least maybe have it commented out? Either way needs to be consistent though since the other part won't work without this function.

cpabbot (changes requested):
Also left one comment about the array of nulls which seems fishy.

cpabbot (commented):
LGTM, will wait to approve until API deployment and testing

cpabbot (approved):
Great work. We have some bugs to fix but this is excellent work getting us up to date!

PR #1721, opened 2023-03-09
[Rec-IM] Removed luxon

cpabbot:
Replaced all luxon uses in recim with date-fns

### DIFF
--- a/src/views/RecIM/components/Helpers/index.js
+++ b/src/views/RecIM/components/Helpers/index.js
@@ -1,14 +1,14 @@
-import { isValid, parseISO, format } from 'date-fns';
+import { isValid, format } from 'date-fns';
 
 const standardDate = (date, includeTime) => {
-  if (!isValid(date)) date = parseISO(date);
+  if (!isValid(date)) date = Date.parse(date);
   let formattedDate = includeTime ? format(date, 'MMM d h:mmaaa') : format(date, 'MMM d');
   return formattedDate;
 };
 
 const formatDateTimeRange = (startDateTime, endDateTime) => {
-  if (!isValid(startDateTime)) startDateTime = parseISO(startDateTime);
-  if (!isValid(endDateTime)) endDateTime = parseISO(endDateTime);
+  if (!isValid(startDateTime)) startDateTime = new Date(Date.parse(startDateTime));
+  if (!isValid(endDateTime)) endDateTime = new Date(Date.parse(endDateTime));
 
   // Jan 1 3:00pm - 6:00pm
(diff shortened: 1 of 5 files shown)


src/views/RecIM/components/Helpers/index.js
```diff
@@ -1,35 +1,28 @@
-import { DateTime } from 'luxon';
+import { isValid, parseISO, format } from 'date-fns';
 
 const standardDate = (date, includeTime) => {
-  if (!DateTime.isDateTime(date)) date = DateTime.fromISO(date);
-  let formattedDate = date.monthShort + ' ' + date.day;
-  if (includeTime) {
-    formattedDate += ' ' + date.toLocaleString(DateTime.TIME_SIMPLE);
-  }
+  if (!isValid(date)) date = parseISO(date);
```
EjPlatzer, line 4: Native JS Date already supports parsing ISO (in fact it's the only officially supported format for parsing dates from strings)
```suggestion
  if (!isValid(date)) date = Date.parse(date);
```

src/views/RecIM/components/Helpers/index.js
```diff
@@ -1,35 +1,28 @@
...
-    endDateTime.toLocaleString(DateTime.DATE_FULL);
+  let isOneDay = format(startDateTime, 'MM dd yyyy') === format(endDateTime, 'MM dd yyyy');
   if (isOneDay) {
-    let endTime = startDateTime.toLocaleString(DateTime.TIME_SIMPLE).replace(':00', '');
+    let endTime = format(endDateTime, 'h:mmaaa').replace(':00', '');
     return `${standardDate(startDateTime, true).replace(':00', '')} - ${endTime}`;
   }
   // Jan 1 3:00pm - Jan 31 3:00pm
-  let isOneYear = startDateTime.year === endDateTime.year;
+  let isOneYear = format(startDateTime, 'yyyy') === format(endDateTime, 'yyyy');
```
EjPlatzer, line 20: Might be clearer and easier to compare the years as numbers instead of first converting them to strings:
```suggestion
  let isOneYear = startDateTime.getFullYear() === endDateTime.getFullYear();
```

src/views/RecIM/components/Helpers/index.js
EjPlatzer: You probably don't need to use `parseISO` anywhere, so it would be great if you could update the other uses of it as well.

EjPlatzer (changes requested):
Thanks for making this change! It mostly looks good, I just had two small thoughts on how you could use native JS instead of the date-fns library in a couple places.

EjPlatzer (approved):
LGTM!

PR #1723, opened 2023-03-10
Drop Mail Location from Involvement Profile page

Kevin-Chu58:
This pr removes Mail Location from the Involvement Profile page. If anyone wants to know the mail location of another member, they can simply click the link to other's profile and view it there.

PS: this pr was originally planned to solve the duplicate mail location bug, but turned out we decided to move forward with another approach without Mail Location.

Here is the new UI for Involvement Profile page.

Full screen
image

Mobile view
image

< Admin View >
Full screen
image

Mobile view
image

### DIFF
--- a/src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.jsx
+++ b/src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.jsx
@@ -154,10 +154,6 @@ const MemberListItem = ({
   let content;
   let options;
 
-  let mailLoc = Number(member.Mail_Location)
-    ? `Box #${member.Mail_Location}`
-    : member.Mail_Location || null;
-
   if (isAdmin || isSiteAdmin) {
     const disabled = participationDescription === 'Guest' || participationDescription === 'Member';
     // Can't make guests or members a group admin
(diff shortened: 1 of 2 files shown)


Kevin-Chu58:
Not sure why the avatars in admin view look so cluttered, might because of the option component that messed up margin and padding.

bennettforkner:
@ChrisImagineer Confirming that you are alright with us removing mailboxes from the involvement membership list? Evan and I both think it is almost unused and our student developers say they can't really think of a use case for it where there isn't another simple solution.
Mailboxes will stay on people search and the individual profile page, of course.

ChrisImagineer:
That’s fine.
It was originally there to provide a GroupAdvisor (faculty/staff) or GroupLeader a quick way to send campus mail to students individually for their roster group. Personal notes or group, personalized invitations or other hard-copy information/books, etc.

I think it does create a loss for faculty/staff, many who don’t have access to reports.gordon, and for faculty who only have course rosters. I’m not sure how to solve that, or which may complain of the loss.  Its okay to remove that mailstop group feature, and we’ll see what happens.

On 03/10/23, 09:25, "Bennett F Forkner" ***@***.******@***.***>> wrote:


@ChrisImagineer<https://github.com/ChrisImagineer> Confirming that you are alright with us removing mailboxes from the involvement membership list? Evan and I both think it is almost unused and our student developers say they can't really think of a use case for it where there isn't another simple solution.
Mailboxes will stay on people search and the individual profile page, of course.

EjPlatzer:
Thanks for weighing in @ChrisImagineer!

> I think it does create a loss for faculty/staff, many who don’t have access to reports.gordon, and for faculty who only have course rosters.

To be clear, all faculty and staff will still be able to view any user's Mailstop on that user's profile. For example, if I look at the roster for an involvement I'm in, I can click on a member to be taken to their profile, where I can view their mailstop. So it seems the only slight regression here is that, if you wanted to find the mailstop of every member of an involvement, you'd have to look at each user's profile one-by-one, instead of seeing them all listed on directly on the involvement page.


> Its okay to remove that mailstop group feature, and we’ll see what happens.

We'll plan to move forward with this for now.

EjPlatzer (approved):
LGTM!

We might consider reformatting this, not that the info in each row is so sparse. But for now, this addresses the issue well

PR #1724, opened 2023-03-10
[Rec-IM] Fixed standardDate bug

cpabbot:


src/views/RecIM/components/Helpers/index.js
```diff
@@ -1,7 +1,12 @@
 import { isValid, format } from 'date-fns';
 
 const standardDate = (date, includeTime) => {
-  if (!isValid(date)) date = Date.parse(date);
+  if (!isValid(date)) date = new Date(Date.parse(date)); // try parsing if invalid
+  if (!isValid(date)) {
+    // if still invalid (perhaps null)
+    console.log('standardDate called with an invalid or null date');
```
amos-cha, line 7:
```suggestion

```

src/views/RecIM/components/Helpers/index.js
```diff
@@ -1,7 +1,12 @@
...
+    return;
+  }
```
amos-cha, line 9:
```suggestion
  if(!date) return;
  if (!isValid(date)) date = Date.parse(date);
```

amos-cha: I don't know what's the point of double parsing if the issue is that the date passed is null.

src/views/RecIM/views/Match/index.jsx
```diff
@@ -81,7 +81,9 @@ const Match = () => {
       <>
         <Grid container spacing={4}>
           <Grid item xs={6} textAlign="right">
-            <Typography className={styles.subtitle}>{standardDate(match?.Time, true)}</Typography>
+            <Typography className={styles.subtitle}>
+              {match && standardDate(match.Time, true)}
+            </Typography>
```
amos-cha, line 86:
```suggestion
            <Typography className={styles.subtitle}>{standardDate(match?.Time, true)}</Typography>
```

amos-cha: issue is solved if the helper.jsx returns null if match.time is null

amos-cha (changes requested):

src/views/RecIM/components/Helpers/index.js
cpabbot: Potentially it could be an invalid date but not be null. This handles any invalid input. The initial parse is in case it was in some kind of valid format.

src/views/RecIM/views/Match/index.jsx
cpabbot: Helper.jsx provides error checking, but this will also prevent our calling it with null in the first place. Kinda like API handling redundant with UI is what I was thinking

src/views/RecIM/components/Helpers/index.js
amos-cha: ok, still looks incredibly weird. most likely we'll have to take the console.log out. but good find.

amos-cha (approved):
still looks weird to me but ok

src/views/RecIM/components/Helpers/index.js
cpabbot: Yeah we can probably remove that idk, just thought it could be helpful for debugging but probably not necessary.

PR #1725, opened 2023-03-10
[Rec-IM] Match Attendance

cpabbot:
## Match Attendance

I'm not ecstatic about the styling but I think this is pretty solid especially for now; feel free to give feedback if there's anything you think would make design more intuitive, but I think this works pretty well. I considered a few different approaches but went with this for simplicity and clarity.

I also considered combining the parameters to pass down into one `attendanceDetails` object to reduce the number of parameters, but I thought those parameters might end up getting used for other contexts in the future so I left them as is. It is a large number of parameters, but I don't think it hurts; I could even leave a comment to document what the different parameters are used for if desired.

Admin view
image

User view
image

### DIFF
--- a/src/services/http.ts
+++ b/src/services/http.ts
@@ -53,7 +53,12 @@ const patch = <TResponse>(
 ): Promise<TResponse> =>
   makeRequest(endpoint, 'PATCH', JSON.stringify(body), setContentTypeJSON(headers));
 
-const del = <TResponse>(endpoint: string): Promise<TResponse> => makeRequest(endpoint, 'delete');
+const del = <TResponse>(
+  endpoint: string,
+  body: Object = '',
+  headers = new Headers(),
+): Promise<TResponse> =>
+  makeRequest(endpoint, 'delete', JSON.stringify(body), setContentTypeJSON(headers));
 
 const apiBaseURL = import.meta.env.DEV ? '/' : (import.meta.env.VITE_API_URL as string);
 
--- a/src/services/recim/match.ts
+++ b/src/services/recim/match.ts
@@ -90,10 +90,16 @@ type Attendance = {
 const createMatch = (newMatch: UploadMatch): Promise<CreatedMatch> =>
   http.post('recim/matches', newMatch);
 
-const createMatchAttendance = (
+const updateMatchAttendance = (
   attendance: UploadMatchAttendance,
   matchID: number,
 ): Promise<Attendance[]> => http.put(`recim/matches/${matchID}/attendance`, attendance);
+
+const updateAttendance = (matchID: number, attendance: Attendance): Promise<Attendance> =>
+  http.post(`recim/matches/${matchID}/attendance`, attendance);
+
+const removeAttendance = (matchID: number, attendance: Attendance): Promise<Attendance> =>
+  http.del(`recim/matches/${matchID}/attendance`, attendance);
 
 const getMatchAttendance = (matchID: number): Promise<MatchAttendance> =>
   http.get(`recim/matches/${matchID}/attendance`);
@@ -133,7 +139,9 @@ export {
   getMatchTeamStatusTypes,
   updateMatchStats,
   updateMatch,
-  createMatchAttendance,
+  updateMatchAttendance,
+  updateAttendance,
+  removeAttendance,
   getMatchAttendance,
   deleteMatchCascade,
   getSurfaces,
--- a/src/views/RecIM/components/List/Listing/index.jsx
+++ b/src/views/RecIM/components/List/Listing/index.jsx
@@ -11,6 +11,7 @@ import {
   Menu,
   MenuItem,
   Button,
+  Switch,
 } from '@mui/material';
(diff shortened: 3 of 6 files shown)


src/services/http.ts
```diff
@@ -53,7 +53,12 @@ const patch = <TResponse>(
 ): Promise<TResponse> =>
   makeRequest(endpoint, 'PATCH', JSON.stringify(body), setContentTypeJSON(headers));
 
-const del = <TResponse>(endpoint: string): Promise<TResponse> => makeRequest(endpoint, 'delete');
+const del = <TResponse>(
+  endpoint: string,
+  body: Object = '',
+  headers = new Headers(),
+): Promise<TResponse> =>
+  makeRequest(endpoint, 'delete', JSON.stringify(body), setContentTypeJSON(headers));
```
amos-cha, line 61: @EjPlatzer Is this good enough for `nullable` as you mentioned? Or do you want it to be `?` nullable?

src/services/recim/match.ts
```diff
@@ -90,11 +90,17 @@ type Attendance = {
 const createMatch = (newMatch: UploadMatch): Promise<CreatedMatch> =>
   http.post('recim/matches', newMatch);
 
-const createMatchAttendance = (
+const updateMatchAttendance = (
```
amos-cha, line 93: Good change! I was considering the renaming but was going to change to `put` but `update` is definitely better!

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -22,7 +24,8 @@ import MoreHorizIcon from '@mui/icons-material/MoreHoriz';
 import ClearIcon from '@mui/icons-material/Clear';
 import CheckIcon from '@mui/icons-material/Check';
 import { editTeamParticipant, respondToTeamInvite } from 'services/recim/team';
-import { getActivityTypes } from 'services/recim/activity';
+// import { getActivityTypes } from 'services/recim/activity';
```
amos-cha, line 27:
```suggestion
```

Remove unused type?

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -351,13 +354,23 @@ const TeamListing = ({ team, invite, match, setTargetTeamID, callbackFunction })
...
+  initialAttendance,
+  teamID,
+  matchID,
+}) => {
+  const { teamIDParam } = useParams(); // @TODO will fail if not on team page
   const [avatar, setAvatar] = useState();
   const [name, setName] = useState();
   const [anchorEl, setAnchorEl] = useState();
   const moreOptionsOpen = Boolean(anchorEl);
+  const [didAttend, setDidAttend] = useState(initialAttendance != null);
```
amos-cha, line 373:
```suggestion
  const [didAttend, setDidAttend] = useState(initialAttendance !== null);
```
?

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -351,13 +354,23 @@ const TeamListing = ({ team, invite, match, setTargetTeamID, callbackFunction })
...
+  minimal,
+  callbackFunction,
+  showParticipantOptions,
+  withAttendance,
+  isAdmin,
+  initialAttendance,
+  teamID,
+  matchID,
+}) => {
+  const { teamIDParam } = useParams(); // @TODO will fail if not on team page
```
amos-cha, line 368: If you're passing teamID in, why not just have a conditional like
`_teamID = teamID ?? teamIDParam `
(not sure which way the logic goes)

cpabbot, line 368: Note that this was preexisting; I simply renamed it to teamIDParam and left a comment to not trust that in the wrong context.

amos-cha (changes requested):
Got a few questions, willing to approve if they're all answered! Looks great otherwise !

src/views/RecIM/components/List/Listing/index.jsx
cpabbot: != allows for undefined, so shouldn't matter, but generally != > !== when null checking.

src/views/RecIM/components/List/Listing/index.jsx
cpabbot: Used in likewise commented out code

src/views/RecIM/components/List/Listing/index.jsx
cpabbot: Considered this, and I fixed it a bit but kept the two variables separate rather than creating a third and causing unneeded chaos.

cpabbot: Also good catch on the name breaking the thing, fixed it!

amos-cha (approved):
Looks good! 
I have an idea locking the attendance so that admins don't accidentally click on someone on mobile. This would also take advantage of our put route. But these are ideas to be fleshed out later and definitely nonblocking :)

PR #1727, opened 2023-03-11
[Rec-IM] Team participant upgrades and fixes

amos-cha:
Tackles the following:

Team Page
--
- Participants are **deleted** if activity is still in registration period 
- Participants are made **inactive** if activity is out of registration period
^ this is done so that participants cannot jump teams as well as not locking themselves into a team during the registration period.

- Clicking off menu will not refresh the page
- Clicking any of the updates will refresh via callback

image
Member reinstation after made inactive

image
Old methodology now ACTUALLY works

image
Invited participants cant be removed or promoted

Bug Fixes
--
- Home page no longer pulls inactive participation as a part of home teams

### DIFF
--- a/src/services/recim/series.ts
+++ b/src/services/recim/series.ts
@@ -99,7 +99,7 @@ const putSeriesSchedule = async (schedule: UploadSeriesSchedule): Promise<Series
   http.put(`recim/series/schedule`, schedule);
 
 const scheduleSeriesMatches = async (seriesID: number): Promise<Match[]> =>
-  http.post(`recim/series/${seriesID}/autoschedule`);
+  http.post(`recim/series/${seriesID}/autoschedule`, {});
 
 const deleteSeriesCascade = async (seriesID: number): Promise<CreatedSeries> =>
   http.del(`recim/series/${seriesID}`);
--- a/src/services/recim/series.ts
+++ b/src/services/recim/series.ts
@@ -92,15 +92,17 @@ const getAllSeries = (): Promise<Series[]> => http.get(`recim/series`);
 const editSeries = (seriesID: number, updatedSeries: PatchSeries): Promise<CreatedSeries> =>
   http.patch(`recim/series/${seriesID}`, updatedSeries);
 
+const getSeriesSchedule = (seriesID: number): Promise<SeriesSchedule> =>
+  http.get(`recim/series/${seriesID}/schedule`);
+
 const putSeriesSchedule = async (schedule: UploadSeriesSchedule): Promise<SeriesSchedule> =>
   http.put(`recim/series/schedule`, schedule);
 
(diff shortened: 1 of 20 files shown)


src/services/recim/series.ts
```diff
@@ -99,7 +99,7 @@ const putSeriesSchedule = async (schedule: UploadSeriesSchedule): Promise<Series
   http.put(`recim/series/schedule`, schedule);
 
 const scheduleSeriesMatches = async (seriesID: number): Promise<Match[]> =>
-  http.post(`recim/series/${seriesID}/autoschedule`);
+  http.post(`recim/series/${seriesID}/autoschedule`, {});
```
silasw2, line 102: Do we need a body tag for this?

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -365,15 +369,20 @@ const ParticipantListing = ({
...
   const moreOptionsOpen = Boolean(anchorEl);
   const [didAttend, setDidAttend] = useState(initialAttendance != null);
 
+  const handleClickOff = () => {
+    setAnchorEl(null);
+  };
+
   const handleClose = () => {
     setAnchorEl(null);
+    callbackFunction((val) => !val);
```
silasw2, line 385: nice implementation of this!

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -409,12 +424,29 @@ const ParticipantListing = ({
...
+    await editTeamParticipant(parseInt(teamIDParam), editedParticipant);
+    handleClose();
+  };
+
+  const handleMakeInactive = async () => {
     let editedParticipant = {
       Username: participant.Username,
       RoleTypeID: 6,
     }; // Role 6 is inactive
+    await editTeamParticipant(parseInt(teamIDParam), editedParticipant);
```
silasw2, line 441: was the parse int necessary here? would it work without?

silasw2 (approved):
LGTM! Ask me to re-review when you have the other PR ready!

src/services/recim/series.ts
amos-cha: yep, the other PR adds a comment in there stating `temporary solution`. There is a new parameter to handle Round Robin Match Cap & Number of Ladder Matches

src/views/RecIM/components/List/Listing/index.jsx
amos-cha: I'm not sure if it's necessary as I haven't tried without it. It's based on the boiler plate set up with the old promote logic from when the horizon menu was first made.

amos-cha:
Add minor new implementation:
image
Co-captains now can be demoted to member!

silasw2 (approved):
LGTM!

PR #1730, opened 2023-03-14
[Rec-IM] Add Deletes

cpabbot:
## Add Deletes
- Activity delete
- Team delete
- Match delete

image
image
image

### DIFF
--- a/src/components/GordonDialogBox/index.jsx
+++ b/src/components/GordonDialogBox/index.jsx
@@ -41,6 +41,7 @@ const GordonDialogBox = ({
       <DialogTitle
         className={titleClass ? titleClass : styles.gc360_gordondialogbox_title}
         id="alert-dialog-title"
+        sx={severity ? { bgcolor: `${severity}.main`, pb: '0' } : { bgcolor: 'primary.main' }}
       >
         {severity ? (
           <Alert variant="filled" severity={severity}>
@@ -65,7 +66,7 @@ const GordonDialogBox = ({
           <Button
             variant="contained"
             onClick={buttonClicked}
-            color="primary"
+            color={severity ?? 'primary'}
             disabled={isButtonDisabled}
           >
             {buttonName ?? 'Okay'}
--- a/src/views/RecIM/views/Activity/index.jsx
+++ b/src/views/RecIM/views/Activity/index.jsx
@@ -307,14 +307,6 @@ const Activity = () => {
               setOpenTeamForm={(bool) => setOpenTeamForm(bool)}
               activityID={activityID}
             />
-            <MatchForm
-              closeWithSnackbar={(status) => {
-                handleTeamFormSubmit(status, setOpenMatchForm);
-              }}
-              openMatchForm={openMatchForm}
-              setOpenMatchForm={(bool) => setOpenMatchForm(bool)}
-              activity={activity}
-            />
             <GordonDialogBox
               title="Admin Settings"
               fullWidth
--- a/src/views/RecIM/views/Activity/index.jsx
+++ b/src/views/RecIM/views/Activity/index.jsx
@@ -21,6 +21,14 @@ import { formatDateTimeRange } from '../../components/Helpers';
 import GordonDialogBox from 'components/GordonDialogBox';
 import defaultLogo from 'views/RecIM/recim_logo.png';
 
+const getNumMatches = (seriesArray) => {
+  let n = 0;
+  seriesArray.forEach((series) => {
+    n += series.Match?.length ?? 0;
+  });
+  return n;
+};
+
 const Activity = () => {
   const navigate = useNavigate();
   const { activityID } = useParams();
@@ -349,11 +357,15 @@ const Activity = () => {
                 severity="error"
               >
                 <br />
-                <Typography>
-                  Are you sure you want to permanently delete this activity: '{activity.Name}'?{' '}
-                  <br />
-                  This action cannot be undone.
+                <Typography variant="body1">
+                  Are you sure you want to permanently delete this activity: '{activity.Name}'?
                 </Typography>
+                <Typography variant="body1">
+                  This includes <b>{activity.Team?.length ?? 0} teams</b> and
+                  <b> {activity.Series?.length ?? 0} series</b>, with a total of{' '}
+                  <b>{getNumMatches(activity.Series)} matches</b>.
+                </Typography>
+                <Typography variant="body1">This action cannot be undone.</Typography>
               </GordonDialogBox>
             )}
           </Grid>
--- a/src/views/RecIM/views/Activity/index.jsx
+++ b/src/views/RecIM/views/Activity/index.jsx
@@ -1,8 +1,7 @@
 import { Grid, Typography, Card, CardHeader, CardContent, Button, IconButton } from '@mui/material';
 import AddCircleRoundedIcon from '@mui/icons-material/AddCircleRounded';
 import { useEffect, useState } from 'react';
-import { useParams } from 'react-router';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useParams } from 'react-router-dom';
 import { useUser } from 'hooks';
 import GordonLoader from 'components/Loader';
 import GordonUnauthorized from 'components/GordonUnauthorized';
@@ -334,7 +333,7 @@ const Activity = () => {
                   </Grid>
                   <Grid item>
                     <Button color="error" variant="contained" onClick={handleDelete}>
-                      Delete this team
+                      Delete this activity
                     </Button>
                   </Grid>
                 </Grid>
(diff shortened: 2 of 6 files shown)


src/views/RecIM/views/Activity/index.jsx
```diff
@@ -1,23 +1,24 @@
 import { Grid, Typography, Card, CardHeader, CardContent, Button, IconButton } from '@mui/material';
 import AddCircleRoundedIcon from '@mui/icons-material/AddCircleRounded';
 import { useEffect, useState } from 'react';
-import { useParams } from 'react-router';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useParams } from 'react-router-dom';
```
amos-cha, line 4: ty for the update!

amos-cha (approved):
Good feature implementation. I would prefer a small rework on the confirmation as I don't know how I feel about using the web browser's confirmation as our confirmation. But since these are all admin tools, they should not be blocking

silasw2:
I would not use an alert for this. Using dialog box would be best.

amos-cha:
oo wow spicy i like the new confirmation box

src/components/GordonDialogBox/index.jsx
```diff
@@ -41,6 +41,7 @@ const GordonDialogBox = ({
       <DialogTitle
         className={titleClass ? titleClass : styles.gc360_gordondialogbox_title}
         id="alert-dialog-title"
+        sx={severity ? { bgcolor: `${severity}.main`, pb: '0' } : { bgcolor: 'primary.main' }}
```
amos-cha, line 44:
```suggestion
        sx={{ bgcolor: `${severity ?? 'primary'}.main`, pb: '0' }}
```
a very nice change! but wouldn't this be the same as ^
I could be wrong

src/views/RecIM/views/Match/index.jsx
```diff
@@ -89,6 +95,14 @@ const Match = () => {
     setOpenMatchForm(false);
   };
 
+  const handleDelete = () => {
+    deleteMatchCascade(matchID);
+    setOpenConfirmDelete(false);
+    setOpenSettings(false);
+    navigate(`/recim/activity/${match.Activity.ID}`);
```
amos-cha, line 102: Hypothetically we could start pre-emptively sending data back to our pages for snackbar information
i.e
```
let res = deleteMatchCascade(matchID); //pretty sure this returns something or at least it should
. . .
navigate(`/recim/activity/${match.Activity.ID}`, {"match" : res}); //pass the prop back. I don't think it'll break anything
```

src/views/RecIM/views/Match/index.jsx
```diff
@@ -174,9 +195,7 @@ const Match = () => {
           </Grid>
           <Grid item xs={2}>
             <img
-              src={
-                match?.Activity.Team.find((t) => t.ID === match?.Team[1]?.ID).Logo ?? defaultLogo
-              }
+              src={match?.Team.find((t) => t.ID === match?.Team[1]?.ID)?.Logo ?? defaultLogo}
```
amos-cha, line 198: good catch, I think I do this in my PR as well

src/views/RecIM/views/Activity/index.jsx
```diff
@@ -288,6 +313,49 @@ const Activity = () => {
...
+                title="Admin Settings"
+                fullWidth
+                open={openSettings}
+                cancelButtonClicked={() => setOpenSettings(false)}
+                cancelButtonName="Close"
+              >
+                <br />
+                <Grid container alignItems="center" justifyContent="space-between">
+                  <Grid item>
+                    <Typography>Permanently delete the activity '{activity.Name}'</Typography>
```
amos-cha, line 327: I think I would prefer for there to be a bit more information on the Activity. Especially a check for if any games/series have started. It's a bit nitpicky so I would consider it non-blocking and we can obviously add it later.

src/views/RecIM/views/Team/index.jsx
```diff
@@ -78,31 +83,53 @@ const Team = () => {
     setOpenTeamForm(false);
   };
 
+  const handleDelete = () => {
+    deleteTeam(teamID);
+    setOpenConfirmDelete(false);
+    setOpenSettings(false);
+    navigate(`/recim/activity/${team.Activity.ID}`);
+    // @TODO add snackbar
+  };
```
amos-cha, line 92: same comment as above, but obviously we can handle it in another PR

src/views/RecIM/views/Team/index.jsx
```diff
@@ -78,31 +83,53 @@ const Team = () => {
...
+                    setOpenTeamForm(true);
+                  }}
+                  className={styles.editIconButton}
+                  sx={{ ml: 1 }}
+                >
+                  <EditIcon />
+                </IconButton>
+              )}
+            </Typography>
+            {teamRecord()}
```
amos-cha, line 118:
```suggestion
            {teamRecord()}
```
This unfortunately doesn't work the way we want it to. I'll rework it in a separate PR unless you want to.

The issue lies in the fact that teamRecord pulls from the first available series' record. Which is not an accurate team record. There are 2 ways we could potentially handle this:

1. We can do a summation on ALL the series records and make a global Activity Record (which might make some teams appear worse because they made the playoffs but didn't do well) This is an easy fix, just do a foreach and add them together
2. **my preference** We implement a way to switch between matches of a series. (could be by tabs or some other way) Assume we use tabs, if I select `series#1` I'll get my `series#1` record and Schedule will show the matches of that series. This also means that we would need to have a `Match History` implementation on the Team Page that shows previous games which might also need to be sorted. This is relatively complex but I feel would be the coolest. (Possibly something to work on down the line)

amos-cha: Regardless, non-blocking

amos-cha (approved):
Good work on the new confirmation window! I left a few comments that are probably worth a read but they're more or less non-blocking. Feel free to tackle them if you feel like it or for the sake of time, just make an issue and get onto the next issue i suppose

src/components/GordonDialogBox/index.jsx
cpabbot: Almost but I only want the zero bottom padding when there's a severity; your suggested change would have zero bottom padding all the time.

This is not the ideal way to do this but I don't want to mess with Gordon Dialog Box too much since I don't know all the use cases and don't want to refactor it.

amos-cha: all g

src/views/RecIM/views/Match/index.jsx
cpabbot: Probably true, I'ma leave this for now until we have snackbar stuff implemented and have a better idea of what's going on.

src/views/RecIM/views/Match/index.jsx
cpabbot: Thought so since it worked on our deployed branch. Had to fix bc it crashed the page.

src/views/RecIM/views/Activity/index.jsx
cpabbot: Yeah, I don't think it'd be too hard but I think I'm going to leave it for later if we want it in the future.

src/views/RecIM/views/Team/index.jsx
amos-cha: I just realized this wasn't actually a part of your PR rather just reformatting. NVM but I think the logic still stands and we probably need to make an issue for it.

src/views/RecIM/views/Activity/index.jsx
cpabbot: Changed my mind lol I'm doin it

amos-cha: lol ok 👍🏼

src/views/RecIM/views/Team/index.jsx
cpabbot: Gotcha, yeah let's handle that in a separate PR since this was just a rearrangement! Sounds good though

src/views/RecIM/views/Activity/index.jsx
```diff
@@ -246,48 +279,85 @@ const Activity = () => {
...
+              activityID={activityID}
+            />
+            <MatchForm
+              closeWithSnackbar={(status) => {
+                handleTeamFormSubmit(status, setOpenMatchForm);
+              }}
+              openMatchForm={openMatchForm}
+              setOpenMatchForm={(bool) => setOpenMatchForm(bool)}
+              activity={activity}
+            />
```
amos-cha, line 317: We have 2 match forms?

amos-cha (changes requested):
potentially a small merge error?

src/views/RecIM/views/Activity/index.jsx
cpabbot: Great catch!!

amos-cha (approved):
🔥

PR #1732, opened 2023-03-18
Recim image first hotfix

Kevin-Chu58:
While the image is null, the UI still throws an error when you clicked the `submit` button. That's because of the following code, where on line 337 compares `cropperImageData` and `null` value. But the operator is wrong here. When we use `!==`, this compares the data type on both sides, and since `undefined` is not the same type as `null`, the code will go through line 338 which the cropper is `undefined` in `cropperRef`. So the correct operator is `!=`, where both `undefined` and `null` are false, thus the code goes through line 339 instead of line 338.

image


amos-cha (approved):
Thank you for the fast fix!

PR #1733, opened 2023-03-18
[Rec-IM] Show/Sort Series Teams by Series Records + Styling Update

amos-cha:
Team Records
--

Comparison (OLD)
image

- Teams now have a Win/Loss count (Tie is unavailable currently as the calculation needs a rework)
- Activity font on the listing is smaller, and location has been moved below the team name (the benefit of this is that we can fit longer team/activity names and only need a smaller space for the listing/invite)

image
- `Home` shows an all encompassing Activity Record (a calculation done on summation of Series Records)
- `Activity` page shows Series Records individually and sorts them by Most Wins

image
- Activity see's teams by `Series` (thus individual series records)
- Tab is autofocused on first `ongoing` series (handled by time), defaulted to first tab

image
- Series tabs are centered **UNTIL** they need to be scrollable
- MUI does not allow for scrollable and centered, had to do an interesting workaround using `Mui-Box`

Design Rework
--

### Invites
image
image
- Accept/Reject colors use available Gordon Colors
- Accept/Reject reduced to `IconButton`
- Updated collapse size to ensure buttons dont get overlayed

Comparison (OLD)
image

- Invites tab is **ALWAYS** available. 
- Badge disappears if there is no invite
image

### Match Listing
image
- Displays **Final** over completed games and Score
- Displays **Forfeited** over forfeited games
- `vs.` if not a completed status
- Updated space available for Team names. This causes a small issue if the name is VERY VERY long, however, without the increased in dedicated size for the score, the name pushes up against the score. Also currently the Team names encroach on the space dedicated for Time and Surface (below) 

Comparison (Old)
image

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -5,7 +5,7 @@
 # VITE_ANALYTICS_ID=UA-3559968-3
 
 # @TRAIN
-# REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+# VITE_API_URL=https://360ApiTrain.gordon.edu/
 
 # @RECIM
-REACT_APP_API_URL=https://360ApiRecIM.gordon.edu/
+VITE_API_URL=https://360ApiRecIM.gordon.edu/
--- a/src/_vars.scss
+++ b/src/_vars.scss
@@ -5,9 +5,11 @@
 // Ensure that the relative path (i.e. `../../`, etc.) is correct when importing from a nested
 // component.
 
-// For use in a .js file, see `src/theme.js`.
+// For use in a .ts file, see `src/theme.ts`.
 $primary-blue: #014983;
+$primary-blue-tint: #016bc0;
 $primary-cyan: #00aeef;
+$primary-cyan-shade: #0098d0;
 $secondary-green: #b2bb1c;
 $secondary-yellow: #fdb913;
 $secondary-orange: #de571f;
--- a/src/views/RecIM/components/Forms/MatchForm/index.jsx
+++ b/src/views/RecIM/components/Forms/MatchForm/index.jsx
@@ -120,11 +120,12 @@ const MatchForm = ({ closeWithSnackbar, openMatchForm, setOpenMatchForm, activit
 
   const currentInfo = useMemo(() => {
     if (match) {
+      console.log(match);
       //I tried using inbuild javascript functions but I can't wrap my head around multiple
       //filters. You are welcome to improve on the logic below if you so desire.
       var teamIDs = [];
       match.Team.forEach((team) =>
-        teamIDs.push(match.Series.TeamStanding.find((_team) => team.ID === _team.TeamID).Name),
+        teamIDs.push(match.Series.TeamStanding.find((_team) => team.ID === _team.TeamID)?.Name),
       );
       return {
         StartTime: match.StartTime,
@@ -216,7 +217,7 @@ const MatchForm = ({ closeWithSnackbar, openMatchForm, setOpenMatchForm, activit
     });
     return updatedFields;
   }
-
+  console.log(match, activity);
   const handleConfirm = () => {
     setSaving(true);
 
@@ -232,7 +233,7 @@ const MatchForm = ({ closeWithSnackbar, openMatchForm, setOpenMatchForm, activit
     matchRequest.TeamIDs.forEach((value) => {
       if (activity) idArray.push(activity.Team.find((team) => team.Name === value).ID);
       else if (match)
-        idArray.push(match.Series.TeamStanding.find((team) => team.Name === value).TeamID);
+        idArray.push(match.Series.TeamStanding.find((team) => team.Name === value)?.TeamID);
     });
     matchRequest.TeamIDs = idArray;
 
--- a/src/views/RecIM/components/List/index.jsx
+++ b/src/views/RecIM/components/List/index.jsx
@@ -68,7 +68,6 @@ const MatchList = ({ matches, activityID }) => {
 // setTargetTeamID is used for edit Match teams
 const TeamList = ({ teams, match, series, invite, setInvites, setTargetTeamID }) => {
   const navigate = useNavigate();
-  if (!teams?.length && !match && !series) return <Typography>No teams to show.</Typography>;
 
   const handleInviteResponse = (response, activityID, teamID) => {
     if (response === 'accepted') {
--- a/src/views/RecIM/components/List/index.jsx
+++ b/src/views/RecIM/components/List/index.jsx
@@ -58,7 +58,9 @@ const ParticipantList = ({
 const MatchList = ({ matches, activityID }) => {
   if (!matches?.length || !matches[0]) return <Typography>No matches to show.</Typography>;
   let content = matches.map((match) => (
-    <MatchListing key={match.ID} match={match} activityID={activityID} />
+    <MatchListing key={match?.ID} match={match} activityID={activityID} />
+    // I have no idea why, but on ladder matches, match can't be found
+    // despite it showing up on the debugger.
   ));
 
   return <List dense>{content}</List>;
--- a/src/views/RecIM/components/List/index.jsx
+++ b/src/views/RecIM/components/List/index.jsx
@@ -1,9 +1,11 @@
 import { List, Typography } from '@mui/material';
 import { ActivityListing, MatchListing, ParticipantListing, TeamListing } from './Listing';
 import { useNavigate } from 'react-router-dom';
+import styles from './List.module.css';
 
 const ActivityList = ({ activities, showActivityOptions }) => {
-  if (!activities?.length) return <Typography>No activities to show.</Typography>;
+  if (!activities?.length)
+    return <Typography className={styles.secondaryText}>No activities to show.</Typography>;
   let content = activities.map((activity) => (
     <ActivityListing
       key={activity.ID}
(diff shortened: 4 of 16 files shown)


src/_vars.scss
```diff
@@ -5,9 +5,11 @@
 // Ensure that the relative path (i.e. `../../`, etc.) is correct when importing from a nested
 // component.
 
-// For use in a .js file, see `src/theme.js`.
+// For use in a .ts file, see `src/theme.ts`.
 $primary-blue: #014983;
+$primary-blue-tint: #016bc0;
 $primary-cyan: #00aeef;
+$primary-cyan-shade: #0098d0;
```
amos-cha, line 12: These two are pulled from theme.js 
There should not be a problem with this I assume.

src/views/RecIM/components/Forms/MatchForm/index.jsx
```diff
@@ -216,7 +217,7 @@ const MatchForm = ({ closeWithSnackbar, openMatchForm, setOpenMatchForm, activit
     });
     return updatedFields;
   }
-
+  console.log(match, activity);
```
amos-cha, line 220:
```suggestion
```
whoops

src/views/RecIM/components/Forms/MatchForm/index.jsx
```diff
@@ -120,11 +120,12 @@ const MatchForm = ({ closeWithSnackbar, openMatchForm, setOpenMatchForm, activit
 
   const currentInfo = useMemo(() => {
     if (match) {
+      console.log(match);
```
amos-cha, line 123:
```suggestion
```
whoops

.env.production
```diff
@@ -5,7 +5,7 @@
 # VITE_ANALYTICS_ID=UA-3559968-3
 
 # @TRAIN
-# REACT_APP_API_URL=https://360ApiTrain.gordon.edu/
+# VITE_API_URL=https://360ApiTrain.gordon.edu/
 
 # @RECIM
-REACT_APP_API_URL=https://360ApiRecIM.gordon.edu/
+VITE_API_URL=https://360ApiRecIM.gordon.edu/
```
cpabbot, line 11: Curious why this wasn't a problem before?

src/views/RecIM/components/List/index.jsx
```diff
@@ -76,6 +76,28 @@ const TeamList = ({ teams, match, invite, setInvites, setTargetTeamID }) => {
...
+            LossCount: team.LossCount,
+            TieCount: team.TieCount,
+          },
+        ],
+      }),
+    );
+  }
+
   let content = match
     ? match.Team.map((team) => (
```
cpabbot, line 102: Not from the PR but relevant: instead of mapping match.Team and having to pass in match, why doesn't the TeamList just get called with team={match.Team}? Then we don't need any of this match prop stuff.

src/views/RecIM/components/List/index.jsx
```diff
@@ -64,9 +64,9 @@ const MatchList = ({ matches, activityID }) => {
   return <List dense>{content}</List>;
 };
 // setTargetTeamID is used for edit Match teams
-const TeamList = ({ teams, match, invite, setInvites, setTargetTeamID }) => {
+const TeamList = ({ teams, match, series, invite, setInvites, setTargetTeamID }) => {
   const navigate = useNavigate();
-  if (!teams?.length && !match) return <Typography>No teams to show.</Typography>;
+  if (!teams?.length && !match && !series) return <Typography>No teams to show.</Typography>;
```
cpabbot, line 69: This doesn't exactly make sense logically. I think what you're going for is ORs not ANDs, but I don't think match nor series needs to be a requirement; just team length. We should return "no teams to show" if and only if teams doesn't exist / has length 0.

src/views/RecIM/views/Home/index.jsx
```diff
@@ -70,7 +74,7 @@ const Home = () => {
       setInvites(await getTeamInvites());
       if (profile) {
         setParticipant(await getParticipantByUsername(profile.AD_Username));
-        setMyTeams(await getParticipantTeams(profile.AD_Username));
+        setParticipantTeams(await getParticipantTeams(profile.AD_Username));
```
cpabbot, line 77: What's your thinking behind this change? To me, `myTeams` makes more sense to me than `participantTeams` as to what it is exactly, but I don't mind it.

cpabbot (changes requested):
Great work on this!! Love all the visual changes and work here! Left a few comments.

.env.production
amos-cha: It was, but we don't use production except on deployment server. I had to make that adjustment and I'm getting tired of manually replacing it so I made it a change here to be pushed in.

src/views/RecIM/components/List/index.jsx
amos-cha: I think you're getting half of it, but I'm pretty sure that this is actually an all encompassing check to make sure that the team list is properly propagated. the `&& !series` checks to make sure that we're mapping proper data. However, I'm actually questioning why we even do this check at all. I think it was originally set in place when the only way we could propagate TeamList was with teams.length. Since now there are more ways to do this, we might be able to remove this entire conditional check.

src/views/RecIM/components/List/index.jsx
amos-cha: this was due to the `MatchStats` form needing to know the Match's basic data as well as specifically the specific team's data. Thus we also need the `match` (the line right below this, I think its line103) has that specification.

src/views/RecIM/views/Home/index.jsx
amos-cha: This is mainly due to the fact that I wanted to actual "My Teams" card to be called `myTeams` which conflicted with the then useState. This was just a choice between which I thought was more relevant to the term `my teams`

src/views/RecIM/components/List/index.jsx
cpabbot: Ah I did not realize we did this that way; I checked the code and now understand. Using TeamList for that is kind of strange but I think I understand the thought process and I suppose it is technically a list of teams. This is fine for now then.

src/views/RecIM/components/List/index.jsx
cpabbot: Think about if there was null teams but match existed - this logic would fail would it not? And yeah I think if match or series fails we can handle that elsewhere; this should just check if there are any teams to show or not.
```suggestion
  if (!teams?.length) return <Typography>No teams to show.</Typography>;
```

src/views/RecIM/views/Home/index.jsx
cpabbot: This makes sense, I might rename myInvites to myInvitesTab and myTeams to myTeamsTab; might be more consistent? Don't have to though, but I feel like that would be most consistent (considering also how we label the card as myTeamsCard).

src/views/RecIM/components/List/index.jsx
amos-cha: im pretty sure this conditional is a redundant check to ensure that teamslist has something passed into it

src/views/RecIM/views/Home/index.jsx
amos-cha: either way works

src/views/RecIM/components/List/index.jsx
amos-cha: I'm just deleting the check, just tested. It works perfectly without it.

cpabbot (dismissed):

src/views/RecIM/components/List/index.jsx
```diff
@@ -58,15 +58,16 @@ const ParticipantList = ({
...
+    // despite it showing up on the debugger.
   ));
 
   return <List dense>{content}</List>;
 };
 // setTargetTeamID is used for edit Match teams
-const TeamList = ({ teams, match, invite, setInvites, setTargetTeamID }) => {
+const TeamList = ({ teams, match, series, invite, setInvites, setTargetTeamID }) => {
   const navigate = useNavigate();
-  if (!teams?.length && !match) return <Typography>No teams to show.</Typography>;
```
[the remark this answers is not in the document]
amos-cha: I'm so confused on what you want. You requested that `!series` be removed because the logic fails. But then you want !match to stay in there? I think there's some logic flow that might not make sense to you at this point because if you're concerned about the fact that this is a good check to simply exist, we should be putting `!series` in there too. 

Also I think you believe that I tossed that `!series` conditional inside of this if statement without thinking and that I'm doing this simply because it happens everywhere else. This is not the case, I add that `!series` conditional because otherwise, teams in the series tab drop down **DO NOT RENDER** because I do not propagate the team or match field.  Once again, I beg the question, what do you mean that the boolean logic does not make sense and that I need an "or"

Just a reminder:
Home page -> TeamList(**team**, _no match_, _no series_)
MatchStatsEdit -> TeamList(_no team_, **match**, _no series_)
Activity page -> TeamList(_no team_, _no match_, **series**)

So, the question stands, **do you want this check in there or not**. If you want the check, I'm putting `!series` in there as well. If you want to sanitize BEFORE we put anything into TeamList then we should remove the check.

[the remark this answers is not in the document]
cpabbot: Oh I think I understand better, thanks for laying that all out like that! I did not realize that the team param was not always used to get the teams, which makes sense sorta but was not obvious to me. Idk if this is ideal but I'ma just approve this now, I've harassed you enough haha :)

cpabbot (approved):
LGTM, thanks for all your hard work!!

PR #1734, opened 2023-03-18
Fac/Staff Personal Info Toggle Visual Fix

charlielee144r:
Continuation from #1709.

This PR fixes state logic for showing which selection (from private or public) personal information currently is. Previously, the logic for setting the state of the toggle always returned false when there was no Home phone no. added (it seems), which automatically set the toggle to 'true' (which is shown as public by the toggler). So whether or not you set your profile as private or public with the toggle, it would show that it was set to public (see https://github.com/gordon-cs/gordon-360-ui/issues/1653).


charlielee144r:
Fixes #1653

EjPlatzer (approved):
LGTM! Ditto what Bennett said

PR #1735, opened 2023-03-18
Involvement Tabs

sofistonehouse:


sofistonehouse (commented):
This mostly fixes the layout issues, but when I try to shrink the screen to a phone size, the tab titles are too big and almost run into each other, so I'm not sure what to do about that

src/views/InvolvementsAll/components/Requests/index.jsx
```diff
@@ -136,8 +136,9 @@ const Requests = ({ profile, session }) => {
   }
 
   return (
-    <Grid item xs={12} lg={8}>
-      <Card className={styles.requests}>{content}</Card>
+    <Grid item xs={12} lg={12}>
+      {/* <Card className={styles.requests}></Card> */}
```
EjPlatzer, line 140: Before we merge, you should make sure to remove any code that you commented out.

src/views/InvolvementsAll/components/Requests/index.jsx
```diff
@@ -46,7 +46,7 @@ const Requests = ({ profile, session }) => {
   } else if (involvementsLeading?.length > 0) {
     content = (
       <>
-        <CardHeader title="Membership Requests" className={styles.requests_header} />
+        {/* <CardHeader title="Membership Requests" className={styles.requests_header} /> */}
 
         <CardContent>
```
EjPlatzer, line 51: Since you got rid of the surrounding `Card` component, you can remove the `CardContent` as well.

EjPlatzer: Also, it looks like you need to remove the `CardHeader` and `CardContent` in the `else` block for when no requests are found (~lines 108, 115).

src/views/InvolvementsAll/index.jsx
```diff
@@ -173,87 +170,117 @@ const InvolvementsAll = () => {
...
-                    {types.map((type) => (
-                      <MenuItem value={type} key={type}>
-                        {type}
-                      </MenuItem>
-                    ))}
-                  </Select>
-                </FormControl>
-              </Grid>
             </Grid>
           </CardContent>
```
EjPlatzer, line 174: I think that, rather than ending here, this `CardContent` element should wrap the entire `TabContext` as well (and you should remove any `CardContent`s within the `TabPanel`s below).

EjPlatzer (changes requested):
This is looking good! Just a couple bits of cleanup.

src/views/InvolvementsAll/InvolvementsAll.module.scss
```diff
@@ -0,0 +1,14 @@
+@import '../../vars';
+
+.tab {
+  //background-color: $primary-blue;
```
bennettforkner, line 4: Make sure to delete these comments before you are actually ready for merge!

src/views/InvolvementsAll/index.jsx
```diff
@@ -80,6 +84,10 @@ const InvolvementsAll = () => {
     navigate(`?session=${value}`);
   };
 
+  const handleChange = (event, newValue) => {
```
bennettforkner, line 87: I would rename this to be a little bit more descriptive of what change it is handling. Perhaps `handleChangeTab(...)`?

src/views/InvolvementsAll/index.jsx
```diff
@@ -173,86 +170,122 @@ const InvolvementsAll = () => {
...
+                    value="1"
+                  />
+                  <Tab
+                    sx={{ border: '2px solid', borderColor: gordonColors.secondary.neutral }}
+                    label="Requests"
+                    className={styles.tab}
+                    value="2"
+                  />
+                  <Tab
+                    // label={`${involvementSessionText} Involvements`}
```
bennettforkner, line 197: Make sure to remove commented-out code

bennettforkner (changes requested):

bennettforkner:
Hey @sofistonehouse could you add a few screenshots to the description above so that we can all see what these changes look like? It would be good to have at least 1 on desktop and 1 on mobile. Thanks!

src/views/InvolvementsAll/index.jsx
```diff
@@ -173,86 +169,104 @@ const InvolvementsAll = () => {
...
-                  <InvolvementsGrid
-                    involvements={myInvolvements}
-                    sessionCode={selectedSession}
-                    noInvolvementsText={myInvolvementsNoneText}
-                  />
+                  profile && (
+                    <Grid item xs={12} lg={12}>
+                      {loading ? (
+                        <GordonLoader />
+                      ) : (
```
bennettforkner, line 195: You've nested two GordonLoaders inside of conditional logic on this page. It can look kind of glitchy while loading when you do that. I would combine them so that line 188 looks like:
`{loadingProfile || loading ? (`
Then you can remove the logic in lines 193-195

src/views/InvolvementsAll/index.jsx
```diff
@@ -173,86 +169,104 @@ const InvolvementsAll = () => {
...
+                  xs={12}
+                  lg={12}
+                  xl={12}
+                  container
+                  alignItems="center"
+                  justifyContent="center"
+                >
+                  {!isOnline ? null : loadingProfile ? (
+                    <GordonLoader />
+                  ) : (
```
bennettforkner, line 218: I think this is already inside of a check for `loadingProfile`. Correct me if I'm misreading the indentations.

src/views/InvolvementsAll/index.jsx
```diff
@@ -173,86 +169,104 @@ const InvolvementsAll = () => {
...
+                              {type}
+                            </MenuItem>
+                          ))}
+                        </Select>
+                      </FormControl>
+                    </Grid>
+                  </Grid>
+                  {loading ? (
+                    <GordonLoader />
+                  ) : (
```
bennettforkner, line 260: If you follow my instructions for combining `loading` and `loadingProfile` from above, you can remove this logic.

bennettforkner (changes requested):

Luke-W-Hart:
@bennettforkner @EjPlatzer I know it's been a while since this has been worked on, just trying to finish it up! 
It seems like each of the conditions are in each of their respective tab panels, so I think the logic on lines 212-216 and 254-262 should be kept. Does that make sense or should it be changed?

PR #1738, opened 2023-03-20
[Rec-IM] Show Series Schedule + Visual Updates (Menus) + Merge Develop

amos-cha:


amos-cha:
> > In the future, merging from develop (or for anything like this) should be kept isolated in its own PR. Here it is difficult for us to now review and understand what changes were actually made by the Rec-IM team. Just for future note. But whenever you mark this ready I can review.
> 
> I second this, it is hard to keep what you changed straight with the develop merge. If possible, could your re-base your commits to a different branch so that we can handle them separately?

I did this in an attempt to see if it fixed the issues regarding not showing names due to profiles being broken. Turns out it was a train issue and not us. Thus, I wasn't bothered enough to revert the issue.

PR #1743, opened 2023-03-26
[Rec-IM] Form cleanup

silasw2:
This is a bit of a doozy PR, but significantly improves code quality and readability. It also refactors some of the way we handle forms in general with the main difference being the Edit Match Stats page. Will post picture updates when available.


amos-cha (approved):
Looked over and gone over in person. Excellent work! If there is nothing left to add, this would be great to add to Rec-IM ASAP!

amos-cha (approved):
Sentiment from before has not changed

PR #1746, opened 2023-03-27
[Rec-IM] 'Join' Instead of 'Create Team' for Individual Activities

cpabbot:
# 'Join Activity' instead of 'Create Team' for solo activities
Two functions:
1. 'Join Activity'
   - automatically create a team with one's own name
   - prevent images or names from being edited
2. 'Add Participant'
   - for admins
   - currently gives invite status, so it will show up in the user's invites


src/views/RecIM/components/Forms/ImageOptions/index.jsx
```diff
@@ -29,15 +29,15 @@ const ImageOptions = ({
...
   const [imageWidth, setImageWidth] = useState();
   const [imageHeight, setImageHeight] = useState();
   const cropperRef = useRef();
-  const [loading, setLoading] = useState();
+  const [loading, setLoading] = useState(false);
 
   // load ImageOptions
   useEffect(() => {
-    setLoading(true);
+    // setLoading(true);
```
cpabbot, line 40: Not ideal of course. This file needs to be refactored but will do in a separate PR. Reason for changes needed is because of the && logic removal from the Activity/Team pages which was covering up this loading bug.

amos-cha: fair, make it an issue

src/views/RecIM/components/Forms/InviteParticipantForm/index.jsx
```diff
@@ -4,17 +4,22 @@ import GordonDialogBox from 'components/GordonDialogBox';
...
 import styles from './InviteParticipantForm.module.css';
+import userService from 'services/user';
+import { useUser } from 'hooks';
 
 const InviteParticipantForm = ({
   closeWithSnackbar,
   openInviteParticipantForm,
   setOpenInviteParticipantForm,
   teamID,
+  soloTeam,
```
amos-cha, line 19:
```suggestion
  individualSport,
```

Can we make this name change? I don't really like soloTeam, it sounds odd.

amos-cha (changes requested):
I would really like the name change to be done, soloTeam just sounds weird

src/views/RecIM/components/Forms/InviteParticipantForm/index.jsx
cpabbot: I can change that

amos-cha (dismissed):
👍🏼

amos-cha (approved):

PR #1749, opened 2023-03-28
Replace unescape function call

### LINKED ISSUE #512: Functions escape() and unescape() are deprecated. Remove them and ensure no problems result. (see issue comment for details.)
https://stackoverflow.com/questions/30631927/converting-to-base64-in-javascript-without-deprecated-escape-call

charlielee144r:
This PR removes the last(?) unescape deprecated function call.

Fixes #512


charlielee144r:
I wasn't quite able to test if this was working correctly due to some other issues I found. Will keep as draft for now.

charlielee144r:
I was able follow the trail for each time this function is called and it seems that I can't reach the else statement. Is it possible that it never has an opportunity to even reach the else?

bennettforkner:
> I was able follow the trail for each time this function is called and it seems that I can't reach the else statement. Is it possible that it never has an opportunity to even reach the else?

It's definitely possible. Note that the conditional is based off of the content encoding of the uploaded file data. That would be set by the upload process (the image uploader in most cases) which I think is consistent across the site. I wouldn't like to remove it entirely just in case it caused issues, but your changes are unlikely to cause any problems if we are unlikely to even hit that case. I'm going to approve for that reason.

bennettforkner (approved):
LGTM. See my previous comment as well.

PR #1750, opened 2023-03-28
[Rec-IM] Display match winner with styling

cpabbot:
image
image


amos-cha (approved):
I like this first iteration, I had a couple ideas about it but this is good at the moment!

PR #1752, opened 2023-03-29
[Rec-IM] Home Page hang

amos-cha:
There was a bug that caused non-participants to load infinitely on the home page due to `Team Invites` and `Participant Teams` loading (since they weren't participants). Logic is moved after we check for a participant; if `null` we know that they haven't signed up for rec-im (which is fine). 

Next thing to do is to move the `Waiver` to `Accept Team Invite` and `Join an Activity`


cpabbot (approved):
LGTM! Good fix!

Kevin-Chu58 (approved):
LGTM!

PR #1764, opened 2023-04-04
Centralizing themes

sharonpatta:
Most components have been changed to use a theme variable, rather than hard-coding in a color. (Components that have not yet been changed: "Results" header for people search, Timesheets info)


src/app.jsx
```diff
@@ -30,25 +32,27 @@ const App = () => {
...
-                {routes.map((route) => (
-                  <Route key={route.path} path={route.path} element={route.element} />
-                ))}
-              </Routes>
-            </>
-          </main>
-        </section>
-      </Router>
-    </ErrorBoundary>
+    <CssVarsProvider theme={theme}>
```
EjPlatzer, line 35: Since you started working on this, all the `Provider`-type components were relocated to `src/components/Providers/index.tsx`. Could you please rebase (or merge) with the latest `develop` and move this to that file?

src/components/EventList/index.jsx
```diff
@@ -62,7 +64,13 @@ const EventList = ({ events }) => {
 
   return (
     <Card>
-      <CardHeader title={header} className={styles.header} />
+      <CardHeader
+        title={header}
+        className={styles.header}
+        style={{
+          backgroundColor: theme.colorSchemes.light.palette.primary[300],
+        }}
```
EjPlatzer, line 72: This is not exactly what we want. The end goal is that, when the MUI Theme is updated (e.g. because the user selected Dark Mode), every component that uses theme colors will change styles.

What you have currently here will not update when the MUI Theme is updated, it will only change when the code in the `src/theme.ts` file is updated (which obviously cannot be done by user input). Notice, for example, that you've hard-coded `theme.colorSchemes.light` here - so what would happen when the UI switches to using `theme.colorSchemes.dark` instead?

There are a couple of correct ways to do this. The simplest, for our current arrangement, is to use the new `--mui-*` CSS variables to style this component via CSS (similar to how it was previously being styled by the `styles.header` class).

The other way, if we want to put the styling info into the JSX file (which is debatable) is to use the `theme` variable injected by MUI (via the Provider at the root of the component tree). That is, instead of using the static `theme` object that we created in `src/theme.ts` but that will never be updated, we want to use the dynamic `theme` variable that MUI is managing and that can be changed by, for example, a Dark Mode toggle.

There are basically two ways to access this dynamic `theme` variable:
1. Using the `sx` prop of MUI components, as in this example: https://mui.com/system/palette/#color
2. With the `useTheme` hook: https://mui.com/material-ui/customization/theming/#accessing-the-theme-in-a-component

src/components/Header/index.jsx
```diff
@@ -153,7 +153,7 @@ const GordonHeader = ({ onDrawerToggle }) => {
 
   return (
     <section className={styles.gordon_header}>
-      <AppBar className={styles.app_bar} position="static">
+      <AppBar className={styles.app_bar} position="static" color="primary">
```
EjPlatzer, line 156: According to the [`AppBar` API docs](https://mui.com/material-ui/api/app-bar/), `"primary"` is the default value of the `color` prop, so I think this isn't needed.
```suggestion
      <AppBar className={styles.app_bar} position="static">
```

src/components/Profile/components/Identification/Identification.module.scss
```diff
@@ -155,7 +155,7 @@ $image-side-width-desktop: 90px;
               text-decoration: none;
             }
             &_container {
-              color: $primary-blue;
+              color: var(--mui-palette-primary-300);
```
EjPlatzer, line 158: Is it possible to just use `--mui-palette-primary` here instead of specifying which shade of the primary color we want to use?

EjPlatzer (changes requested):
Thanks for working on this! I can tell you made some great progress, although some of these changes are not quite in the correct direction.

.env.production
```diff
@@ -1,8 +1,8 @@
...
 
 # @PROD
-# VITE_API_URL=https://360Api1.gordon.edu/
-# VITE_ANALYTICS_ID=UA-3559968-3
+VITE_API_URL=https://360Api1.gordon.edu/
+VITE_ANALYTICS_ID=UA-3559968-3
 
 # @TRAIN
-VITE_API_URL=https://360ApiTrain.gordon.edu/
+# VITE_API_URL=https://360ApiTrain.gordon.edu/
```
bennettforkner, line 8: Make sure to revert this before having it in an open PR!

src/components/Profile/components/MembershipsList/components/MembershipInfoCard/index.jsx
```diff
@@ -15,6 +16,9 @@ const MembershipInfoCard = ({ myProf, membership, onTogglePrivacy }) => {
           className={`gc360_text_link ${
             showPrivate ? styles.private_membership : styles.public_membership
           }`}
+          style={{
+            color: theme.colorSchemes.light.palette.primary[300],
+          }}
```
bennettforkner, line 21: We try to avoid inline styles wherever possible. Can this be updated to be in an scss file instead?

PR #1765, opened 2023-04-05
[Rec-IM] Change Requests

cpabbot:
ISG Change Request and Minor Improvements

Haven't done everything, but going to stop for now and see if we can get these things in!

Notes:
- [x] ISG change requests (some)
- [x] Remove "registration closed" blue tag
- [x] Fix issue of conflicting green tags
- [x] Make Rec-IM slogan text smaller
- [ ] Ensure headers stick on one line (smaller logos or text on activity page etc.)
- [ ] Make the headings on the dropdowns tiny text
- [x] Swap tuner icon to more (3 dots) icon
- [x] Consider swapping orange buttons to blue text inline (like surfaces)
- [x] Create "an" Activity
- [x] Fix bug where second team's score is not bolded when they win
- [x] Swap tabs for teams to filter dropdown
- [x] Only show active activities on the home page
- [x] Delete all matches in a series (without deleting the series itself)
- [x] Swap delete confirmations


amos-cha (approved):
I don't see anything glaring. It looks good and I've had enough time to make sure functionality is fully in place since it's currently deployed. Good work :)

PR #1767, opened 2023-04-08
[Rec-IM] Invite navigation fix

Kevin-Chu58:
A really simple fix for the invite navigation. Just check the boolean state of `invite` to setup the `to` property.


amos-cha (approved):
Awesome, good find!

PR #1768, opened 2023-04-08
[Rec-IM] Fix form validation

EjPlatzer:
I have refactored error validation for the RecIM forms. Instead of passing errors up and down the component tree needlessly, we simply perform the validation where needed based on static validation rules. This approach works really well almost across the board, and has resulted in simple error handling and overall shorter and simpler code.

If, in the future you need a more complex validation rule, you could add something like a `otherValidations` property to the field, and update the function `validateFieldFromUpdatedInfo` to use it.

The only place where the error status was being used outside of the `Form` component was in `EditMatchStatsForm`. I understand why the code is the way it is, but I personally think you might want to rework that form slightly for better user experience, Regardless, I was able to make it work by having that component repeat the validation.


amos-cha (changes requested):
So I've found a couple bugs:

This occurs when we try to type things into the `date` fields:
image

Whereas it used to be an error:
image

Besides this error, I don't think there should be anything else big worth noting? In response to the UI Update, I think that could be valid. However, we spent a decent time brainstorming this and this was the best concept we had (scrollable centered tabs) which visually doesn't look too bad for 2 teams but more importantly will work for our Ladder Match set-up. The visual update could be relatively low on priority as it is a purely admin feature.

EjPlatzer:
> This occurs when we try to type things into the `date` fields: image


Good catch! I didn't test everything else after fixing `EditMatchStats`, should be fixed now.

amos-cha (changes requested):
One more thing, I didnt catch:
image
Number of Teams should not be a required field (I think we need to accept the empty string)

Also, I don't know if this is in the scope of your PR but I noticed a bug that I didn't check before: 
image
I'm not sure why we're able to put in a character string here



Besides that, I think everything else is good!

src/views/RecIM/components/Forms/SeriesForm/index.jsx
```diff
@@ -43,103 +58,61 @@ const SeriesForm = ({
...
+          helperText: '*Optional',
+          required: false,
+        },
+        {
+          label: 'Number of Teams',
+          name: 'numberOfTeamsAdmitted',
+          type: 'number',
+          min: 1,
+          helperText: '*Invalid Number',
+          required: true,
```
EjPlatzer, line 103: Per Amos, this should not be required, we can allow empty
```suggestion
          min: 0,
          helperText: '*Invalid Number',
```

EjPlatzer:
> One more thing, I didnt catch: image Number of Teams should not be a required field (I think we need to accept the empty string)

It was marked as required before, but I guess the validation wasn't treating it that way. I have removed the `required` attribute and set the minimum to 0, so now it should allow an empty value but not negative values (although you'll still need server-side validation as always).

> Also, I don't know if this is in the scope of your PR but I noticed a bug that I didn't check before: image I'm not sure why we're able to put in a character string here

I'm not sure what form that is, but if I had to guess, it's probably a `type="text"` input instead of `type="number"`, so the browser allows you to type whatever you want. I fixed one improper type in this PR but I was doing my best to avoid the existing behavior.

amos-cha (approved):
Good stuff, thank you so much for looking over it!

PR #1769, opened 2023-04-10
[Rec-IM] schedule surfaces

amos-cha:
Handles changes from here: https://github.com/gordon-cs/gordon-360-api/pull/885
+ API bug fixes

### DIFF
--- a/src/services/recim/recim.ts
+++ b/src/services/recim/recim.ts
@@ -1,4 +1,4 @@
-import { BaseActivity } from './activity';
+import { CreatedActivity } from './activity';
 import http from '../http';
 
 export type Lookup = {
@@ -7,7 +7,6 @@ export type Lookup = {
 };
 
 type Account = {
-  GordonID: string;
   FirstName: string;
   LastName: string;
   Email: string;
@@ -17,7 +16,7 @@ type Account = {
 
 type ActivityReport = {
   NumberOfParticipants: number;
-  Activity: BaseActivity;
+  Activity: CreatedActivity;
 };
 
 type NewParticipantReport = {
--- a/src/services/recim/recim.ts
+++ b/src/services/recim/recim.ts
@@ -1,4 +1,46 @@
+import { BaseActivity } from './activity';
+import http from '../http';
+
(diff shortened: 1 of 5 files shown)


src/services/recim/recim.ts
```diff
@@ -1,4 +1,46 @@
+import { BaseActivity } from './activity';
+import http from '../http';
+
 export type Lookup = {
   ID: number;
   Description: string;
 };
+
+type Account = {
+  GordonID: string;
```
cpabbot, line 10: I don't think even admins are supposed to be seeing GordonIDs are they?

bennettforkner: @cpabbot correct. Only faculty and staff are permitted to see Gordon ID Numbers. And that should only be on their 360 profile. It shouldn't be needed anywhere else as far as we are aware.

src/services/recim/recim.ts
```diff
@@ -1,4 +1,46 @@
...
+  FirstName: string;
+  LastName: string;
+  Email: string;
+  ADUsername: string;
+  AccountType: string;
+};
+
+type ActivityReport = {
+  NumberOfParticipants: number;
+  Activity: BaseActivity;
```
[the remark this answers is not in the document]
amos-cha: having the report return a whole piece of activity data is not important (i.e, its matches, teams, series...) and it doesn't make sense to have so much data being passed around. If we wanted to make it more detailed, we still pass the activityID which can be used to query any of the activities.

src/services/recim/recim.ts
amos-cha: Considering this report will be locked behind AuthGroup of RecIM Director, this should be okay no? Unless they do not count as FacStaff

bennettforkner: The RecIM director has no special need to see Gordon ID numbers, so s/he should not get any special access to that.

Principle of least privilege dictates that we only grant what is actually necessary, no unnecessary access.

cpabbot (approved):
LGTM, good work! We will continue to work on reporting features going forward.

PR #1770, opened 2023-04-10
Update `xlsx` dependency

EjPlatzer:
I have updated our `xlsx` package to the latest version. Note that the `xlsx` [package on the public npm registry](https://www.npmjs.com/package/xlsx) is no longer the authoritative version and is not being updated. The official source for the package is now their own [cdn](https://cdn.sheetjs.com/) (and also the official source code is in their own Gitea instance, with the GitHub repo being deprecated).

I have tested that the updated package works as expected.

I fixed a regression in the Clifton Strengths uploader caused by a bad merge during the switch to Vite (my bad!).

I have also refactored the `SpreadsheetUploader` component. The only behavioral difference is that the preview of uploaded data now has a `max-height` of `40vb` (aka 40% of the height/block-direction of the viewport). This means that the data preview is scrollable when it is longer than that size, while the dialog buttons remain visible at all times. This also allowed me to use the dialog's buttons only, whereas previously there were other buttons at the top of the dialog content so that the user didn't have to scroll to the bottom of the preview in order to submit.

Here are some picture to demonstrate:
image
image
image
image

The other refactoring was purely code quality: simplifying syntax in a few places and re-working the file upload callback slightly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,7 +50,7 @@
         "react-to-print": "^2.14.1",
         "react-visibility-sensor": "^5.1.1",
         "url-regex-safe": "^2.1.0",
-        "xlsx": "^0.17.3"
+        "xlsx": "https://cdn.sheetjs.com/xlsx-0.19.2/xlsx-0.19.2.tgz"
       },
       "devDependencies": {
         "@types/lodash": "^4.14.176",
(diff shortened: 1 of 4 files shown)


bennettforkner (dismissed):

EjPlatzer:
I realized that the display of date columns was not super nice - it was showing "the number of days since 1900-01-01", which is a fine format to store dates but not at all human readable.

I added the `cellDates: true` option of `xlsx.read` so that it will transform these date numbers into JS date objects when deserializing a spreadsheet. This allows to format the dates more helpfully:

image

I then had to update the `submitCliftonStrengths` function to handle a proper JS date instead of the old format. While I was at it, I reworked the component slightly so that it stores data in the state and renders that data to JSX, instead of storing JSX as state.

bennettforkner (approved):

PR #1772, opened 2023-04-11
[Rec-IM] Image Upload Fix

Kevin-Chu58:
A simple one line fix that prevents the default image got sent to API every single time you drop a custom image in cropper.


amos-cha (approved):
👍🏼

PR #1773, opened 2023-04-14
Update visuals for involvements 'change image' use case

### LINKED ISSUE #1747: Involvements picture scaled too big to see options when trying to change the image.
As an admin of an involvement, trying to change the image for the involvement works (you can change the image) but the existing image is scaled too big to see anything behind it. 

Image

charlielee144r:
This PR just adds some styles and moves some code around so that 
1. Users can see the text to 'drag and drop or click image to choose image' which was previously covered up by the existing involvements photo (see #1747 example).
2. Users can no longer see the text asking to select an image when there is actually no functionality to do so (see example below).
Example:
Before:
image
After:
image

Fixes #1747

### DIFF
--- a/src/views/InvolvementProfile/InvolvementProfile.module.scss
+++ b/src/views/InvolvementProfile/InvolvementProfile.module.scss
@@ -7,3 +7,7 @@
     cursor: pointer;
   }
 }
+
+.update_image_components {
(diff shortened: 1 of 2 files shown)


bennettforkner (approved):
LGTM!

PR #1774, opened 2023-04-14
Add a Mailto link in the search results page

### LINKED ISSUE #1194: PeopleSearch: Mailto Link in Results View
It might be helpful for the email address in the search results list to be clickable (that is, a "mailto:" link).  This would save clicking the line and then clicking the email address in the profile page.

charlielee144r:
Implements a use case where you can easily start emailing someone from the people search results list by adding a mailing icon/button for each result listed. 
This is only implemented in Desktop view for now because of the limited space for such a use case on a mobile view.

Fixes #1194

### DIFF
--- a/src/views/PeopleSearch/components/PeopleSearchResult/PeopleSearchResult.module.scss
+++ b/src/views/PeopleSearch/components/PeopleSearchResult/PeopleSearchResult.module.scss
@@ -38,4 +38,6 @@
 
 .mailingIconContainer {
   flex: 1;
+  height: 100px;
+  justify-content: center;
 }
--- a/src/views/PeopleSearch/components/PeopleSearchResult/PeopleSearchResult.module.scss
+++ b/src/views/PeopleSearch/components/PeopleSearchResult/PeopleSearchResult.module.scss
@@ -1,6 +1,6 @@
 @import '../../../../vars';
 
-.resultContainer {
+.result_container {
   display: flex;
   flex-direction: row;
   align-items: center;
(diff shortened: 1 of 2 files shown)


charlielee144r:
Here is what the current styling looks like:
image

bennettforkner:
Hi @charlielee144r , looking good so far! On the design side, I would recommend giving a little bit more left & right padding to the button as well as decreasing the border-radius.

charlielee144r:
Current status of screen with all edits:
image

src/views/PeopleSearch/components/PeopleSearchResult/index.tsx
```diff
@@ -81,42 +88,72 @@ const PeopleSearchResult = ({ person, lazyLoadAvatar }: Props) => {
...
+              color: gordonColors.neutral.grayShades[50],
+              height: '100%',
+              width: 40,
+              borderRadius: 2,
+            }}
+          />
+        </a>
+      </CardActionArea>
+    </div>
+  );
```
bennettforkner, line 108: You can actually set this to null if it is mobile view. A nice shortcut for specifying "only show this thing if this other thing is true" is:
```suggestion
  const emailIcon = !isMobileView && (
    <div className={styles.mailingIconContainer}>
      <CardActionArea className={styles.mailCardAction}>
        <a href={`mailto:${person.Email}`}>
          <MailOutlineIcon
            sx={{
              color: gordonColors.neutral.grayShades[50],
              height: '100%',
              width: 40,
              borderRadius: 2,
            }}
          />
        </a>
      </CardActionArea>
    </div>
  );
```

src/views/PeopleSearch/components/PeopleSearchResult/PeopleSearchResult.module.scss
```diff
@@ -1,5 +1,11 @@
 @import '../../../../vars';
 
+.resultContainer {
```
bennettforkner, line 3:
```suggestion
.result_container {
```
Also here (see below comment)

bennettforkner (changes requested):
Good job on this Charlie, it's a great feature to add here.

I have a few small code quality suggestions in my comments.

bennettforkner (approved):
LGTM!

PR #1776, opened 2023-04-17
[RecIM] Sports!

silasw2:
This PR allows for the creation of sports on the admin page, with the following layout.
image

Here is the corresponding form that appears when the "add a sport" button is clicked
image

A rule book now appears on the top right corner of the match page, with a clickable button to bring up a rule tooltip,
image

Let me know if it looks good! The only thing I can mention, is that I could see the rules button opening a card with the rules if they were expected to be long. But I think that the tooltip will work for RecIM's purposes.

### DIFF
--- a/src/services/recim/sport.ts
+++ b/src/services/recim/sport.ts
@@ -35,6 +35,6 @@ const getSportByID = async (ID: number): Promise<Sport> => http.get(`recim/sport
 const editSport = async (ID: number, updatedSport: PatchSport): Promise<Sport> =>
   http.patch(`recim/series/${ID}`, updatedSport);
 
-const deleteSport = async (ID: number) => console.log('temp');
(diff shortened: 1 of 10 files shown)


src/services/recim/sport.ts
```diff
@@ -33,6 +33,8 @@ const getAllSports = async (): Promise<Sport[]> => http.get(`recim/sports`);
 const getSportByID = async (ID: number): Promise<Sport> => http.get(`recim/sports/${ID}`);
 
 const editSport = async (ID: number, updatedSport: PatchSport): Promise<Sport> =>
-  http.patch(`recim/series/${ID}`, updatedSport);
+  http.patch(`recim/sports/${ID}`, updatedSport);
```
amos-cha, line 36: whoops thanks

amos-cha (approved):
Fantastic work!

PR #1777, opened 2023-04-17
[Rec-IM] match tabs

amos-cha:
desktop:
image
We can now display views with more than 300 matches reasonably.

mobile:
image

Also fixes Schedule Edit bug.


src/views/RecIM/components/List/index.jsx
```diff
@@ -66,15 +69,84 @@ const ParticipantList = ({
 };
 
 const MatchList = ({ matches, activityID }) => {
+  const [selectedDay, setSelectedDay] = useState(0);
+  useEffect(() => {
```
silasw2, line 73: More for curiosity here, but do we have to u?e a useEffect? Also, what happens if the index is -1? Wouldn't we want it to go to the next available day with matches instead of just the first tab?

silasw2 (dismissed):
I think looks good, left a comment that might need addressing if you want to take a look.

src/views/RecIM/components/List/index.jsx
amos-cha: Hmmm, this is a good point. The only thing about that is that I struggle with thinking of an implementation that would be able to handle all edge cases without doing a 7 day for-loop to check to make sure there aren't future matches. I could look into it.

This is definitely an implementation I would want and within the scope of this PR. I'll try to tackle it today and ask for re-review!

amos-cha:
> This is fantastic, I'm stoked for this update!
> 
> I realized one thing that should be changed I think: if there is only one day, can we not show the tabs?

I accidentally handled that change in https://github.com/gordon-cs/gordon-360-ui/pull/1779

src/views/RecIM/components/List/index.jsx
amos-cha: handled

cpabbot (approved):
LGTM

PR #1779, opened 2023-04-18
[Rec-IM] teamlisting update

amos-cha:
image

Main updates
--
1) listings are ordered by multiple tiebreakers (Wins -> Reverse loss (to prevent 0:0 being below 0:1) -> Sportsmanship)
2) Tie condition is fixed in new API
3) Sportsmanship Rating is now returned within Team Record to be used in listings


cpabbot (commented):
This looks great, I'll approve once the base branch is merged and this base branch has been updated accordingly.

One note is I'm wondering if it would look better to show records in a format like "14-3" or "14-2-1" (in the case of a tie), which is the standard way I've seen records shown. This format has the advantage of being more explicit though about wins/losses/ties so idk which is better. We could also always just survey opinions and consider that tweak for the future.

amos-cha:
> This looks great, I'll approve once the base branch is merged and this base branch has been updated accordingly.
> 
> One note is I'm wondering if it would look better to show records in a format like "14-3" or "14-2-1" (in the case of a tie), which is the standard way I've seen records shown. This format has the advantage of being more explicit though about wins/losses/ties so idk which is better. We could also always just survey opinions and consider that tweak for the future.

I agree, I think we should just ask admins. I'm basing my logic off Soccer which has WTL

src/services/recim/team.ts
```diff
@@ -50,12 +50,14 @@ export type TeamMatchHistory = {
 };
 
 export type TeamRecord = {
-  ID: number;
+  SeriesID: number;
```
amos-cha, line 53: this was originally just straight up wrong. I don't know how I didn't catch it. I suppose it's because we don't use the seriesID for anything right now

src/views/RecIM/views/Activity/index.jsx
```diff
@@ -385,10 +385,10 @@ const Activity = () => {
         ) : (
           <Grid container justifyContent="center" spacing={2}>
             <Grid item container justifyContent="center" spacing={2}>
-              <Grid item xs={12} md={7}>
+              <Grid item xs={12} md={6.5}>
                 {scheduleCard}
               </Grid>
-              <Grid item direction={'column'} xs={12} md={5}>
+              <Grid item direction={'column'} xs={12} md={5.5}>
```
amos-cha, line 391: sizing had to be re-adjusted to fit the image

cpabbot (approved):
LGTM! Love the images and thanks for adding sportsmanship!

Should wait to merge until after `recim-match-tabs` has been merged.

cpabbot (approved):

PR #1780, opened 2023-04-18
[Rec-IM] Create & Edit custom participant in Admin page

Kevin-Chu58:
This pr adds a new form that allows creating custom participants, and updating their info in the Admin page.
Also improves the loading speed of Admin page by replacing `GetProfileInfo()` with a lighter method.

API pr: https://github.com/gordon-cs/gordon-360-api/pull/891.

New `Create Participant` Button
image

`Create Participant` View
image

Only `Custom Participant` has the `MoreOption` Button at the righthand side
And they don't have a pic for now
image

Menu displayed when clicked the `MoreOption` Button
image

`Edit Participant` View
image

Quick Search in RecIM team page
* Admins can search both Gordon participants and non-Gordon participants
image
image

**Participant List Checklist**
* let me know if I missed any
- [x] Admin page - participant list
- [x] Team page - roster
- [x] Team page - invite list
- [x] Match page - attendance list

### DIFF
--- a/src/services/recim/participant.ts
+++ b/src/services/recim/participant.ts
@@ -2,26 +2,30 @@ import http from '../http';
 import { Team } from './team';
 import { Lookup } from './recim';
 
-export type Participant = {
+export type Participant = CustomParticipant & {
   Username: string;
-  Email: string;
   Role: string;
-  AllowEmails: boolean;
-  SpecifiedGender: string;
   GamesAttended: number;
   Status: string;
   Notification: ParticipantNotification[];
   IsAdmin: boolean;
   IsCustom: boolean;
-  FirstName: string;
-  LastName: string;
 };
 
 type CustomParticipant = {
   AllowEmails: boolean;
+  Email: string;
   SpecifiedGender: string;
   FirstName: string;
   LastName: string;
+};
+
+type BasicInfo = {
+  FirstName: string;
+  LastName: string;
+  UserName: string;
+  Nickname: string;
+  MaidenName: string;
 };
 
 type PatchParticipantActivity = {
@@ -103,6 +107,16 @@ const getParticipantStatusTypes = (): Promise<Lookup[]> =>
 const getParticipantActivityPrivTypes = (): Promise<Lookup[]> =>
   http.get(`recim/participants/lookup?type=activitypriv`);
 
+const getAccountsBasicInfo = async (
+  query: string,
+): Promise<[searchTime: number, searchResults: BasicInfo[]]> => {
+  const searchStartTime = Date.now();
+  // Replace period or space with a slash: 'first.last' or 'first last' become 'first/last'
+  const searchQuery = query.toLowerCase().trim().replace(/\.|\s/g, '/');
+  const searchResults: BasicInfo[] = await http.get(`recim/participants/search/${searchQuery}`);
+  return [searchStartTime, searchResults];
+};
+
 const sendNotification = (
   username: string,
   notification: UploadParticipantNotification,
@@ -141,6 +155,7 @@ export {
   getParticipantStatusHistory,
   getParticipantStatusTypes,
   getParticipantActivityPrivTypes,
+  getAccountsBasicInfo,
   sendNotification,
   editCustomParticipant,
   editParticipantAdmin,
--- a/src/services/recim/participant.ts
+++ b/src/services/recim/participant.ts
@@ -6,10 +6,22 @@ export type Participant = {
   Username: string;
   Email: string;
   Role: string;
+  AllowEmails: boolean;
+  SpecifiedGender: string;
   GamesAttended: number;
   Status: string;
   Notification: ParticipantNotification[];
   IsAdmin: boolean;
+  IsCustom: boolean;
+  FirstName: string;
+  LastName: string;
+};
+
+type CustomParticipant = {
+  AllowEmails: boolean;
+  SpecifiedGender: string;
+  FirstName: string;
+  LastName: string;
 };
 
 type PatchParticipantActivity = {
@@ -69,6 +81,11 @@ type CreatedParticipantActivity = {
 const createParticipant = (username: string): Promise<Participant> =>
   http.put(`recim/participants/${username}`);
 
+const createCustomParticipant = (
+  username: string,
+  newcustomParticipant: CustomParticipant,
+): Promise<Participant> => http.put(`recim/participants/${username}/custom`, newcustomParticipant);
+
 const getParticipants = (): Promise<Participant[]> => http.get(`recim/participants`);
 
 const getParticipantByUsername = (username: string): Promise<Participant> =>
@@ -92,6 +109,12 @@ const sendNotification = (
 ): Promise<CreatedParticipantNotification> =>
   http.post(`participants/${username}/notifications`, notification);
 
+const editCustomParticipant = (
+  username: string,
+  updatedCustomParticipant: CustomParticipant,
+): Promise<Participant> =>
+  http.patch(`recim/participants/${username}/custom/update`, updatedCustomParticipant);
+
 const editParticipantAdmin = (username: string, isAdmin: boolean): Promise<Participant> =>
   http.patch(`recim/participants/${username}/admin`, isAdmin);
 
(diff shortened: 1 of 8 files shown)


src/services/recim/participant.ts
```diff
@@ -6,10 +6,22 @@ export type Participant = {
...
+  FirstName: string;
+  LastName: string;
+};
+
+type CustomParticipant = {
+  AllowEmails: boolean;
+  SpecifiedGender: string;
+  FirstName: string;
+  LastName: string;
 };
```
amos-cha, line 25: Take a look at what we do for `activity`.ts, we can cut down on some of these repeated pieces of data. Also for the record @Kevin-Chu58's API PR MAKES a username that ensures no duplicates. However, I can see a small chance that there could be overlap if a Gordon User is added i.e:

silas.white (exists)
we add another silas.white -> silas.white2
Gordon adds another silas.white -> silas.white2 (conflict)

src/views/RecIM/components/Forms/ParticipantForm/index.jsx
```diff
@@ -0,0 +1,122 @@
...
+        LastName: participant.LastName,
+        SpecifiedGender: participant.SpecifiedGender,
+        AllowEmails: participant.AllowEmails,
+      };
+    }
+    return {
+      FirstName: '',
+      LastName: '',
+      SpecifiedGender: 'U',
+      AllowEmails: true,
```
amos-cha, line 59: We have an "AllowEmails" but we don't have a way of storing emails of non-gordon participants. I think we either need another column for emails OR you need to set it always to false.

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -241,34 +243,42 @@ const ParticipantListing = ({
...
-        setAvatar(preferredImage || defaultImage);
+      if (!participant.IsCustom) {
+        if (participant.Username) {
+          const { def: defaultImage, pref: preferredImage } = await user.getImage(
+            participant.Username,
+          );
+          setAvatar(preferredImage || defaultImage);
+        }
+      } else {
+        // TODO - Depdent on how we want to store their Pics
```
amos-cha, line 273:
```suggestion
        // TODO - Depends on how we want to store their Pics
```

amos-cha (changes requested):
Small questions I need to have addressed. Besides that it looks good.

amos-cha:
> Should we somehow change the wording to perhaps say "add a non-Gordon participant" or in some way clarify that this is only for making custom participants, not for adding someone to Rec-IM generally?

This isn't COMPLETELY necessary. If we were to fill in the exact data aka, `username`, `firstname`, `lastname` of a gordon student. It would still work perfectly fine since the only "gordon" feature is that we link to Gordon profiles

src/services/recim/participant.ts
Kevin-Chu58: The autogenerated username algorithm is now fixed. We will always add a suffix `.custom` after each non-Gordon participant username.

src/views/RecIM/components/QuickSearch/index.jsx
```diff
@@ -0,0 +1,294 @@
...
+  );
+};
+
+GordonQuickSearch.propTypes = {
+  customPlaceholderText: PropTypes.string,
+  disableLink: PropTypes.any,
+  onSearchSubmit: PropTypes.func,
+};
+
+export default GordonQuickSearch;
```
amos-cha, line 294: hmmm i don't know if we should be ripping out the entire gordon search? I can't think of another solution off the top of my head but it feels a little sketchy?

Benefits are obviously that Rec-IM stays modular but this is definitely a repeated component?
thoughts @silasw2  @cpabbot

amos-cha (commented):
I think we need a couple levels of approval for this. I want to keep Rec-IM modular but it feels weird to be ripping/copying exact components/functions from the rest of 360?

src/views/RecIM/components/QuickSearch/index.jsx
Kevin-Chu58: True. Most part is the same, except we also need to return `FirstName` and `LastName` otherwise the invite list shows `undefined undefined` for all people.
`    if (key === 'Enter' && suggestionList && suggestionList.length > 0) {
      if (sIndex === -1) {
        theChosenOne = suggestionList[0].UserName;
        firstName = suggestionList[0].firstName;
        lastName = suggestionList[0].lastName;
      } else {
        theChosenOne = suggestionList[sIndex].UserName;
        firstName = suggestionList[sIndex].firstName;
        lastName = suggestionList[sIndex].lastName;
      }
      // If prop set to disable link, then trigger the onSearchSubmit callback function
      // Else, redirect the user to the selected profile page
      disableLink
        ? onSearchSubmit(theChosenOne, firstName, lastName)`

amos-cha: Right, I agree that we have a unique piece of the component, but it is, in essence, identical. I'm just wondering if we should handle it this way or if we should modify the actual GordonQuickSearch which would interlink 360 and Rec-IM further

Kevin-Chu58:
> This is looking great! One other quick note: can you change the gender dropdown to have full options "male, female, other/unspecified" (idk what to use for third but one of those maybe?) rather than just the letters?

image
@cpabbot Updated!

amos-cha:
> > This is looking great! One other quick note: can you change the gender dropdown to have full options "male, female, other/unspecified" (idk what to use for third but one of those maybe?) rather than just the letters?
> 
> image @cpabbot Updated!

sharelock 🧑‍🍳

amos-cha (dismissed):
I'm willing to approve, but wait for @silasw2 or @cpabbot to also approve the change.

Kevin-Chu58:
> > > This is looking great! One other quick note: can you change the gender dropdown to have full options "male, female, other/unspecified" (idk what to use for third but one of those maybe?) rather than just the letters?
> > 
> > 
> > image @cpabbot Updated!
> 
> sharelock 🧑‍🍳

I am making a pun using the share lock and exclusive lock used to maintain the ACID of database, a daddy joke :)

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -378,21 +404,38 @@ const ParticipantListing = ({
...
+            <ListItemAvatar>
+              <Avatar
+                src={`data:image/jpg;base64,${avatar}`}
+                className={minimal ? styles.avatarSmall : styles.avatar}
+                variant="rounded"
+              ></Avatar>
+            </ListItemAvatar>
+            <ListItemText primary={name} secondary={participant.Role} />
+          </ListItemButton>
+        )}
```
amos-cha, line 438: This chunk of code seems odd. Why does a custom participant not need the `<ListItemButton>`. Also, if the nesting is required, can you just make a helper function that wraps the sub component? It's the exact same piece of code.

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -412,6 +455,32 @@ const ParticipantListing = ({
             )}
           </Menu>
         )}
+        {isAdminPage && participant.IsCustom && (
+          <Menu
+            open={moreOptionsCustomOpen}
+            onClose={handleClickOff}
+            anchorEl={anchorCustomEl}
```
amos-cha, line 462: can we rename
```
moreOptionsCustomOpen
anchorCustomEl
```
to something else? Those variable names are not very descriptive. moreParticipantOptions? participantAnchorEl?

src/views/RecIM/components/List/Listing/index.jsx
```diff
@@ -412,6 +455,32 @@ const ParticipantListing = ({
...
+            }}
+            transformOrigin={{
+              vertical: 'top',
+              horizontal: 'right',
+            }}
+          >
+            <MenuItem
+              dense
+              onClick={() => {
+                editDetails(participant);
```
amos-cha, line 475:
```suggestion
                editDetails(participant);
```
editDetails of what? descriptive?

amos-cha (changes requested):
Few changes to be addressed

amos-cha (approved):
Forced to approve. API now fails without new UI

PR #1781, opened 2023-04-19
[Rec-IM] finally fixed invite redirect when reject

Kevin-Chu58:
Though this pr has nothing to do with the branch name `gotoButton`, it does resolve the redirection when click the reject team invite button, which has been tested and worked properly.


amos-cha (approved):
👍🏼

PR #1783, opened 2023-04-25
Cocurricular transcript styling adjustments

sofistonehouse:


sofistonehouse:
Not styled correctly on mobile but I will continue to work on that

src/views/CoCurricularTranscript/Components/Activity/index.tsx
```diff
@@ -12,14 +12,28 @@ type Props = {
...
-        <div className={styles.leadership_line}>
-          <div className={styles.organization_role}>Leader</div>
-          <div className={styles.date}>{formatDuration(leaderSessions)}</div>
+        <div className={styles.participation_line}>
+          <div className={styles.organization_role}>Leader ({formatDuration(leaderSessions)})</div>
+        </div>
+      )}
+      {memberSessions.length > 0 && (
+        <div className={styles.participation_line}>
+          <div className={styles.organization_role}>Member ({formatDuration(memberSessions)})</div>
```
bennettforkner, line 36: For consistency and readability, it's probably best to format these the same as `Advisor` above.
```suggestion
          <div className={styles.organization_role}>
            Leader ({formatDuration(leaderSessions)})
          </div>
        </div>
      )}
      {memberSessions.length > 0 && (
        <div className={styles.participation_line}>
          <div className={styles.organization_role}>
            Member ({formatDuration(memberSessions)})
          </div>
```

src/views/CoCurricularTranscript/Components/Activity/index.tsx
```diff
@@ -66,3 +80,24 @@ const formatDuration = (sessionRecords: MembershipHistorySession[]) => {
...
+      // If this is the first session, initialize interval.
+      interval = new MembershipInterval(session);
+    } else {
+      // Extend the interval.
+      interval.extendTo(session);
+    }
+  });
+
+  return interval?.toString();
+};
```
bennettforkner, line 103: This is a lot of repeated code for doing something only slightly different from `formatDuration(sessions)`.

Perhaps you could eliminate `formatFinalDuration(...)` and update the `formatDuration(sessions)` function to the following:

```
const formatDuration = (sessionRecords: MembershipHistorySession[], summarizeDates: bool = false) => {
...
    } else if (interval.consecutiveWith(session) || summarizeDates) {
...
  return intervalDescriptions.join(', ');
};
```

Then you could just call `formatDuration(sessions, true)` instead of `formatFinalDuration(...)`

bennettforkner (changes requested):

PR #1784, opened 2023-04-26
[Rec-IM] email options

amos-cha:
image

This branch is based off https://github.com/gordon-cs/gordon-360-ui/pull/1765, please merge that before merging this.


src/services/recim/activity.ts
```diff
@@ -59,7 +59,7 @@ const getActivityByID = (ID: number): Promise<Activity> => http.get(`recim/activ
 
 const getActivities = (active: boolean, time: String): Promise<Activity[]> => {
   if (time) return http.get(`recim/activities?active=${active}&time=${time}`);
-  if (active) return http.get(`recim/activities?active=${active}`);
+  if (typeof active !== 'undefined') return http.get(`recim/activities?active=${active}`);
```
silasw2, line 62: This creates a lot of confusion in this function. Active is a boolean operator, so with this current implementation, if active is false, it will only get the active activities.

src/services/recim/match.ts
```diff
@@ -123,6 +123,13 @@ const updateMatch = (ID: number, updatedMatch: PatchMatch): Promise<CreatedMatch
 const deleteMatchCascade = async (matchID: number): Promise<CreatedMatch> =>
   http.del(`recim/matches/${matchID}`);
 
+// shortcut service to delete a list of matches
+const deleteMatchList = async (matches: Match[]) => {
```
silasw2, line 127: Truthfully, I don't understand why we would have this. Can't we just loop in whatever function we want to delete the match with?

src/views/RecIM/components/Header/index.jsx
```diff
@@ -56,10 +56,10 @@ const Header = ({ match, team, activity, admin, children }) => {
           {match && (
             <RecIMBreadcrumb>
               {largeWidth
-                ? `Match: ${match?.Team[0]?.Name ?? 'TBD'} vs 
-              ${match?.Team[1]?.Name ?? 'TBD'}`
-                : `Match: ${truncate(match?.Team[0]?.Name ?? 'TBD')} vs ${truncate(
-                    match?.Team[1]?.Name ?? 'TBD',
+                ? `Match: ${match.Team?.[0]?.Name ?? 'TBD'} vs
```
silasw2, line 59: Does doing Match.Team.[0] work as an index?

src/views/RecIM/views/Admin/index.jsx
```diff
@@ -14,16 +14,16 @@ import {
   SurfaceList,
   SportList,
 } from '../../components/List';
-import { getActivities } from '../../../../services/recim/activity';
-import { getTeams } from '../../../../services/recim/team';
-import { getParticipants } from '../../../../services/recim/participant';
-import { deleteSurface, getSurfaces } from '../../../../services/recim/match';
+import { getActivities } from 'services/recim/activity';
```
silasw2, line 17: Oh nice! I really have no clue how that got in here in the first place.

silasw2 (changes requested):
Some things need to be addressed, but overall pretty good.

src/services/recim/activity.ts
amos-cha: this was not my implementation. This is from @cpabbot's PR.

src/services/recim/match.ts
amos-cha: It's simply a shortcut for a loop I can foresee being used in other place in the future:
- delete all activity matches
- admin page delete matches

This is definitely future proofing so potentially I could remove it but I can already see myself adding it back in within the next month or two.

src/views/RecIM/components/Header/index.jsx
amos-cha: not my implementation

amos-cha:
I'm closing this PR because for whatever reason it doesn't recognize the `unique` changes and has become hard to decipher what changes are mine and what are from the branch I created this PR from.

Issue #1790, opened 2023-05-05
Cleanup remaining uses of deprecated @mui/styles package


EjPlatzer:
When we upgraded to MUIv5, we added the `@mui/styles` package which allowed us to keep using the old, MUIv4 styling API so that we didn't have to change the styling of many components. However, that package was intended as a transitional tool, and is not being maintained in the long term.

Please find all references to the `@mui/styles` package and update each component to use the same styling solution as the rest of the project.

Below are all the references in every index.jsx file:
- [x] src/components/GordonTooltip/index.jsx
- [x] src/components/Header/components/NavAvatarRightCorner/index.jsx
- [x] src/components/SchedulePanel/index.jsx
- [x] src/components/PWAInstructions/index.jsx
- [x] src/views/InvolvementProfile/components/Membership/components/AdminCard/components/RequestsRecieved/index.jsx
- [x] src/views/Timesheets/index.jsx
- [x] src/views/Timesheets/components/ShiftItem/index.jsx

Centerville1:
Looked through the code for references, no fixes yet.

Below are all the references I found:
- src/components/GordonTooltip/index.jsx
- src/components/Header/components/NavAvatarRightCorner/index.jsx
- src/components/SchedulePanel/index.jsx
- src/compontnet/PWAInstructions/index.jsx
- src/views/InvolvementProfile/components/Membership/components/AdminCard/components/RequestsRecieved/index.jsx
- src/views/Timesheets/index.jsx
- src/views/Timesheetscomponents/ShiftItem/index.jsx

bennettforkner:
@Centerville1, I added your references to the original description above with checkboxes so that we can track which have been fixed.

andrew-wzj:
We mainly completed on cleaning up deprecated `@mui/styles`,but we couldn't really test `RequestsRecieved/index.jsx` is working.

Issue #1789, opened 2023-05-05
Update Chart.js


EjPlatzer:
Chart.js is on version 4, but we are still using version 2. Someone ought to work through the migration guides posted on the chart.js website for [Version 2 to Version 3](https://www.chartjs.org/docs/latest/migration/v3-migration.html) and [Version 3 to Version 4](https://www.chartjs.org/docs/latest/migration/v4-migration.html).

Refer also to the `react-chartjs-2` package that wraps chart.js for use in React, which has it's own migration guides: https://react-chartjs-2.js.org/docs/migration-to-v4

amos-cha:
https://github.com/gordon-cs/gordon-360-ui/pull/2297

PR #1791, opened 2023-05-05
Update packages

EjPlatzer:
I have updated almost all our package dependencies. Each package that had breaking changes has been reviewed and tested.

The following packages have not been updated to the latest major version:
1. `react` and `react-dom` are still at version 17, because `@mui/styles` hasn't been updated to support react 18 as a peer dependency. The `@mui/styles` package is intended as a temporary tool for transitioning from the MUIv4 JSS styling solution to the new CSS-in-JS styling solutions in MUIv5, and as such the package is not actively maintained. We should make an effort to refactor the ~5 components that still use `@mui/styles` for styling, but that is beyond the scope of this PR. I have created #1790 to track this issue.
2. `chart.js` and `react-chartjs-2` are stuck at version 2, but chart.js is up to version 4. The breaking changes between these versions are extensive and in-depth, so I decided they belong as a separate pull request. I have created #1789 to track this issue.


EjPlatzer:
I will keep this as a draft until we have the Rec-IM features stabilized into production.

bennettforkner (approved):
LGTM

bennettforkner (approved):
LGTM (still)!

PR #1793, opened 2023-05-10
Update Google Analytics to V4

bennettforkner:
This PR updates the version of our Google Analytics package to work with their new version 4 which will replace Universal Analytics on July 1st.

The changes are pretty minimal and simple but we needed a new tag code to use on the site as well. Most of the new features are in the analytics dashboard itself.


EjPlatzer (approved):
LGTM!

PR #1794, opened 2023-05-16
Revert to login via redirect

EjPlatzer:
I changed the fallback for acquiring a token to use the pop up method in #1792. However, that has caused some issues in Train that I didn't notice locally. Logging in via pop up does not work on browsers unless the Pop up privilege has been explicitly granted to the site. Redirect is therefore more reliable.


bennettforkner (approved):
LGTM!

PR #1796, opened 2023-05-19
Improve error logging for ID Uploader

EjPlatzer:
The error logging within the ID Uploader service depends on deprecated navigator APIs (platform, appVersion, vendor) - as well as the `navigator.userAgent` that is gradually being deprecated and quickly becoming less than informative - for useful error logging. This is causing problems because the data extracted from these old APIs is now appearing in logs as repeated "unknown"s, which is not very helpful.

Chromium-based browsers (which hold >70% market share of browsers) already support the (currently experimental) User Agent Data interface that is planned to take the place of these old interfaces<sup>1</sup>. As such, I have updated the logging logic to take from the User Agent Data API if it exists in the browser, and to fallback to the old user agent otherwise.

I also updated the `postCroppedImage` method of the ID Uploader to use the new `logging.parseUserAgentData` method. In the process, I refactored it slightly to be more readable and performant.

<sup>1</sup> - https://caniuse.com/mdn-api_navigator_useragentdata

### DIFF
--- a/src/services/logging.ts
+++ b/src/services/logging.ts
@@ -16,7 +16,7 @@ declare global {
   }
 }
 
-const postLogMessage = (message: string) => http.post('log', message);
+const post = (message: string) => http.post('log', message);
 
 // @TODO: Make this a strcutured component of all logs, e.g. with new columns for OS, Version, etc.
 const parseUserAgentData = () => {
@@ -29,7 +29,7 @@ const parseUserAgentData = () => {
 };
 
 const errorLogService = {
-  postLogMessage,
+  post,
   parseUserAgentData,
 };
 
--- a/src/services/logging.ts
+++ b/src/services/logging.ts
@@ -16,7 +16,10 @@ declare global {
   }
 }
 
-const post = (message: string) => http.post('log', message);
+const post = (message: string) => {
+  const userAgentData = parseUserAgentData();
+  http.post('log', `${userAgentData}: ${message}`);
+};
 
 // @TODO: Make this a strcutured component of all logs, e.g. with new columns for OS, Version, etc.
 const parseUserAgentData = () => {
--- a/src/services/logging.ts
+++ b/src/services/logging.ts
@@ -16,8 +16,9 @@ declare global {
   }
 }
 
-const postErrorMessage = (message: string) => http.post('log', message);
+const postLogMessage = (message: string) => http.post('log', message);
 
+// @TODO: Make this a strcutured component of all logs, e.g. with new columns for OS, Version, etc.
 const parseUserAgentData = () => {
   if (navigator.userAgentData) {
     const { brands, platform, mobile } = navigator.userAgentData;
@@ -28,7 +29,7 @@ const parseUserAgentData = () => {
 };
 
 const errorLogService = {
-  postErrorMessage,
+  postLogMessage,
   parseUserAgentData,
 };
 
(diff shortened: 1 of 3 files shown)


src/services/logging.ts
bennettforkner, line 1: I might call this the `log` service, but not a big concern

src/services/logging.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// @TODO: Make this a strcutured component of all logs, e.g. with new columns for OS, Version, etc.
+const parseUserAgentData = () => {
+  if (navigator.userAgentData) {
+    const { brands, platform, mobile } = navigator.userAgentData;
+    const { brand, version } = brands[0];
+    return `${platform} running ${brand} version ${version} on ${mobile ? 'mobile' : 'non-mobile'}`;
+  }
+  return navigator.userAgent;
+};
```
bennettforkner, line 29: I think it would make sense for user agent data to be tracked for every log entry, do you think?

If so, I would rather have this be called inside of the `post` function rather than before it.

src/views/IDUploader/index.jsx
```diff
@@ -74,23 +74,25 @@ class IDUploader extends Component {
...
-      this.setState({ submitDialogOpen: true });
-    } catch (error) {
-      logMessage += `, but image failed to post with error: ${error}`;
-      if (attemptNumber < 5) {
-        this.postCroppedImage(croppedImage, attemptNumber + 1);
-      } else {
-        this.setState({ errorDialogOpen: true });
+  async postCroppedImage(croppedImage) {
+    const profile = await user.getProfileInfo();
+    const userAgentData = logging.parseUserAgentData();
```
bennettforkner, line 79: This variable name does not fully describe what the function returns. I think this might be slightly clearer.
```suggestion
    const userAgentString = logging.parseUserAgentData();
```

src/views/IDUploader/index.jsx
```diff
@@ -74,23 +74,25 @@ class IDUploader extends Component {
...
+
+    while (!postedSuccessfully && attemptNumber < 5) {
+      try {
+        await user.postIDImage(croppedImage);
+        this.setState({ submitDialogOpen: true });
+        logging.post(logIntro + ` succeeded on #${attemptNumber}`);
+        postedSuccessfully = true;
+      } catch (error) {
+        const errorDetails = JSON.stringify(error);
+        logging.post(logIntro + ` failed #${attemptNumber} with error: ${errorDetails}`);
```
bennettforkner, line 92:
```suggestion
        logging.post(logIntro + ` failed on #${attemptNumber} with error: ${errorDetails}`);
```

bennettforkner (changes requested):

src/services/logging.ts
EjPlatzer: I agree that it would be great to get it inside all logs (see my comment). I suppose we could just put this at the front of every log message, which is a good enough solution for now. The advantage of using the new API is that this info can't really identify a unique user, so we don't necessarily need to worry about tracking concerns. 

I'll make that change.

src/views/IDUploader/index.jsx
EjPlatzer: Hmm, to me the purpose of the variable name is to describe what the variable represents, not what kind of data it is - that's what we have typing and intellisense for. I'm not sure why `userAgentString` is clearer or more descriptive then `userAgentData`, and in fact the former seems somewhat misleading because it implies that we're talking about the `navigator.userAgent` string (as that's what most people mean when they say user agent string), which ideally we are not.

EjPlatzer:
@bennettforkner I am going to merge this, so that we can hopefully accrue some helpful log info over the weekend. We can always clean up this implementation in a future pull request.

bennettforkner:
Yea no problem. No breaking issues or anything, just some suggested cleanup.

PR #1799, opened 2023-05-25
Use ReactGA properly

EjPlatzer:
When we updated to the `react-ga4` package, someone updated an import without actually checking that the import was still valid. It was not still valid, and it caused unauthenticated users to see an error page on the home screen, breaking the site.

I have updated the import to use the new syntax from the new package.


bennettforkner:
"someone" lol. I'm sorry @EjPlatzer .

Although I think I just missed changing that import to ReactGA (I didn't update the import)

PR #1802, opened 2023-05-30
Add instructions to use ssh when cloning repository.

jsenning:
It has worked well for developers to clone this repository using SSH rather than HTTPS.  The developer instructions have been modified to direct developers to use this method.  Instructions are provided create a new key (if they don't already have one) and install it on Github.

### DIFF
--- a/docs/Developer's Guide.md
+++ b/docs/Developer's Guide.md
@@ -21,7 +21,16 @@ This guide walks you through the basics of developing the Gordon 360 frontend, a
 
 Here are the steps to setup the frontend for development:
 
-1. Clone this repository to the machine that you will develop on.
+1. Make sure the SSH key for your account is installed on Github.  To check this:
+
+   - Log into [Github](https://github.com).  Click on the user menu in the upper right corner and select "Settings".
+   - Click on "SSH and GPG keys" and determine if the key for the account you are using is already installed.
+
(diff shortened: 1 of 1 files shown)


bennettforkner (approved):
I am fine with this, but we might want to document both options that GitHub offers if we are going to document them at all. Visual Studio/VSCode, for example, allows you to clone the repos via HTTP/s through their own interface.

bennettforkner:
We don't have a requirement that our school-year 360 developers use SSH for the git connection, but this documentation would make it seem like we are requiring them to use that method. That's my main concern. If there's a requirement specific to the practicum, I would think that should be documented separately in the practicum resources rather than in 360 dev as a whole.

jsenning:
@bennettforkner I understand and appreciate your comment.  I've changed the doc to be more general, but suggest that summer practicum students use the SSH method.

bennettforkner (approved):

jsenning:
I suppose it's just what we're used to and it's what Russ has his students set up when they learn to use github in their CS classes.  Perhaps it's because we're often working from the command line rather than an IDE like VS code...

bennettforkner:
Oh, good to know. If that's true, it might be more familiar to the student developers as well and we could keep it consistent. I'm up for whatever. Thoughts @EjPlatzer ?

russtuck:
Yes, I teach the git command line first, both because it's a good foundation, and because it's sometimes necessary to fix things up.  Using ssh with a key has been the most trouble-free and secure way to do that (though I see that the github CLI might now offer an equivalently easy and secure option).  I'm glad vscode works so well with both ssh and https.

PR #1805, opened 2023-05-31
Fix Google Analytics Tracking Code

bennettforkner:
The GA4 property does not seem to be tracking 360 traffic properly. I tested it locally and the issue was fixed when I specified the analytics ID in the env file.


EjPlatzer (changes requested):
Are you trying to get the analytics ID to only be available in production? This would add it in Train as well.

russtuck (approved):
LGTM.  We'll need to be careful merging this during the release.

Issue #1808, opened 2023-05-31
Create a transcript for guest in Guest View


ArabellaJi:
This is a sub-issue for issue #580.

Basically, we need to create a transcript for the guest using some mocked up data and add a Transcript Tab or link in the Guest View. See more details in issue #580.

Issue #784, opened 2020-06-19
Navigation "Active" status issues


cpabbot:
Somehow the execution of our "active" tab switching is not very robust; I'm guessing it as something to do with the animation (which should not negatively impact functionality). I originally thought maybe it might also have to do with data loading (if you click something before the page has fully loaded) but I do not think so based on brief testing.

Here are a couple symptom issues so you can see for yourself:
- If you click one of your involvements from your public profile, "involvements" tab should be active but is not.
- If you switch back and forth between tabs several times and use browser 'back' arrows, the active tab often gets stuck. This occurs on all tabs but experientially more often on "involvements" and less often on "events" which I would pay attention for clues investigating root cause.

cpabbot:
Also just clicking back breaks it as well

cpabbot:
Appears to be working now. If someone can confirm that would be great @EjPlatzer, but I am closing this issue now.

cpabbot:
This is still an issue. Will update description with a little more detail.

bennettforkner:
Is this really still an issue @cpabbot @EjPlatzer ? I am not experiencing either of the symptoms. Perhaps this issue was resolved in changes already pushed somewhere?

PR #1817, opened 2023-06-01
Guest Transcript

ArabellaJi:
I've added a Transcript link in the Navigation Menus and finished the header and sub-header of the Guest Transcript. It looks like this now:

image

This PR is for #1808.


ArabellaJi:
Now the Guest Transcript page looks like this:

image

src/views/CoCurricularTranscript/Components/Activity/index.tsx
```diff
@@ -12,6 +12,15 @@ type Props = {
 
 const Activity = ({ description, sessions }: Props) => {
   const leaderSessions = sessions.filter((s) => s.Participation === Participation.Leader);
+  console.log(sessions);
```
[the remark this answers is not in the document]
ArabellaJi: Will do!

src/views/CoCurricularTranscript/Components/Activity/index.tsx
```diff
@@ -12,6 +12,15 @@ type Props = {
...
+  if (sessions[0] == null) {
+    return (
+      <div className={styles.experience_transcript_activities}>
+        <div className={styles.organization_role}>role</div>
+        <div className={styles.date}> date </div>
```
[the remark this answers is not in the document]
ArabellaJi: Sorry I just saw these feedback. Will work on them now.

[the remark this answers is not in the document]
bennettforkner: I think I just commented because I forgot to actually forgot to finish the review. My comments were all `Pending`!

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -36,14 +44,155 @@ const CoCurricularTranscript = () => {
...
+      MembershipID: 6547,
+      SessionCode: '202306',
+      Participation: participation1,
+    };
+    const activity1: MembershipHistory = {
+      ActivityCode: '360',
+      ActivityDescription: '360.gordon.edu',
+      ActivityImagePath: '',
+      Sessions: [session1],
+      LatestDate: '2023-06-07T00:00:00',
```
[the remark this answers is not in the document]
ArabellaJi: I would love to, but can we really put a relative date rather than a specific date and time since the API should bring back the specific date and time for the web to display?

src/views/CoCurricularTranscript/Components/Activity/index.tsx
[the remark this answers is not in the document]
ArabellaJi: hahaha

ArabellaJi:
I spoke with Dr. Tucker @russtuck and he came up with an interesting idea. Since we have already created transcripts for our guests, why not give them a complete profile like this:

image

How do you think? @bennettforkner @EjPlatzer @ChrisImagineer

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -36,14 +44,155 @@ const CoCurricularTranscript = () => {
...
+  if (loadingProfile) {
     return <GordonLoader />;
   }
 
   if (!profile) {
-    return <GordonUnauthorized feature={'your experience transcript'} />;
+    const participation1: Participation = Participation.Member;
+    const session1: MembershipHistorySession = {
+      MembershipID: 6547,
+      SessionCode: '202306',
```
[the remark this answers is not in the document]
bennettforkner: Yes, but that won't look good in a few years since it will be old data, we'll need task to maintain it. I would recommend connecting it to the current term rather than a specific one.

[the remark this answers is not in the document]
ArabellaJi: Oh good idea. Will do!

EjPlatzer:
> I am concerned with this idea because incoming/new students do in fact have some valid data that we want to show them on their profile. The first thing they do in 360 is upload their ID photo, so they would be able to have everything that other people have on their profile except the victory promise and involvements. At that point, I think it would be confusing to have example information for the victory promise/involvements since it would be mixed in with real data.

This wouldn't apply to new students who are authenticated, only unauthenticated guests I believe.

bennettforkner:
> This wouldn't apply to new students who are authenticated, only unauthenticated guests I believe.

Oh good point. So what is the use case for this then? Prospective student looking deep into our systems? I'm not sure how valuable it would be to have their own profile example. I would think most people would just prefer to view other peoples' real profiles as an example rather than a fake one that doesn't really have information pertinent to them.

I could see the benefit of a more marketing/admissions -oriented page where we walk them through what it is like to have a 360 profile and show them ways they would interact with the site. But I don't think a fake profile would necessarily be the best way to do that.

Thoughts @ChrisImagineer ?

(We should probably create a separate issue in the project board for this as this topic is out of the scope of this PR)

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -5,30 +5,31 @@ import { Participation } from 'services/membership';
 import { TranscriptItems } from 'services/transcript';
 
 const participation1: Participation = Participation.Member;
+const currentYear: string = new Date().getFullYear().toString();
```
bennettforkner, line 8: This means that participation2 will often end up being the coming fall instead of the past one. That doesn't really make sense if it is a future involvement. I would recommend setting everything back a year (subtract 1 from the year).

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+};
+export const transcriptItems1: TranscriptItems = {
+  honors: [],
+  experiences: [],
+  service: [],
+  activities: [activity1, activity2],
+};
+export const student1: StudentProfileInfo = {
+  fullName: 'First M Last',
+  Majors: ['Undesided'],
```
bennettforkner, line 42:
```suggestion
  Majors: ['Undecided'],
```

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Minors: [],
+  Advisors: [],
+  ID: '50232456',
```
bennettforkner, line 45:
```suggestion
  ID: '12345678',
```
I'm guessing this is your ID number. You shouldn't make that public and it's unfortunate that it's now in the permanent GitHub history if so.

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+  HomeStreet1: '255 Grapevine Rd',
+  HomeStreet2: '',
+  HomeCity: 'Wenham',
+  HomeState: 'MA',
+  HomePostalCode: '',
+  HomeCountry: 'AM',
+  HomePhone: '',
+  HomeFax: '',
+  KeepPrivate: '',
+  Barcode: '72639526863451',
```
bennettforkner, line 68: Not sure if this is your barcode, but it should not be.

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Email: 'first.last@gordon.edu',
+  Gender: 'F',
+  AD_Username: 'first.last',
+  show_pic: 0,
+  preferred_photo: 0,
+  Country: 'America',
```
bennettforkner, line 74: I don't think this is a valid country. I believe it would be either `US` or `United States of America`.

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -20,7 +28,7 @@ const SectionTitle: { [Key in keyof TranscriptItems]: string } = {
 const CoCurricularTranscript = () => {
   const [loading, setLoading] = useState(true);
   const [transcriptItems, setTranscriptItems] = useState<TranscriptItems | undefined>();
-  const { profile, loading: loadingProfile } = useUser();
+  let { profile, loading: loadingProfile } = useUser();
```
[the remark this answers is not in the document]
bennettforkner: Ah interesting. Okay, that makes sense. I am concerned as to what this will do when you switch from the experience transcript to, say, the homepage. Will that still use the example data that you put in here as the profile?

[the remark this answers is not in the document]
bennettforkner: If so, you might benefit from using a copy of the profile instead of the actual useUser profile.

src/views/CoCurricularTranscript/guestTranscript.tsx
ArabellaJi: Will do!

src/views/CoCurricularTranscript/guestTranscript.tsx
ArabellaJi: It is not my ID but thank you! And I think 12345678 does look better as a mocked up ID.

src/views/CoCurricularTranscript/guestTranscript.tsx
ArabellaJi: Okay! Changing it now

src/views/CoCurricularTranscript/guestTranscript.tsx
ArabellaJi: Okay! Changing it now

src/views/CoCurricularTranscript/index.tsx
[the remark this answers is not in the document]
ArabellaJi: Yeah, if I understand what you are saying, the data that I wrote would only been used when the users go into the transcript page. If the user switch from the transcript page back to the homepage, for now, they will only see the welcome massage like this:

image

bennettforkner:
These two most recent commits doesn't seem related to your pull request in my understanding. What is the purpose of them?

ArabellaJi:
Sorry for replying late @bennettforkner! Right, the two most recent commits contain something that is not related to this pull request. That is because I was working on the issue #784 and forget to change the branch... So I accidentally committed those changes to this PR too in the second to last commit, and the most recent commit was to remove those unrelated changes. You can only look at the changes in guestTranscript.tsx file in the second to last commit (https://github.com/gordon-cs/gordon-360-ui/pull/1817/commits/fb93dff928babec25c9ed44a07e02e341890bd95), those are the only codes that need to be reviewed. Thanks!
Sorry for the confusion! Let me know if there are any places that are unclear.

src/views/CoCurricularTranscript/index.tsx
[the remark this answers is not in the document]
bennettforkner: I would think that would be fine.

src/views/CoCurricularTranscript/index.tsx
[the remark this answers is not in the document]
bennettforkner: Great, that answers that!

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
+import { MembershipHistory, MembershipHistorySession } from 'services/user';
+import { StudentProfileInfo } from 'services/user';
+import { Class } from 'services/peopleSearch';
+import { Participation } from 'services/membership';
+import { TranscriptItems } from 'services/transcript';
+
+const participation1: Participation = Participation.Member;
+const currentYear: number = new Date().getFullYear();
+const session1: MembershipHistorySession = {
+  MembershipID: 2904,
```
bennettforkner, line 10: I would throw a negative number into here rather than using a random ID that might be taken. That could cause unexpected behavior and confusion later. Perhaps just `-1` if that works?

ArabellaJi: Right! I just changed it.

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -36,12 +36,15 @@ const CoCurricularTranscript = () => {
     loadTranscript();
   }, [profile]);
 
-  if (loading || loadingProfile) {
```
bennettforkner, line 39: Why did you remove loading here? Is that not still a valid measure?

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+export const student1: StudentProfileInfo = {
+  fullName: 'First M Last',
+  Majors: ['Undecided'],
+  Minors: [],
+  Advisors: [],
+  ID: '12345678',
+  Title: '',
+  FirstName: 'First',
+  MiddleName: '',
+  LastName: 'Last',
```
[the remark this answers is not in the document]
ArabellaJi: Will do!

src/views/CoCurricularTranscript/index.tsx
bennettforkner: So the problem is that if `transcriptService.getItems(...` fails, nothing catches the error and sets loading to false.

Instead of your updates to lines 44-47, I would add a `.catch((error) => {...})` to the call out to the service on line 31.

bennettforkner: This is an issue because the transcript request fails, so loading is never set to false. I would add a `.catch(...)`, which will set loading to false, to the transcript request on line 31 and then put loading back in here.

ArabellaJi: I see. I add an else statement outside `if (loading || loadingProfile)`, and it solves the problem. Is this what you mean to do? However, I failed to use the `!useIsAuthenticated()` on line 40, it seems like it makes the `profile` on line 61 may be equal to null, which is not allowed. I won't have this problem if I use `!profile()`. Is there a reason that we should use `useIsAuthenticated` instead of `profile`?

image

EjPlatzer: Using `!profile` as the first condition is not ideal because it will render the example profile for a signed-in user while their actual profile loads (which will either be confusing or just flashes of different content that doesn't look good).

You can't call `useIsAuthenticated()` like that because it's a Hook - you have to obey the Rules of Hooks (see the [orange Pitfall in the React Docs on hooks](https://react.dev/learn/state-a-components-memory#meet-your-first-hook)).

What you can do is something like the following:
```javascript
const isAuthenticated = useIsAuthenticated();
const {profile: userProfile, loading: loadingProfile) = useUser();
const profile = isAuthenticated ? userProfile : exampleProfile;
...
const loadTranscript = () => {
  if (isAuthenticated) {
    if (!profile) return;
    setLoading(true);
    // load transcript
  } else {
    setTranscript(exampleTranscript);
  }
  setLoading(false);
}
...
if (loading || loadingProfile) {
  return <GordonLoader />;
} else {
// render transcript
}
```

ArabellaJi:
Sorry I made too many commits to try to restore many things back to their original states... The two most useful commits are the "changed to use useIsAuthenticated() ([021e8e4](https://github.com/gordon-cs/gordon-360-ui/pull/1817/commits/021e8e4a7eb9949a6357537923965da4a95b2b06))" and the "hide the Experience Transcript button when the user is logged in ([c1d316e](https://github.com/gordon-cs/gordon-360-ui/pull/1817/commits/c1d316e20313b9159dc994d2b0b23ed6b333c158))". Maybe you can just look at these two files.

src/components/Header/components/NavButtonsRightCorner/index.jsx
```diff
@@ -50,6 +50,16 @@ const GordonNavButtonsRightCorner = ({ onClose, openDialogBox, open, anchorEl })
...
 
+  const transcriptButton = !isAuthenticated ? (
+    <GordonNavButton
+      unavailable={!isOnline ? 'offline' : null}
+      onLinkClick={onClose}
+      openUnavailableDialog={openDialogBox}
+      linkName={'Experience Transcript'}
+      linkPath={'/transcript'}
+    />
+  ) : null;
```
bennettforkner, line 61: This is shorthand for the same thing:
```suggestion
  const transcriptButton = !isAuthenticated && (
    <GordonNavButton
      unavailable={!isOnline ? 'offline' : null}
      onLinkClick={onClose}
      openUnavailableDialog={openDialogBox}
      linkName={'Experience Transcript'}
      linkPath={'/transcript'}
    />
  );
```

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+  HomeState: 'MA',
+  HomePostalCode: '',
+  HomeCountry: 'US',
+  HomePhone: '',
+  HomeFax: '',
+  KeepPrivate: '',
+  Barcode: '12312312312312',
+  Email: 'first.last@gordon.edu',
+  Gender: 'F',
+  AD_Username: 'first.last',
```
bennettforkner, line 71: If you change the other names, these should also be example.student technically.

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -54,7 +52,7 @@ const CoCurricularTranscript = () => {
                 Gordon College Experience Transcript
               </Typography>
             }
-            subheader={<SubHeader profile={profile} />}
+            subheader={<SubHeader profile={profile != null ? profile : student1} />}
```
bennettforkner, line 55: More shorthand for the same thing:
```suggestion
            subheader={<SubHeader profile={profile ?? student1} />}
```

src/components/Header/components/NavButtonsRightCorner/index.jsx
EjPlatzer: Technically, these are not equivalent. If the user is authenticated, the original code sets `transcriptButton` to `null`, whereas your suggestion will set `transcriptButton` to `false`. However, because JSX treats `null` and `false` the same way for rendering (i.e. "Don't render anything for this element"), they are functionally equivalent. I just wanted to clarify that this isn't some JS voodoo, it's a shortcut specific to how JSX renders HTML from JS values.

bennettforkner: So yes, it's shorthand for the same thing with JSX.

Thanks for the helpful details!

src/views/CoCurricularTranscript/guestTranscript.tsx
ArabellaJi: Right! I forgot them

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
+import { MembershipHistory, MembershipHistorySession } from 'services/user';
+import { StudentProfileInfo } from 'services/user';
+import { Class } from 'services/peopleSearch';
+import { Participation } from 'services/membership';
+import { TranscriptItems } from 'services/transcript';
+
+const participation1: Participation = Participation.Member;
```
EjPlatzer, line 7: I don't think this needs to be an intermediate variable. You could just replace the one spot that it's referenced with the `Participation.Member` value directly.

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Participation: participation1,
+};
+const activity1: MembershipHistory = {
+  ActivityCode: 'ISO',
+  ActivityDescription: 'International Student Organization',
+  ActivityImagePath: '',
+  Sessions: [session1],
+  LatestDate: currentYear - 1 + '-12-15T00:00:00',
+};
+const participation2: Participation = Participation.Member;
```
EjPlatzer, line 21: Same as above, you can just use the `Participation.Member` value directly where this variable is referenced.

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Sessions: [session2],
+  LatestDate: currentYear + '-05-10T00:00:00',
+};
+export const transcriptItems1: TranscriptItems = {
+  honors: [],
+  experiences: [],
+  service: [],
+  activities: [activity1, activity2],
+};
+export const student1: StudentProfileInfo = {
```
EjPlatzer, line 40: I think we could rename this for more clarity, maybe
```suggestion
export const exampleStudentProfile: StudentProfileInfo = {
```

src/views/CoCurricularTranscript/guestTranscript.tsx
```diff
@@ -0,0 +1,116 @@
...
+  Participation: participation2,
+};
+const activity2: MembershipHistory = {
+  ActivityCode: '360',
+  ActivityDescription: '360.gordon.edu',
+  ActivityImagePath: '',
+  Sessions: [session2],
+  LatestDate: currentYear + '-05-10T00:00:00',
+};
+export const transcriptItems1: TranscriptItems = {
```
EjPlatzer, line 34: Could we rename this for more clarity, such as
```suggestion
export const exampleTranscriptItems: TranscriptItems = {
```

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -20,30 +21,27 @@ const SectionTitle: { [Key in keyof TranscriptItems]: string } = {
...
+      }
       setLoading(true);
-
-      await transcriptService.getItems(profile.AD_Username).then(setTranscriptItems);
-
-      setLoading(false);
-    };
-
-    loadTranscript();
+      transcriptService.getItems(profile.AD_Username).then(setTranscriptItems);
```
EjPlatzer, line 34: This actually won't work exactly as expected currently. Because you aren't `await`ing these `Promise`s, the browser will start the `getItems` request and then continue on before it's finished, executing `setLoading(false)` potentially **before** the `Promise` from the `getItems` request has resolved, meaning we haven't actually finished loading data yet.

To fix that, you can put the code inside the `useEffect` into an async function and invoke that function at the end of the `useEffect`.

EjPlatzer (changes requested):
This looks great! Thanks for working on this. Just a couple minor suggestions for cleanup

src/views/CoCurricularTranscript/index.tsx
EjPlatzer: The best way to do this is to define the async function within the useEffect, and then call it immediately after it's defined.

It should look something like this:
```javascript
useEffect(() => {
  const loadTranscript = () => {
    if (isAuthenticated) {
      ...
      const transcriptItems = await transcriptSerivce.getItems();
      setTranscriptItems(transcriptItems);
    } else {
      setTranscriptItems(transcriptItems1);
    }
    setLoading(false);
  }
  loadTranscript();
}, [isAuthenticated, profile]);
```
or if you still prefer the `.then` syntax:
```javascript
useEffect(() => {
  const loadTranscript = () => {
    if (isAuthenticated) {
      ...
      await transcriptSerivce.getItems().then(setTranscriptItems);
    } else {
      setTranscriptItems(transcriptItems1);
    }
    setLoading(false);
  }
  loadTranscript();
}, [isAuthenticated, profile]);
```

ArabellaJi: Sounds good lol! It works

EjPlatzer (approved):
LGTM!

ArabellaJi:
> LGTM!

Okayy!

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -20,30 +21,31 @@ const SectionTitle: { [Key in keyof TranscriptItems]: string } = {
...
-
-      await transcriptService.getItems(profile.AD_Username).then(setTranscriptItems);
-
+      if (isAuthenticated) {
+        if (!profile) {
+          return;
+        }
+        const transcriptItems = await transcriptService.getItems(profile.AD_Username);
+        setTranscriptItems(transcriptItems);
+        setLoading(true);
```
EjPlatzer, line 36: Having `setLoading(true)` here is confusing. We want the page to be loading while we're waiting on data, right? So shouldn't we set loading to true before we start loading data, i.e. above the `await` line?

EjPlatzer (approved):
LGTM!

Issue #974, opened 2021-02-10
Profile Links: Allow input without https or www or both


cpabbot:
I think this would make adding profile links more accessible and should be fairly simple. I would change the prefix values to be lists of acceptable url prefixes (to include all combinations or https, www, etc.) and this would also eliminate the hard coding I did for handshake (two types of prefixes for that).

andrew-wzj:
Quick fix to this issue would be deleting string  'www.' from prefix values which doesn't cause any problem on redirecting to profile page.
The problem for handshake links (two types of prefixes) is that links with gordon. and app. could redirect to profile, and we just need to delete /user in prefix values.

PR #1820, opened 2023-06-01
S23 allow various prefix inputs for links

mla04762:
[#974](https://github.com/gordon-cs/gordon-360-ui/issues/974)


bennettforkner:
Hi @mla04762 -- I think you created this branch with the recim branch as your base. When you create a new branch, you should start with the one that you are merging into rather than something else. That's why you have 86 files changed in this pull request and it cannot be merged.

I would recommend creating a new branch from `develop` and using `git cherry-pick` to pull your commits out of this branch.

https://git-scm.com/docs/git-cherry-pick

mla04762:
(faulty branch with wrong base): https://github.com/gordon-cs/gordon-360-ui/pull/1831 is the updated PR

Issue #546, opened 2019-07-18
Implement Add to Calendar buttons for events


joshuaRogers24:

andrew-wzj:
I am interested in adding calendar button for event, and I wonder to discuss the design in detail. Also question for how to access user's calendar.

andrew-wzj:
https://github.com/gordon-cs/gordon-360-ui/pull/1822
This is the linked pull requests

mla04762:
@bennettforkner We weren't able to find a way to get two separate time's for the start and end time for a event (we could only get one string that combined the two times as a range), is there a way to get this from the api?

bennettforkner:
> @bennettforkner We weren't able to find a way to get two separate time's for the start and end time for a event (we could only get one string that combined the two times as a range), is there a way to get this from the api?

https://github.com/gordon-cs/gordon-360-ui/blob/develop/src/services/event.ts

The event service shows the structure of our event objects in typescript. There is a start and end date (and time). I'm not 100% sure where you are getting events without separate dates.

mla04762:
We didn't see that until now, we will look into that. Thank You!

bennettforkner:
I recommend tracing back the calls from EventItem backwards so that you can find exactly what type is being passed to it.

PR #1822, opened 2023-06-01
Add Calendar to Event

andrew-wzj:
Implement Add to Calendar buttons for events #546 
I just add the calendar button for event so user can easily add the specific event into the calendar.
Work done: event title and location sync with eventItem
Work need done: conversion for date and time into right format. (Description could be easily sync)


andrew-wzj:
The build check failed since import { AddToCalendarButton } is not part of original dependency of gordon 360. The

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -1,10 +1,14 @@
...
 import { useState } from 'react';
 import styles from './EventItem.module.css';
+import { AddToCalendarButton } from 'add-to-calendar-button-react';
 
 const EventItem = ({ event }) => {
   const [expanded, setExpanded] = useState(false);
 
+  
+
+
```
bennettforkner, line 11: I don't think this extra whitespace is necessary. Generally we try to limit it to one new line if any.

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -1,10 +1,14 @@
-import { CardContent, Collapse, Grid, Typography } from '@mui/material';
+import { Button, CardContent, Collapse, Grid, Typography } from '@mui/material';
 import { useState } from 'react';
 import styles from './EventItem.module.css';
+import { AddToCalendarButton } from 'add-to-calendar-button-react';
```
bennettforkner, line 4: Is this a new package? I don't see a new listing in `package.json`. Am I missing something?

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -32,6 +36,18 @@ const EventItem = ({ event }) => {
...
+          <Button onClick={() => alert(event.date +" " +event.timeRange)}
+          >name</Button>
+        <AddToCalendarButton
+          name= {event.title}
+          options={['Apple','Google']}
+          location= {event.location}
+          startDate="2023-06-04"
+          endDate="2023-06-04"
+          startTime="10:15"
+          endTime="23:30"
```
bennettforkner, line 48: If this pull request is open and ready for review, these dates should be populated with the values from the event, not testing values.

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -32,6 +36,18 @@ const EventItem = ({ event }) => {
...
+          timeZone="America/Indiana/Indianapolis"
```
bennettforkner, line 49: This is not our timezone. That would be an hour behind us.

I believe you would want "America/New York" since all of our events thus far are in that timezone.

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -32,6 +36,18 @@ const EventItem = ({ event }) => {
           <Typography type="caption" className={styles.descriptionText}>
             {event.Description || 'No description available'}
           </Typography>
+          <Button onClick={() => alert(event.date +" " +event.timeRange)}
+          >name</Button>
```
bennettforkner, line 40: Since this PR is open now, you shouldn't have any testing code left behind in it.

bennettforkner (commented):
Looks like a great addition to 360!

Remember that you can always open a pull request as a draft until you have removed all of your testing code. Otherwise, it will imply that you are ready for reviews like this! :)

bennettforkner:
> The build check failed since import { AddToCalendarButton } is not part of original dependency of gordon 360. The 

All you would need to do is run an `npm install [package name]` in a terminal window to install the package in your branch. Then, it will show up in package.json as well.

I will note that all package dependencies should go through proper review before being added to 360. That means that @russtuck and/or @jsenning should do a more in-depth review of the security and health of the package before adding it.

src/components/EventList/components/EventItem/index.jsx
andrew-wzj: Yes! This is the only package I added

bennettforkner: See my later comment concerning this.

src/components/EventList/components/EventItem/index.jsx
andrew-wzj: Thanks for reminder

package.json
```diff
@@ -12,6 +12,8 @@
     "@mui/material": "^5.10.7",
     "@mui/styles": "^5.10.7",
     "@mui/x-date-pickers": "^5.0.3",
+    "add-to-calendar-button": "^2.2.9",
+    "add-to-calendar-button-react": "^2.2.7",
```
bennettforkner, line 16: I wouldn't think you need both of these. Can you chose one and only use that one?

package.json
```diff
@@ -58,7 +60,7 @@
     "eslint-plugin-jsx-a11y": "^6.5.1",
     "eslint-plugin-react": "^7.27.1",
     "husky": "^7.0.4",
-    "jsdoc": "^3.6.7",
+    "jsdoc": "^4.0.2",
```
bennettforkner, line 63: Be careful when randomly updating packages. I would avoid doing it in this PR and revert your change here.

We usually try to update packages intentionally and look through all of the changes before updating. Especially with a new major version (3.x.x to 4.x.x, for example) there can be a lot of changes that would break some of the features we use.

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -1,10 +1,59 @@
...
+    const times = timeString.split(' - ');
+    const startTime = times[0];
+    const endTime = times[1];
+  
+    const convertSingleTime = (singleTime) => {
+      const [time, period] = singleTime.split(' ');
+      const [hours, minutes] = time.split(':');
+      let convertedHours = parseInt(hours, 10);
+      
+      if (period.toLowerCase() === 'pm' && convertedHours !== 12) {
```
bennettforkner, line 32: This seems like a lot of work just to get datetime objects in a specific format. There are libraries that will do this for you instead of you reinventing the wheel a bit. We have a package installed in 360 called `moment`.

See:
https://momentjs.com/docs/#/parsing/
https://momentjs.com/docs/#/displaying/format/

Additionally, why do you need to format the date/time at all. I would be very surprised if `AddToCalendarButton` can not take in a date object.

package.json
andrew-wzj: Sure! The second one is what we only need.

src/components/EventList/components/EventItem/index.jsx
andrew-wzj: The startDate and endDate only accept format YYYY-MM-DD, so I have to call convertDate in AddToCalendarButton

andrew-wzj: Thanks for the useful links! I would notice that next time

bennettforkner: > The startDate and endDate only accept format YYYY-MM-DD, so I have to call convertDate in AddToCalendarButton

My proposal is that you don't reinvent the wheel by creating a `convertDate` function. That work is already done for you. Please see the links that I sent.

bennettforkner: > Thanks for the useful links! I would notice that next time

This was a comment on your current PR while I am requesting be addressed before merge. This should be considered now, not next time. I welcome @amos-cha 's feedback as well.

EjPlatzer: > This seems like a lot of work just to get datetime objects in a specific format. There are libraries that will do this for you instead of you reinventing the wheel a bit. We have a package installed in 360 called `moment`.
> 
> 
> 
> See:
> 
> https://momentjs.com/docs/#/parsing/
> 
> https://momentjs.com/docs/#/displaying/format/
> 
> 
> 
> Additionally, why do you need to format the date/time at all. I would be very surprised if `AddToCalendarButton` can not take in a date object.

Moment is deprecated. We use date-fns instead, which has an equally powerful format function

EjPlatzer: Sure, the date-fns docs for the `format` function are here: https://date-fns.org/v2.30.0/docs/format

package.json
russtuck, line 16: I looked at these repos.  the -react version is a small wrapper around the base.  Both appear to be actively developed and supported. The -react repo is a one-person (jekuer) effort since last Oct, but is ~100 lines and wouldn't be hard to pull inside.

add-to-calendar-button has been actively developed for over 2 years, with multiple developers, but with the -react developer as the most prolific.  It appears to contain about 2000-3000 lines of code.  It has >1100 stars and >19,000 weekly npm installs, so it's fairly widely used.  The lead developer (jekuer) say he is in Germany.   There's a simple website, add-to-calendar-button.com.

This all looks reasonable to me, with no red flags, though it's still a small team effort, not one from a well-known organization.

We looked at solving this problem (creating links to standard calendars) a few years ago, and it was too much work for the time available.  So it's nice to find a package to do it.

bennettforkner: Awesome, thanks for looking into it!

src/components/EventList/components/EventItem/index.jsx
andrew-wzj: @EjPlatzer Is the date-fns already in the 360 package as moment? Or should I install it instead

bennettforkner: You can confirm that it is installed in `package.json`.
image

mla04762:
The most recent commit of our branch contains the add-to-calendar package that uses a web component version of the package which allows us to not downgrade or upgrade our react version to make it compatible. We also were able to use the date-fns package to format the date into the appropriate formate for the add-to-calendar-button instead of having to create a whole new function for it :)

package-lock.json
```diff
@@ -63,7 +64,7 @@
         "eslint-plugin-jsx-a11y": "^6.5.1",
         "eslint-plugin-react": "^7.27.1",
         "husky": "^7.0.4",
-        "jsdoc": "^3.6.7",
+        "jsdoc": "^4.0.2",
```
russtuck, line 67: This change should be un-done automatically when you revert the change in package.json as Bennett requested. 

Other changes marked "revert" below should also go away at the same time. 
 But check that they do.  I think the ones marked "keep" will stay.

package-lock.json
```diff
@@ -379,9 +380,9 @@
       }
     },
     "node_modules/@babel/parser": {
-      "version": "7.20.7",
-      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.20.7.tgz",
-      "integrity": "sha512-T3Z9oHybU+0vZlY9CiDSJQTD5ZapcW18ZctFMi0MOAl/4BjFF4ul7NVSARLdbGO5vDqy9eQiGTV0LtKfvCYvcg==",
+      "version": "7.22.4",
```
russtuck, line 383: Revert these lines

package-lock.json
```diff
@@ -1281,6 +1282,18 @@
         "@jridgewell/sourcemap-codec": "^1.4.10"
       }
     },
+    "node_modules/@jsdoc/salty": {
```
russtuck, line 1285: keep: it's new, presumably needed by the package you added

package-lock.json
```diff
@@ -2245,6 +2258,18 @@
         "acorn": "^6.0.0 || ^7.0.0 || ^8.0.0"
       }
     },
+    "node_modules/add-to-calendar-button": {
```
russtuck, line 2261: keep this

package-lock.json
```diff
@@ -5378,12 +5403,13 @@
       }
     },
     "node_modules/jsdoc": {
-      "version": "3.6.11",
-      "resolved": "https://registry.npmjs.org/jsdoc/-/jsdoc-3.6.11.tgz",
-      "integrity": "sha512-8UCU0TYeIYD9KeLzEcAu2q8N/mx9O3phAGl32nmHlE0LpaJL71mMkP4d+QE5zWfNt50qheHtOZ0qoxVrsX5TUg==",
+      "version": "4.0.2",
+      "resolved": "https://registry.npmjs.org/jsdoc/-/jsdoc-4.0.2.tgz",
```
russtuck, line 5407: revert

package-lock.json
```diff
@@ -8422,6 +8441,14 @@
       "integrity": "sha512-w89qg7PI8wAdvX60bMDP+bFoD5Dvhm9oLheFp5O4a2QF0cSBGsBX4qZmadPMvVqlLJBBci+WqGGOAPvcDeNSVg==",
       "dev": true
     },
+    "node_modules/timezones-ical-library": {
```
russtuck, line 8444: keep

package-lock.json
```diff
@@ -9258,9 +9285,9 @@
       }
     },
     "@babel/parser": {
-      "version": "7.20.7",
-      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.20.7.tgz",
-      "integrity": "sha512-T3Z9oHybU+0vZlY9CiDSJQTD5ZapcW18ZctFMi0MOAl/4BjFF4ul7NVSARLdbGO5vDqy9eQiGTV0LtKfvCYvcg=="
+      "version": "7.22.4",
```
russtuck, line 9288: revert

package-lock.json
```diff
@@ -9826,6 +9853,15 @@
         "@jridgewell/sourcemap-codec": "^1.4.10"
       }
     },
+    "@jsdoc/salty": {
```
russtuck, line 9856: keep

package-lock.json
```diff
@@ -10484,6 +10520,14 @@
       "peer": true,
       "requires": {}
     },
+    "add-to-calendar-button": {
```
russtuck, line 10523: keep

package-lock.json
```diff
@@ -12785,12 +12829,13 @@
       }
     },
     "jsdoc": {
-      "version": "3.6.11",
-      "resolved": "https://registry.npmjs.org/jsdoc/-/jsdoc-3.6.11.tgz",
-      "integrity": "sha512-8UCU0TYeIYD9KeLzEcAu2q8N/mx9O3phAGl32nmHlE0LpaJL71mMkP4d+QE5zWfNt50qheHtOZ0qoxVrsX5TUg==",
+      "version": "4.0.2",
```
russtuck, line 12832: revert

package-lock.json
```diff
@@ -15064,6 +15102,11 @@
       "integrity": "sha512-w89qg7PI8wAdvX60bMDP+bFoD5Dvhm9oLheFp5O4a2QF0cSBGsBX4qZmadPMvVqlLJBBci+WqGGOAPvcDeNSVg==",
       "dev": true
     },
+    "timezones-ical-library": {
```
russtuck, line 15105: keep

package.json
```diff
@@ -58,7 +59,7 @@
...
```
russtuck, line 62: Why didn't you revert this as Bennett requested?  Please do that now, then request a new review.

russtuck (changes requested):
package.json still needs fixing, which should be easy.  The rest looks fine.

russtuck (approved):
LGTM
Please do one final test and add a note that it still works.

mla04762:
We did a final check and everything seems to work fine.

PR #1830, opened 2023-06-02
Added Confetti with Subscribe and Join

jhammond4:
When a user clicks the subscribe button, confetti like the confetti from the Happy Birthday banner pops up. The same happens when someone clicks okay to send a request to join a club. I also changed the confetti to be Gordon's main colors, but this can easily be switched back

### DIFF
--- a/src/components/GordonConfetti/index.jsx
+++ b/src/components/GordonConfetti/index.jsx
@@ -5,13 +5,13 @@ const GordonConfetti = ({ active, colorOption, colors }) => {
   let config = {
     angle: 90,
     spread: 360,
-    startVelocity: 40,
+    startVelocity: 20,
     elementCount: 200,
-    dragFriction: 0.12,
-    duration: 6000,
-    stagger: 3,
-    width: '10px',
-    height: '10px',
+    dragFriction: 0.001, //some numbers were adjusted for the confetti effects
+    duration: 10000,
+    stagger: 5,
+    width: '30px',
+    height: '60px',
     perspective: '500px',
   };
 
@@ -22,8 +22,8 @@ const GordonConfetti = ({ active, colorOption, colors }) => {
       '#ddd',
       gordonColors.primary.blue,
       gordonColors.primary.cyan,
-      gordonColors.secondary.orange,
-      gordonColors.secondary.green,
+      //gordonColors.secondary.orange,    //these are commented out so the confetti is Gordon's primary colors
+      //gordonColors.secondary.green,
     ];
   } else if (colorOption === 'GordonBlue') {
     config.colors = ['#ddd', gordonColors.primary.blue, gordonColors.primary.cyan];
--- a/src/components/GordonDialogBox/index.jsx
+++ b/src/components/GordonDialogBox/index.jsx
@@ -1,6 +1,8 @@
 import { Button, Dialog, DialogActions, DialogContent, DialogTitle } from '@mui/material';
 import { Alert, AlertTitle } from '@mui/material';
 import styles from './GordonDialogBox.module.css';
+import GordonConfetti from 'components/GordonConfetti'; //GordonConfetti and useState are imported for the confetti
+import { useState } from 'react';
 
 /**
  * @param {Object} props the component props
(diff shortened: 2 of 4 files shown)


src/components/GordonConfetti/index.jsx
```diff
@@ -22,8 +22,8 @@ const GordonConfetti = ({ active, colorOption, colors }) => {
       '#ddd',
       gordonColors.primary.blue,
       gordonColors.primary.cyan,
-      gordonColors.secondary.orange,
-      gordonColors.secondary.green,
+      //gordonColors.secondary.orange,    //these are commented out so the confetti is Gordon's primary colors
+      //gordonColors.secondary.green,
```
bennettforkner, line 26: Hey @jhammond4 -- there is a configuration setting that you can pass in which will make the confetti be only Gordon primary colors. See lines 28-30.

Please don't change the global Confetti configuration since it affects the Happy Birthday message as well.

src/components/GordonConfetti/index.jsx
```diff
@@ -5,13 +5,13 @@ const GordonConfetti = ({ active, colorOption, colors }) => {
...
-    dragFriction: 0.12,
-    duration: 6000,
-    stagger: 3,
-    width: '10px',
-    height: '10px',
+    dragFriction: 0.001, //some numbers were adjusted for the confetti effects
+    duration: 10000,
+    stagger: 5,
+    width: '30px',
+    height: '60px',
```
bennettforkner, line 14: Were these changes specifically requested or approved by Chris Carlson? He originally approved the confetti mechanics as they are now. I wouldn't want to change it globally without consulting him.

src/components/GordonDialogBox/index.jsx
```diff
@@ -31,48 +33,66 @@ const GordonDialogBox = ({
...
+    <>
+      <div style={{ position: 'fixed', zIndex: 999999, top: -60, left: '50vw' }}>
+        <GordonConfetti active={confetti} colorOption="Gordon" />
+      </div>
+      <Dialog
+        className={styles.gc360_gordondialogbox}
+        {...otherProps}
+        aria-labelledby="alert-dialog-title"
+        aria-describedby="alert-dialog-description"
       >
```
bennettforkner, line 53: The GordonDialogBox component is used in a lot more places where we _do not_ want confetti to appear. I would suggest applying the confetti to the specific page in question, not here on this component.

src/components/GordonConfetti/index.jsx
bennettforkner: You can pass in `colors` or `colorOption` to the GordonConfetti component and it will use the colors that you specify.

src/components/GordonConfetti/index.jsx
jhammond4: I just changed them myself when I was messing around with the confetti to see what would happen-I can change them back to their originals

src/components/GordonConfetti/index.jsx
jhammond4: Got it-I can switch this back as well.

src/views/InvolvementProfile/components/Membership/components/NonMemberButtons/index.jsx
```diff
@@ -74,70 +83,82 @@ const NonMemberButtons = ({
...
     <>
-      <CardActions>
-        {isGuest ? (
-          <Button variant="contained" color="primary" onClick={onUnsubscribe}>
-            Unsubscribe
-          </Button>
-        ) : (
+      <Card>
+        <div style={{ position: 'fixed', zIndex: 999999, top: -60, left: '50vw' }}>
+          <GordonConfetti active={confetti} colorOption="Gordon" />
```
bennettforkner, line 90: You were trying to get this to be only the GordonBlue color palette before. Do you still want it to be just blue and cyan? If so, you can make it `colorOption="GordonBlue"` instead of `colorOption="Gordon"`

jhammond4: That makes sense. I'm good with leaving it as 'Gordon' for the multicolor confetti, it would fit better for the various clubs that aren't explicitly Gordon-specific like you said.

bennettforkner (commented):
Suggesting that you remove these comments as they do not necessarily provide information about your implementation of the code. In these cases comments are not necessary IMO.

russtuck (approved):
LGTM, and it looks like you have addressed all of Bennett's comments.

PR #1831, opened 2023-06-02
S23 allow various prefix for links (new commit on develop as base instead of recim)

mla04762:
#1820


mla04762:
Need to fix some errors before pushing

bennettforkner:
Instead of closing this PR, you can just convert it to a draft if you would like. It's a small button in the right-hand sidebar.

src/components/Profile/components/Identification/components/SocialMediaLinks/components/LinksDialog/index.jsx
```diff
@@ -23,9 +23,17 @@ const LinksDialog = ({ links, createSnackbar, onClose, setLinks }) => {
...
-      value === '' || value.indexOf(prefix) === 0 || (prefix2 && value.indexOf(prefix2) === 0);
+    const { prefix } = socialMediaInfo[platform];
+    let isValid = value === '';
+    console.log('isValid' + isValid);
+    if (!isValid && prefix) {
+      for (let i = 0; i < prefix.length; i++) {
+        if (value.indexOf(prefix[i]) === 0) {
+          isValid = true;
+        }
+      }
```
bennettforkner, line 34: FYI, I know this is a draft, but normal `for` loops are prone to errors, so we tend to prefer functions like `.map(`, `.filter(`, `.some(`, or `.foreach(`.

`.some(...)` checks if any of the elements in the collection return true for the logic in the lamda function.

For example:
```suggestion
      isValid = prefix.some(p =>
        value.indexOf(p) === 0);
```

andrew-wzj:
The problem of making social media format without both https and www, it cause the page doesn't redirect to the social media page but search in 360 locally by changing url to localhost/facebook.com/....

andrew-wzj:
The expected change here is to matching user input url with each valid prefix in the list. Later, the further improvement could be if user didn't type https or www or both, it would help user to save urls in right format.

jsenning (approved):
LGTM

PR #1836, opened 2023-06-05
[Rec-IM] remove co captain

antoniavonto:
Remove promote to co-captain.
Implemented promote to captain, if you promote to captain, it sets the previous captain back to member.
image

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -4,7 +4,7 @@
 # VITE_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
- VITE_API_URL=https://360apirecim.gordon.edu/
+ VITE_API_URL=https://360ApiTrain.gordon.edu/
 
 # @LOCALHOST
 # VITE_API_URL=http://localhost:51626/
--- a/.env.development
+++ b/.env.development
@@ -4,7 +4,7 @@
 # VITE_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
- VITE_API_URL=https://360ApiTrain.gordon.edu/
+ VITE_API_URL=https://360apirecim.gordon.edu/
 
 # @LOCALHOST
 # VITE_API_URL=http://localhost:51626/
--- a/src/views/RecIM/components/List/index.jsx
+++ b/src/views/RecIM/components/List/index.jsx
@@ -88,8 +88,8 @@ const ParticipantList = ({
         teamID={teamID}
         callbackFunction={(bool) => callbackFunction(bool)}
         showParticipantOptions={
-          showParticipantOptions || //I changed this from &&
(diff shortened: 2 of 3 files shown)


.env.development
```diff
@@ -4,7 +4,7 @@
 # VITE_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
- VITE_API_URL=https://360ApiTrain.gordon.edu/
+ VITE_API_URL=https://360apirecim.gordon.edu/
```
bennettforkner, line 7: Please make sure to revert this change before merging into recim or develop.

src/views/RecIM/components/List/index.jsx
```diff
@@ -42,6 +42,28 @@ const ParticipantList = ({
...
   callbackFunction,
 }) => {
+  const [newCaptain, setNewCaptain] = useState();
+  const [teamId, setTeamID] = useState();
+
+  useEffect(() => {
+    // find the old capitain to demote
+    participants.map(async (participant) => {
+      if (
+        participants.length > 0 &&
```
bennettforkner, line 52: If you're inside this lambda function, you already know that participants.length is at least 1.
```suggestion
```

src/views/RecIM/components/List/index.jsx
```diff
@@ -42,6 +42,28 @@ const ParticipantList = ({
...
+        participant !== newCaptain &&
+        teamId !== undefined
+      ) {
+        let editedParticipant = {
+          Username: participant.Username,
+          RoleTypeID: 3,
+        }; // Role 3 is member
+        await editTeamParticipant(teamId, editedParticipant);
+      }
+      callbackFunction((bool) => !bool);
```
bennettforkner, line 63: I wouldn't name a variable after a common type, `boolean`. That will confuse other developers. It would be better to do something simple like a character.
```suggestion
      callbackFunction((b) => !b);
```

src/views/RecIM/components/List/index.jsx
```diff
@@ -64,8 +88,8 @@ const ParticipantList = ({
         teamID={teamID}
         callbackFunction={(bool) => callbackFunction(bool)}
         showParticipantOptions={
-          showParticipantOptions &&
-          participant.Role !== 'Team-captain/Creator' &&
+          showParticipantOptions ||
```
antoniavonto, line 91: showParticipantOptions was returning false so then the buttons to promote to captain and remove from team were not showing for me. It worked when I changed to "||", however I am not sure if it would be wrong for other parts of the code.

amos-cha (changes requested):
Good work! Just minor things then it should be ready to go :)

PR #1837, opened 2023-06-05
Fix the "tab is not highlighted correctly sometimes" Issue

ArabellaJi:
This PR is for #784 to fix the issue that the tabs are not highlighted correctly sometimes.

### DIFF
--- a/src/components/Header/index.jsx
+++ b/src/components/Header/index.jsx
@@ -12,7 +12,7 @@ import GordonDialogBox from 'components/GordonDialogBox';
 import { useDocumentTitle, useNetworkStatus, useWindowSize } from 'hooks';
 import { projectName } from 'project-name';
 import { forwardRef, useEffect, useState } from 'react';
-import { NavLink, Route, Routes } from 'react-router-dom';
+import { NavLink, Route, Routes, useLocation } from 'react-router-dom';
 import routes from 'routes';
 import { authenticate } from 'services/auth';
 import { windowBreakWidths } from 'theme';
@@ -39,27 +39,26 @@ const GordonHeader = ({ onDrawerToggle }) => {
    * The checks use regular expressions to check for matches in the url.
    */
   const updateTabHighlight = () => {
-    let currentPath = location.pathname;
     // Tab url regular expressions must be listed in the same order as the tabs, since the
     // indices of the elements in the array on the next line are mapped to the indices of the tabs
-    let urls = [
+    const urls = [
       /^\/$/,
       /^\/involvements\/?$|^\/activity/,
       /^\/events\/?$/,
-      /^\/people$/,
+      /^\/people$|^\/myprofile|^\/profile/,
       /^\/timesheets$/,
     ];
-    setTabIndex(false);
+    let currentPath = useLocation().pathname;
+    let tab = false;
     for (let i = 0; i < urls.length; i++) {
       if (urls[i].test(currentPath)) {
-        setTabIndex(i);
+        tab = i;
       }
     }
+    if (tab !== tabIndex) {
+      setTabIndex(tab);
+    }
   };
-
-  useEffect(() => {
-    updateTabHighlight();
-  }, [location.pathname]);
 
   useEffect(() => {
     if (width < windowBreakWidths.breakMD) {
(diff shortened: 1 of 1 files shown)


bennettforkner (approved):
LGTM -- Assuming you tested this properly.

Ah, sorry. I see this is a draft.

src/components/Header/index.jsx
```diff
@@ -39,28 +39,27 @@ const GordonHeader = ({ onDrawerToggle }) => {
...
+    const urls = [
       /^\/$/,
       /^\/involvements\/?$|^\/activity/,
       /^\/events\/?$/,
-      /^\/people$/,
+      /^\/people$|^\/myprofile|^\/profile/,
       /^\/timesheets$/,
     ];
-    setTabIndex(false);
+    let currentPath = useLocation().pathname;
```
EjPlatzer, line 51: The `useLocation` function provided by React Router is a Hook. As I mentioned in a different review, Hooks have to obey certain rules in React in order for everything to work properly, which are detailed in the [React Docs](https://react.dev/warnings/invalid-hook-call-warning).

The relevant one here is "always use Hooks at the top level of your React function", which specifically includes "Don’t call Hooks nested functions". The `updateTabHighlight` function is one such nested function, and is not at the "top level" of the function, so this call to `useLocation` violates that rule.

It works okay in this case, but there are many subtle ways that this code could change which would break React, so we shouldn't leave it as is.

Fixing this is actually really simple - just move the invocation of `useLocation` outside of `updateTabHighlight`. For Example:
```javascript
const location = useLocation();
...
const updateTabHighlight = () => {
  ...
  const currentPath = location.pathname;
  ...
};
```

That will fix the most severe issue that we're calling a hook at not the top level.

src/components/Header/index.jsx
```diff
@@ -152,88 +151,85 @@ const GordonHeader = ({ onDrawerToggle }) => {
...
-
-          <div className={styles.center_container}>
-            <Tabs
-              textColor="inherit"
-              indicatorColor="secondary"
-              centered
-              value={tabIndex}
-              onChange={(event, value) => setTabIndex(value)}
+    <>
+      {updateTabHighlight()}
```
EjPlatzer, line 155: I have a couple of comments about this.

Firstly, it's a little bit odd to call the function here, inside of the return statement and nested in the JSX Fragment. You could achieve exactly the same effect by calling the function anywhere else in the component after it's defined, so it's unnecessarily confusing to put it here.

Secondly, it's not best practice to define a function and then call it inside the body of a component. That is because, every time the component renders, it will recreate the function and then call it again. That is causing the component to do a lot of unnecessary work. Also, notice that you're calling `setTabIndex` inside the function, which means that every time this component renders, if the tab index should change, you're immediately changing the state, which will cause another re-render.

More specifically, when `location` changes (because the user navigated to a new page), `useLocation` will update - which counts as a state update the same way that calling `setState` causes a state update. That will cause the Header component to re-render with the new value of `location`. Then, while the component is re-rendering, it will re-evaluate the definition of `updateTabHighlight` and re-create the function. Finally, the component will reach the `return` statement and begin interpreting the JSX Fragment - at which point it reaches this line and runs `updateTabHighlight`. Because `location` has changed, there's a good chance that `tab !== tabIndex` will be true, so we will call `setTabIndex`, which will cause the component to re-render again.

All that to say - we're doing more work than is necessary in a round-about way here. Because `tabIndex` can be calculated based only on `location`, it doesn't need to be it's own state variable, it can just be a regular javascript variable.

There a few different ways to clean this up, but I think the best way is actually by creating a custom hook, which would look something like this:
```javascript

const TabUrlPatterns = [
      /^\/$/,
      /^\/involvements\/?$|^\/activity/,
      /^\/events\/?$/,
      /^\/people$|^\/myprofile|^\/profile/,
      /^\/timesheets$/,
    ];

const useTabHighlight = () => {
  const location = useLocation();
  const [tabIndex, setTabIndex] = useState(0);

  useEffect(() => {
    const matchedIndex = TabUrlPatterns.findIndex((pattern) => pattern.test(location));
    setTabIndex(matchedIndex); // This won't cause an update if the new value is the same as the old value
  }, [location]);

  return tabIndex;
};
```

Then you can replace all the tab highlight code in Header with this:
```javascript
  const tabIndex = useTabHighlight();
```

EjPlatzer (changes requested):
Thanks so much for working on this Arabella! I'm super excited to hear that you've come up with a working solution!

However, I have some concerns about this solution, which are expressed in my comments below. Basically, it comes down to the fact that, with the current version of the code, you're kinda hacking React in ways that work right now but are unstable and could cause issues for us down the line. It all comes back to re-rendering in React, so if you'd like to learn more about that, this article is a fantastic deep dive into rendering: https://www.joshwcomeau.com/react/why-react-re-renders/

src/components/Header/index.jsx
ArabellaJi: That makes sense! Thank you for writing this whole paragraph for me! I did the changes you suggested, and it works perfectly. There is only one thing that I don't understand. If I use `location` here in the code `const matchedIndex = TabUrlPatterns.findIndex((pattern) => pattern.test(location));`, the highlight is not working at all, which means there is no highlight on any of the tabs no matter which tab I press. However, if I change the `location` to `location.pathname`, then `const matchedIndex = TabUrlPatterns.findIndex((pattern) => pattern.test(location.pathname));` works very well. I don't understand why simply adding a `.pathname` solve the problem, but I believe basically the problem is solved.

EjPlatzer: Sorry, that was my bad using `location`. The `location` returned by `useLocation` is an object like this:
```typescript
interface Location {
  pathname: string;
  search: string;
  hash: string;
  state: unknown;
  key: string;
}
```
So naturally, that object won't match any of the Regular Expressions that are searching for a string.

src/components/Header/index.jsx
```diff
@@ -23,43 +23,45 @@ import styles from './Header.module.css';
...
-      if (urls[i].test(currentPath)) {
-        setTabIndex(i);
-      }
-    }
-  };
-
-  useEffect(() => {
-    updateTabHighlight();
-  });
+  console.log('tabIndex: ' + tabIndex);
```
EjPlatzer, line 64: Make sure to remove this console log, we don't want any extra in production.

src/components/Header/index.jsx
ArabellaJi: I see! Thanks!

src/components/Header/index.jsx
```diff
@@ -23,43 +23,43 @@ import styles from './Header.module.css';
...
+  /^\/people$|^\/myprofile|^\/profile/,
+  /^\/timesheets$/,
+];
+
+/**
+ * Update the tab highlight indicator based on the url
+ *
+ * The checks use regular expressions to check for matches in the url.
+ */
+const updateTabHighlight = () => {
```
EjPlatzer, line 41:
```suggestion
const useTabHighlight= () => {
```

Can we change the name of this to clarify that it's a custom hook.

ArabellaJi: I see. Will do!

EjPlatzer (approved):

Issue #1287, opened 2021-07-19
People Search:  Searching Alumni by graduate year


jsenning:
When searching for alumni in Advanced People Search, it would be helpful to search by major and/or year(s).  Ideally we could enter a range of years (e.g. 1990-1999) in a text field.

cpabbot:
> Ideally we could enter a range of years

Haha sounds like we need a slider!

Search by major is already possible for alumni. Search by minors is currently not possible but for a separate reason marked here https://github.com/gordon-cs/gordon-360-api/issues/591.

bennettforkner:
Noting here that we would also want to be able to search current/incoming students by preferred or expected grad year.

bennettforkner:
Related to #1615 as people search is not currently showing alumni before 2002.

PR #1844, opened 2023-06-06
Implement people search by graduation year

antoniavonto:
As seen below, we have a visual representation of being able to search alumni based on the graduation year. For this to work, we also made changes to the backend API. We have been able to implement search based on graduation year but haven't been able to implement searching based on a range of years (potentially a slider). See: #1287 

image


antoniavonto:
See: https://github.com/gordon-cs/gordon-360-api/pull/913

bennettforkner:
Why is this a separate search field from "Class Year"? Are they not the same thing?

I imagine we should be defaulting to preferred no matter what anyway.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -80,13 +81,14 @@ const searchPageTitle = (
 const defaultSearchParams: PeopleSearchQuery = {
   includeStudent: true,
   includeFacStaff: true,
-  includeAlumni: false,
+  includeAlumni: true,
```
bennettforkner, line 84: I think this should stay as false. I get that it's confusing that you need to check this and also enter a year, but we want to show only current community members unless the user requests alumni.

.env.development
```diff
@@ -7,4 +7,4 @@
  VITE_API_URL=https://360ApiTrain.gordon.edu/
 
 # @LOCALHOST
-# VITE_API_URL=http://localhost:51626/
+# VITE_API_URL=http://localhost:51631/
```
bennettforkner, line 10: Please revert this change before merge.

bennettforkner (commented):
> I have changed the naming before seeing your previous message about the name "Academic year". I changed from "class year" to "class standing" and from "preferred graduation year" to "graduation year." What would be the most clarifying?

I like "class standing" better. Cool!

I still think it should all be within the Student/Alumni section though.

antoniavonto:
We added the switch component to toggle between the two different search options based on graduation year or a range of grad years. We also made changes to the API to make the range of graduation years to work. There are still some things that we would like to improve for this project before finalizing it.  
image
image

bennettforkner:
> We added the switch component to toggle between the two different search options based on graduation year or a range of grad years. We also made changes to the API to make the range of graduation years to work. There are still some things that we would like to improve for this project before finalizing it.

Since this is already visually a switch, I would just call it "Search by Year Range"

bennettforkner:
Also can you please address my suggestion that this all go under a single Student/Alumni section? `Student/Alumni` is one section now. You don't see that as the header because you are of person type "student":
image

antoniavonto:
We have updated/changed all necessary suggestions we received. We fixed the bug so that someone would not be able to search for the class standing and graduation year parameters at the same time. The images below are the updated visual changes. 
image
image
for mobile:
image

amos-cha:
@antoniavonto your merge with develop created duplicate services. Fix this before we can proceed with reviews!

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -139,6 +150,13 @@ const SearchFieldList = ({ onSearch }: Props) => {
   const [departments, setDepartments] = useState<string[]>([]);
   const [buildings, setBuildings] = useState<string[]>([]);
   const [halls, setHalls] = useState<string[]>([]);
+  const [userProvidedYear, setUserProvidedYear] = useState(new Date().getFullYear());
```
EjPlatzer, line 153: This doesn't need to be a state variable, since it's never updated. You can make it a constant outside the component.

EjPlatzer: Also, it's probably clearer to name it `currentYear`, since this variable is not actually user provided.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -139,6 +150,13 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+  const [graduationYearRange, setGraduationYearRange] = useState<number[]>([
+    1889,
+    userProvidedYear,
+  ]);
```
EjPlatzer, line 157: I think that this doesn't need to be it's own state value, since we're already storing the initial year and final year in the search params. You can just provide defaults when the search param values are empty strings.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -421,6 +481,49 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+                      label="Search by Year Range"
+                      labelPlacement="end"
+                    ></FormControlLabel>
+                  ) : (
+                    <FormControlLabel
+                      control={<Switch onChange={handleSwitchChange}></Switch>}
+                      label="Search by Graduation Year"
+                      labelPlacement="end"
+                    ></FormControlLabel>
+                  )}
```
EjPlatzer, line 526: We can condense this conditional case - since the FormControlLabel is always rendered and it's content is almost exactly the same, except the label:
```suggestion
                 <FormControlLabel
                      control={<Switch onChange={handleSwitchChange} />}
                      label={switchYearRange ? "Search by Year Range" : "Search by Graduation Year"}
                      labelPlacement="end"
                   />
```

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -85,12 +92,16 @@ const defaultSearchParams: PeopleSearchQuery = {
...
   residence_hall: '',
-  class_year: '',
+  class_standing: '',
+  graduation_year: '',
   home_town: '',
   state: '',
   country: '',
   department: '',
   building: '',
+  //graduation_year_range: '',
```
amos-cha, line 102:
```suggestion
  //graduation_year_range: '',
```
Is this meant to be commented out?

src/views/PeopleSearch/components/SearchFieldList/index.tsx
amos-cha, line 157:
```suggestion
  const [graduationYearRange, setGraduationYearRange] = useState<number[]>([
    1889,
    userProvidedYear,
  ]);
```

why 1889?

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -421,6 +481,49 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+                        value: (userProvidedYear - i).toString(),
+                        label: (userProvidedYear - i).toString(),
+                      }))}
+                      Icon={FaCalendarTimes}
+                      disabled={!searchParams.includeAlumni}
+                      select
+                    />
+                  ) : (
+                    <Grid item width={225}>
+                      <Slider
```
EjPlatzer, line 499: I like the idea of a slider here, but I'm concerned it may be the wrong choice in terms of UX. Since the range is several hundred years, it takes a lot of effort to get the value exactly what you want it to be. Consider someone on a touch screen who has slightly shaky hands - they would probably have great difficult setting this to be exactly 1970-1974, for example. I think in the interest of UX, we might need to change this to two separate year inputs.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -245,19 +263,61 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+      setSearchParams((sp) => ({
+        ...sp,
+        ...{ initial_year: '', final_year: '' },
+      }));
+    } else {
+      setSearchParams((sp) => ({
+        ...sp,
+        ...{ initial_year: '', final_year: '' },
+      }));
+    }
```
amos-cha, line 317: Is there a reason the if/else block has the exact same block of code?

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -421,6 +481,49 @@ const SearchFieldList = ({ onSearch }: Props) => {
                     select
                     disabled={!searchParams.includeStudent}
                   />
+                  {switchYearRange == true ? (
+                    <SearchField
+                      name="graduation_year"
+                      value={searchParams.graduation_year}
+                      updateValue={handleUpdate}
+                      options={Array.from({ length: userProvidedYear - 1889 + 1 }, (_, i) => ({
```
amos-cha, line 489: again why 1889?

amos-cha (changes requested):

EjPlatzer (changes requested):
This is awesome! I have several notes on how the code can be cleaned up. The most notable change I'm recommending is that the `Slider` component is probably not ideal for this use case, although I like the idea.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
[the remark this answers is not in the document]
antoniavonto: I forgot to remove this comment, thanks for pointing it out!

antoniavonto: I forgot to remove this comment. Thanks for pointing it out!

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: was the year the school was founded

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: No reason just mistake, just removed after testing it.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: It is the year the school was founded

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -245,19 +263,61 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+        ...{ graduation_year: '' },
+      }));
+    }
+
     setSearchParams((sp) => ({
       ...sp,
       [event.target.name]:
         event.target.type === 'checkbox' ? event.target.checked : event.target.value,
     }));
+  };
```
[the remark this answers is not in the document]
antoniavonto: Did not know about that, thank you!

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -149,7 +167,7 @@ const SearchFieldList = ({ onSearch }: Props) => {
       // Only students and facstaff search students by default - alumni aren't allowed to search students
       includeStudent: isStudent || isFacStaff,
       // Only alumni search alumni by default
-      includeAlumni: isAlumni,
+      // includeAlumni: isAlumni,
```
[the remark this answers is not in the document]
antoniavonto: I am not quite sure, but if not commented out the checkbox for including searching for alumni will be set checked as the default.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -139,6 +150,13 @@ const SearchFieldList = ({ onSearch }: Props) => {
   const [departments, setDepartments] = useState<string[]>([]);
   const [buildings, setBuildings] = useState<string[]>([]);
   const [halls, setHalls] = useState<string[]>([]);
+  const [userProvidedYear, setUserProvidedYear] = useState(new Date().getFullYear());
+  const [graduationYearRange, setGraduationYearRange] = useState<number[]>([
+    1889,
+    userProvidedYear,
+  ]);
+  const [switchYearRange, setSwitchYearRange] = useState(true);
+  const [studentSearch, setStudentSearch] = useState(Boolean);
```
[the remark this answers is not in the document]
antoniavonto: Not actually used studentSearch, just removed it

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: Yes that is true, thank you! Just changed that

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -245,19 +263,61 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+  };
 
   const handleEnterKeyPress = (event: KeyboardEvent<HTMLDivElement>) => {
     if (event.key === 'Enter') {
       search();
     }
   };
 
+  const handleSliderChange = (event: Event, newValue: number | number[]) => {
+    setGraduationYearRange(newValue as number[]);
```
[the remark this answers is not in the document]
antoniavonto: When I try making it as:
image
I get an error with the on change of the slider that says: Type 'number' is not assignable to type 'number[ ]'. This leads me to assume that newValue might not be always a number[ ], but I cannot see why.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: That is true, it would make it more concise. Just changed it, thank you!

src/views/PeopleSearch/components/SearchFieldList/index.tsx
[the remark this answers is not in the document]
antoniavonto: I will not get too into this issue since we are thinking of moving away from the use of the slide

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: I agree with you the slider is probably not the best option for this feature. We were thinking of some options and wanted your opinion. 

1. Keeping the slide, but also allowing some other way of changing the year range.
2. Making a select that allow you to choose two dates for the range
3. Making two text boxes for the initial and end year (Not my preferred option since people might type other things and it might not be intuitive for the user)

amos-cha: I honestly don't know if there is a really good option. I think in terms of mobile usability I would argue that a location where we could select a year would be optimal. However, I don't know if the practicality of a year selector is much better (if any better) than the live updating of the slider.

.env.development
```diff
@@ -8,3 +8,4 @@
 
 # @LOCALHOST
 # VITE_API_URL=http://localhost:51626/
+
```
amos-cha, line 11:
```suggestion
```

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -149,7 +162,7 @@ const SearchFieldList = ({ onSearch }: Props) => {
       // Only students and facstaff search students by default - alumni aren't allowed to search students
       includeStudent: isStudent || isFacStaff,
       // Only alumni search alumni by default
-      includeAlumni: isAlumni,
+      // includeAlumni: isAlumni,
```
amos-cha, line 165: reminder to uncomment or remove

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -421,6 +474,41 @@ const SearchFieldList = ({ onSearch }: Props) => {
                     select
                     disabled={!searchParams.includeStudent}
                   />
+                  {switchYearRange == true ? (
+                    <SearchField
+                      name="graduation_year"
+                      value={searchParams.graduation_year}
+                      updateValue={handleUpdate}
+                      options={Array.from({ length: currentYear - 1889 + 1 }, (_, i) => ({
+                        value: (currentYear - i).toString(),
```
amos-cha, line 483:
```suggestion
                      options={Array.from({ length: currentYear - 1889 + 1 }, (_, i) => ({
                        value: (currentYear - i).toString(),
```

I would suggest making a note somewhere that 1889 is the establish date of Gordon (thus earliest possible year)

amos-cha (changes requested):

amos-cha (dismissed):
Looks good, I think it's best to leave figuring out the UX as an issue. I don't think it is a necessarily blocking change.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: I could, instead of unsing the set function to change graduationYearRange to just separate straight the newValue, but then it would be hard to implement the visualization of the hands of the slider changing.
image
image

russtuck: @EjPlatzer are you satisfied with leaving this as it is?

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -40,12 +41,18 @@ import {
...
 import { compareByProperty, searchParamSerializerFactory } from 'services/utils';
 import { gordonColors } from 'theme';
 import SearchField, { SelectOption } from './components/SearchField';
+import addressService from 'services/address';
+import Slider from '@mui/material/Slider';
+import Switch from '@mui/material/Switch';
+
+function valuetext(value: number) {
+  return '${value}';
+}
```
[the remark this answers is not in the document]
russtuck: @antoniavonto this looks like it's still a to-do item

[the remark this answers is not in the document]
russtuck: Nevermind, it looks done.  Sorry.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -245,19 +263,61 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
 
   const handleEnterKeyPress = (event: KeyboardEvent<HTMLDivElement>) => {
     if (event.key === 'Enter') {
       search();
     }
   };
 
+  const handleSliderChange = (event: Event, newValue: number | number[]) => {
+    setGraduationYearRange(newValue as number[]);
+    var values = graduationYearRange.toString().split(',');
```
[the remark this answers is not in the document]
russtuck: @antoniavonto Have you addressed and tested the reliability concern?

russtuck:
It looks like this is close.  I commented on what appear to be the remaining issues.

antoniavonto:
@amos-cha

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -415,6 +465,41 @@ const SearchFieldList = ({ onSearch }: Props) => {
                     select
                     disabled={!searchParams.includeStudent}
                   />
+                  {switchYearRange == true ? (
```
amos-cha, line 468:
```suggestion
                  {switchYearRange ? (
```

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -239,19 +249,59 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+      final_year: values[1],
+    }));
+    setSearchParams((sp) => ({
+      ...sp,
+      class_standing: '',
+    }));
+  };
+
+  const handleSwitchChange = () => {
+    if (switchYearRange === true) {
```
amos-cha, line 295:
```suggestion
    if (switchYearRange) {
```

amos-cha (changes requested):
Almost, then it looks good

amos-cha (approved):
👍🏼

antoniavonto:
@amos-cha

amos-cha (dismissed):

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -260,6 +260,7 @@ const SearchFieldList = ({ onSearch }: Props) => {
       setSearchParams((sp) => ({
         ...sp,
         class_standing: '',
+        minor: '',
```
antoniavonto, line 263: Alumni's Minors are not saved, so the results would always be empty if graduation year was searched together with a minor

antoniavonto, line 263: Alumni's Minors are not saved, so the results would always be empty if graduation year was searched together with a minor

amos-cha (approved):

src/views/PeopleSearch/components/SearchFieldList/index.tsx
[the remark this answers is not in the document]
antoniavonto: @EjPlatzer Since we created a new issue for finding a new solution in the place of the slider, is it ok to leave this like this?

PR #1845, opened 2023-06-07
Use `date-fns` `formatDistance` instead of custom

EjPlatzer:
The `date-fns` package that we use for simple date utilities has a `formatDistance` (and `formatDistanceToNow`) function that creates a string saying how far away the passed-in date is from now in human-friendly language. I previously created a custom implementation of such a function. It is better to use the `date-fns` implementation, so that we don't have to maintain it ourselves. Their implementation also does more thorough error-checking and has better support for old browsers (and will be updated as browser tech updates).


bennettforkner (approved):
LGTM. Assuming this is tested and works

Issue #1196, opened 2021-06-18
Use new logo for people with no photo


russtuck:
Neal Ericsson has a new "lion" logo, which we should use as an update for the current one when a person has no photo.

ChrisImagineer:
How about using the Mascot head for a missing photo
![Mascot_head-solo](https://user-images.githubusercontent.com/43452258/135862718-6774dc08-e1a7-43eb-add2-6e0628081169.jpg)
.

josh-peters-2357:
@ChrisImagineer Chris, this image is pretty cool but it is not transparent. Thinking ahead to when we might add dark mode, there will be no easy fix, so it would be easier to just have transparency now. Is there a copy of this picture like that?

cpabbot:
Here it is with transparency. That being said, this might look good with a white background regardless.

![135862718-6774dc08-e1a7-43eb-add2-6e0628081169-removebg-preview](https://user-images.githubusercontent.com/17221652/147250821-b1f0d9f2-c280-45ab-a5bf-fc9b049c4500.png)

russtuck:
Looking at who has the default photo, it looks like mostly people who haven't been on campus (on-line program, incoming students).  So the current picture, which doesn't draw attention, might be appropriate.  Or maybe something even more in the background.  If this is correct, is it best to leave it alone and not spend more time on it?

ChrisImagineer:
Best to leave it alone, and I could have a conversation with the grad programs to see what they’d like to do.
It seems more like an onboarding issue for welcoming new grad students. So this might just be a soft-solve, not a technical one. But well spotted as a UX issue.
Chris

On 06/16/23, 14:08, "Russ Tuck" ***@***.******@***.***>> wrote:


Looking at who has the default photo, it looks like mostly people who haven't been on campus (on-line program, incoming students). So the current picture, which doesn't draw attention, might be appropriate. Or maybe something even more in the background. If this is correct, is it best to leave it alone and not spend more time on it?

ArabellaJi:
If someone continues to work on this in the future, here's some code that might be helpful.
UI: https://github.com/gordon-cs/gordon-360-ui/pull/1846
API: https://github.com/gordon-cs/gordon-360-api/pull/914

russtuck:
I think the current default image is appropriate.  There's work referenced on how to make a change if needed in the future.

PR #1846, opened 2023-06-07
Change the default Gordon photo to the Lion's Head

ArabellaJi:
This is a draft pull request for issue #1196. Previously, 360 had been using the default Gordon image on the back end when users did not upload their profile photo. However, we now want to put the default Gordon image into the public/images path on the front end and replace it with the new image (Lion's Head) we want.

- [x] store the default Gordon image in the front-end and call it where needed
- [x] replace the default Gordon image with the new lion's head image
- [ ] scale the image properly and test it in different situations (How will it look like in different places (in profile, in the right corner of header, ...)? How does scaling the default Gordon image affect the way other resume images are presented and how can this be avoided?)


ArabellaJi:
Now the profile looks like this: :monocle_face:

image

EjPlatzer:
> This issue is on hold. See more discussions in #1196.

@ArabellaJi Would it make sense to close this PR and the API one for now? We can always re-open them later if they become active again, and they're referenced on the issue so the work won't disappear.

ArabellaJi:
@EjPlatzer That makes sense. Will do!

Issue #879, opened 2020-08-23
Schedule Fixes


russtuck:
(Per administrator ChrisImagineer's request) Make the Schedule viewable for a Student Profile. Consider making it viewable for Faculty, but optional to display for Staff/Administration. Make it not viewable on an Alumni profile.

Remove public toggle; student information and safety/security. Administrator's decision.

(lavenwinkle) The calendar lists quad 4 classes along with quad 3 classes right now, even though quad 4 classes have not started.

(russtuck) It's almost time for the Spring Semester, so I'm updating my schedule with my new office hours. But it still shows last fall's classes. Spring classes would be much more useful!

It would also be more useful to see other people's coming class schedules, not their old ones.

Maybe switch over at the halfway point between terms? Switch from fall to spring on Jan 1?

The "schedule" part of the profile page was designed to capture faculty office hours.  However, it is currently broken so it's impossible to add office hours.  (It's also not obvious how to delete them, and I'm not sure if they are editable.)

In addition, there are severe limits on the "Description" field which limit its use for remote students.  It silently deletes parts of URLs (everything after "?"), so it breaks a Zoom link with password.  It also only allows one URL, so it doesn't allow both a sign-up page and a Zoom link.  And the total length makes it hard to describe both in-person and online options.

Priorities:
1) Allow adding events (office hours)
2) Allow longer description with multiple full URLs
3) Improve event editing and deleting UI

#### Related Issues
#1839 
#1827 
#1326

andrew-wzj:
api related issues
 1. fix the description ( the reason why the results not immediately visible)
 2. handle qual courses
 3. (optional) add term selection

bennettforkner:
> @bennettforkner The edit description function seems to work correctly on Gordon360 (not the 360train). Is there a way we could get the code from the current 360 to fix the function for editing the description function on the schedule panel?

Not sure what you mean by `the current 360`. There are two primary branches in both 360 repository. `develop` corresponds to 360train.gordon.edu and `master` corresponds to 360.gordon.edu.

bennettforkner:
You can view the differences between train and production here: https://github.com/gordon-cs/gordon-360-ui/compare/master...develop

It looks like Evan made some changes to the schedule that have not yet been released to `master`:
![image](https://github.com/gordon-cs/gordon-360-ui/assets/43764277/c59f5e71-bb1b-4c93-9589-4def27d6173c)

PR #1847, opened 2023-06-07
S23 schedule fixes

andrew-wzj:
- [partially done ] expand schedule panel by default, since it's only visible when not empty (nice, but optional: make expand/contract sticky, so it stays until the user changes it)
  this broken feature could be fixed in better way, but today we just use defaultExpanded as the solution.

- [ done] only show 8am-9pm, Mon-Sat

#879


.env.development
```diff
@@ -7,4 +7,4 @@
  VITE_API_URL=https://360ApiTrain.gordon.edu/
 
 # @LOCALHOST
-# VITE_API_URL=http://localhost:51626/
+#  VITE_API_URL=http://localhost:51626/
```
amos-cha, line 10: revert

amos-cha (changes requested):

PR #1849, opened 2023-06-08
SSH port forwarding port order correction and other tweaks

jsenning:
The order of ports in the "ssh -R" port forwarding command was reversed so that was updated.  Other changes were made to make terms consistent.

### DIFF
--- a/docs/Developer's Guide.md
+++ b/docs/Developer's Guide.md
@@ -108,24 +108,26 @@ Sometimes, you would like to connect the frontend on your local computer to the 
 
 1. Your local machine must be configured as an SSH host.
 
-   [Windows Installation](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse?tabs=gui).
+  - Windows [Instructions](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse?tabs=gui).
 
-   - If you are on Windows 11+, you may need to update the firewall to allow inbound connections on a public network.
+    - If you are on Windows 11+, you may need to update the firewall to allow inbound connections on a public network.
       - Search for `Check Firewall Status` in the start menu
       - Click `Turn Windows Defender Firewall on or off`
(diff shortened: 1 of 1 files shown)


docs/Developer's Guide.md
```diff
@@ -125,24 +125,20 @@ Sometimes, you would like to connect the frontend on your local computer to the
 
 2. Create the SSH tunnel
 
-   After setting up the ssh server on your host machine, find your host IP or DNS address. Open a command prompt or terminal window and enter the following:
+   After setting up the ssh server on your local machine, find your host IP or DNS address.  Open a command prompt or terminal window and enter `ipconfig` (Windows), `hostname` and/or `ifconfig` (MacOS), or `hostname -I` and/or `hostname -f` (Linux).  You will need to know this so you can connect to your local machine from the virtual machine.
+   
+   Log into the VM and open a Git Bash shell or a command prompt/powershell window. Enter the following command:
```
bennettforkner, line 130: `hostname` works on Windows as well. That is why I wrote hostname for all of them. I would rather not confuse people with a bunch of different cases if we don't have to.

jsenning: Okay.  Further down the instructions said to find the IP number or hostname using ipconfig on Windows, ifconfig on mac and linux so I just expanded that and moved it up.  On linux ifconfig has been replaced by ip (which I don't like nearly as well as the output is not as clean).  Also the hostname command on Linux may or may not return the FQDN (depending on how the host name was set), so the "-f" option is helpful.  Together this just means that "hostname" by itself does not work for all platforms.
I'll make some adjustments to the doc.

jsenning: I've tried commands on Windows and Linux, as well as tried the hostname and ifconfig commands on MacOS.  I think some more detail is required than that provided by just hostname on the various platforms since we need an FQDN or IP number.  The latest revisions are even a bit longer but hopefully clearer.

bennettforkner (approved):
LGTM. Good catch on the reversal of those ports!

PR #1854, opened 2023-06-09
S23 Home Page Headers

jhammond4:
Changes made to add headers to home page tabs for consistency. The CLWCreditDaysLeft file may not be correct-I'm not able to see this tab so I have no clue if it looks how it's supposed to


jhammond4:
I found out the CLWCreditDaysLeft tab is incorporated into the DaysLeft tab which I was unaware of, so disregard what I said about it here.

src/views/Home/components/CLWCreditsDaysLeft/index.jsx
```diff
@@ -158,7 +158,13 @@ const CLWCreditsDaysLeft = () => {
       <CardContent>
         <Grid container direction="row" alignItems="center">
           <Grid item xs={7} align="left">
-            <CardHeader title={currSessionDescription} />
+            <CardHeader
+              title={currSessionDescription}
+              style={{
+                backgroundColor: gordonColors.primary.blue,
+                color: gordonColors.neutral.grayShades[50],
+              }}
```
EjPlatzer, line 166: Please do not use inline styles. Instead, it would be better to add a CSS class for the header that can be re-used across the site.

You can add a class (such as "gc360_card_header") into `src/app.global.scss` and it will work no matter what component you use it from.

src/views/Home/components/NewsCard/index.jsx
```diff
@@ -20,12 +21,25 @@ const NewsCard = () => {
...
             </Grid>
             <Grid item xs={5} align="right">
-              <Button variant="contained" color="secondary" component={Link} to="/news">
+              <Button
+                variant="contained"
+                //color="secondary"
+                style={{
+                  backgroundColor: gordonColors.primary.cyan,
+                  color: gordonColors.neutral.grayShades[50],
+                }}
```
EjPlatzer, line 30: The `color=secondary` prop makes this button cyan in a way that is consistent with the rest of the site. Is there a reason you replaced that with these inline styles?

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -6,6 +6,7 @@ import session from 'services/session';
 import { gordonColors } from 'theme';
 import styles from '../Doughnut.module.css';
 import { Card, CardContent, CardHeader, Grid, Typography } from '@mui/material';
+import { NoBackpackSharp } from '@mui/icons-material';
```
EjPlatzer, line 9: This looks like an unused import that may have been accidentally added. Please remove it if you don't plan to use it.

EjPlatzer (changes requested):
Thanks @jhammond4 for working on making our styling more consistent!

I left a couple of comments about best practices for styling and other small things.

I also want to point out that we sometimes intentionally leave card headers neutral colored - too much of our primary color can lead to the page looking busy and over-saturated with content, defeating the purpose of highlighting specific content with the blue headers. There's a principle in UI Design called the 60-30-10 rule that says that your page should be 60% neutral, 30% primary theme color, and 10% accent color.

src/views/Home/components/DaysLeft/index.jsx
jhammond4: I'm not sure where it came from-thanks for the catch, I'll remove it

src/views/Home/components/NewsCard/index.jsx
jhammond4: If I remember right at this point I was doing a lot in inline styles and saw a few other places using inline-I'll switch it back and keep it as it was

jhammond4:
With the 60/30/10 rule, do you want me to leave the headers as is? I thought it looked better with the headers because it makes each section stand out more, but I didn't take into account if it'd be too much of one color

EjPlatzer:
> With the 60/30/10 rule, do you want me to leave the headers as is? I thought it looked better with the headers because it makes each section stand out more, but I didn't take into account if it'd be too much of one color

I won't tell you which is better - that's up to you and the team to decide. I just wanted to mention the rule and let you know that sometimes less is more, but I don't know whether that applies in this case or not. It would he helpful to include screenshots in the PR so we can see the before and after - especially on different screen sizes.

src/views/Home/components/CLWCreditsDaysLeft/index.jsx
```diff
@@ -59,7 +59,7 @@ const CLWCreditsDaysLeft = () => {
         // Allow different tooltips for different datasets within the same pie;
         callbacks: {
           // Code taken from https://github.com/chartjs/Chart.js/issues/1417
-          label: function(item, data) {
+          label: function (item, data) {
```
russtuck, line 62:
```suggestion
          label: function(item, data) {
```
The extra space looks like it might have been accidental.

russtuck:
@jhammond4 is styling well known enough now to address the comments?

jhammond4:
I'm planning to return to this to make sure the styling is done consistently with the new css format.

jhammond4:
Screenshot from 2023-06-30 14-27-45
Screenshot from 2023-06-30 15-35-47
Here is a before and after picture for the tabs with headers.

jhammond4:
Screenshot from 2023-06-30 16-37-21
This is the mobile view-I forgot to add it with the others

jhammond4:
Closing this, as it is being consolidated into https://github.com/gordon-cs/gordon-360-ui/pull/1980.

PR #1855, opened 2023-06-12
Event page header

jhammond4:
Added a header to the event page tab. I moved the Attended CL&W button down by the buttons by the search bar. I also added a function that checks the screen's width and accordingly modifies how long the search bar is so all of the buttons can fit beside it without overflowing to an extra row. The color of "Gordon" in the title was dark blue, so I switched it to cyan to go along with similar titles on other pages.

### DIFF
--- a/src/views/Events/index.jsx
+++ b/src/views/Events/index.jsx
@@ -23,7 +23,6 @@ import Media from 'react-media';
 import gordonEvent, { EVENT_FILTERS } from 'services/event';
 import { gordonColors } from 'theme';
 import { useLocation, useNavigate } from 'react-router';
-import { NoBackpackSharp } from '@mui/icons-material';
 
 const Events = () => {
   const [open, setOpen] = useState(false);
--- a/src/views/Events/index.jsx
+++ b/src/views/Events/index.jsx
@@ -110,31 +110,31 @@ const Events = () => {
     setURLParams(!includePast, filters);
   };
 
-  /*const searchWidth = () => {
+  function searchBarLength() {
     if (width >= 1800 && width <= 1920) {
-      return 5.87;
+      return 7.19;
     } else if (width >= 1700 && width < 1800) {
-      return 6.39;
+      return 6.89;
     } else if (width >= 1600 && width < 1700) {
-      return 6.91;
+      return 6.55;
     } else if (width >= 1500 && width < 1600) {
-      return 7.43;
+      return 6.31;
     } else if (width >= 1400 && width < 1500) {
-      return 7.95;
-    } else if (width >= 1300 && width < 1400) {     //I tried to make a function to check the window size and adjust the search bar's width accordingly-I can't get it to work. I may come back to it eventually
-      return 8.47;
+      return 7.15;
+    } else if (width >= 1300 && width < 1400) {
+      return 6.76;
     } else if (width >= 1200 && width < 1300) {
-      return 8.99;
+      return 6.3;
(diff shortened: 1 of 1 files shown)


src/views/Events/index.jsx
```diff
@@ -109,6 +109,32 @@ const Events = () => {
...
+    } else if (width >= 1100 && width < 1200) {
+      return 6.93;
+    } else if (width >= 1000 && width < 1100) {
+      return 6.39;
+    } else if (width >= 920 && width < 1000) {
+      return 5.87;
+    } else {
+      return 5.87;
+    }
+  }
```
EjPlatzer, line 136: Can you explain the goal of this function? It looks like you're trying to dynamically resize the search bar, but I'm not sure what the significance of each size is - were you trying to prevent it from overflowing onto multiple lines? If so, the simpler way to do that is to pass `true` as the value of the `xs` prop for the Grid item in questions. With `xs={true}`, the Grid item will dynamically resize itself to fill the leftover space without overflowing.

src/views/Events/index.jsx
```diff
@@ -141,34 +167,175 @@ const Events = () => {
...
-                  <CardHeader title={searchPageTitle} />
+          <Card style={{ padding: '0 0vw' }}>
+            <CardHeader
+              title={searchPageTitle}
+              style={{
+                backgroundColor: gordonColors.primary.blue,
+                color: gordonColors.neutral.grayShades[50],
+              }}
+            />
+            <Card style={{ padding: '0 3vw' }}>
```
EjPlatzer, line 178: Is this nested card needed?

EjPlatzer: I only ask because it seems odd to have a Card with a header, and then instead of content have another card which has content,

src/views/Events/index.jsx
```diff
@@ -141,34 +167,175 @@ const Events = () => {
...
+              <CardContent>
+                <Grid container direction="row" alignItems="center">
+                  <Grid item xs={4} />
+                  <Grid item xs={4} align="center"></Grid>
+                  <Grid item xs={4} align="right"></Grid>
+                </Grid>
```
EjPlatzer, line 184: This whole `Grid` container is defunct now, since it only held the heading before, right?

EjPlatzer (dismissed):
This looks nice! I left a couple comments on ways this could be cleaner.

src/views/Events/index.jsx
jhammond4: That was the idea of this-I didn't know about that, it'd be much simpler than this function. I'll look into it!!

jhammond4:
I'm planning to return to this as well to make sure it is consistent with the new css styling and address these fixes.

src/views/Events/index.jsx
jhammond4: If I remember right, I did this this way to make sure the cards stayed separate. There may be a better way to accomplish the same thing-I didn't experiment much with it.

src/views/Events/index.jsx
jhammond4: If the cards stay separate, I think that'd be the case. I'll keep this in mind

jhammond4:
Screenshot from 2023-07-03 13-25-52
Screenshot from 2023-07-03 15-08-09
Here's some before and after screenshots I forgot about.

jhammond4:
Closing this, as it is being consolidated into https://github.com/gordon-cs/gordon-360-ui/pull/1980.

PR #1856, opened 2023-06-12
People Page Header

jhammond4:
Added a header to the main tab on the people page to be more consistent with other tabs on various pages and changed the color of "Gordon" on the header to cyan for consistency as well

### DIFF
--- a/src/views/PeopleSearch/components/SearchFieldList/index.tsx
+++ b/src/views/PeopleSearch/components/SearchFieldList/index.tsx
@@ -44,9 +44,9 @@ import { useNavigate } from 'react-router-dom';
 import { AuthGroup } from 'services/auth';
 import peopleSearchService, { Class, PeopleSearchQuery, SearchResult } from 'services/peopleSearch';
 import { compareByProperty, searchParamSerializerFactory } from 'services/utils';
-import { gordonColors } from 'theme';
 import SearchField, { SelectOption } from './components/SearchField';
 import addressService from 'services/address';
+import styles from './SearchFieldList.module.css';
 
 /**
  * A Regular Expression that matches any string with any alphanumeric character `[a-z][A-Z][0-9]`.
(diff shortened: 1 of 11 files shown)


src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -72,7 +72,7 @@ const relationship_statuses = [
 const searchPageTitle = (
   <>
     Search the
-    <b style={{ color: gordonColors.primary.blue }}> Gordon </b>
+    <b style={{ color: gordonColors.primary.cyan }}> Gordon </b>
```
EjPlatzer, line 75: Thanks again for paying attention to consistency issues. However, if this is inconsistent with other pages (which I know it currently is with the involvements page), then it is the other pages that need to change, not this one. 

This text was previously cyan, but we changed it to blue because the cyan on white is insufficient color contrast for accessibility.

jhammond4: A lot of pages don't use a header at the top of each page, but on pages with multiple tabs (like Involvements or Events) I thought it made more sense for each tab to have a header so they all looked the same, then I wanted to add headers across the site so it is the same everywhere. Should I leave them in or leave them as is?

For the cyan color, there are other places on the website that use cyan on white (like Involvements or Events), but this is the only page I could find with a dark blue "Gordon" on white. Should I still change it back to dark blue?

EjPlatzer: Yeah, we should definitely leave this one as blue. We should probably also make an issue to find everywhere that has cyan on white because that is not accessible to people with certain disabilities.

russtuck:
@jhammond4 is this still and active PR?  If so, adding before and after screenshots and addressing the comments so far will help it move forward.  If not, consider making it a draft if you plan to work on it later, or closing it if another PR will serve the same purpose.

jhammond4:
I'm planning to return to this later, I would like to make sure the css styling used in consistent with the new format.

jhammond4:
Screenshot from 2023-07-05 09-51-00
Screenshot from 2023-07-05 09-42-27
Here's before and after pictures of the page

PR #1864, opened 2023-06-13
S23 CSS Theme Palette Implementation

Centerville1:
First progress on converting colors to use CSS variables and the theme palette.


Centerville1:
@EjPlatzer after talking with Dr. Tuck, he suggested we try and merge our progress into develop as in smaller parts as we go, to try to reduce merge conflicts with everyone else's changes (if we waited until we were finished).  So let me know if you see any changes needed, we've only changed a few components over so far, and some of those probably need to be converted to CSS later, instead of using in-line.

src/_vars.scss
```diff
@@ -54,3 +58,41 @@ $break-xl: 1536px;
...
+// $neutral-black-opacity20: #00000033;
+// $neutral-black-opacity15: #00000026;
+// $neutral-black-opacity125: #00000020;
+// $neutral-black-opacity10: #0000001a;
+
+// $dark-text-color: rgba(0, 0, 0, 0.87);
+
+// $wellness-red: #a30000;
+// $wellness-green: #005500;
+// $wellness-yellow: #ffcc00;
```
EjPlatzer, line 98: This seems unnecessary, since you still have the original values above.

src/app.global.scss
```diff
@@ -43,24 +43,25 @@ a {
...
   &:hover {
-    color: $primary-cyan !important;
+    color: var(--mui-palette-secondary-main) !important;
   }
 }
 
 @media (max-width: $break-md) {
   .gc360_a_link {
-    color: $primary-cyan !important;
+    color: var(--mui-palette-secondary-train) !important;
```
EjPlatzer, line 64: Is this supposed to be `main` instead of `train`?
```suggestion
    color: var(--mui-palette-secondary-main) !important;
```

src/components/GordonDialogBox/GordonDialogBox.module.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../vars';
+
+.gc360_gordondialogbox {
```
EjPlatzer, line 3: You created this file, but didn't edit the GordonDialogBox component at all? Or did you not commit the changes to the component yet?

src/components/GordonDialogBox/GordonDialogBox.module.scss
```diff
@@ -0,0 +1,18 @@
...
+    color: $neutral-white;
+    text-align: center;
+  }
+  &_content :global(.MuiDialogTitle-root) {
+    padding: 8px 20px 8px 16px;
+  }
+  &_actions {
+    justify-content: space-evenly;
+  }
+}
```
EjPlatzer, line 18: Also, since this is a CSS module, we don't the very specific and long naming scheme.

It would be easier and clearer to flatten the CSS selectors and give them simpler, more clear names, e.g.
```suggestion
.dialog {
  display: grid;
  grid-template-rows: auto auto auto;
  overflow-y: auto;
}

.title {
    background-color: var(--mui-palette-primary-main);
    color: $neutral-white;
    text-align: center;
}

.content :global(.MuiDialogTitle-root) {
    padding: 8px 20px 8px 16px;
}

.actions {
    justify-content: space-evenly;
}
```

src/components/Header/Header.module.scss
```diff
@@ -65,6 +65,7 @@
     .app_bar {
       background-image: url('./gc_seal_top_64.png');
       height: $header-height-sm;
+      background-color: var(--mui-palette-primary-main);
```
EjPlatzer, line 68: I think you missed the `$neutral-white` at the bottom of this component.

src/components/Header/index.jsx
```diff
@@ -140,7 +140,7 @@ const GordonHeader = ({ onDrawerToggle }) => {
 
   return (
     <section className={styles.gordon_header}>
-      <AppBar className={styles.app_bar} position="static">
+      <AppBar className={styles.app_bar} position="static" color="primary">
```
EjPlatzer, line 143: I believe this is unnecessary, since `primary` is the default value of the `color` prop for `AppBar`.

src/components/NavButton/index.jsx
```diff
@@ -25,7 +25,13 @@ const GordonNavButton = ({
 }) => {
   const link =
     unavailable !== null || linkPath === null ? (
-      <ListItem divider={divider} button onClick={onLinkClick} disabled={unavailable !== null}>
+      <ListItem
+        divider={divider}
+        button
+        onClick={onLinkClick}
+        disabled={unavailable !== null}
+        className="gc360_link"
```
EjPlatzer, line 33: What was the reasoning behind applying the link styles to these? I'm not against the change necessarily, but I think it's outside the scope of this PR - this is a change to the expected styles of an element, whereas this PR should mostly be pure refactoring with on actual changes to the page.

src/components/Profile/components/Identification/Identification.module.scss
```diff
@@ -27,7 +27,7 @@ $image-side-width-desktop: 90px;
   box-shadow: 0px 2px 1px -1px $neutral-black-opacity20, 0px 1px 1px 0px $neutral-black-opacity15,
     0px 1px 3px 0px $neutral-black-opacity125;
   &_header {
-    background-color: $primary-blue;
+    background-color: var(--mui-palette-primary-300);
```
EjPlatzer, line 30: I know this file is massive, but it seems to me that you ought to do an entire component in one go, rather than replacing it piece meal.

src/components/Profile/components/Identification/index.jsx
```diff
@@ -63,8 +64,8 @@ const Identification = ({ profile, myProf, isOnline, createSnackbar }) => {
       },
       cancelButton: {
         backgroundColor: 'white',
-        color: gordonColors.primary.blue,
-        border: `1px solid ${gordonColors.primary.blue}`,
+        color: 'var(--mui-palette-primary-main)',
```
EjPlatzer, line 67: I think you missed a reference to `gordonColors` on line 76 below.

src/components/Profile/components/Identification/index.jsx
```diff
@@ -49,11 +50,11 @@ const Identification = ({ profile, myProf, isOnline, createSnackbar }) => {
   // Styles used throughout this component
   const style = {
     button: {
-      background: gordonColors.primary.blue,
+      background: 'var(--mui-palette-primary-main)',
       color: 'white',
```
EjPlatzer, line 54: For this component to actually be ready for the dark mode switch, all of these hard-coded `white` colors will also need to be replaced. I'm not saying you need to do that now, but it seems pointless to change only part of a component's styles, and then have to come back to it again later.

src/components/Profile/components/MembershipsList/components/MembershipInfoCard/index.jsx
```diff
@@ -15,6 +16,9 @@ const MembershipInfoCard = ({ myProf, membership, onTogglePrivacy }) => {
           className={`gc360_text_link ${
             showPrivate ? styles.private_membership : styles.public_membership
           }`}
+          style={{
+            color: newTheme.colorSchemes.light.palette.primary[300],
+          }}
```
EjPlatzer, line 21: We definitely shouldn't add **new** inline styles - and also this one is hard-coding the wrong value. I know you probably didn't add this, but it needs to be removed before we merge.

src/views/About/index.jsx
```diff
@@ -15,6 +15,8 @@ const About = () => {
     versionService.getVersion().then(setVersion);
   }, []);
 
+  const getColor = (cssVar) => getComputedStyle(document.documentElement).getPropertyValue(cssVar);
```
EjPlatzer, line 18: This is now unused in the file, please delete it.

src/views/Home/components/CLWCreditsDaysLeft/index.jsx
```diff
@@ -17,6 +25,8 @@ const CLWCreditsDaysLeft = () => {
   const [loading, setLoading] = useState(true);
   const [currSessionDescription, setCurrSessionDescription] = useState('');
 
+  const getColor = (cssVar) => getComputedStyle(document.documentElement).getPropertyValue(cssVar);
```
EjPlatzer, line 28: We probably don't want to commit this file as is, since this is not an acceptable final solution. Please undo these changes for now until we find a working solution for chart.js

src/views/Timesheets/components/ShiftItem/index.jsx
```diff
@@ -26,7 +26,7 @@ import styles from './ShiftItem.module.css';
 
 const CustomTooltip = withStyles((theme) => ({
   tooltip: {
-    backgroundColor: theme.palette.common.black,
+    backgroundColor: theme.colorSchemes.light.palette.common.black,
```
EjPlatzer, line 29: This is hard-coding the light theme, we shouldn't merge it.

src/views/Timesheets/index.jsx
```diff
@@ -39,7 +39,7 @@ const withNoSeconds = (date) => set(date, { seconds: 0, milliseconds: 0 });
 
 const CustomTooltip = withStyles((theme) => ({
   tooltip: {
-    backgroundColor: theme.palette.common.black,
+    backgroundColor: theme.colorSchemes.light.palette.common.black,
```
EjPlatzer, line 42: As above, we shouldn't merge anything that hard-codes the light theme.

src/views/PeopleSearch/index.tsx
```diff
@@ -10,6 +10,7 @@ import { SearchResult } from 'services/peopleSearch';
 import PeopleSearchResult from './components/PeopleSearchResult';
 import SearchFieldList from './components/SearchFieldList';
 import styles from './PeopleSearch.module.scss';
+import { newTheme } from 'theme';
```
EjPlatzer, line 13: This is unused in this file, please remove it.

src/views/Home/components/NewsCard/index.jsx
```diff
@@ -12,7 +13,7 @@ const NewsCard = () => {
   }, []);
 
   return (
-    <Card>
+    <Card className={styles.card}>
```
EjPlatzer, line 16: Is this necessary? The Card should have default styles, so this is either irrelevant or else you're changing the styles of this card specifically, which I don't think is in the scope of this PR.

src/views/Home/components/DiningBalance/index.jsx
```diff
@@ -6,21 +6,26 @@ import session from 'services/session';
...
-let dollarsColor = gordonColors.secondary.yellow;
-let guestColor = gordonColors.secondary.orange;
-let emptyColor = gordonColors.neutral.lightGray;
-
 const DiningBalance = () => {
   const [loading, setLoading] = useState(true);
   const [diningInfo, setDiningInfo] = useState(null);
   const [[daysRemaining, daysInSession], setDaysLeft] = useState([null, null]);
 
+  const getColor = (cssVar) => getComputedStyle(document.documentElement).getPropertyValue(cssVar);
```
EjPlatzer, line 19: As above, this is not a good final solution, so I don't think we should merge it.

EjPlatzer (changes requested):
Great work on this y'all! Really exciting to see so much progress toward a unified site with editable theming!

I left several comments mostly on small cleanups to get this high quality before merging.

src/components/GordonDialogBox/GordonDialogBox.module.scss
Centerville1: This might have been created as part of Sharon's branch, I haven't looked at this in detail yet, so correct, this is unfinished

src/components/Header/Header.module.scss
Centerville1: Yeah, I haven't gone through header fully yet(just with my initial experimentation, but this will still work for now, will be changed when we get to that component.

src/components/Header/index.jsx
Centerville1: Correct, yep, I'll fix that, again, this is still a mix of legacy and new code at this point, we have a document tracking which changes we have finished, and are doing internal code reviews to try to ensure completeness as much as possible.

src/components/Profile/components/Identification/Identification.module.scss
Centerville1: Yep, most of the individual changes were just part of my experimentation, we're now going through whole components at once to change, and since this PR I've actually fully completed this component, so I'll include it in the next merge.

src/components/Profile/components/Identification/index.jsx
Centerville1: Yep, those changes will be in the next PR.  Sorry this first one is kind of piecemeal, most of the changes here are just from my experimentation to learn how to do this.  Now that all three of us are intentionally working on whole components, future PRs will be much more organized.

EjPlatzer: No worries, that makes sense.

src/components/Profile/components/Identification/index.jsx
Centerville1: Yep, that's already been changed since this PR, it'll be in the next one

src/components/Profile/components/MembershipsList/components/MembershipInfoCard/index.jsx
Centerville1: Fair enough, yeah, I think this was from before I was working on it, so I'll revert it until we get to this file for real

src/views/About/index.jsx
Centerville1: getColor was still being used, but it was unneeded, so I'll change it.

src/views/Home/components/CLWCreditsDaysLeft/index.jsx
Centerville1: Fair enough, I'll undo those for now :+1:

src/views/Home/components/NewsCard/index.jsx
Centerville1: Yep, it is necessary, the card by default gets a color from the MUI palette, which isn't changed by out theme palette.  For all of the cards, I'm having to add a className to them to reference the CSS theme variables for their background colors...

src/views/Timesheets/components/ShiftItem/index.jsx
Centerville1: Yeah, I think this is legacy from Sharon's work, reverting for now :+1:

src/views/Timesheets/index.jsx
Centerville1: Reverting for now :+1:

src/_vars.scss
Centerville1: Yeah, sorry, meant to get rid of that

src/components/GordonDialogBox/GordonDialogBox.module.scss
Centerville1: We'll look at this in detail when we get to it... is it okay to leave in for now?

src/components/GordonDialogBox/GordonDialogBox.module.scss
Centerville1: Yeah, I know you mentioned that would be good to change, now that we use modules.  I'll try to work on that at some point, but that'll be a bit of a project.

src/components/NavButton/index.jsx
Centerville1: Well, we tested with dark mode setting silly colors just to see what isn't referencing it right, and the cards are set to a default dark gray, rather than the yellow we had set in our dark mode for testing... so, it didn't work...

Centerville1: Because of that, since this PR I've made changes to a bunch of cards like that, which will be included in future commits for the above reasons, unless there's something I'm missing

src/views/Home/components/DiningBalance/index.jsx
Centerville1: Fair , I removed all uses of getComputedStyle for now

src/components/GordonConfetti/index.jsx
```diff
@@ -26,7 +26,11 @@ const GordonConfetti = ({ active, colorOption, colors }) => {
       gordonColors.secondary.green,
     ];
   } else if (colorOption === 'GordonBlue') {
-    config.colors = ['#ddd', gordonColors.primary.blue, gordonColors.primary.cyan];
+    config.colors = [
+      '#ddd',
+      newTheme.vars.palette.primary[300],
+      newTheme.vars.palette.secondary[300],
```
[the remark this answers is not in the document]
Centerville1: Good point, I've reverted this for now until we get a chance to look at it in detail.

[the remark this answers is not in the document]
Centerville1: Also, as I said in other places, our future commits will be much more organized, this just has a lot of initial work and experiments, which will be standardized in the future.

src/app.jsx
```diff
@@ -29,6 +29,9 @@ const App = () => {
     historyRef.current.listen(() => analytics.onPageView());
   }, []);
 
+  // console.log(theme.vars.palette.neutral.main);
+  // console.log(theme.vars);
```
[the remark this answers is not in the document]
Centerville1: Done 👍

src/components/GordonDialogBox/GordonDialogBox.module.scss
russtuck: If it's not used, I don't think we should add it.  It will require attention to delete later, and distract people looking for things from what they really want.  Save it somewhere (or just leave it in whatever branch it came from).  But don't merge it into develop.

EjPlatzer: I think that it would be better to remove it from the pull request so that we aren't merging unused code that will need to be reviewed later.

src/_vars.scss
```diff
@@ -7,6 +7,7 @@
 
 // For use in a .js file, see `src/theme.js`.
 
+// DEPRECATED by theme palette update, colors here should be removed eventually in the future!
```
russtuck, line 10: As we discussed in person, every color needs a (single) name.  That name should be used in the theme definitions.  If you want to delete this name eventually, then create a new better name and use that.  (In fact, use the good name to define the old name, if possible).   In general, each hex code should only appear once anywhere.

src/_vars.scss
```diff
@@ -7,6 +7,7 @@
...
 $primary-blue: #014983;
```
russtuck, line 11: Please find all uses of this hex code, at least in this pull request, and replace them with the appropriate name.  Do the same for all the rest of the hex codes below.  (Exception: if there are two logically distinct names and uses for the same hex code, presumably used in different contexts, that's probably OK.)

src/views/About/index.jsx
```diff
@@ -101,7 +100,7 @@ const About = () => {
         <Typography variant="subtitle1">
           Found a bug?
           <a href="mailto:cts@gordon.edu?Subject=Gordon 360 Bug">
-            <Button style={{ color: gordonColors.primary.cyan }}>Report to CTS</Button>
+            <Button style={{ color: 'var(--mui-palette-primary-main)' }}>Report to CTS</Button>
```
EjPlatzer, line 103: Two things here:
1. You appear to have switched the color of this `Button` from cyan to blue.
2. You shouldn't need to pass this as an inline style, you can just pass the `color` prop to the button.

src/_vars.scss
```diff
@@ -16,15 +17,15 @@ $secondary-red: #b53228;
...
-$neutral-white: #ffffff;
-$neutral-white-opacity80: #ffffffcd;
-$neutral-white-opacity50: #ffffff80;
-$neutral-white-opacity30: #ffffff4d;
-$neutral-light-gray: #ebeaea;
-$neutral-light-gray2: #dddddd;
-$neutral-gray: #808080;
-$neutral-gray2: #cccccb;
-$neutral-gray3: #adadad;
+$neutral-white: #ffffff; //Setting: header buttons, light event cards, person profile(both pub and priv), about and help card header text, people search icon,
```
russtuck, line 20: Please put each comment above its color, unless it is very short, so there are no lines over 80 characters long.

src/views/Home/components/NewsCard/index.jsx
EjPlatzer: Hmm, that seems to me like you might've configured the dark theme wrong then. As I understand it, MUI has a default dark theme which should effect all components when you switch themes, so you don't need to, for example, restyle all MUI components (which kinda defeats the purpose of using MUI in the first place).

Adding a class to EVERY `Card` component in the codebase is not an acceptable solution, so this might need further investigation.

src/_vars.scss
```diff
@@ -44,6 +45,9 @@ $wellness-red: #a30000;
 $wellness-green: #005500;
 $wellness-yellow: #ffcc00;
 
+// NOT DEPRECATED
```
russtuck, line 48: Good comment, but please say it in a way that will make sense a year from now (after the theme changes are done and forgotten)

src/_vars.scss
russtuck: Why is it still here?

EjPlatzer (changes requested):

src/app.global.scss
russtuck: Please commit the suggestion.

russtuck (commented):
Left a few comments, and as discussed, will hold off reviewing more since you're working on it and Evan is reviewing.

src/_vars.scss
Centerville1: As we discussed, the _vars file will be deprecated by this update, and is only still here to ensure backwards compatibility while we transition.  Because of that, there's no need that I can see to updated the colors here with a different naming scheme.

src/_vars.scss
Centerville1: See below comment for why outdated

src/_vars.scss
Centerville1: We don't actually need these comments any more, so I'll just remove them...

src/_vars.scss
Centerville1: I think the comments beneath this line are fairly clear, except for a typo I left, I'll commit the fixed typo, hopefully it's clear enough

src/components/GordonDialogBox/GordonDialogBox.module.scss
Centerville1: So, we'll probably be adding a CSS file to Gordon Dialog Box anyway...  so the attention to check it later will already be given.  Would it be sufficient to add a comment to the CSS file for now, since it will likely be used later @EjPlatzer @russtuck ?

src/views/Home/components/NewsCard/index.jsx
Centerville1: So, MUI does have a default dark theme, which works, but, at least the way I set up the themes, our theme doesn't override the card color from MUI.  I don't know if you have any idea what I might need to change to make our dark mode override the MUI properly.

One thing I do know, a lot of MUI components like buttons have color props that are set to MUI them setting, like color='primary,' and they DO change with our defined theme colors...  For testing purposes, I set all of the dark theme colors to yellow, so we can easily see which components have been changed to reference the correct theme CSS variables...  Buttons which have MUI color props do turn yellow in dark mode, but the cards stay the default MUI dark mode dark gray color.

I don't know if that helps you understand this at all @EjPlatzer , I'll try to read up on how all these themes interact, because I clearly don't understand it fully...

Centerville1: In the meantime, I have been changing other cards in this way, which seems to work fine, but maybe I don't understand the problem with this.  
Is the problem that defining a CSS class for them overrides all the default MUI styles, thus defeating the purpose?  
I guess I would expect that defining a CSS class only overrides the MUI styles for CSS properties that we include manually, but that might be a faulty assumption on my part...

EjPlatzer: Maybe a clearer way to say this is that, rather than overriding the MUI default again each time we use the MUI component (`Card` in this case), we should override the component's styling once at the theme level so that every `Card` across the entire site automatically gets overridden in a consistent way without us having to do any extra work in each place that `Card` is used.

I think perhaps the piece that you're missing is that all of the MUI theme is configurable, not just the palette. So if you want `Card` to have different styles, you can declare the `Card` styles for each theme mode inside the base theme, and it will apply to every `Card`.

The question I would ask is - do you need to change the `Card` styling? You mentioned you're using some yellow colors just for testing, but you only applied those colors to the palette colors, right? So it's expected that the MUI components that don't reference our palette colors are still using the default dark mode theme. Is there something wrong with the default dark mode theme? If so, you should change it at the theme level as outlined above. If not, you shouldn't change it all - just let the theme be as is.

Does that make sense?

Centerville1: Ah, got it, that is helpful.  I think we probably want the cards to reference our palette colors, because we'll probably want to customize the color it chooses, especially in dark mode, rather than the default... but yes, that makes total sense that there is an easier way that I didn't understand, thanks!

src/components/ThemeModeSwitcher/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+  useColorScheme,
+} from '@mui/material/styles';
+import styles from './modeSwitcher.module.css';
+
+export const ModeSwitcher = () => {
+  const { mode, setMode } = useColorScheme();
+  const [mounted, setMounted] = useState(false);
+
+  //Need to figure out how to properly reference the theme provided by the CssVarsProvider context.
+  console.log('Called mode switcher');
```
amos-cha, line 18:
```suggestion
  //Need to figure out how to properly reference the theme provided by the CssVarsProvider context.
  console.log('Called mode switcher');
```

ditto

src/components/ThemeModeSwitcher/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+  if (!mounted) {
+    return null;
+  }
```
amos-cha, line 25:
```suggestion
  useEffect(() => {
    setMounted(true);
  }, []);
  if (!mounted) {
    return null;
  }
  ```
  In what case will this component not be mounted?

**Review by amos-cha — CHANGES_REQUESTED** [review]

File: src/components/ThemeModeSwitcher/index.jsx (hunk 27 above)
Comment (Centerville1) [reply src/components/ThemeModeSwitcher/index.jsx:25]: Probably never, that mode switch button was just copied from the MUI docs, so it's probably designed to cover more cases than we have...  I'll remove the useEffect

File: src/components/GordonDialogBox/GordonDialogBox.module.scss (hunk 3 above)
Comment (Centerville1) [reply src/components/GordonDialogBox/GordonDialogBox.module.scss:3]: > So, we'll probably be adding a CSS file to Gordon Dialog Box anyway... so the attention to check it later will already be given. Would it be sufficient to add a comment to the CSS file for now, since it will likely be used later @EjPlatzer @russtuck ?

@EjPlatzer thoughts on this?

Comment (EjPlatzer) [reply src/components/GordonDialogBox/GordonDialogBox.module.scss:3]: > > So, we'll probably be adding a CSS file to Gordon Dialog Box anyway... so the attention to check it later will already be given. Would it be sufficient to add a comment to the CSS file for now, since it will likely be used later @EjPlatzer @russtuck ?
> 
> @EjPlatzer thoughts on this?

If you're going to add it later, than it can be included in that pull request. It creates a needlessly confusing history for the file to have it created in git before it's referenced anywhere in the project. I think it should be removed before this PR is merged. We should, in general, strive for clean pull requests that are self-contained and leave the code cleaner than it was before.

Comment (Centerville1) [reply src/components/GordonDialogBox/GordonDialogBox.module.scss]: Okay, I've deleted it...

File: src/components/NavButton/index.jsx (hunk 7 above)
Comment (EjPlatzer) [reply src/components/NavButton/index.jsx:33]: Ok, I see what you mean now. In one sense, the current code is correct because these aren't all links. But in a more pragmatic sense it probably does make sense to style them all consistently.

File: src/components/NavButton/index.jsx [hunk 28]
```diff
@@ -35,7 +41,7 @@ const GordonNavButton = ({
       </ListItem>
     ) : (
       <NavLink end to={linkPath} onClick={onLinkClick} className="gc360_link">
-        <ListItem divider={divider} button>
+        <ListItem divider={divider} button className="gc360_link">
```
EjPlatzer, line 44: This added class is redundant with the class on the `NavLink`. You should only need one of these.

src/views/About/index.jsx
Centerville1: Good point, I've made that change :+1:

src/components/ThemeModeSwitcher/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <button
+      variant="outlined"
+      onClick={() => {
+        if (mode === 'light') {
+          setMode('dark');
+        } else {
+          setMode('light');
+        }
+      }}
```
EjPlatzer, line 24: This is actually technically using stale data to update `setMode`. The right way to do this kind of conditional update based on previous state in React is with the callback version of `useState`:
```suggestion
      onClick={() => setMode(prevMode => prevMode === 'light' ? 'dark' : 'light')}
```

src/components/NavButton/index.jsx
Centerville1: So, the link buttons were still displaying as different colors again when I only had them on the NavLink component, I can remove them from the navLink component, which seems to work from testing, so I'll commit that for review.

src/components/ThemeModeSwitcher/index.jsx
Centerville1: Interesting, this code was from the MUI docs for this feature, so I guess they've got some room for improvement, I'll make that change :slightly_smiling_face:

EjPlatzer: @Centerville1 It's possible that the custom hook doesn't have a callback format.

Your code is also not correct, so that would be the issue if the callback does work. The inner arrow function for the callback doesn't actually return anything. The way that arrow functions work, if you include braces after the arrow, then the function body is a statement that returns nothing unless you use an explicit return statement. Conversely, if you don't use braces for the body of the function, it's an expression, meaning that whatever value the expression evaluates to  is the value returned by the function.

For example, this logs `undefined` because the body of `addOne` is a statement with no `return` statement.
```javascript
const addOne = (x) => { x + 1 };
console.log(addOne(1));
```
Whereas this logs `2`
```javascript
const addOne = (x) => x + 1;
console.log(addOne(1));
```

Centerville1: > @Centerville1 It's possible that the custom hook doesn't have a callback format.
> 
> Your code is also not correct, so that would be the issue if the callback does work. The inner arrow function for the callback doesn't actually return anything. The way that arrow functions work, if you include braces after the arrow, then the function body is a statement that returns nothing unless you use an explicit return statement. Conversely, if you don't use braces for the body of the function, it's an expression, meaning that whatever value the expression evaluates to is the value returned by the function.
> 
> For example, this logs `undefined` because the body of `addOne` is a statement with no `return` statement.
> 
> ```js
> const addOne = (x) => { x + 1 };
> console.log(addOne(1));
> ```
> 
> Whereas this logs `2`
> 
> ```js
> const addOne = (x) => x + 1;
> console.log(addOne(1));
> ```

I tried it with and without brackets, I only just added the brackets when it didn't work without them.  I couldn't find anything online about useColorScheme having a callback function, although tbf there isn't that much documentation in general.

EjPlatzer: Gotcha, then it probably doesn't have a callback form, which is fine. The way you had it before isn't awful.

Centerville1: Thanks to Amos we were able to do it a little simpler, using ternary operators to do it, so should be a little simpler.

src/components/ThemeModeSwitcher/index.jsx
```diff
@@ -0,0 +1,25 @@
+/* Temporary component for testing theme implementation and dark mode.  Used in the right side
+nav menu currently.
+
+@TODO convert to some final implementation
+*/
+import { useEffect, useState } from 'react';
```
EjPlatzer, line 6: This looks like an unused import. Was it added purposefully?

Centerville1: Nope, got automatically added when we were experimenting, I'll get rid of it :+1:

EjPlatzer (approved):
There are still some opportunities for cleanup, but given this is being actively worked on, I think we're at a place where this is ready to merge. Great work y'all!

Centerville1:
> There are still some opportunities for cleanup, but given this is being actively worked on, I think we're at a place where this is ready to merge. Great work y'all!

Great, thanks Evan,

Centerville1:
> There are still some opportunities for cleanup, but given this is being actively worked on, I think we're at a place where this is ready to merge. Great work y'all!

Great, thanks Evan!  Are you able to merge this, I don't have merge permissions, and you may need to force a merge anyway, since there are requested changes from Amos (which I did address)

PR #1866, opened 2023-06-13
implementing the dialog box for allowing fac/staff to edit their office location 

antoniavonto:
not working yet, progress stopped due to the need of having a staff view to test and continue writing.
see: https://github.com/gordon-cs/gordon-360-api/pull/917


antoniavonto:
image
image
image

src/services/user.ts
```diff
@@ -238,8 +243,14 @@ const getAdvisors = (username: string): Promise<StudentAdvisorInfo[]> =>
 
 const getMailboxCombination = () => http.get('profiles/mailbox-combination/');
 
+const getBuildings = (): Promise<string[]> => http.get(`advancedsearch/buildings`);
+
 const setMobilePhoneNumber = (value: number) => http.put(`profiles/mobile_phone_number/${value}/`);
 
+const updateOfficeLocation = (OfficeLocation: OfficeLocationQuery) => {
+  return http.patch(`profiles/office_location`, OfficeLocation);
+};
```
amos-cha, line 252: 2 things:

1) I'm pretty sure your route is an `async` route
2) if so, you'll either need to add a `Promise< [YOUR OBJECT NAME] >` wrapper 

OR 

Just don't return anything, but then you might not know if the update succeeded 🤷🏼

amos-cha (changes requested):

src/services/user.ts
EjPlatzer: They're already returning a promise, it's just an inferred type.

amos-cha: hmmm valid, but is it not better to define the return type?

antoniavonto: We based updateOfficeLocation from setMobilePhoneNumber which was also a inferred type that did not define the return type:
image

antoniavonto: We based updateOfficeLocation from setMobilePhoneNumber, which also returned a promise without defining the return type:
image

amos-cha: sounds good, if Evan is good with it then 👍🏼

EjPlatzer: It definitely is better as a rule to define the return type, because currently it's being inferred as `Promise<unknown>`, so if the route is actually returning data than we should document that type. If it isn't actually returning data, you can just make it `Promise<undefined>`.

EjPlatzer:
> @EjPlatzer What is your opinion on combining @antoniavonto & @ArabellaJi 's `Update Office Loc` & `Update Office Hours` forms into a single form with a single update route. I feel like they both are incredibly lightweight and could easily fit on a single form.

@amos-cha I don't think that makes sense because theses are separate data points that the user will not consistently be updating together.

src/components/Profile/components/OfficeInfoList/UpdateOfficeLocationDialog/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+      createSnackbar('Your office location will update within a couple hours.', 'success');
+    } catch {
+      createSnackbar('Office location failed to update. Please contact CTS.', 'error');
+    }
+    setOpen(false);
+  };
+
+  const createSnackbar = (message, severity) => {
+    setSnackbar({ message, severity, open: true });
+  };
```
amos-cha, line 28: is there any specific benefit to having a helper function that simply adds an `open` tag

amos-cha (changes requested):

src/components/Profile/components/OfficeInfoList/UpdateOfficeLocationDialog/index.jsx
antoniavonto: answered by Evan in the edit office hours pull request

russtuck (changes requested):
One small change and I think it will be ready.  

Please also merge from develop and retest, so it will be ready to merge.

russtuck (approved):
Very nice.

In testing, I noticed that the pop-up has no initial setting when I first do it, but if I do it again, it's pre-filled with the current (just set) data.  It would be great if it were auto-filled with the current data the first time it pops up.    But it's not required.  It's fine like it is.

antoniavonto:
> Very nice.
> 
> In testing, I noticed that the pop-up has no initial setting when I first do it, but if I do it again, it's pre-filled with the current (just set) data. It would be great if it were auto-filled with the current data the first time it pops up. But it's not required. It's fine like it is.

@russtuck would you like to make this a new issue?

antoniavonto:
@amos-cha

amos-cha (approved):
👍🏼

PR #1870, opened 2023-06-15
Term selector after conversion

mla04762:
https://github.com/gordon-cs/gordon-360-api/issues/915

We added a term selector for the schedule panel where users can view their course schedule based on which academic term they choose from the drop down. 
termselect

@amos-cha also helped us resolve some issues on the existing code as well as the schedule panel re-rendering every time the user selects a term to view.


src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
```diff
@@ -43,8 +43,13 @@ const GordonScheduleCalendar = (props) => {
...
-    let currentSession = await session.getCurrent();
-    setCurrentSession(currentSession);
+    if (props.term === '') {
+      let currentSession = await session.getCurrent();
+      setCurrentSession(currentSession);
+    } else {
+      let currentSession = await session.get(props.term);
+      setCurrentSession(currentSession);
+    }
     setLoading(false);
```
amos-cha:
```suggestion
    let currentSession = props.term === '' ? await session.getCurrent() : await session.get(props.term);
    setCurrentSession(currentSession)
    setLoading(false);
```

amos-cha (changes requested):
Code quality. Make sure you test this, I'm not sure exactly what happens if you await within a ternary

amos-cha:
@mla04762 please handle your merge conflicts!

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
```diff
@@ -1,29 +1,20 @@
...
 import session from 'services/session';
-// @TODO CSSMODULES - Schedule Calendar needs work but left as normal for now
 import './ScheduleCalendar.css';
 
-export default class GordonScheduleCalendar extends Component {
-  constructor(props) {
-    super(props);
+const GordonScheduleCalendar = (props) => {
+  const [loading, setLoading] = useState(true);
+  const [selectedEvent, setSelectedEvent] = useState(null);
```
amos-cha, line 12:
```suggestion
  const [selectedEvent, setSelectedEvent] = useState();
```
useState defaults null

src/components/Profile/index.jsx
```diff
@@ -14,6 +14,7 @@ import {
   SchedulePanel,
   VictoryPromise,
 } from './components';
+import { Markup } from 'interweave';
```
amos-cha, line 17:
```suggestion
import { Markup } from 'interweave';
```
what is the use of this import?

src/services/schedule.ts
```diff
@@ -56,29 +56,47 @@ function makeScheduleCourses(schedule: CourseSchedule[]): ScheduleEvent[] {
...
-        id: eventId,
-        title: course.CRS_CDE + ' in ' + course.BLDG_CDE + ' ' + course.ROOM_CDE,
-        start: beginTime.toDate(),
-        end: endTime.toDate(),
-        resourceId: day,
-      };
-      eventArray.push(courseEvent);
+    // Added a if statement for classes that do not have a start and end time
+    if (course.ROOM_CDE === 'ASY') {
+      for (const day of summerMeetingDays) {
```
amos-cha, line 65: `for` loops are great. However, javascript has a lot of better ways of handling arrays/objects. In this case, please look at `Array.prototype.map` method.

src/services/schedule.ts
```diff
@@ -56,29 +56,47 @@ function makeScheduleCourses(schedule: CourseSchedule[]): ScheduleEvent[] {
...
+    } else {
+      const beginTime = moment(course.BEGIN_TIME, 'HH:mm:ss')
+        .set('y', today.year())
+        .set('M', today.month())
+        .set('d', today.day());
+      const endTime = moment(course.END_TIME, 'HH:mm:ss')
+        .set('y', today.year())
+        .set('M', today.month())
+        .set('d', today.day());
+      for (const day of getMeetingDays(course)) {
```
amos-cha, line 87: same as above

src/services/schedulecontrol.ts
```diff
@@ -9,11 +9,8 @@ type ScheduleControl = {
...
   http.put(`schedulecontrol/privacy/ ${makePrivate ? 'Y' : 'N'}`); // 'Y' = private, 'N' = public
 
-// TODO: Convert route to send description via request body, so that encoding is unnecessary
-async function setScheduleDescription(Description: string): Promise<void> {
-  const replaced = Description.replace(/\//g, 'SlSh').replace(/:/g, 'CoLn').replace(/\./g, 'dOT');
-  const encoded = encodeURIComponent(replaced);
-  await http.put('schedulecontrol/description/' + encoded);
+async function setScheduleDescription(description: string): Promise<void> {
+  await http.put('schedulecontrol/description/', description);
 }
```
amos-cha, line 14:
```suggestion
async function setScheduleDescription(description: string): Promise<void> => http.put('schedulecontrol/description/', description);

```

amos-cha (changes requested):

src/components/Profile/index.jsx
mla04762: We left out some code from testing, it is removed now

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
```diff
@@ -1,27 +1,20 @@
+import React, { useEffect, useState } from 'react';
 import GordonLoader from 'components/Loader';
 import Moment from 'moment';
-import { Component, Fragment } from 'react';
+import { Fragment } from 'react';
```
amos-cha, line 4:
```suggestion
```

replace the location of where `Fragment` is used with
```
<>
...
</>
```

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
```diff
@@ -32,92 +25,80 @@ export default class GordonScheduleCalendar extends Component {
...
+        localizer={localizer}
+        min={dayStart}
+        max={dayEnd}
+        step={15}
+        timeslots={4}
+        defaultView="day"
+        view={['day']}
+        onDoubleClickEvent={(event) => {
+          props.handleDoubleClick(event);
+        }}
```
amos-cha, line 89:
```suggestion
        onDoubleClickEvent={(event) => props.handleDoubleClick(event)}
```

src/components/Profile/components/SchedulePanel/index.jsx
```diff
@@ -20,6 +26,10 @@ import { gordonColors } from 'theme';
 import EditDescriptionDialog from './components/EditDescriptionDialog';
 import GordonScheduleCalendar from './components/ScheduleCalendar';
 import styles from './ScheduleHeader.module.css';
+import scheduleService from 'services/schedule';
+
```
amos-cha, line 30:
```suggestion
```

src/components/Profile/components/SchedulePanel/index.jsx
```diff
@@ -34,7 +44,34 @@ const GordonSchedulePanel = (props) => {
...
+        const { SessionCode: currentSessionCode } = await sessionService.getCurrent();
+        setCurrentAcademicSession(currentSessionCode);
+      }
+    };
+    loadPage();
+  }, [sessionFromURL]);
+
+  const handleSelectSession = async (value) => {
+    setSelectedSession(value);
+    value = sessionService.decodeSessionCode(value);
```
amos-cha, line 72:
```suggestion
    value = sessionService.decodeSessionCode(value);
```

Does something happen here by reference? You've already `setSelectedSession` so what is going on here?

amos-cha (changes requested):

src/components/Profile/components/SchedulePanel/index.jsx
andrew-wzj: I remember previously when @mla04762 and I copy `handleSelectSession`from`involvementsAll``index.jsx`, and we replaced `navigate` with `reloadHandler()` so the schedule panel would auto change the schedule by term selected.

andrew-wzj: This line code doesn't affect anything with schedule result, so I think we should delete it.

andrew-wzj:
Reminder: add code for displaying summer class "ASY"

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
russtuck: It looks like this was resolved by not accepted.  Please either accept the suggestion or explain why not (or point out where I'm missing the change.)

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
```diff
@@ -1,29 +1,19 @@
...
-      selectedEvent: null,
-      isDoubleClick: false,
-      currentSession: [],
-    };
-    this.eventInfo = [];
-  }
-
-  customEventPropGetter = (event, start, end, isSelected) => {
+  const customEventPropGetter = (event, start, end, isSelected) => {
     if (event.id > 1000) {
```
russtuck, line 17: I know you didn't change this line, but it needs a comment to explain the meaning of the magic number 1000.  

I wonder if this function is even needed any more.  Can it be deleted?

src/services/schedule.ts
```diff
@@ -67,19 +77,29 @@ function makeScheduleCourses(schedule: CourseSchedule[]): ScheduleEvent[] {
...
       .set('M', today.month())
       .set('d', today.day());
-    for (const day of getMeetingDays(course)) {
-      const courseEvent = {
-        id: eventId,
+
+    const meetingDays = getMeetingDays(course);
+
+    if (course.ROOM_CDE === 'ASY') {
+      return summerMeetingDays.map((day) => ({
```
russtuck, line 84: It's not just summer courses that are ASY.  There are also some during other terms.  Please rename summerMeetingDays.  (Maybe "asyncMeetingDays"?)  I haven't looked at all the uses, so don't blindly take my suggestions.

russtuck (commented):
I'm not finished, but here are a few comments.

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
mla04762: I think we might have confused another error for this and marked it as resolved, we have correctly changed it to the suggestion.

.env.development
```diff
@@ -4,7 +4,7 @@
 # VITE_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
- VITE_API_URL=https://360ApiTrain.gordon.edu/
+  VITE_API_URL=https://360ApiTrain.gordon.edu/
```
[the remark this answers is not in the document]
mla04762: I actually found out that we committed changes to the .env file after backend testing, we reverted the change back to the original.

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
mla04762: the line `if (event.id > 1000) {` was used in both the schedulePanel file and the scheduleCalendar file. The usage in both files became irrelevant because we no longer have custom schedules or users being able to edit their schedule by double clicking (because official course schedules cannot be edited). I removed this from both files.

src/services/schedule.ts
mla04762: This is a good point, we changed the variable name and re-pushed it.

russtuck:
Issues found in testing:
* Editing the schedule note doesn't change it.
* "Show Schedule" and "Hide Schedule" don't make sense now that they are in a header box.  Just say "Class Schedule" (or "Course Schedule")
* Is default open/close state of schedule what you intended?

src/components/Profile/components/SchedulePanel/index.jsx
```diff
@@ -1,451 +1,242 @@
...
+                title={isExpanded ? 'Hide Schedule' : 'Show Schedule'}
+              />
+            </AccordionSummary>
+            <AccordionDetails>
+              <Grid container direction="row" justifyContent="center" align="left" spacing={4}>
+                {props.isOnline && (
+                  <Grid container direction="row" item xs={12} lg={12} spacing={2}>
+                    <Grid item lg={1}></Grid>
+                    <Grid item xs={4} lg={1} align="left" className={styles.officeHourText}>
+                      <Markup content="Office Hours: " />
```
russtuck, line 189: (We discussed this in person, but recording here.) The title and the edit pop-up should agree on what this is called. Let's use "Public Office Hours Note", as in the pop-up, at least while there is also another Office Hours link.

russtuck (changes requested):
I'm not sure I've looked at the code carefully enough, but I did find a few more things.

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.jsx
mla04762: We got confused and changed the schedule panel index instead of the schedule calendar, both files have been changed.

src/components/Profile/components/SchedulePanel/index.jsx
```diff
@@ -1,451 +1,242 @@
...
-          start: startTime[0] + ':' + startTime[1],
-          end: endTime[0] + ':' + endTime[1],
-          resourceId: slotInfo.resourceId,
-        });
+      const schedule = await scheduleService.getSchedule(searchedUser.AD_Username, props.term);
+      setEventInfo(scheduleService.makeScheduleCourses(schedule));
+      if (scheduleControlInfo) {
+        setDescription(
+          scheduleControlInfo.Description
+            ? scheduleControlInfo.Description.replace(new RegExp('SlSh', 'g'), '/')
```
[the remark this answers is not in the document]
mla04762: We decided to leave this regexp code for now. Although we don't need it anymore, the data in the database still has the description stored as the regex (before changing it).

russtuck (approved):
LGTM

PR #1871, opened 2023-06-15
implemented the dialog box for changing the fac/staff office hours

antoniavonto:
Implemented a Dialog box for fac/staff to be able to change their office hours.
see:https://github.com/gordon-cs/gordon-360-api/pull/922


antoniavonto:
image
image

amos-cha:
This is a personal opinion, @EjPlatzer can agree or disagree, but I feel that having a single line text box is not the best way of changing information. Especially if that line can exceed the size of that line. 

I suggest either give the update box hold structured information:
Monday: `time selector` to `time selector`
Tuesday: `time selector` to `time selector`
...

Where no input or cleared input would be considered no availability that day.

Granted these are professors who shouldn't type in something like: "I had a salad for lunch" in place of office hours, but sterilization is always a good thing.

antoniavonto:
> This is a personal opinion, @EjPlatzer can agree or disagree, but I feel that having a single line text box is not the best way of changing information. Especially if that line can exceed the size of that line.
> 
> I suggest either give the update box hold structured information: Monday: `time selector` to `time selector` Tuesday: `time selector` to `time selector` ...
> 
> Where no input or cleared input would be considered no availability that day.
> 
> Granted these are professors who shouldn't type in something like: "I had a salad for lunch" in place of office hours, but sterilization is always a good thing.

I suppose this could work, and we could implement that. The way that the fac/staff can change their office hours in go is just by text, I believe because some put more details like you can book me link or where to meet if not in their office. I think it would be nice to ask the opinion of someone that uses this feature in go. Maybe instead of just one line it could be a larger text box.

ArabellaJi:
> > This is a personal opinion, @EjPlatzer can agree or disagree, but I feel that having a single line text box is not the best way of changing information. Especially if that line can exceed the size of that line.
> > I suggest either give the update box hold structured information: Monday: `time selector` to `time selector` Tuesday: `time selector` to `time selector` ...
> > Where no input or cleared input would be considered no availability that day.
> > Granted these are professors who shouldn't type in something like: "I had a salad for lunch" in place of office hours, but sterilization is always a good thing.
> 
> I suppose this could work, and we could implement that. The way that the fac/staff can change their office hours in go is just by text, I believe because some put more details like you can book me link or where to meet if not in their office. I think it would be nice to ask the opinion of someone that uses this feature in go. Maybe instead of just one line it could be a larger text box.

That's true. Perhaps we can have both the time selector and the test box?

EjPlatzer:
The problem with adding structured data is that the database we're updating has a simple `VARCHAR(75)` column for office hours, meaning we can only store simple text up to 75 characters.

We can discuss the possibility of storing more structured or meaningful data, since we're moving this to entirely within 360, but that's a bigger question that's probably outside the scope of this initial pull request.

amos-cha:
> The problem with adding structured data is that the database we're updating has a simple `VARCHAR(75)` column for office hours, meaning we can only store simple text up to 75 characters.
> 
> We can discuss the possibility of storing more structured or meaningful data, since we're moving this to entirely within 360, but that's a bigger question that's probably outside the scope of this initial pull request.

https://github.com/gordon-cs/gordon-360-api/issues/927

src/services/user.ts
```diff
@@ -238,8 +243,15 @@ const getAdvisors = (username: string): Promise<StudentAdvisorInfo[]> =>
...
 const getMailboxCombination = () => http.get('profiles/mailbox-combination/');
 
+const getBuildings = (): Promise<string[]> => http.get(`advancedsearch/buildings`);
+
 const setMobilePhoneNumber = (value: number) => http.put(`profiles/mobile_phone_number/${value}/`);
 
+const updateOfficeLocation = (OfficeLocation: OfficeLocationQuery) => {
+  return http.patch(`profiles/office_location`, OfficeLocation);
+};
+const updateOfficeHours = (value: string) => http.patch(`profiles/office_hours/${value}/`);
```
amos-cha, line 253: issue referenced from API PR here

amos-cha (changes requested):

src/components/Profile/components/OfficeInfoList/UpdateOfficeHoursDialog/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+      await userService.updateOfficeHours(hours);
+      createSnackbar('Your office hours will update within a couple hours.', 'success');
+    } catch {
+      createSnackbar('Office hours failed to update. Please contact CTS.', 'error');
+    }
+    setOpen(false);
+  };
+
+  const createSnackbar = (message, severity) => {
+    setSnackbar({ message, severity, open: true });
```
amos-cha, line 24: feel free to push back on me here: is there any real benefit to having a helper function that simply adds an extra `open: true` tag?

src/components/Profile/components/OfficeInfoList/UpdateOfficeLocationDialog/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+      createSnackbar('Your office location will update within a couple hours.', 'success');
+    } catch {
+      createSnackbar('Office location failed to update. Please contact CTS.', 'error');
+    }
+    setOpen(false);
+  };
+
+  const createSnackbar = (message, severity) => {
+    setSnackbar({ message, severity, open: true });
+  };
```
amos-cha, line 28: same as above

src/services/user.ts
```diff
@@ -237,8 +242,15 @@ const getAdvisors = (username: string): Promise<StudentAdvisorInfo[]> =>
 
 const getMailboxCombination = () => http.get('profiles/mailbox-combination/');
 
+const getBuildings = (): Promise<string[]> => http.get(`advancedsearch/buildings`);
+
 const setMobilePhoneNumber = (value: number) => http.put(`profiles/mobile_phone_number/${value}/`);
 
+const updateOfficeLocation = (OfficeLocation: OfficeLocationQuery) => {
+  return http.patch(`profiles/office_location`, OfficeLocation);
+};
```
amos-cha, line 251:
```suggestion
const updateOfficeLocation = (OfficeLocation: OfficeLocationQuery) => http.patch(`profiles/office_location`, OfficeLocation);

```

amos-cha (changes requested):

src/components/Profile/components/OfficeInfoList/UpdateOfficeHoursDialog/index.jsx
EjPlatzer: I believe this is copied from the other instances of snackbars, which I mostly wrote and/or refactored.

The reason I chose to create this helper function is to encapsulate the behavior of "create / update the current snackbar so that it's like this". The main use case is when the snackbar is shared by a whole deep component tree, so the top-level parent displays it and passes this simple helper down to the deeply nested children. I think it's helpful to have the simple interface (with some extra type safety), so the consumer only has to think about what message and severity the snackbar has, without worrying about implementation details like the `open` state.

Also, the goal is that one day we have a single, site-wide snackbar that can be used by any component (probably through a context-consuming hook), so providing a simple, encapsulated interface should make that adoption path smoother in the future.

antoniavonto: Just made the necessary changes to give more clarity

src/components/Profile/components/OfficeInfoList/UpdateOfficeLocationDialog/index.jsx
antoniavonto: Just made the necessary changes to give more clarity

antoniavonto:
@amos-cha

amos-cha (dismissed):
👍🏼

amos-cha (approved):

PR #1872, opened 2023-06-16
Add patch route to http.ts

amos-cha:
Currently all our updates use `http put` which is not correct terminology when updating elements of a resource and not the entire resource. 

This update has already been in the `recim` branch, but as the branch is not ready to be merged, it would be beneficial for the practicum students to have access to the patch route.

### DIFF
--- a/src/services/http.ts
+++ b/src/services/http.ts
@@ -45,7 +45,8 @@ const post = <TResponse>(
   headers = new Headers(),
 ): Promise<TResponse> =>
   makeRequest(endpoint, 'post', JSON.stringify(body), setContentTypeJSON(headers));
-// PATCH is required to be in all caps, http services automatically capitalizes headers for post,put,get,del... but not patch
+
+// PATCH is required to be in all caps.  http services automatically capitalizes headers for post,put,get,del... but not patch.
 const patch = <TResponse>(
   endpoint: string,
   body: Object = '',
--- a/src/services/http.ts
+++ b/src/services/http.ts
@@ -46,6 +46,14 @@ const post = <TResponse>(
 ): Promise<TResponse> =>
   makeRequest(endpoint, 'post', JSON.stringify(body), setContentTypeJSON(headers));
 
+const patch = <TResponse>(
+  endpoint: string,
+  body: Object = '',
+  headers = new Headers(),
+): Promise<TResponse> =>
+  makeRequest(endpoint, 'PATCH', JSON.stringify(body), setContentTypeJSON(headers));
+
+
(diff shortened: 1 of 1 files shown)


src/services/http.ts
```diff
@@ -46,6 +46,14 @@ const post = <TResponse>(
 ): Promise<TResponse> =>
   makeRequest(endpoint, 'post', JSON.stringify(body), setContentTypeJSON(headers));
 
+const patch = <TResponse>(
+  endpoint: string,
+  body: Object = '',
+  headers = new Headers(),
+): Promise<TResponse> =>
+  makeRequest(endpoint, 'PATCH', JSON.stringify(body), setContentTypeJSON(headers));
```
russtuck, line 54: Why is it "PATCH" in all caps here, but "post" in lowercase above?  I think this is important, and was hard to figure out.  So it deserves a comment here.  If nothing else, it might prevent someone n the future from "cleaning it up" and breaking things mysteriously.

src/services/http.ts
```diff
@@ -174,6 +182,7 @@ const httpUtils = {
   del,
   get,
   post,
+  patch,
```
russtuck, line 185: Is the order here significant?  It seems a little odd (as in illogical) to have patch between post and postImage, though i assume the order doesn't matter except for readability.

russtuck (changes requested):
This makes sense (and I'm eager to see RecIM merged).  I think it needs a comment before being merged.

src/services/http.ts
```diff
@@ -46,6 +46,14 @@ const post = <TResponse>(
 ): Promise<TResponse> =>
   makeRequest(endpoint, 'post', JSON.stringify(body), setContentTypeJSON(headers));
```
amos-cha, line 48:
```suggestion
// PATCH is required to be in all caps, http services automatically capitalizes headers for post,put,get,del... but not patch
```

src/services/http.ts
amos-cha, line 185:
```suggestion
  get,
  patch,
  post,
```

src/services/http.ts
```diff
@@ -45,6 +45,14 @@ const post = <TResponse>(
   headers = new Headers(),
 ): Promise<TResponse> =>
   makeRequest(endpoint, 'post', JSON.stringify(body), setContentTypeJSON(headers));
+// PATCH is required to be in all caps, http services automatically capitalizes headers for post,put,get,del... but not patch
```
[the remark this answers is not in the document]
russtuck: Just add a blank line and periods for clarity

russtuck (changes requested):
Great.  Just suggestings a format change for the comment.

russtuck (approved):

PR #1874, opened 2023-06-16
[COLORS] Changes to theme mode implementation

ahwnsals1234:


src/views/EnrollmentCheckIn/components/CompletedCheckIn/CompletedCheckIn.module.scss
```diff
@@ -0,0 +1,5 @@
+.completed_checkIn {
+  &_congrats_note {
```
Centerville1, line 2: Since these are using modules now, you don't have to worry about making super unique names for the CSS classes.  I.E, for this one you can just call it congrats_note, without the completed_checkIn { & part...

src/views/Events/Events.module.scss
```diff
@@ -0,0 +1,25 @@
+.event_page {
+  &_searchPageTitle {
```
Centerville1, line 2: Same thing here, we should get rid of the & nested classes where they aren't needed.

src/views/InvolvementProfile/InvolvementProfile.module.scss
```diff
@@ -11,3 +11,10 @@
 .update_image_components {
   margin-top: 15px;
 }
+
+.involvementProfile_buttons {
+  &_removeImage {
```
Centerville1, line 16: ditto

src/views/InvolvementProfile/components/Membership/components/AdminCard/AdminCard.module.scss
```diff
@@ -0,0 +1,6 @@
+.AdminCard_page {
+  &_cardHeader {
```
Centerville1, line 2: ditto

src/views/InvolvementProfile/components/Membership/components/AdminCard/components/RequestsReceived/index.jsx
```diff
@@ -66,7 +62,11 @@ const RequestsReceived = ({ onAddMember }) => {
               />
 
               <ListItemSecondaryAction>
-                <Button style={redButton} onClick={() => onDeny(request.RequestID)} size="small">
+                <Button
+                  className={styles.requestReceived_deny_button}
```
Centerville1, line 66: If you want a button that uses the color scheme with main as the background, and contrastText color, you can just use a color='error.'  You probably don't need it in this case, but there may be some buttons where externall CSS isn't necessary, which might simplify things

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.jsx
```diff
@@ -19,13 +19,11 @@ import { useParams } from 'react-router';
...
 import { stripDomain } from 'services/utils';
-import { gordonColors } from 'theme';
 import RequestsReceived from './components/RequestsReceived';
 import { ConflictError, NotFoundError } from 'services/error';
+import styles from './AdminCard.module.css';
 
 const headerStyle = {
-  backgroundColor: gordonColors.primary.blue,
-  color: '#FFF',
   padding: '10px',
```
Centerville1, line 27: You should also add the other CSS tags into the external CSS, like padding.  There should be know style objects defined inside of code, they should be fully moved to separate CSS.

src/views/InvolvementProfile/components/Membership/components/AdminCard/index.jsx
```diff
@@ -90,7 +88,11 @@ const AdminCard = ({ createSnackbar, isSiteAdmin, involvementDescription, onAddM
   return (
     <>
       <Card>
-        <CardHeader title="Manage Roster" style={headerStyle} />
+        <CardHeader
+          title="Manage Roster"
+          style={headerStyle}
```
Centerville1, line 93: You shouldn't have both a style and a className tag.  If you have this, this is a hint that you need to move some more tags into the external CSS.

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/index.jsx
```diff
@@ -22,7 +22,7 @@ import { useUser } from 'hooks';
 import { useEffect, useState } from 'react';
 import membership from 'services/membership';
 import userService from 'services/user';
-import { gordonColors } from 'theme';
+import styles from './MemberListItem.module.css';
 
 const rowComponentStyle = {
   margin: '10px 0px',
```
Centerville1, line 28: Same thing here, these margins should be moved into CSS classes.

src/views/Timesheets/components/SavedShiftsList/index.jsx
```diff
@@ -175,7 +178,7 @@ export default class SavedShiftsList extends Component {
     );
 
     let header = (
-      <Grid item xs={12} style={styles2.headerStyle}>
+      <Grid item xs={12} style={styles2.headerStyle} className={styles.timesheets_headerStyle}>
```
Centerville1, line 181: Same thing here, should move everything from styles2 into the CSS class you created

src/views/Timesheets/components/ShiftItem/index.jsx
```diff
@@ -421,7 +420,7 @@ export default class ShiftItem extends Component {
     );
     let deleteButton = (
       <IconButton onClick={this.handleDeleteButtonClick} size="large">
-        <DeleteForeverOutlinedIcon style={{ color: gordonColors.secondary.red }} />
+        <DeleteForeverOutlinedIcon className={styles.timeshifts_dButton} />
```
Centerville1, line 423:
```suggestion
        <DeleteForeverOutlinedIcon className={styles.timesheets_dButton} />
```
I think you have a typo here, but this will probably be changed by getting rid of the &s in the CSS class.  It also might be good to have more descriptive classnames than just dbutton and ibutton, maybe delete_button and icon_button (or whatever makes more sense)

Centerville1 (changes requested):
Looks pretty good, just a couple changes and refactoring, shouldn't be too bad :+1:

src/views/EnrollmentCheckIn/components/CompletedCheckIn/index.jsx
```diff
@@ -1,13 +1,11 @@
 import { Button, Grid, Typography } from '@mui/material';
 import { useUser } from 'hooks';
 import { Link } from 'react-router-dom';
-import { gordonColors } from 'theme';
 // @TODO CSSMODULES - outside directory
```
Centerville1, line 4: You should be able to remove this comment now, since we've added CSS modules to this

src/views/Events/Events.module.scss
```diff
@@ -0,0 +1,26 @@
+_searchPageTitle {
+  color: var(--mui-palette-secondary-main);
+}
+_event_filter_icon {
+  color: var(--mui-palette-neutral-900);
+  font-size: 20;
+}
+_event_filter_list_icon {
+  color: var(--mui-palette-neutral-900);
```
Centerville1, line 9: Did you test this color?  I don't think our neutral 900 is the same as the old gray 900, so I just wanted to make sure you looked at it

src/views/EventsAttended/index.jsx
```diff
@@ -8,13 +8,6 @@ import { Link } from 'react-router-dom';
 import event from 'services/event';
 import { gordonColors } from 'theme';
```
Centerville1, line 9: This looks like an unused import

src/views/EventsAttended/index.jsx
```diff
@@ -8,13 +8,6 @@ import { Link } from 'react-router-dom';
...
 
-const style = {
```
Centerville1, line 11: You removed this style, but on lines 33 and 55 you still have buttons that use this.  You should be able to just add the color='secondary' prop to the buttons to get the same effect

src/views/InvolvementProfile/components/Membership/components/AdminCard/components/RequestsReceived/index.jsx
```diff
@@ -66,7 +61,7 @@ const RequestsReceived = ({ onAddMember }) => {
               />
 
               <ListItemSecondaryAction>
-                <Button style={redButton} onClick={() => onDeny(request.RequestID)} size="small">
+                <Button color="error" onClick={() => onDeny(request.RequestID)} size="small">
```
Centerville1, line 64: Nice use of the color prop!

src/views/InvolvementProfile/InvolvementProfile.module.scss
```diff
@@ -11,3 +11,26 @@
 .update_image_components {
   margin-top: 15px;
 }
+
+_removeImage {
```
Centerville1, line 15: You need to title CSS classes starting with a dot, so these probably aren't working right now (did you test it to make sure it looked right?)  You should also probably get rid of the underscores here.  They aren't needed anymore since we aren't using the nested styles. 
```suggestion
.removeImage {
```

src/views/InvolvementProfile/components/Membership/components/MemberList/MemberList.module.scss
```diff
@@ -0,0 +1,5 @@
+_headerStyle {
```
Centerville1, line 1: Again here 
```suggestion
.headerStyle {
```

src/views/InvolvementProfile/components/Membership/components/MemberList/components/MemberListItem/MemberListItem.module.scss
```diff
@@ -0,0 +1,23 @@
+_leaveButton {
```
Centerville1, line 1: ditto

src/views/InvolvementProfile/index.jsx
```diff
@@ -303,10 +297,9 @@ const InvolvementProfile = () => {
                       <div className={styles.photoUploader} {...getRootProps()}>
                         <input {...getInputProps()} />
                         <img
-                          className="rounded_corners"
+                          className={'rounded_corners' + styles._imageDrop}
```
Centerville1, line 300: What is this doing?  is this actually referencing a CSS class called rounded_corners_imageDrop(which is what 'rounded_corners' + styles._imageDrop is doing)?

src/views/InvolvementProfile/index.jsx
```diff
@@ -325,10 +318,7 @@ const InvolvementProfile = () => {
                   <Cropper
                     ref={cropperRef}
                     src={preview}
-                    style={{
-                      maxWidth: maxCropPreviewWidth(),
-                      maxHeight: maxCropPreviewWidth() / cropperData.aspectRatio,
-                    }}
+                    className={styles._cropper / cropperData.aspectRatio}
```
Centerville1, line 321: here they were dividing two numbers, I think this is a case where we'll have to leave the style tag, since they are using functions and variables that they compute in the code (maxCropPreviewWidth() and cropperData.aspectRatio are both Javascript, not usable in CSS)

src/views/InvolvementsAll/components/Requests/components/RequestSent/index.jsx
```diff
@@ -2,24 +2,24 @@ import { Button, Divider, Grid, Typography } from '@mui/material';
...
-  };
-
   let cancel;
   if (member.Status === 'Pending') {
     cancel = (
-      <Button variant="outlined" size="small" style={button} onClick={handleCancel}>
+      <Button
+        variant="outlined"
+        size="small"
+        className={styles._cancel_button}
```
Centerville1, line 20: You can probably use the color='error' here, and get rid of the CSS.  Just check what that looks like.

src/views/Timesheets/components/SavedShiftsList/SavedShiftsList.module.scss
```diff
@@ -16,3 +16,26 @@
     }
   }
 }
+_RButton {
```
Centerville1, line 19: ditto, and maybe make this a more descriptive name
```suggestion
.RButton {
```

src/views/Timesheets/components/SavedShiftsList/index.jsx
```diff
@@ -175,31 +174,31 @@ export default class SavedShiftsList extends Component {
     );
 
     let header = (
-      <Grid item xs={12} style={styles2.headerStyle}>
+      <Grid item xs={12} className={styles._headerStyle}>
         <div>
           <Grid container direction="row">
             <Grid item xs={3}>
-              <Typography className="disable_select" variant="body2" style={styles2.headerItem}>
+              <Typography className={'disable_select' + styles.headerItem} variant="body2">
```
Centerville1, line 181: Again here, what was the class disable select, and is this now referencing a class called 'disable_selectheaderItem?  This won't reference both classes, I think it will try to concatenate their names.

And ditto for the other places in this file you do the same

src/views/Timesheets/components/ShiftItem/ShiftItem.module.scss
```diff
@@ -20,3 +20,31 @@
     margin-top: -8px;
   }
 }
+_delete_button {
```
Centerville1, line 23: Same thing here 
```suggestion
.delete_button {
```

Centerville1 (changes requested):

src/views/InvolvementProfile/index.jsx
```diff
@@ -411,8 +400,8 @@
             <Grid item align="center">
               {ActivityBlurb && <Typography>{ActivityBlurb}</Typography>}
               {ActivityURL?.length !== 0 && (
-                <Typography>
-                  <a href={ActivityURL} className="gc360_text_link" style={{ fontWeight: 'bold' }}>
+                <Typography className={styles.activityURL}>
+                  <a href={ActivityURL} className="gc360_text_link">
```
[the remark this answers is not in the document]
Centerville1: @EjPlatzer we got this warning about the involvement profile website link being potentially vulnerable to cross-site scripting.  It doesn't look like there is any code in the frontend to prevent scripting, so might need to be looked into.

Centerville1 (approved):
LGTM

src/views/InvolvementProfile/index.jsx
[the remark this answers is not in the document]
EjPlatzer: This is the true in the sense that a user with sufficient permissions to change the ActivityURL (i.e. a group admin or a site admin) can edit it to be a malicious URL (see the current URL for the 360 involvement in Train as an example).

There's a question of tradeoffs here - the goal of this field is to display a user-provided URL, which will never truly be safe because a malicious user can input a URL that redirects to e.g. a Man In The Middle phishing page. On the other hand, it's possible we could sanitize some of the most obviously harmful URLs (although sanitization is notoriously difficult). It's worth exploring more, but I would also say this isn't an imminent security threat for our threat model.

PR #1885, opened 2023-06-20
New logo on header with home button

jhammond4:
Changed a few files so the header now uses the new logo. Because it now looks more like a clickable home button, I added an invisible home button behind it. I talked to the design team about it and they were okay with doing this-I sent them what it looks like to see what they say and if there are any changes to be made. I also checked and made sure it works on the 3 main screen sizes and mobile.


src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -6,6 +6,7 @@ import session from 'services/session';
 import { gordonColors } from 'theme';
 import styles from '../Doughnut.module.css';
 import { Card, CardContent, CardHeader, Grid, Typography } from '@mui/material';
+import { NoBackpackSharp } from '@mui/icons-material';
```
amos-cha, line 9: what is this import?

src/views/Home/components/NewsCard/index.jsx
```diff
@@ -20,12 +21,25 @@ const NewsCard = () => {
               Today's Student News
             </Grid>
             <Grid item xs={5} align="right">
-              <Button variant="contained" color="secondary" component={Link} to="/news">
+              <Button
+                variant="contained"
+                //color="secondary"
```
amos-cha, line 26: if unused, please remove!

amos-cha (changes requested):

jhammond4:
Really the only files that are important to the header are the ones under the header folder. I think I just changed the other files (NewsCard, CLWCreditsDaysLeft, DaysLeft, and SearchFieldList) and committed them here even though they have their own separate prs. The header files are really the only necessary ones for this pr

jhammond4:
Screenshot from 2023-07-05 15-28-37
Screenshot from 2023-07-05 15-28-50
Screenshot from 2023-07-05 15-29-03
Screenshot from 2023-07-05 15-29-24
Pictures of each screen size with the new header

src/components/Header/index.jsx
```diff
@@ -8,7 +8,7 @@ import {
   Work as WorkIcon,
 } from '@mui/icons-material';
 import { AppBar, Button, IconButton, Tab, Tabs, Toolbar, Typography } from '@mui/material';
-import GordonDialogBox from 'components/GordonDialogBox';
+import GordonDialogBox from 'components/GordonDialogBox/index';
```
amos-cha, line 11: is there a specific reason this was changed?

src/components/Header/index.jsx
```diff
@@ -68,20 +68,33 @@ const GordonHeader = ({ onDrawerToggle }) => {
...
+          title={'Credentials Needed'}
+          buttonClicked={() => setDialog(null)}
+          buttonName={'Okay'}
+        >
+          This feature is unavailable while not logged in. Please log in to access it.
+        </GordonDialogBox>
+      );
+    } else {
+      return null;
+    }
```
amos-cha, line 97: Is there a reason this was changed from the previous version? The logic looks identical?

amos-cha (commented):
Got a couple questions

src/components/Header/index.jsx
jhammond4: I don't remember why it was changed (I don't remember changing it, maybe it was from a merge? I'm not sure) but regardless I can switch it back

src/components/Header/index.jsx
jhammond4: I don't remember why this was changed either but I can switch it back

amos-cha (approved):
looks good!

russtuck (approved):

src/components/Header/Header.module.css
EjPlatzer, line 1: This file should not be committed. The css files are auto-generated from the .scss files, so they should not be tracked by git.

src/components/Header/index.jsx
```diff
@@ -151,7 +151,15 @@ const GordonHeader = ({ onDrawerToggle }) => {
...
-
+          &nbsp;&nbsp; &nbsp;&nbsp;
+          <Tab //this is an "invisible" button behind the new logo that links back to the home page
+            className={styles.tab}
+            label=" "
+            component={ForwardNavLink}
+            disableRipple={true}
+            to="/"
+            onChange={(event, value) => setTabIndex(value)}
+          />
```
EjPlatzer, line 162: This is not the best way to do this. We should have an actual image tag in the markup that is wrapped in a link, rather than a background image that is covered by an invisible Tab. Also, I suspect this breaks tab highlighting.

EjPlatzer (commented):
Commenting here but please do not add this to the release.

src/components/Header/index.jsx
jhammond4: Got it. I can look into that. I also made it so clicking on the tab resets the tab highlight to Home, which I can keep intact if necessary. I know there is talk of modifying the header tabs, so it may need to be changed

PR #1888, opened 2023-06-20
S23 theme palette implementation 2nd PR

Centerville1:


src/theme.ts
```diff
@@ -289,3 +289,18 @@ export const newTheme = extendTheme({
...
+
+extendTheme({
+  components: {
+    MuiCard: {
+      styleOverrides: {
+        root: ({ theme }) => ({
+          backgroundColor: White,
+          [theme.getColorSchemeSelector('dark')]: { backgroundColor: TestTool },
+        }),
+      },
```
Centerville1, line 303: This is my initial experimentation with overriding card styling, not working currently, something to do with my use of theme, I tried newTheme and theme here, based directly on the docs I found, but it doesn't seem to be working, more investigation needed...

src/theme.ts
```diff
@@ -289,3 +289,18 @@ export const newTheme = extendTheme({
...
   },
 });
+
+console.log(newTheme.components);
+
+extendTheme({
+  components: {
+    MuiCard: {
+      styleOverrides: {
+        root: ({ theme }) => ({
```
Centerville1, line 299: This is my initial experimentation with overriding card styling, not working currently, something to do with my use of theme, I tried newTheme and theme here, based directly on the docs I found, but it doesn't seem to be working, more investigation needed...

src/theme.ts
```diff
@@ -289,3 +289,18 @@ export const newTheme = extendTheme({
...
+          backgroundColor: White,
+          [theme.getColorSchemeSelector('dark')]: { backgroundColor: TestTool },
```
EjPlatzer: I think within the theme, you want to use the MUI theme colors, not the external colors. For example: 
```suggestion
          backgroundColor: theme.vars.palette.neutral.light,
```

EjPlatzer: Cited from the [MUI docs](https://mui.com/material-ui/experimental-api/css-theme-variables/customization/#components)

Centerville1: This is outdated, I ended up adding the following lines to the dark theme, which overrides the colors used by cards and some other MUI text components:
```    
dark: {
      // Palette for dark theme mode -------------------------------------------------------
      palette: {
        background: {
          paper: TestTool, // Card Colors
        },
        text: {
          primary: TestToolContrast, // Various MUI components and text
        },
```

src/components/Profile/components/MembershipsList/index.jsx
```diff
@@ -95,7 +95,7 @@ const MembershipsList = ({ username, myProf, createSnackbar }) => {
 
   const transcriptButton = myProf && (
     <Grid container justifyContent="center">
-      <Link className="gc360_link" to="/transcript">
+      <Link to="/transcript">
```
Centerville1, line 98: This was removed to get rid of the gray box that appeared due to reusing this style from the right side nav, we may want to migrate the styles used in right side nav links to their own CSS file, instead of global at some point

Issue #1880, opened 2023-06-17
Fix Profile Advisors Bug


amos-cha:
Currently the `getProfileInfo` hook which uses `AD_Username` aka `username` uses a bandaid fix devised by @bennettforkner to handle the error of getting Advisor details. This is not a big issue in most cases as this would normally only occur on the profile page. However, with the incoming merge of `recim` (if not done already) we will get upwards of 300 errors regarding our `401 Unauthorized`.

I haven't investigated enough to come up with a proper solution, but off the top of my head, I would suggest either:
1) Make a more lightweight version of the route that can be called by `Unauthorized` users that the UI can handle with checking if the profile you're looking at is your own or not.
2) Make the check in the API and do a `try catch` block to not include advisor data
3) Do both for redundancy

![Image](https://user-images.githubusercontent.com/78386128/246621806-fdd76029-0733-4abd-a9a8-6cd6b8750e86.png)

amos-cha:
https://github.com/gordon-cs/gordon-360-ui/pull/1893

PR #1893, opened 2023-06-21
S23 fix profile advisors

amos-cha:
This is a permanent fix to the bandage fix by @bennettforkner 

https://github.com/gordon-cs/gordon-360-ui/issues/1880

### DIFF
--- a/src/components/Profile/components/PersonalInfoList/index.jsx
+++ b/src/components/Profile/components/PersonalInfoList/index.jsx
@@ -10,7 +10,6 @@ import {
   Switch,
   Typography,
 } from '@mui/material';
-import { Link } from 'react-router-dom';
 import { IconButton, Button } from '@mui/material';
 import LockIcon from '@mui/icons-material/Lock';
 import VisibilityIcon from '@mui/icons-material/Visibility';
@@ -46,6 +45,7 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }) => {
   );
   const [openAlumniUpdateForm, setOpenAlumniUpdateForm] = useState(false);
   const [mailCombo, setMailCombo] = useState();
+  const [advisorsList, setAdvisorsList] = useState([]);
   const [showMailCombo, setShowMailCombo] = useState(false);
   const isStudent = profile.PersonType?.includes('stu');
   const isFacStaff = profile.PersonType?.includes('fac');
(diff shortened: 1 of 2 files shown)


src/components/Profile/components/PersonalInfoList/index.jsx
```diff
@@ -10,7 +10,6 @@ import {
   Switch,
   Typography,
 } from '@mui/material';
-import { Link } from 'react-router-dom';
```
amos-cha, line 13: was never used

src/components/Profile/components/PersonalInfoList/index.jsx
```diff
@@ -87,12 +87,16 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }) => {
 
   // FacStaff spouses are private for private users
   const isSpousePrivate = isFacStaff && keepPrivate && profile.SpouseName !== PRIVATE_INFO;
-
   useEffect(() => {
     async function loadMailboxCombination() {
```
EjPlatzer, line 91: Can you rename this function, since it's no longer just loading the mailbox combo?

EjPlatzer (changes requested):

EjPlatzer (changes requested):
Great work! One small cleanup

EjPlatzer (approved):
LGTM!

src/components/Profile/components/PersonalInfoList/index.jsx
```diff
@@ -326,11 +330,11 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }) => {
   const advisors =
     (myProf || canViewAcademicInfo) && isStudent ? (
       <ProfileInfoListItem
-        title={profile.Advisors?.length > 1 ? 'Advisors:' : 'Advisor:'}
+        title={advisorsList?.length > 1 ? 'Advisors:' : 'Advisor:'}
         contentText={
-          profile.Advisors?.length < 1
+          advisorsList.length < 1
```
bennettforkner, line 335: No `?` here? I see one above.

amos-cha: `advisorList` is initialized as `[]`, the API also returns an empty array if none are found so it's never null

PR #1899, opened 2023-06-23
[COLORS] Quick Search and Header conversion

Centerville1:


src/theme.ts
```diff
@@ -135,6 +135,7 @@ let GordonBlueHalfOpacity = '#01498382';
 
 // Secondary
 let ScottieCyan = '#00AEEF';
+let ScottieCyan10PercentOpacity = '#00AEEF1A';
```
amos-cha, line 138: Perhaps a better naming scheme

amos-cha (approved):

PR #1905, opened 2023-06-26
Revert "Fix the "tab is not highlighted correctly sometimes" Issue"

ArabellaJi:
Reverts gordon-cs/gordon-360-ui#1837


russtuck (approved):
I'm sorry we haven't figured out how to fix this yet, but I agree it's wise to revert it so we can push the other improvements to prod.  We'll keep trying to fix this so we can release it next week.

PR #1906, opened 2023-06-26
Fix PeopleSearch params not updating on navigation

EjPlatzer:
For posterity - somehow, the `popstate` event listener was not being triggered within `SearchFieldList` when the user navigated in history. The fix was to add the event listener within a separate useEffect that only runs on mount, rather than inside the one that runs everytime `location` from React-Router updated.

Somehow, this bug in the event listener didn't appear until the Header component called `useLocation`. We have no idea why that is the case. :shrug:

### DIFF
--- a/src/components/Header/index.jsx
+++ b/src/components/Header/index.jsx
@@ -40,13 +40,12 @@ const TabUrlPatterns = [
  */
 const useTabHighlight = () => {
   const location = useLocation();
-  let currentPath = location.pathname;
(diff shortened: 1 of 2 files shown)


amos-cha (approved):
👍🏼  thank you for the find!

ArabellaJi:
Ah I just found one thing: when I search for person A and then person B, and then use the back arrow to go back, the first time it wouldn't work (URL changed, but text field didn't change). Then if I press the front arrow and then press the back arrow, it works fine.

PR #1909, opened 2023-06-26
People Search: Read params from URL directly in event listener

EjPlatzer:
Previously, we listened for the `popstate` event and updated a `ref` when the event happened. Simultaneously, we had a useEffect that would run whenever `location` changed, which would update the search params from the URL *if* the `ref` said to. This *almost* worked, except that, on the very first `popstate` event, the ref would not update before the `useEffect` ran, meaning we wouldn't read the params from the URL the first time that the user triggered `popstate` after `SearchFieldList` mounted.

The solution is to simply perform the update (i.e. read the params from the URL) directly inside the event listener callback. The only times we need to read the params from the URL are when the component first mounts (in case we're mounting with a paramful URL, i.e. because the user opened 360 from a deeplink), and when the `popstate` event is triggered. So there's really no need to have this indirection of a ref and a callback watching `location`, since we don't depend on the `location` but are simply reacting to the event.

### DIFF
--- a/src/views/PeopleSearch/components/SearchFieldList/index.tsx
+++ b/src/views/PeopleSearch/components/SearchFieldList/index.tsx
@@ -155,9 +155,6 @@ const SearchFieldList = ({ onSearch }: Props) => {
   );
   const [searchParams, setSearchParams] = useState(initialSearchParams);
 
-  // Used to only read search params from URL on first load (and on back/forward navigate via event listener)
-  const shouldReadSearchParamsFromURL = useRef(true);
-
   const [loading, setLoading] = useState(true);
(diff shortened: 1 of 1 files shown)


russtuck (approved):
This is a complicated piece of code to fix (and review).  It uses (or stops using) useEffect, useCallback, useMemo, useRef, and useState.  Thanks for diving in and figuring it out!  

What you've done makes sense, and also works when I test it.

amos-cha (commented):
I don't think I'm in a place to approve since I don't particularly understand what is going on but if it has been verified by @russtuck to work then I can't complain!

Excellent find :)

russtuck:
I retested after the merge.

PR #1922, opened 2023-06-28
[COLOR] changed color styles to CSS for more files 

ahwnsals1234:


jsenning (dismissed):
These look good to me.

Centerville1 (changes requested):
lgtm so far!  We'll wait to merge this until we get our main PR merged to develop, so marking it for changes to prevent accidental merging.

ahwnsals1234:
I deleted the OfflineBanner file from 'src/components/OfflineBanner' because considering that it was only being called from the file itself, I assumed it is obsolete.

src/components/EventList/components/EventItem/EventItem.module.scss
```diff
@@ -9,25 +9,25 @@
...
-  color: $neutral-white;
+  .event_column,
+  .descriptionText {
+    color: var(--mui-palette-primary-contrastText);
   }
 }
 
 :global(.gc360_event_list) > section {
-  background-color: $neutral-white;
+  background-color: var(--mui-palette-primary-contrastText);
```
Centerville1, line 19: This should probably be neutral light, since it's a background color, and not text(I assume)

src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
```diff
@@ -1,9 +1,8 @@
 import { Polar } from 'react-chartjs-2';
 import { toTitleCase } from 'services/utils';
 import { Colors, VictoryPromiseCategory, VictoryPromiseColor } from 'services/victoryPromise';
-import { gordonColors } from 'theme';
 
-const colors = new Array<VictoryPromiseColor | typeof gordonColors.neutral.lightGray>(4);
+const colors = new Array<VictoryPromiseColor | '#dbdbda'>(4);
```
Centerville1, line 5: Since this color is used in two places, it might be better to save this color to a variable in this file and then use that, up to you though.

src/components/Profile/components/VictoryPromise/components/VictoryPromiseIcons.tsx
```diff
@@ -23,8 +22,8 @@ const VictoryPromiseIcon = ({
   category: VictoryPromiseCategory;
   active: boolean;
 }) => {
-  const primary = active ? Colors[category] : gordonColors.primary.blue;
-  const secondary = active ? gordonColors.neutral.grayShades[50] : gordonColors.neutral.gray;
+  const primary = active ? Colors[category] : '#014983';
+  const secondary = active ? '#ffffff' : '#CCCCCB';
```
Centerville1, line 26: Same thing here with the reused colors, if might be better to save the reused ones to a variable.

Also, just to double check, 'var(--mui-...)' inside quotes didn't work here, right?

src/components/Profile/components/VictoryPromise/components/VictoryPromiseIcons.tsx
```diff
@@ -44,10 +43,8 @@ const VictoryPromiseIcon = ({
   );
 };
 
-type PrimaryColor = VictoryPromiseColor | typeof gordonColors.primary.blue;
-type SecondaryColor =
-  | typeof gordonColors.neutral.gray
-  | (typeof gordonColors.neutral.grayShades)[50];
+type PrimaryColor = VictoryPromiseColor | '#014983';
```
Centerville1, line 46: I'm not convinced this is the same, since it was doing a typeof call before.  Could you tell what this was doing previously?

Centerville1 (changes requested):
Looks pretty good to me!  Just a couple changes and questions I had.

Centerville1 (commented):
lpgtm imho tbh, I'll approve this once we get our main branch merged to develop

Centerville1 (approved):
Good to go

Centerville1 (approved):

Issue #1139, opened 2021-05-26
ApartApps accepts multiple hall ranks of same value


cpabbot:
Fix using current UI: When the user clicks 'save' or 'submit,' there should be a snackbar error that pops up (similar to if you were to try and add two of the same hall) and informs the user that two halls cannot have the same rank.

![image](https://user-images.githubusercontent.com/17221652/119680199-f427ca00-be0e-11eb-9e4a-1129d8b5d1c1.png)

Better alternative:
Make these ListItems draggable so that way the user can intuitively rearrange the order; this way, no two halls can ever have the same rank and hallrank = index always. There does not appear to be a MUI implementation for this, so you may have to get a little creative :)

_Note: if you resolve this issue with the original UI, please make another issue with the "Better alternative" part so that can be a future improvement!_

bennettforkner:
It is worth noting that this might not actually be an issue. I'm not 100% sure how Ethan M/Housing would like it, but it makes sense to me that if someone does not consider one hall better than another that they could rank them the same. It could be a system issue, though I don't see any mention of that from @cpabbot.

First step is probably to do some testing to figure out what happens when you rank two halls equally. Then we can figure out what the end-user (Ethan/Housing) would like to happen.

andrew-wzj:
![Image](https://github.com/gordon-cs/gordon-360-ui/assets/98111145/ed12a6e1-a489-460b-b9b4-9fea2b72dee1)

PR #1923, opened 2023-06-28
Not allow duplicate ranks for halls

jaeminlee0515:
This was from issue #1139 where a user could add several halls to have the same ranks. Now it allows it so that you can only have one rank for each hall.

### DIFF
--- a/src/views/ApartmentApp/components/StudentApplication/index.jsx
+++ b/src/views/ApartmentApp/components/StudentApplication/index.jsx
@@ -477,6 +477,19 @@ const StudentApplication = ({ userProfile }) => {
             ['HallRank', 'HallName'], // Sort halls by rank and name
           ),
         }));
+      }
+      if (hallRankValue !== applicationDetails.ApartmentChoices[index].HallRank) {
+        // Display an error if the selected rank is already in the list
+        createSnackbar(`Rank ${String(hallRankValue)} is already in the list.`, 'info');
+
+        // Leave the ApartmentChoice array unchanged, but refresh the sorting.
+        setApplicationDetails((prevApplicationDetails) => ({
+          ...prevApplicationDetails,
+          ApartmentChoices: sortBy(
+            prevApplicationDetails.ApartmentChoices,
+            ['HallRank', 'HallName'], // Sort halls by rank and name
+          ),
+        }));
       } else {
         const newApartmentChoice = {
           ApplicationID: applicationDetails.ApplicationID,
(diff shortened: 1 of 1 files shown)


jaeminlee0515:
-Fixed transparency issue bug at the bottom bar of the application with the guidance of @Centerville1.

src/views/ApartmentApp/components/StudentApplication/components/HallChoiceList/index.jsx
```diff
@@ -46,7 +47,21 @@ const HallChoiceList = ({
...
+      setClickCount(apartmentChoices?.length);
+    }
+  }, [apartmentChoices]);
+
+  const handleAddHallClick = () => {
+    // Only continue if click count is less than 5
+    if (clickCount < 5) {
+      onHallAdd();
+      setClickCount((prevClickCount) => prevClickCount + 1);
+    }
```
amos-cha, line 63:
```suggestion
    if (apartmentChoices?.length >= 5) {
      setClickCount(5);
    } else {
      setClickCount(apartmentChoices?.length);
    }
  }, [apartmentChoices]);

  const handleAddHallClick = () => {
    // Only continue if click count is less than 5
    if (clickCount < 5) {
      onHallAdd();
      setClickCount((prevClickCount) => prevClickCount + 1);
    }
```
I do not suggest using a click counter. This can easily be done by checking the number of components you're rendering in the hall list.

src/views/ApartmentApp/components/StudentApplication/index.jsx
```diff
@@ -561,6 +572,11 @@ const StudentApplication = ({ userProfile }) => {
   const deleteApartmentApplication = async () => {
     setDeleting(true);
     setDeleteButtonAlertTimeout(null);
+    if (!applicationExists) {
+      setApplicationDetails(loadApplication);
+      setDeleting(false);
+      return;
+    }
```
amos-cha, line 579: it would be beneficial for readers to have a comment explaining why this condition is needed and why you need to short the function.

src/views/ApartmentApp/components/StudentApplication/index.jsx
```diff
@@ -623,7 +639,7 @@ const StudentApplication = ({ userProfile }) => {
           // Quick fix since the API is trying
           //  to validate the public profile object but there is no Hall property
           let applicationDetailsNoProfiles = applicationDetails;
-          applicationDetailsNoProfiles.Applicants.forEach((a) => a.Profile.Hall = '');
+          applicationDetailsNoProfiles.Applicants.forEach((a) => (a.Profile.Hall = ''));
```
amos-cha, line 642:
```suggestion
          applicationDetailsNoProfiles.Applicants.forEach((a) => a.Profile.Hall = '');
```

src/views/ApartmentApp/components/StudentApplication/components/HallChoiceList/index.jsx
```diff
@@ -46,7 +47,20 @@ const HallChoiceList = ({
...
     loadApartmentHalls();
-  }, []);
+
+    if (apartmentChoices?.length >= 5) {
+      setClickCount(5);
+    } else {
+      setClickCount(apartmentChoices?.length);
+    }
+  }, [apartmentChoices]);
+
```
amos-cha, line 57:
```suggestion
    if (apartmentChoices?.length >= 5) {
      setClickCount(5);
    } else {
      setClickCount(apartmentChoices?.length);
    }
  }, [apartmentChoices]);

```

I still do not like the idea of counting clicks. It is not a very robust system. Please use an implementation where we count the number of rendered components

amos-cha (changes requested):

amos-cha (approved):
Great work
giphy

PR #1925, opened 2023-06-28
S23 fix attended clw

Centerville1:
Fix for production attended cl&w events button error.


EjPlatzer (changes requested):
@Centerville1 It looks like you may have included changes from other branches in this PR. Please make sure to only include the changes relevant to this fix, so that we can review each set of changes in their own PR.

Centerville1:
Yep, my bad, I'll make a new PR.

PR #1926, opened 2023-06-28
Fixed cl&w Credits Button Bug

Centerville1:
Added check for empty date strings to add-to-calendar-button.  Potentially should be merged to master to fix bug there, so let me know how you want me to approach that...


EjPlatzer (approved):
The code looks good!

Please rename the PR before merging, it should describe the code changes. The current title only makes sense in the context of an unrelated PR, and isn't relevant to this PR directly.

As far as getting this fix into production, we could do a hotfix release once this is merged into develop and tested there. That said, I don't think this is necessarily a big enough issue to warrant a hotfix, since no one is checking their Attended CL&W events during the summer.

PR #1935, opened 2023-06-29
S23/add UI version to about page

jsenning:
Display UI deployment datetime and Git SHA at bottom of About page.  A parallel PR in the API fixes a bug so the API build datetime and Git SHA is also available to be displayed.

**NOTE:** Currently the UI version information is automatically created upon checkout using the script .husky/post-checkout.  This should probably become a GitHub action that is triggered by deployment.


jsenning:
Lint/Build test fails to find src/views/About/version.json.  This file needs to be automatically created by deployment.  Perhaps the correct solution is to keep it in the repo and overwrite it with the correct version during deployment.

jsenning:
When this PR started the UI version information is automatically created upon checkout using the script .husky/post-checkout.  
This has been changed and the version file is automatically created using a "prebuild" script listed in package.json.

russtuck (approved):
LGTM

PR #1939, opened 2023-06-30
[COLOR] S23 theme palette implementation

Centerville1:


src/views/ApartmentApp/index.jsx
```diff
@@ -57,6 +60,8 @@ const ApartApp = () => {
         <div className={'student_apartment_application'}>
           <StudentApplication userProfile={profile} />
         </div>
+        // MJ Note, does this code still get used/should we automate it somehow?
+        // Also should this be seperated into its own component?
```
Centerville1, line 64: @EjPlatzer there's a block of commented out code for apartApp, which contains the page to display after apartApps are closed for the year...  How does this work currently, is that something that should be automated or accessible outside of un-commenting the code?  Is this block still used, since it has a date from 2022 in it currently

ahwnsals1234 (approved):

amos-cha (approved):

ahwnsals1234 (approved):

PR #1942, opened 2023-06-30
Alumni checkbox will only be set checked by default if the user is only an alumni

antoniavonto:
Alumni checkbox will only be set checked by default if the user is only an alumni.
Taking in consideration that there are people that fall into many categories (seniors for example), we discussed what would be the best idea for the default.


amos-cha (approved):
LGTM, don't merge until after release

russtuck (approved):
LGTM

Issue #1927, opened 2023-06-28
people search department drop down order


andrew-wzj:
The "department" drop down menu for employees in people search has an odd order, because a seemingly random set of departments are listed as "Office of ...", "Center for", or "Department of".  It might be easier to find things if a these prefixes were shown on the end, like "Athletics, Office of".  

This could be done with regular expressions in the UI, or the API.  It could also be done in the database, but that might have undesirable consequences like changing how the department name appears in other places.  Once an approach is determined, we can convert this to an issue in the appropriate repo.

antoniavonto:
https://github.com/gordon-cs/gordon-360-ui/pull/1944

PR #1944, opened 2023-07-03
arranging drop down of departments in people search

antoniavonto:
Now the drop down of the departments in people search are easier to find, once the "Office of", "Department of", and "Center for" were put at the end of each item. Now the order is arranged by the main name of the department.
issue: https://github.com/gordon-cs/gordon-360-ui/issues/1927
before:
image

after:
image


antoniavonto:
@amos-cha

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -245,6 +250,21 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+        dep[i] = dep[i].replace(/^Office of /, '') + ' (Office of)';
+      } else if (/^Center for /.test(d)) {
+        dep[i] = dep[i].replace(/^Center for /, '') + ' (Center for)';
+      } else if (/^Department of /.test(d)) {
+        dep[i] = dep[i].replace(/^Department of /, '') + ' (Department of)';
+      }
+    });
+    console.log(dep);
+    return dep;
+  };
```
EjPlatzer, line 266: Wouldn't it make more sense to do this within the `peopleSearchService`? It seems to me that the service code, which handles retrieving and formatting code, should be the single place where this transformation from internal form to user interface form happens.

src/services/peopleSearch.ts
```diff
@@ -75,6 +75,20 @@ export type PeopleSearchQuery = {
...
+const getOldName = (dep: string) => {
+  if (dep.endsWith('(Office of)')) {
+    dep = 'Office of ' + dep.replace(/ \(Office of\)/, '');
+    console.log(dep);
+  } else if (dep.endsWith('(Center for)')) {
+    dep = 'Center for ' + dep.replace(/ \(Center for\)/, '');
+    console.log(dep);
+  } else if (dep.endsWith('(Department of)')) {
+    dep = 'Department of ' + dep.replace(/ \(Department of\)/, '');
+    console.log(dep);
```
EjPlatzer, line 87: Please remove these console logs before merging.

EjPlatzer (changes requested):

src/views/PeopleSearch/components/SearchFieldList/index.tsx
antoniavonto: That is right, Thank you!

src/services/peopleSearch.ts
```diff
@@ -75,16 +76,27 @@ export type PeopleSearchQuery = {
...
+    if (/^Office of /.test(d)) {
+      dep[i] = dep[i].replace(/^Office of /, '') + ' (Office of)';
+    } else if (/^Center for /.test(d)) {
+      dep[i] = dep[i].replace(/^Center for /, '') + ' (Center for)';
+    } else if (/^Department of /.test(d)) {
+      dep[i] = dep[i].replace(/^Department of /, '') + ' (Department of)';
+    }
+  });
+  return dep;
+};
```
EjPlatzer, line 91: We can simplify this function a little bit more:
1. `Promise.all()` is only needed when you want to await a list of promises in parallel. To await one Promise, you do not need it:
```suggestion
const getRenamedDepartments = async () => {
  const dep = await getDepartments();
  dep.map((d, i) => {
    if (/^Office of /.test(d)) {
      dep[i] = dep[i].replace(/^Office of /, '') + ' (Office of)';
    } else if (/^Center for /.test(d)) {
      dep[i] = dep[i].replace(/^Center for /, '') + ' (Center for)';
    } else if (/^Department of /.test(d)) {
      dep[i] = dep[i].replace(/^Department of /, '') + ' (Department of)';
    }
  });
  return dep;
};
```
2. `Array.map()` constructs a new array with the same length as the original, where each element is the result of the callback. For example, `const squares = [1, 2, 3].map(x => { return x * x; })` sets `squares` to `[2, 4, 6]`. So the way you're currently using it is wasteful because you're mutating the original array `dep` while also creating a new array of the same length that is filled with `undefined` elements (since a function/callback that doesn't return anything implicitly returns `undefined` in Javascript). The two ways to do this are to either:
   1. Use `Array.forEach` instead, which will loop through the elements with constructing a new array, so you can mutate the original array with no waste.
   2. Use `map` as intended. This way has some upsides because you don't have to worry about indices and mutation, you just declare the new value from the old value, which makes it much less error prone. It is technically less performant, but the difference is irrelevant unless you're mapping millions of elements constantly.
```suggestion
const getRenamedDepartments = async () => {
  const dep = await getDepartments();
  const updatedDeps = dep.map((d) => {
    if (/^Office of /.test(d)) {
      return d.replace(/^Office of /, '') + ' (Office of)';
    } else if (/^Center for /.test(d)) {
      return d.replace(/^Center for /, '') + ' (Center for)';
    } else if (/^Department of /.test(d)) {
      return d.replace(/^Department of /, '') + ' (Department of)';
    }
  });
  return updatedDeps;
};
```
or more succinctly
```suggestion
const getRenamedDepartments = async () => {
  const dep = await getDepartments();
  return dep.map((d) => {
    if (/^Office of /.test(d)) {
      return d.replace(/^Office of /, '') + ' (Office of)';
    } else if (/^Center for /.test(d)) {
      return d.replace(/^Center for /, '') + ' (Center for)';
    } else if (/^Department of /.test(d)) {
      return d.replace(/^Department of /, '') + ' (Department of)';
    }
  });
};
```
3. You could, if you want, use `.then` instead of `await`:
```suggestion
const getRenamedDepartments = () =>
  getDepartments().then(dep => dep.map((d) => {
    if (/^Office of /.test(d)) {
      return d.replace(/^Office of /, '') + ' (Office of)';
    } else if (/^Center for /.test(d)) {
      return d.replace(/^Center for /, '') + ' (Center for)';
    } else if (/^Department of /.test(d)) {
      return d.replace(/^Department of /, '') + ' (Department of)';
    }
  }));
```
And you can use the `map` utility function in `src/services/utils.ts` to make this slightly less verbose:
```suggestion
const getRenamedDepartments = () =>
  getDepartments().then(map((d) => {
    if (/^Office of /.test(d)) {
      return d.replace(/^Office of /, '') + ' (Office of)';
    } else if (/^Center for /.test(d)) {
      return d.replace(/^Center for /, '') + ' (Center for)';
    } else if (/^Department of /.test(d)) {
      return d.replace(/^Department of /, '') + ' (Department of)';
    }
  }));
```
4. You could actually collapse all three cases into one only slightly more complicated regular expression:
```suggestion
// Match any department whose name starts with one of the prefixes "Office of ", "Department of " or "Center for "
const departmentPrefixRegex = /^(Office of|Department of|Center for) (.*)$/;
const getRenamedDepartments = () =>
  // Match known prefixes and, if matched, move the prefix to the end of the department name in parentheses
  getDepartments().then(map((department) => department.replace(departmentPrefixRegex, "$2 ($1)"));
```

antoniavonto: For some reason the .map without dep[i] = dep[i] is not working, so I changed for the forEach. 

Also when using the big regular exp /^(Office of|Department of|Center for) (.*)$/ and then trying to replace(departmentPrefixRegex, "$2 ($1)")); is also not working, even making more changes to the syntax. Is it ok to not be that concise?

russtuck:
This is working well.  However, two "departments" use "Office of the", and these end up at the very end, under "the".  For example, "Office of the Registrar" shows up as "the Registrar (Office of)" and it would be more useful as "Registrar (Office of the)".

src/services/peopleSearch.ts
```diff
@@ -75,6 +77,32 @@ export type PeopleSearchQuery = {
   involvement: string;
 };
 
+const getRenamedDepartments = async () => {
+  const dep = await getDepartments();
+  dep.forEach((d, i) => {
```
russtuck, line 82: Something to think about, but NOT required...
Create a getNewName() method just like getOldName(), and have getRenamedDepartments call getNewName().  

Pro: The two conversions are very similar, so it's even easier to see that they reverse each other (and that could be easily tested with a unit test).  And the forEach would have a very simple body.

Con: It might be little more code.

russtuck (changes requested):
Nice work!  Please just fix the "office of the" (e.g. Registrar).

src/services/peopleSearch.ts
antoniavonto: Just resolved the (office of the) case, thank you for pointing it out!

Th only thing about calling both get old name and get new name in the getRenamedDepartments is that we only want to get the old name after the function search is called, but the new name is needed before calling the search function.

src/services/peopleSearch.ts
```diff
@@ -75,6 +77,36 @@ export type PeopleSearchQuery = {
...
+      dep[i] = dep[i].replace(/^Office of the /, '') + ' (Office of the)';
+    } else if (/^Office of /.test(d)) {
+      dep[i] = dep[i].replace(/^Office of /, '') + ' (Office of)';
+    } else if (/^Center for /.test(d)) {
+      dep[i] = dep[i].replace(/^Center for /, '') + ' (Center for)';
+    } else if (/^Department of /.test(d)) {
+      dep[i] = dep[i].replace(/^Department of /, '') + ' (Department of)';
+    }
+  });
+  console.log(dep);
```
amos-cha, line 93:
```suggestion
```

src/services/peopleSearch.ts
```diff
@@ -75,6 +77,36 @@ export type PeopleSearchQuery = {
...
+    dep = 'Office of the ' + dep.replace(/ \(Office of the\)/, '');
+  } else if (dep.endsWith('(Office of)')) {
+    dep = 'Office of ' + dep.replace(/ \(Office of\)/, '');
+  } else if (dep.endsWith('(Center for)')) {
+    dep = 'Center for ' + dep.replace(/ \(Center for\)/, '');
+  } else if (dep.endsWith('(Department of)')) {
+    dep = 'Department of ' + dep.replace(/ \(Department of\)/, '');
+  }
+  return dep;
+};
```
amos-cha, line 108: is there a specific use case for this? I feel like we could just `getDepartments` again right? Feel free to correct me if I'm wrong. (or you could map `renamedDepartments` with the original `getDapartments` if you need both at the same time)

amos-cha (changes requested):

src/services/peopleSearch.ts
```diff
@@ -1,4 +1,6 @@
+import { useState } from 'react';
 import http from './http';
+import { forEach } from 'lodash';
```
amos-cha, line 3:
```suggestion
import { forEach } from 'lodash';
```
can you check if this import is being used? it shouldn't be as `typescript`/`javascript` has a inbuilt method for their array prototypes

src/services/peopleSearch.ts
amos-cha:
```
const getRenamedDepartments = async () => {
  const dep = await getDepartments();
  const mappedResults = [];
  dep.forEach((d) => {
    let departmentName = { name: d, formattedName: d }
    if (/^Office of the /.test(d)) {
      departmentName.formattedName = d.replace(/^Office of the /, '') + ' (Office of the)';
    } else if (/^Office of /.test(d)) {
      departmentName.formattedName = d.replace(/^Office of /, '') + ' (Office of)';
    } else if (/^Center for /.test(d)) {
      departmentName.formattedName = d.replace(/^Center for /, '') + ' (Center for)';
    } else if (/^Department of /.test(d)) {
      departmentName.formattedName = d.replace(/^Department of /, '') + ' (Department of)';
    };
    mappedResults.push(departmentName);
  });
  return mappedResults;
};
```
you could probably replace the first function with something along the lines of this. I didn't write this function with linting or checking for syntax, so this could potentially be pseudo code.

Then outside of the service, when you pass the query back to the `search` route, you can just pass the original `name`

src/services/peopleSearch.ts
```diff
@@ -74,6 +74,29 @@ export type PeopleSearchQuery = {
...
+      departmentName.label = d.replace(/^Office of /, '') + ' (Office of)';
+    } else if (/^Center for /.test(d)) {
+      departmentName.label = d.replace(/^Center for /, '') + ' (Center for)';
+    } else if (/^Department of /.test(d)) {
+      departmentName.label = d.replace(/^Department of /, '') + ' (Department of)';
+    }
+    mappedResults.push(departmentName);
+  });
+  return mappedResults;
+};
```
EjPlatzer, line 99: Circling back to the regular expression idea, I have tested this code in your branch and I know for a fact that it works:
```suggestion
/**
 * Match any department whose name starts with one of these prefixes:
 *  - "Office of (the) "
 *  - "Department of "
 *  - "Center for "
 */
const DepartmentPrefixRegex = /^(Office of(?: the)?|Department of|Center for) (.*)$/;

const getRenamedDepartments = async (): Promise<SelectOption[]> => {
  const departmentNames = await getDepartments();
  return departmentNames.map<SelectOption>((departmentName) => ({
    value: departmentName,
    label: departmentName.replace(DepartmentPrefixRegex, '$2 ($1)'),
  }));
};
```

I also think that, even despite the regular expression, it's clearer in intent and easier to read and maintain in the long run. That's just my opinion though.

src/services/peopleSearch.ts
```diff
@@ -74,6 +74,29 @@ export type PeopleSearchQuery = {
   relationship_status?: string;
   involvement: string;
 };
+export interface SelectOption {
+  label: string;
+  value: string;
+}
+
+const getRenamedDepartments = async () => {
```
EjPlatzer, line 82: I think we can give this a clearer name, especially since it no longer returns regular strings. Maybe `getDepartmentDropdownOptions`?

src/services/peopleSearch.ts
```diff
@@ -74,6 +74,29 @@ export type PeopleSearchQuery = {
   relationship_status?: string;
   involvement: string;
 };
+export interface SelectOption {
+  label: string;
+  value: string;
+}
```
EjPlatzer, line 80: This interface is already defined in and exported from `views/PeopleSearch/components/SearchFieldList/components/SearchField`. Since that is the type that we are trying to produce, it's better to import it from there than to redefine it here and risk the two becoming inconsistent overtime.

EjPlatzer (commented):
This is looking great! I like the use of the SelectOption type to provide the value and label together from the service!

I had a couple of thoughts on cleanup, and one (re-)suggestion.

EjPlatzer (approved):
LGTM!

amos-cha (approved):
👍🏼

russtuck (approved):
Nice work!  LGTM

PR #1951, opened 2023-07-05
set timezone by currentBrowser

andrew-wzj:


amos-cha (approved):
LGTM, don't merge till after release

PR #1954, opened 2023-07-05
Create UpdateLogo.md

ahwnsals1234:
A rough draft of the documentation


Centerville1:
It looks like this was never completed.  Might be good to document, but probably won't get done at this point in the Summer.  It could be a helpful project to work on documentation for how the PWA is set up overall, and not just for updating splash screens and favicons, since it's somewhat confusing to figure out for the first time.

russtuck (approved):
While it's brief, it's correct and useful, so I'd suggest merging it.  

(It would be great to add more later, but we can go ahead and save this much.)

Centerville1:
> While it's brief, it's correct and useful, so I'd suggest merging it.
> 
> (It would be great to add more later, but we can go ahead and save this much.)

Works for me, I'd be willing to work on adding more to this at some point, potentially during the semester, so I'll keep it in mind. :+1:

PR #1955, opened 2023-07-05
S23 Finalizing Theme Update - Polishing for Dark Mode

Centerville1:
Finishes up theme palette update and applies a numerous spot fixes for dark mode colors

### DIFF
--- a/src/components/EventList/components/EventItem/index.jsx
+++ b/src/components/EventList/components/EventItem/index.jsx
@@ -54,7 +54,7 @@ const EventItem = ({ event }) => {
               description={event.Description}
               onClick={() => setExpanded((e) => !e)}
               lightMode="system" // Set to system @TODO use localstorage setting to update
-              /*currently will just be set to the users system setting, no way of knowing the
+              /* Currently will just be set to the users system setting, no way of knowing the
               360 local set mode. For users in system color scheme, and those who have 
               their system set the same as 360, the button will appear correct, but otherwise the 
               button might be light in dark mode and vice versa.


src/views/Home/components/DiningBalance/index.jsx
```diff
@@ -4,8 +4,8 @@ import { useEffect, useState } from 'react';
 import { Doughnut } from 'react-chartjs-2';
 import session from 'services/session';
 import user from 'services/user';
-import { gordonColors } from 'theme';
 import styles from '../Doughnut.module.css';
+import { newTheme } from 'theme';
```
russtuck, line 8: Consider "theme360" instead of "newTheme" everywhere, since it won't always be new.

src/views/News/components/NewsItem/index.jsx
```diff
@@ -130,10 +128,10 @@ const NewsItem = ({ posting, unapproved, size, handleNewsItemEdit, handleNewsIte
         </Grid>
 
         {/* Collapsable details */}
-        <Collapse in={open} timeout="auto" unmountOnExit style={{ width: '100%' }}>
+        <Collapse in={open} timeout="auto" unmountOnExit className={styles.collapsable}>
```
russtuck, line 131:
```suggestion
        <Collapse in={open} timeout="auto" unmountOnExit className={styles.collapsible}>
```

src/views/Timesheets/index.jsx
```diff
@@ -365,6 +366,8 @@ const Timesheets = (props) => {
     </Button>
   );
 
+  console.log(newTheme);
```
russtuck:
```suggestion
```

russtuck (changes requested):
Looking at dark mode:

- [x] Home tab "days finished" number has too little contrast.
- [x] linkedin logo is too dark 
- [x] Events tab: "Gordon" and "+Filters" need more contrast
- [ ] People tab: checked checkboxes need more contrast; Gordon needs more contrast

Interesting choice of night marsh for header, instead of primary blue, which also seems dark enough (though not quite as dark).  What's the reasoning for that?  Please get an opinion from Steve Dagley on this.  But we don't have to wait on his opinion to merge, since it's a trivial change if he thinks we should make it.

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -47,6 +53,7 @@ const EventItem = ({ event }) => {
               timeZone="currentBrowser"
               description={event.Description}
               onClick={() => setExpanded((e) => !e)}
+              lightMode="system" // Set to system, if we give users
```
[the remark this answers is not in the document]
Centerville1: So, this still needs some work, and I'm not sure how to solve it to be honest...  The add-to-calendar button just takes a mode.  In system, it will be just set to the users system setting, but it has no way of knowing what the 360 local set mode is.  So, for users who are set to system in 360, and those who have their system set the same as 360, the button will appear correct, but otherwise the button might be light in dark mode and vice versa.

I'm thinking of trying to create a global function or hook or something to check if 360 is in dark mode within javascript files, but it doesn't seem like there is an easy way to do that in MUI.  

For now, it works in some cases, and doesn't look to terrible otherwise, and I'm aware of the issue, but I can add a TODO comment to the code here if that is helpful @russtuck

src/theme.ts
```diff
@@ -253,12 +266,12 @@ export const newTheme = extendTheme({
         text: {
           primary: White, // Various MUI components and text
         },
-        Switch: {
-          defaultColor: TestTool, // switch ball off color
-        },
-        common: {
-          onBackground: TestTool, // switch track off color
-        },
+        // Switch: {
```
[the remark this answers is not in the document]
Centerville1: These are overrides for MUI component colors.  I think it's a good idea to leave them in, because they are potentially helpful for future customization.  In this case, if we want to customize the colors of switches when they are turned off (currently a light gray by default)

I will add a comment explaining the purpose for them though

src/components/EventList/components/EventItem/index.jsx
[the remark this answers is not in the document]
russtuck: Your explanation here is really clear.  Please put it in a comment, so it will stay with the code.  (You can include a TODO comment about possibly doing something better.)

src/theme.ts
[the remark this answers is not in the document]
russtuck: OK, a comment explaining the purpose is a good solution.

russtuck:
Almost there.  Just one comment to add (from this conversation)

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -53,7 +53,14 @@ const EventItem = ({ event }) => {
...
               description={event.Description}
               onClick={() => setExpanded((e) => !e)}
-              lightMode="system" // Set to system, if we give users
+              lightMode="system" // Set to system @TODO use localstorage setting to update
+              /*currently will just be set to the users system setting, no way of knowing the
+              360 local set mode. For users in system color scheme, and those who have 
+              their system set the same as 360, the button will appear correct, but otherwise the 
+              button might be light in dark mode and vice versa.
+              For now, it works in some cases, and doesn't look too terrible otherwise
+              Potential fis is to query the localStorage key to get the theme mode setting for the
```
russtuck, line 62:
```suggestion
              Potential fix is to query the localStorage key to get the theme mode setting for the 
```

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -53,7 +53,14 @@ const EventItem = ({ event }) => {
               timeZone="currentBrowser"
               description={event.Description}
               onClick={() => setExpanded((e) => !e)}
-              lightMode="system" // Set to system, if we give users
+              lightMode="system" // Set to system @TODO use localstorage setting to update
+              /*currently will just be set to the users system setting, no way of knowing the
```
russtuck, line 57:
```suggestion
              /* Currently will just be set to the users system setting, no way of knowing the
```

russtuck (dismissed):
Great.  

Since accepting these 2 typo fixes might clear my approval, I think I'll accept them for you and reapprove.

russtuck (approved):

Issue #1957, opened 2023-07-05
User Cannot Save Social Media Link (Only temporarily)


jaeminlee0515:
Found this while looking at [#1879](https://github.com/gordon-cs/gordon-360-ui/issues/1879), where if you put in a valid social media link and click "Submit", it will show that it is saved and added to your profile page. But when you refresh the page or go to a different page and then navigate to the "My Profile" page, the social media link is gone.

Luke-W-Hart:
FIxed by #2229

PR #1959, opened 2023-07-05
Maintain social media link status

jaeminlee0515:
This allows for the user to save a social media link and when the page is refreshed or navigated to a different page, you can now go back to the "My Profile" page, the social media link will be present. #1957


jaeminlee0515:
Archive

PR #1962, opened 2023-07-05
Allow equal margin of timesheet card

jaeminlee0515:
Now it will allow for the margin's of the card to be equal.

Screenshot 2023-07-05 at 14-52-38 Gordon 360

It would now look like the above.


src/views/Timesheets/components/SavedShiftsList/index.jsx
amos-cha, line 1: @EjPlatzer are we planning to still use 360 timesheets? if so, would it be worth it to refactor this to a functional component

amos-cha (commented):
Looks good

src/views/Timesheets/components/SavedShiftsList/index.jsx
EjPlatzer: > @EjPlatzer are we planning to still use 360 timesheets? if so, would it be worth it to refactor this to a functional component

No, I believe I told y'all this at the beginning of the summer when CTS learned it - we are finally moving forward with having student workers use Criterion for timesheets, so this code continues to be in maintenance-only mode.

amos-cha: I remember this conversation ~sort-of.

PR #1969, opened 2023-07-06
S23 my public profile involvements

antoniavonto:
To fix Issue: When you click on view my public profile your private involvements still show up for you
API: https://github.com/gordon-cs/gordon-360-api/pull/951


antoniavonto:
involvements should show up grayed out because the user share those club sections with himself and all people in the same club section can see its members.

src/components/Profile/components/MembershipsList/index.jsx
```diff
@@ -5,6 +5,7 @@ import { Link } from 'react-router-dom';
 import membershipService from 'services/membership';
 import MembershipInfoCard from './components/MembershipInfoCard';
 import styles from './MembershipsList.module.css';
+import MyProfile from 'views/MyProfile';
```
amos-cha, line 8: mis-import?

PR #1980, opened 2023-07-07
S23 page headers

jhammond4:
Adding headers to 360 pages without them for consistency and design. I previously had separate prs for each page, but I consolidated them here.


jhammond4:
Screenshot from 2023-07-07 16-35-34
Screenshot from 2023-07-07 16-46-21
Screenshot from 2023-07-07 16-35-55
Screenshot from 2023-07-07 16-46-41
Screenshot from 2023-07-07 16-37-04
Screenshot from 2023-07-07 16-47-32
Here's some before and after screenshots of the current state

jhammond4:
There are also two things I can't work out how to fix. Any suggestions?
-The Links popup has a blank space at its top which I can't get to disappear. I've tried rearranging and different calls but nothing works.
- The subheader of the Involvements pages won't change color, even though I am using the same logic as I am on the About page.
Screenshot from 2023-07-07 16-38-10
Screenshot from 2023-07-07 16-36-46

russtuck (changes requested):
First, a few impressions from looking at the UI.  

Home tab: looks great.  The card headers look good, don't change the layout, and improve consistency.

Involvements tab:
* the text for in the search boxes doesn't have enough contrast to be legible.  The old font colors seem fine.
* with search moved down, membership requests is front and enter.  This makes it important for it to be small and out of the way when there's no action needed. So, unless there is a request that needs action, please collapse it automatically.  
* I think My Current Involvements would be a more logical first card on this page.  If you agree, let's move it to the top.  (My reasoning is that current involvements are the most commonly used item on the page.  The reason for putting approvals at the top is so they get attention when needed.  But to do that, they need to be almost invisible when not needed.)

Events:
* same comment about search bar text being too faint
* header looks good, and moving "Attended CL&W" down to the search bar line is both more logical and a cleaner layout

People tab:
* same faint search box font
* unchecked boxes are invisible
* latest changes in develop (more search options, profile page changes) haven't been merged - please do that, and check that it doesn't break anything
* I'm not sure the card header is good here, since there's only one card;  it looks like a header in a header.

Others: looks good

I'll share this much, now, and will look at the code next.

src/services/peopleSearch.ts
```diff
@@ -58,6 +58,7 @@ export type SearchResult = SearchResultBase &
   );
 
 export type PeopleSearchQuery = {
+  class_year: string;
```
russtuck: I'm not sure why this is here.  It seems unrelated to what you're doing.  Maybe it's a partial merge from another branch?

src/theme.ts
```diff
@@ -252,6 +255,7 @@ export const newTheme = extendTheme({
         },
         text: {
           primary: White, // Various MUI components and text
+          secondary: White,
```
russtuck, line 255: Is it OK for primary and secondary to be identical?

src/views/About/index.jsx
```diff
@@ -5,14 +5,15 @@ import contributors from './contributors.json';
...
+import styles2 from 'app.module.css';
 
 import { Typography, Grid, Button, Card, CardHeader, CardContent } from '@mui/material';
 
 const About = () => {
   const [versionAPI, setVersion] = useState(null);
 
   useEffect(() => {
-    versionService.getVersion().then(setVersion);
+    //versionService.getVersion().then(setVersion);
```
russtuck, line 16: If this isn't needed, please delete the whole hook.

src/views/Home/components/DiningBalance/index.jsx
```diff
@@ -138,7 +138,7 @@ const DiningBalance = () => {
           style={{ paddingTop: 5, paddingBottom: 10 }}
         >
           <Grid item>
-            <Typography variant="body2" className={styles.label2}>
+            <Typography variant="body2" style={{ color: 'gray', textAlign: 'center' }}>
```
russtuck, line 141: I've seen a lot of changes like this.  Is it really better?  The old code seems cleaner, so I wonder what's driving the change?

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -166,6 +171,9 @@ const SearchFieldList = ({ onSearch }: Props) => {
   );
   const [searchParams, setSearchParams] = useState(initialSearchParams);
 
+  // Used to only read search params from URL on first load (and on back/forward navigate via event listener)
```
russtuck: How did these changes get in this PR?

russtuck (changes requested):
Looked at code.  A few questions and small change requests.

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -233,18 +225,15 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
     };
 
     loadPage();
-
   }, []);
   useEffect(() => {
-    // Read search params from URL on navigate (including first load)
-    if (shouldReadSearchParamsFromURL.current) {
-      const newSearchParams = deserializeSearchParams(new URLSearchParams(location.search));
+    const readSearchParamsFromURL = () => {
```
russtuck: Why is this code changing in this PR?

jhammond4: After a git pull I kept getting errors in this file, so I used the file from develop then made my necessary header changes to it

src/services/peopleSearch.ts
jhammond4: That's from another merge, it should be removed by the latest commit

src/theme.ts
jhammond4: I don't remember what that was for, but I notice now it makes a few unnecessary changes-I'll remove it

src/views/About/index.jsx
jhammond4: I was originally getting errors with this line after a pull, I just forgot to add it back.

jhammond4:
What are your thoughts on the Timesheets header? That one only has the one tab, but I feel like that box is so big it looks a bit weird without it. We could also make the box smaller

russtuck:
This is looking good.  The timesheets header is OK, and especially makes sense when other cards become visible.  I think just check for those other cards, then handle the merge conflicts and test to be sure everything still works after merging from develop.

amos-cha (commented):
I'm not the best for judging css best practice

src/components/Header/index.jsx
```diff
@@ -163,15 +163,9 @@ const GordonHeader = ({ onDrawerToggle }) => {
           >
             <MenuIcon className={styles.menu_button_icon} />
           </IconButton>
-          <Link
-            to="/"
-            component={ForwardNavLink}
-            value={tabIndex}
-            onClick={(event, value) => setTabIndex(value)}
-          >
+          <Link to="/" component={ForwardNavLink} value={tabIndex}>
```
Centerville1, line 166: This fixes the page reloading when clicking the image to go home.

russtuck (approved):
LGTM.  Please merge so we can include this in the release.

One thought for a potential followup: "styles2" isn't a wonderful name.  Would it be clearer to have "styles" be "stylesTimesheet" and "styles2" be just "styles" (since they seem to be the ones shared with other tabs)?

src/views/Timesheets/index.jsx
Centerville1, line 1: This is reverting a lot of changes to develop that shouldn't be, I think this is from the outdated version of develop you were using.

src/views/InvolvementProfile/InvolvementProfile.module.scss
```diff
@@ -11,17 +11,3 @@
...
-}
-
-.imageDrop {
-  max-width: 320px;
-  max-height: 320px;
-  border-radius: 0.5rem;
-}
-
-.activityURL {
-  font-weight: bold;
```
Centerville1, line 26: These classes are still being used in the index file, did you mean to delete them?

src/views/Home/components/NewsCard/index.jsx
```diff
@@ -3,15 +3,14 @@ import { useEffect, useState } from 'react';
 import { Link } from 'react-router-dom';
 import newsService from 'services/news';
 import NewsItem from 'views/News/components/NewsItem';
-import styles from './NewsCard.module.css';
```
Centerville1, line 6: This scss can be deleted here, since it's not being used

src/views/Home/components/Doughnut.module.scss
```diff
@@ -17,6 +17,11 @@
   text-shadow: 0px 1px 6px var(--mui-palette-neutral-A400);
 }
 
+.button {
```
Centerville1, line 20: Use ```color="secondary"``` instead

src/views/Home/components/DiningBalance/index.jsx
```diff
@@ -41,6 +41,7 @@ const DiningBalance = () => {
   } else if (typeof diningInfo !== 'object') {
     //Set color to use when displaying balance based on how low it is...
     const diningBalance = parseInt(diningInfo);
+    let balanceColor = colors.secondary.green;
```
Centerville1, line 44: This line probably shouldn't have been added, colors.secondary.green doesn't exist.

src/views/Events/Events.module.scss
```diff
@@ -1,23 +1,20 @@
...
+
   &_icon {
-    color: var(--mui-palette-neutral-500);
+    color: var(--mui-palette-neutral-700);
     font-size: 20;
   }
-  &_filter_button {
-    font-weight: bold;
+
+  &_button {
```
Centerville1, line 11: This should use ```color="secondary"``` and then remove this class

src/views/BannerSubmission/BannerSubmission.module.scss
```diff
@@ -1,9 +1,9 @@
-.bannerSubmission {
-  &_header {
+.banner_submission {
+  &_upload_button {
     background-color: var(--mui-palette-primary-main);
```
Centerville1, line 3: This should use ```color="primary"``` instead

Issue #1974, opened 2023-07-07
Add grayed out information about involvements privacy


antoniavonto:
If you have one involvement session as private it will still show up for other members of that club and section.
It would be good to show this information in a tooltip or something else for the user.

PR #1982, opened 2023-07-10
added shaded note about involvements privacy

antoniavonto:
issue: https://github.com/gordon-cs/gordon-360-ui/issues/1974

image


russtuck (approved):
LGTM.  Thanks for investigating this and making it clearer.

Issue #1140, opened 2021-05-26
ApartApps General UI Improvements


cpabbot:
- Make the UI of 'Need Help' consistent, maybe change from dropdown entirely to use something more in line with conventional 'Need Help' dialogs (I can't think of an example off hand but you can look it up)
![image](https://user-images.githubusercontent.com/17221652/119681308-f3436800-be0f-11eb-928b-1c9c79e5e638.png)

- Clean up admin view (you can use 360.stafftest account to test this)
   - round avg. points to maybe two decimal places
   - fix alignments (some headers are left aligned while content is centered, etc.)
   - optimize layout - adjust the width sizing of the tables, etc.
   - dropdown is "janky" as @EjPlatzer would say
![image](https://user-images.githubusercontent.com/17221652/119683435-a52f6400-be11-11eb-9850-a95e68527e97.png)

andrew-wzj:
@EjPlatzer Could we have access to 360.stafftest account?

andrew-wzj:
Is this issue already been addressed?

andrew-wzj:
@cpabbot We have saved password for 360.stafftest, but there is a problem on accessing verification code. I wonder who has the authorization to login stafftest account

andrew-wzj:
![Image](https://github.com/gordon-cs/gordon-360-ui/assets/98111145/01f8bd75-8b92-453f-9683-2753bfaf7f7a)

andrew-wzj:
![cytomtmy](https://github.com/gordon-cs/gordon-360-ui/assets/98111145/573e0060-e09b-4a0a-92ac-3327fbd1a28b)
The color for header display differently compared to last screenshot, and the problem of aligning text still exists

andrew-wzj:
round avg. points to maybe two decimal places (already fixed)

andrew-wzj:
![Screenshot from 2023-07-10 15-56-54](https://github.com/gordon-cs/gordon-360-ui/assets/98111145/fe83b38c-93ee-456b-bd69-52cafb3eafce)
This is current view and issues are addressed.

PR #1986, opened 2023-07-10
Align apartapps and refresh app data button center

jaeminlee0515:
#1140


silasw2:
Should probably merge develop again before requesting a review, I can take a look after develop is in!

jaeminlee0515:
> Should probably merge develop again before requesting a review, I can take a look after develop is in!

Will do! Thanks!

silasw2 (approved):
LGTM! I cannot run locally so I can't see the exact changes but as long as it looks good on your end there should not be any issues.

andrew-wzj:
> LGTM! I cannot run locally so I can't see the exact changes but as long as it looks good on your end there should not be any issues.

Awesome!

jaeminlee0515:
> LGTM! I cannot run locally so I can't see the exact changes but as long as it looks good on your end there should not be any issues.

This is what it looked like before the fix:
image

This is what it looks like after the fix:
image

PR #1987, opened 2023-07-10
Revert "Add #1885 (new logo on header) to release"

russtuck:
Jack and Matt want to work on this a little more before merging.


russtuck:
Oops, I should revert the push to develop, instead (and then merge that here).

PR #1988, opened 2023-07-10
Revert "New logo on header with home button"

russtuck:
Reverts gordon-cs/gordon-360-ui#1885

I merged this prematurely, so reverting it for more work.


Centerville1 (approved):
lgtm, be aware that when we go to commit this later (after changes) it could cause some issues since the commits are still in develop.  We'll have to revert this revert after the release.

Issue #1851, opened 2023-06-09
Retain results in People Search


ArabellaJi:
When people search for someone on the People page, there will be a number of results listed (Image 1). Then, if they select one of them, they will go to the Profile page (Image 2). However, if they press the back arrow to go back to the original People page, the text they entered is still there (in this case, 360 is still in the "First Name" field), but the results are gone (Image 3). It would be nice if the results were retained when people came back.

![Image](https://github.com/gordon-cs/gordon-360-ui/assets/78650688/9f33fb2c-158f-4ef2-93fa-9fd723c3f0ec)
![Image](https://github.com/gordon-cs/gordon-360-ui/assets/78650688/5c3070d1-aff8-4a3b-b0b7-c8ec797a12c5)
![Image](https://github.com/gordon-cs/gordon-360-ui/assets/78650688/f7c6a1d3-9a7a-4a6d-bc63-0b0e34f34e98)

antoniavonto:
https://github.com/gordon-cs/gordon-360-ui/pull/1990

antoniavonto:
this issue was solved, but there is still space for efficiency improvement

PR #1990, opened 2023-07-10
People search results after back button

antoniavonto:
Instead of retaining the people search results, this is calling the search function with the retained parameters to simulate retaining the search results.

Issue: https://github.com/gordon-cs/gordon-360-ui/issues/1851


src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -173,8 +173,8 @@ const SearchFieldList = ({ onSearch }: Props) => {
    * Whether the user can search for the current params.
    * This prevents a search with empty params, which freezes the client by trying to render thousands of results
    */
-  const canSearch = useMemo(() => {
-    const { includeStudent, includeFacStaff, includeAlumni, ...criteria } = searchParams;
+  const canSearch = (params: PeopleSearchQuery) => {
+    const { includeStudent, includeFacStaff, includeAlumni, ...criteria } = params;
```
amos-cha, line 177: why was this changed?

antoniavonto: because we are calling the search function with the params taken from the URL before the setSearchParams gets to update. So the search function will call canSearch, but this time it should check if it can search with the params of the URL instead of the searchParams that was not updated yet

EjPlatzer:
Just checking, did we decide to do this instead of retaining search results because retaining search results was too difficult? I have a PR, #1992, that successfully retains search results.

antoniavonto:
> Just checking, did we decide to do this instead of retaining search results because retaining search results was too difficult? I have a PR, #1992, that successfully retains search results.

This was to substitute retain search params, but if there is a solution already I can close this branch and the issue in the board.

antoniavonto:
After talking to @russtuck  and @jsenning , we decided to merge this for now and later on think about an improvement for efficiency. Instead of researching every time to actually retain the search results.

PR #1992, opened 2023-07-10
Save People Search results in history and load on navigation

EjPlatzer:
I have implemented the behavior to save people search results into the browser history, and then load them from the history when the user navigates to a history entry that has saved results.


jaeminlee0515:
Hey @EjPlatzer! I wanted to see how the people search results was saved so I went to your branch and tested it out. It works perfectly for when searching several people and navigating the back button from within people search. The only problem I found is that when you click on a profile and navigate back, it doesn't save the search results. It may be more clear from the below:
[Screencast from 07-11-2023 03:58:53 PM.webm](https://github.com/gordon-cs/gordon-360-ui/assets/112501914/84b02320-078b-447c-817d-25ab0d940700)

Antonia's method allows for both searching several people and navigating the back button from within people search as well as when you click on a profile and navigate back, it shows the results by searching it. Here is the visual below:
[Screencast from 07-11-2023 04:35:27 PM.webm](https://github.com/gordon-cs/gordon-360-ui/assets/112501914/c33adbd5-de15-4e0b-9900-88449b7bc0ad)

Do you know the way in which you would be able to implement saving the search result when you click on a profile and navigate back using your method? Any other insights regarding this?

PR #2001, opened 2023-07-11
S23 change office hours route

mla04762:
We decided to move the Office hours information from the Office info card to the schedule panel. We had to modify the backend code and the ui to send the update to the new database table instead of the schedulecontroller.


jsenning (commented):
This seems good to me but a few thoughts for revisions before we consider this done:
1. It would be helpful to find a way to (by default) display office hours if they are set.  If no office hours are defined then we need a obvious way to set them, but maybe not a big "office hours" heading just to say "None"
2. Most staff don't have a "Course Schedule" so having a heading for that will seem very strange/awkward to them.
3. This may be okay, but it is possible to include HTML in the office hours text field to create active links that are unvetted.

src/components/Profile/components/SchedulePanel/index.jsx
```diff
@@ -76,28 +76,19 @@ const GordonSchedulePanel = (props) => {
...
-          scheduleControlInfo.Description
-            ? // We decided to leave the regex code for now because the data stored in the database
-              // before changing the api is still in the regex form.
-              scheduleControlInfo.Description.replace(new RegExp('SlSh', 'g'), '/')
-                .replace(new RegExp('CoLn', 'g'), ':')
-                .replace(new RegExp('dOT', 'g'), '.')
-            : '',
-        );
-        setModifiedTimeStamp(scheduleControlInfo.ModifiedTimeStamp);
+      if (!profileInfo.PersonType?.includes('stu')) {
```
russtuck: Should be includes 'fac' (since  a faculty is also a student if auding a course).  Note that fac appears to actually be all regular employees (facstaff).

src/components/Profile/components/SchedulePanel/index.jsx
```diff
@@ -162,36 +152,35 @@ const GordonSchedulePanel = (props) => {
...
-              <CardHeader className={styles.accordionHeader} title={'Course Schedule'} />
+              <CardHeader
+                className={styles.accordionHeader}
+                title={description === '' ? 'Schedule' : 'Office Hours & Course Schedule'}
+              />
             </AccordionSummary>
             <AccordionDetails>
               <Grid container direction="row" justifyContent="center" align="left" spacing={4}>
-                {props.isOnline && (
+                {props.isOnline && !profile.PersonType?.includes('stu') && (
```
russtuck: includes 'fac', as above

russtuck (changes requested):
Notes: Office hours now renders html, just like go.gordon.edu.  Since this is only settable by employees, and is never anonymous, and it's quite useful, this seems OK.

Almost there.  Just a couple comments.

russtuck (approved):
(You submitted the changes faster than I submitted the review :-)

PR #2004, opened 2023-07-11
Color Mode Switching and Listening Logic + Appearance Settings Dialog

Centerville1:
Implements the event listeners and browser storage for theme settings.  Adds a bunch of logic for setting the theme color in app.jsx, and handling listeners for system and storage updates.

At this point I'm not totally confident in how robust the logic is.  I encountered very occasional times where it wouldn't work when messing with browser cache and devTools, but I'd value more testing and feedback to help improve this.

This PR also implements a new dialog box, linked from the navMenus, for appearance settings.  (WIP as of 7-11-2023)

Appearance settings button (Would appreciate feedback on button positioning in list):
AppearanceButtonInMenu

Appearance settings menu:
AppearanceSettings


src/app.jsx
```diff
@@ -12,6 +13,55 @@ import routes from './routes';
...
+    setupColorSchemeListeners(localStorage.getItem('colorMode'));
+  };
+
+  useEffect(() => {
+    //Listen for changes in local storage
+    window.addEventListener('storage', onModeUpdate);
+    setupColorSchemeListeners(colorMode);
+    return () => {
+      //Clean up listeners
+      window.matchMedia('(prefers-color-scheme: dark)').removeEventListener('change', () => {});
```
EjPlatzer, line 55: This will not work, you have to pass the same function to `removeEventListener` that you passed to `addEventListener`, otherwise the event listener will not be removed.

Note that it can't be a new function with the same code, it has to be actually the same object in memory, meaning you need to save the function to a variable and use that same variable in both add and remove.

src/app.jsx
```diff
@@ -12,6 +13,55 @@ import routes from './routes';
 import analytics from './services/analytics';
 
 const App = () => {
+  // Code for setting up color schemes
```
EjPlatzer, line 16: This code is, as you feared, cluttering App up. Please extract it a custom hook.

src/app.jsx
```diff
@@ -12,6 +13,55 @@ import routes from './routes';
...
+  // mode listener code from
+  //https://dev.to/gustavogomez/manage-dark-mode-in-reactjs-with-listener-to-detect-dark-mode-in-the-browser-3120
+
+  let colorMode = localStorage.getItem('colorMode') ?? 'system';
+
+  const onSelectMode = (prefersDark) => {
+    prefersDark === 'dark' ? setMode('dark') : setMode('light');
+  };
+
+  const setupColorSchemeListeners = (colorScheme) => {
```
EjPlatzer, line 28: You take `colorScheme` as a param here, but within the function you never use it, using the component-scoped variable `colorMode` (which is never updated) instead.

src/app.jsx
```diff
@@ -12,6 +13,55 @@ import routes from './routes';
...
+
+  const { mode, setMode } = useColorScheme();
+  // mode listener code from
+  //https://dev.to/gustavogomez/manage-dark-mode-in-reactjs-with-listener-to-detect-dark-mode-in-the-browser-3120
+
+  let colorMode = localStorage.getItem('colorMode') ?? 'system';
+
+  const onSelectMode = (prefersDark) => {
+    prefersDark === 'dark' ? setMode('dark') : setMode('light');
+  };
```
EjPlatzer, line 26: The way this is currently used, it's unnecessary, you can just replace it with `setMode`.

src/app.jsx
```diff
@@ -12,6 +13,55 @@ import routes from './routes';
...
+
+  //Callback, when the colorScheme changes in local storage
+  const onModeUpdate = () => {
+    setupColorSchemeListeners(localStorage.getItem('colorMode'));
+  };
+
+  useEffect(() => {
+    //Listen for changes in local storage
+    window.addEventListener('storage', onModeUpdate);
+    setupColorSchemeListeners(colorMode);
```
EjPlatzer, line 52: You aren't fully utilizing events here - you wait for them to trigger, but you don't actually read the data they contain.

```suggestion
  useEffect(() => {
    //Callback, when the colorScheme changes in local storage
    const onModeUpdate = (storageEvent) => {
      if (storageEvent.key === 'colorMode') {
        setupColorSchemeListeners(storageEvent.newValue);
      }
    };

    //Listen for changes in local storage
    window.addEventListener('storage', onModeUpdate);
    setupColorSchemeListeners(localStorage.getItem('colorMode'));
```

src/app.jsx
```diff
@@ -12,6 +13,55 @@ import routes from './routes';
 import analytics from './services/analytics';
 
 const App = () => {
+  // Code for setting up color schemes
+
+  const { mode, setMode } = useColorScheme();
+  // mode listener code from
+  //https://dev.to/gustavogomez/manage-dark-mode-in-reactjs-with-listener-to-detect-dark-mode-in-the-browser-3120
+
+  let colorMode = localStorage.getItem('colorMode') ?? 'system';
```
EjPlatzer, line 22: The 'colorMode' local storage key should be a constant, probably defined in a service, that can be imported where needed.

src/app.jsx
EjPlatzer: After getting deep into reviewing this code, I realized that - I believe - most of it is unnecessary.

We don't need to watch for changes to the user's color scheme preference at the app level. If they change it, it will be because they clicked the toggle, so we can handle the change (AKA call `setMode` and update `localStorage`) within the toggle's callback.

So the only thing that we need to do when the app first renders is read the saved preference if it exists or else set it from the system preference. That only needs to happen once when the App mounts, no event listeners needed.

EjPlatzer (changes requested):

src/app.jsx
Centerville1: Don't we want the color to update whenever the system setting changes if the user has selected system, without needing a reload?  That's why the listener is being used...

EjPlatzer: Ah yes, that's what I was forgetting, thank you! Never mind then, we do need most of this.

EjPlatzer: I won't be here the rest of the week, so I'm just gonna to leave here how I would write the hook, in semi-pseudocode.

We can actually dramatically simplify things if we just read the user's preference when the system color scheme changes, and only update if they are using the system theme. That way we don't need to listen for their preference, we just query at the time that it's relevant. And since the system color scheme doesn't change very often, this is probably actually the best way to do it.

```javascript
const useWatchUsersColorScheme = () => {
  const { setMode } = useColorScheme();

  useEffect(() => {
    const prefersDarkQuery = window.matchMedia('(prefers-color-scheme: dark)');
  
    const onSystemColorSchemeChange = (e) => {
      const colorSchemeSource = localStorage.getItem('colorMode') ?? 'system'
  
      if (colorSchemeSource === 'system') {
        setMode(e.matches ? 'dark' : 'light');
      }
    }
    
    prefersDarkQuery.addEventListener('change', onSystemColorSchemeChange);

    return () => {
      prefersDarkQuery.removeEventListener('change', onSystemColorSchemeChange);
    };
  }, [setMode]);
}
```

Centerville1: Ah, gotcha, well, I'll see what I can do about simplifying it in the morning, thanks for all the help!

src/app.jsx
Centerville1: Fixed by the custom hook

src/app.jsx
Centerville1: Outdated by the custom hook

src/app.jsx
Centerville1: Outdated by the custom hook

src/app.jsx
Centerville1: I was able to simplify it a bunch, and extract it into a custom hook.  Along with a couple other use effects to ensure the color is correct when the app launches it seems to all work well, and I haven't encountered any issues with it from my testing.

src/app.jsx
Centerville1: I ended up storing the key in theme.ts, since that's where the other palette information and settings are stored, rather than making a new service to only store the one constant.  If there is somewhere that makes more sense to store this, let me know.

src/components/PaletteSwitcherDialog/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  // sets the site color scheme based on the setting string passed to it, defaults to light
+  const SetModeFromSetting = (scheme: string) => {
+    setMode(
+      scheme === 'system'
+        ? window.matchMedia('(prefers-color-scheme: dark)').matches
+          ? 'dark'
+          : 'light'
+        : scheme === 'dark'
+        ? 'dark'
+        : 'light',
```
amos-cha, line 73: personal opinion. This is a spot where using ternary operators makes the code significantly more convoluted. having a switch case  or `if/else`  on `scheme` would be better

src/components/PaletteSwitcherDialog/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+      title="Appearance Settings"
+      buttonClicked={handleClose}
+      buttonName="Close"
+    >
+      <FormControl>
+        <FormLabel id="palette-mode-group-label">Color Mode</FormLabel>
+        <RadioGroup
+          name="palette-mode-group"
+          aria-labelledby="palette-mode-group-label"
+          row={true}
```
amos-cha, line 101:
```suggestion
          row
```

src/theme.ts
```diff
@@ -111,6 +111,9 @@ export const windowBreakWidths = {
   breakXL: 1536,
 } as const;
 
+// key for caching user color preference in localStorage
+export const storageColorPreferenceKey = 'colorMode';
```
amos-cha, line 115: camel case is typically used for mutable variables. Which, I'm aware, is not consistently followed throughout our repo. Regardless, with a global constant I would prefer different naming such as `STORAGE_COLOR_PREFERENCE_KEY` or `COLOR_PREFERENCE_KEY`

amos-cha (commented):
logic looks sound. However, it's a preference thing for me at this point

amos-cha (dismissed):
LGTM, however I can only judge simple logic. I think Evan should be the one to handle architecture.

EjPlatzer (approved):

PR #2008, opened 2023-07-12
S23 new logo on header

jhammond4:
Improving with fixes to the last pr that was preemptively merged, this adds the new logo as a button on the main header


Centerville1 (dismissed):
This PR is ready for review.

jhammond4:
Just leaving this here for anyone who sees, we're still waiting on new Gordon 360 logos from design so this should not be merged yet.

src/components/Header/index.jsx
```diff
@@ -137,6 +140,16 @@ const GordonHeader = ({ onDrawerToggle }) => {
     </Button>
   );
 
+  function logoSize() {
```
russtuck, line 143: "logoSize()" sounds like it returns the size of the logo, instead of returning the right sized logo.  How about renaming it something like logoSizedForHeader() or headerLogo().

russtuck (changes requested):

russtuck:
Only one very small change.  Looks good!

russtuck (dismissed):
Thanks.  LGTM

russtuck (approved):
New logo looks great.

PR #2012, opened 2023-07-13
Recim theme palette update

ahwnsals1234:
Working on the Dark mode project color styling change for Rec-im now.


src/theme.ts
```diff
@@ -239,7 +241,8 @@ export const theme360 = extendTheme({
           900: '#005195',
         },
         error: {
-          main: OldSchemeRed,
+          main: TestToolContrast,
```
Centerville1, line 244: This should not be changed, we don't want to use the test tool colors.

src/theme.ts
```diff
@@ -222,6 +222,8 @@ export const theme360 = extendTheme({
           700: '#1C75B3',
           800: '#1365A2',
           900: '#014883',
+          A100: '#013865',
```
Centerville1, line 225: What are these two colors for?

src/views/InvolvementProfile/components/Membership/components/AdminCard/components/RequestsReceived/RequestsReceived.module.scss
```diff
@@ -1,5 +1,5 @@
 .redButton {
-  color: var(--mui-palette-error-main);
+  color: var(--mui-palette-error-dark);
```
Centerville1, line 2: If you want this to be red, can you use ```color="red"``` on the button instead, especially if you make OldSchemeRed error-main again

Centerville1 (changes requested):
Code looks pretty good on a basic inspection, I'll look through it in more detail, but a couple initial things to change for now.

src/theme.ts
ahwnsals1234: Then can we then use the color with a different name? or is this just a color we shouldn't use?

src/theme.ts
ahwnsals1234: I removed both A100 and A200 because A100 was being used in confirmationRows which isn't being used anymore and A200  was a color that can be replaced with primary blue

amos-cha (approved):

PR #2016, opened 2023-07-13
UI version now stored in single JSON (not list of JSON)

jsenning:
This PR fixes one problem and also cleans up a bit of code from my last PR for this.
1. The API version information had some overlapping fields on narrow displays (e.g. mobile devices); this has been fixed.
2. The UI version information was stored as single JSON element in a list - which was not needed.  Changing this meant we could remove a map() and remove a <Fragment>.


save-ui-version.sh
```diff
@@ -1,16 +1,16 @@
...
-[
-  {
-    "commit": "$commit_sha",
-    "date": "$commit_date"
-  }
-]
+{
+  "commit": "$commit_sha",
+  "date": "$commit_date"
+}
```
amos-cha, line 15: I don't think I'm the correct person to ensure that the git commands are correct. Perhaps @russtuck or @EjPlatzer can confirm.

amos-cha (approved):
I'm approving the formatting, but someone else should look over the shell script

save-ui-version.sh
jsenning: The script should work fine (it does on development computers) but won't know for sure the path the version.json file is correct until script is run during deployment.

PR #2036, opened 2023-07-17
[Rec-IM] Update admin permissions

amos-cha:
image

Admin page now has options which allow for specific status'
As of July 17th 2023 -- only the admin assignment option will work

Minor fixes:
Re-adjusted the report icon so that it does not wrap on mobile


src/views/RecIM/views/Admin/index.jsx
```diff
@@ -82,6 +83,7 @@ const Admin = () => {
   const [selectedDateOut, setSelectedDateOut] = useState(null);
   const [openRecimReportBox, setOpenRecimReportBox] = useState();
   const [recimReport, setRecimReport] = useState();
+  const isSuperAdmin = useAuthGroups(AuthGroup.RecIMAdmin) || true;
```
amos-cha, line 86:
```suggestion
  const isSuperAdmin = useAuthGroups(AuthGroup.RecIMAdmin);
```

PR #2041, opened 2023-07-17
Fix alumni search no image case and no email visibility

Centerville1:
Removed the case for no image in identification card, we want the card to still show up even if there is no image, instead of just having a loader.
Added logic to check if a user has an email to identification card and people search result list


jsenning (approved):
LGTM

PR #2044, opened 2023-07-17
Align for words to be centered

jaeminlee0515:


jaeminlee0515:
Before change:
Screenshot 2023-07-18 at 10-41-20 Gordon 360

After change:
Screenshot 2023-07-18 at 10-41-58 Gordon 360

We decided to center the alignment of the "No involvement to display" message and also to remove the list for the note that only has one item or message.

src/components/Profile/components/MembershipsList/index.jsx
```diff
@@ -66,14 +70,12 @@ const MembershipsList = ({ username, myProf, createSnackbar }) => {
...
     <div align="left" className={styles.memberships_card_note}>
-      <Typography>NOTE:</Typography>
-      <ul>
-        <li>
-          <Typography>
-            Shaded areas are visible only to you and other members of the same club session.
-          </Typography>
-        </li>
-      </ul>
+      <Typography>
```
jsenning, line 73: Since there is no styling or links inside this note I think we can dispense with the {' '} and the <span> ... </span> tags.

jsenning (changes requested):

jsenning (approved):
Looks good!

PR #2049, opened 2023-07-18
Quick fix for timesheet shift tooltip colors

Centerville1:


jsenning (approved):
Finally - we've been waiting ages for this :)

PR #2048, opened 2023-07-18
Scss changes for note design 

antoniavonto:
these changes were already pushed previously, not sure why shows the old version in develop
old pull request: https://github.com/gordon-cs/gordon-360-ui/pull/1998
before:
image

now: 
image

PR #2047, opened 2023-07-18
Minor fix of rewording login page title

ahwnsals1234:
Small fix Dr.Senning pointed out. Just had to put space between "Gordon" and "360" in the login page.

PR #2040, opened 2023-07-17
consistency in style and code in Change Office Hours

antoniavonto:
before:
image

after (still multiple line to scroll, but wider):
image

now similar to other edit features:
image

PR #2037, opened 2023-07-17
schedule dialog improvement

andrew-wzj:

PR #2050, opened 2023-07-18
Release 07182023

russtuck:
This Release will merge the following changes into production:
* #1923
* #2010 
* #2011
* #1956
* #1990
* #2016 
* #2018 
* #1960 
* #1981 
* #2009 
* #2008 
* #1955 
* #2024 
* #2032 
* #2030 
* #2004 
* #2033 
* #2039 
* #1993 
* #2026 
* #2041 
* #2034
* #2040
* #2037
* #2047
* #2048


index.css
```diff
@@ -0,0 +1,84 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+}
```
EjPlatzer, line 4: This `index.css` stylesheet is not a CSS module (and I don't think it could be one), so these classes are all global. Also, this stylesheet is outside the regular hierarchy of the code (e.g. in the `src` folder), meaning it will probably be forgotten by future devs. So, we should aim to only include in this file styles that are specific to the non-React code and that will not interact with the React stlying at all.

That means that:
1. Each class needs to be very specifically named (e.g. `splash-screen-icon`)
2. We should not use the `*` selector, because it selects every element on the entire page, which it should not do.

src/components/EventList/components/EventItem/index.jsx
```diff
@@ -47,6 +53,14 @@ const EventItem = ({ event }) => {
               timeZone="currentBrowser"
               description={event.Description}
               onClick={() => setExpanded((e) => !e)}
+              lightMode="system" // Set to system @TODO use localstorage setting to update
```
EjPlatzer, line 56: This will at least be accurate to the user's preference when the component renders, although it still won't cause the component to re-render.

```suggestion
              lightMode={localStorage.getItem('mui-mode')}
```

src/components/Header/index.jsx
```diff
@@ -150,15 +163,15 @@ const GordonHeader = ({ onDrawerToggle }) => {
...
-            disableRipple={true}
+          <Link
             to="/"
-            onChange={(event, value) => setTabIndex(value)}
-          />
+            component={ForwardNavLink}
+            value={tabIndex}
+            onClick={(event, value) => setTabIndex(value)}
+          >
+            <img src={logoSizedForHeader()}></img>
```
EjPlatzer, line 172: We should use [the `picture` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) instead, which is the browser-supported way to load the best version of an image for the user's device.

EjPlatzer (changes requested):

src/components/EventList/components/EventItem/index.jsx
Centerville1: Yeah, although the component re-renders a lot currently, since it's inside an event or a class on the schedule panel, so if the user closes the event/class and reopens it the button will switch to the correct color.  So it's possible the component will be in the wrong setting immediately after the setting changes, but only while the user stays focused on the component it's in.
I've implemented it and tested it.  In the relatively rare case that a user changes their system setting with one of the add-to-calendar buttons open, the buttons will be in the wrong mode until they get rendered again, which I think is fine, since simple navigation or a page reload will fix it quickly.  Plus, even in the wrong mode the buttons are still perfectly functional, and don't look too terrible.

russtuck:
Adds these requested fixes:
https://github.com/gordon-cs/gordon-360-ui/pull/2061
https://github.com/gordon-cs/gordon-360-ui/pull/2059

And these small fixes:
https://github.com/gordon-cs/gordon-360-ui/pull/2049
https://github.com/gordon-cs/gordon-360-ui/pull/2044

And these mostly-content updates:
https://github.com/gordon-cs/gordon-360-ui/pull/2051
https://github.com/gordon-cs/gordon-360-ui/pull/2054

I think this is ready to push to master first thing in the morning.

PR #2058, opened 2023-07-18
S23 gordon 2pi

jsenning:
This changes the Gordon 360 logo in the header to be Gordon 2&pi; on March 14.

Normally the header looks like:
Screenshot from 2023-07-18 16-21-29

On March 14 the header becomes:
Screenshot from 2023-07-18 16-21-45


Centerville1 (commented):
lgtm!  I maybe shouldn't be the only reviewer, but this looks very similar to what's been done before.

russtuck (approved):

PR #2060, opened 2023-07-18
[Rec-IM] auto sched parameters

amos-cha:
Old API feature that was requested by the customer but was never implemented in the UI.
image

Adds use to `Round Robin Match Capacity` and `Ladder Matches`

### DIFF
--- a/src/services/recim/series.ts
+++ b/src/services/recim/series.ts
@@ -87,7 +87,8 @@ type AutoScheduleParameters =
     }
   | {
       numberOfLadderMatches: number;
-    };
+    }
+  | null;
 
 //Series Routes
 const createSeries = async (
--- a/src/services/recim/series.ts
+++ b/src/services/recim/series.ts
@@ -81,10 +81,13 @@ export type BracketInfo = {
   IsLosers: boolean;
 };
 
-type AutoScheduleParameters = {
-  roundRobinMatchCapacity: number;
-  numberOfLadderMatches: number;
-};
+type AutoScheduleParameters =
+  | {
+      roundRobinMatchCapacity: number;
+    }
+  | {
+      numberOfLadderMatches: number;
+    };
 
 //Series Routes
 const createSeries = async (
@@ -115,7 +118,7 @@ const putSeriesSchedule = (schedule: UploadSeriesSchedule): Promise<SeriesSchedu
 const scheduleSeriesMatches = (
   seriesID: number,
   params: AutoScheduleParameters | null,
-): Promise<Match[]> => http.post(`recim/series/${seriesID}/autoschedule`, { ...params });
+): Promise<Match[]> => http.post(`recim/series/${seriesID}/autoschedule`, params);
 
 const deleteSeriesCascade = (seriesID: number): Promise<CreatedSeries> =>
   http.del(`recim/series/${seriesID}`);
--- a/src/services/recim/series.ts
+++ b/src/services/recim/series.ts
@@ -82,8 +82,8 @@ export type BracketInfo = {
 };
 
 type AutoScheduleParameters = {
-  RoundRobinMatchCapacity: number;
-  NumberOfLadderMatches: number;
+  roundRobinMatchCapacity: number;
+  numberOfLadderMatches: number;
 };
 
 //Series Routes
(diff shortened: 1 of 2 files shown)


src/views/RecIM/views/Activity/components/ScheduleList/index.jsx
```diff
@@ -99,13 +115,40 @@ const ScheduleList = ({
...
           return numTeams * 2 - 1;
         default:
           return null;
       }
     };
+
+    let parameterFields = (type, numTeams) => {
+      return (
+        <Box component="form">
+          <FormControl>
```
EjPlatzer, line 129: What are the `Box` and `FormControl` for? You aren't using any of the features of the `form` component. It seems like just the `TextField` will give you exactly the same result.

src/views/RecIM/views/Activity/components/ScheduleList/index.jsx
```diff
@@ -122,16 +165,21 @@ const ScheduleList = ({
...
+        {(series.Type === 'Round Robin' || series.Type === 'Ladder') &&
+          parameterFields(series.Type, series.TeamStanding.length)}
       </Typography>,
     );
     setOpenAutoSchedulerDisclaimer(true);
     closeMenusAndForms();
   };
 
   const handleConfirmAutoSchedule = () => {
+    let parameterLabel =
```
EjPlatzer, line 177:
```suggestion
    const parameterLabel =
```

src/services/recim/series.ts
```diff
@@ -81,6 +81,11 @@ export type BracketInfo = {
   IsLosers: boolean;
 };
 
+type AutoScheduleParameters = {
+  RoundRobinMatchCapacity: number;
+  NumberOfLadderMatches: number;
+};
```
EjPlatzer, line 87: This is not accurate to the data you're actually passing.
```suggestion
type AutoScheduleParameters = {
  RoundRobinMatchCapacity: number;
} | {
  NumberOfLadderMatches: number;
};
```

src/views/RecIM/views/Activity/components/ScheduleList/index.jsx
```diff
@@ -122,16 +165,21 @@ const ScheduleList = ({
...
+      series.Type === 'Round Robin' ? 'roundRobinMatchCapacity' : 'numberOfLadderMatches';
```
EjPlatzer, line 178: The casing of these does not match the `AutoScheduleParameters` type you've declared that they should match.

amos-cha: I forgot to save my `.ts` file whoops

src/services/recim/series.ts
amos-cha: was unaware I could do this. I just fully expected `null` to translate. But it makes sense.

EjPlatzer (dismissed):

amos-cha:
Linting bug ._.

EjPlatzer (dismissed):

src/services/recim/series.ts
```diff
@@ -107,8 +116,10 @@ const getSeriesSchedule = (seriesID: number): Promise<SeriesSchedule> =>
 const putSeriesSchedule = (schedule: UploadSeriesSchedule): Promise<SeriesSchedule> =>
   http.put(`recim/series/schedule`, schedule);
 
-const scheduleSeriesMatches = (seriesID: number): Promise<Match[]> =>
-  http.post(`recim/series/${seriesID}/autoschedule`, {});
+const scheduleSeriesMatches = (
+  seriesID: number,
+  params: AutoScheduleParameters | null,
```
EjPlatzer, line 121: Redundant since `AutoScheduleParameters` can be `null`.
```suggestion
  params: AutoScheduleParameters,
```

src/views/RecIM/views/Activity/components/ScheduleList/index.jsx
```diff
@@ -28,6 +38,8 @@ import { useWindowSize } from 'hooks';
 import { windowBreakWidths } from 'theme';
 import { deleteMatchList } from 'services/recim/match';
 import GordonLoader from 'components/Loader';
+import { Box } from '@mui/system';
+import { FormControl } from '@mui/base';
```
EjPlatzer, line 42: These are now unused, aren't they?

amos-cha: oo yes good catch

PR #2061, opened 2023-07-18
S23 Header fixes

jhammond4:
Changing the img tags in the header and modifying the header buttons in response to Evan's feedback and other people's issues.


jsenning (approved):
LGTM

PR #2066, opened 2023-07-19
Replace people search missing image default color

jsenning:
People search results either show users profile pictures (if they exist) or fills this area with a neutral gray color.  The gray color used worked well in light mode but was a little bright in dark mode.  This PR replaces the fill color with a neutral gray but with 15% opacity so it blends the background color.  The blank images look just about the same in light mode but are much darker in dark mode.


Centerville1 (approved):
lgtm!

PR #2070, opened 2023-07-19
S23 revert revert pr1980 plus add fixes

russtuck:


EjPlatzer (changes requested):

src/views/InvolvementsAll/index.jsx
```diff
@@ -18,7 +18,7 @@ import membershipService, { NonGuestParticipations } from 'services/membership';
 import sessionService from 'services/session';
 import InvolvementsGrid from './components/InvolvementsGrid';
 import Requests from './components/Requests';
-import styles from './InvolvementsAll.module.css';
+import styles from './Involvements.module.css';
```
Centerville1, line 21: If we aren't using InvolvementsAll.module.scss anymore, that file should be removed.

src/views/Home/components/NewsCard/index.jsx
```diff
@@ -27,6 +25,7 @@ const NewsCard = () => {
             </Grid>
           </Grid>
         }
+        className={'gc360_header'}
```
Centerville1:
```suggestion
        className='gc360_header'
```
Brackets aren't needed, so this should be written like this everywhere you use it

src/views/Timesheets/index.jsx
```diff
@@ -310,6 +310,7 @@ const Timesheets = (props) => {
       setClockInOut('Clock In');
       setSelectedDateIn(null);
       setSelectedDateOut(null);
+      setHoursWorkedInDecimal(0);
```
Centerville1, line 313: Was this line added intentionally

PR #2071, opened 2023-07-19
Fix broken public office info

EjPlatzer:
The public profile of any user with Office Info is currently broken in develop because the code is passing an object as a React child. The fix is to pass strings instead.

I also fixed all the linting warnings because of unused code and imports.

### DIFF
--- a/src/components/Profile/components/OfficeInfoList/index.jsx
+++ b/src/components/Profile/components/OfficeInfoList/index.jsx
@@ -1,20 +1,14 @@
-import { useState, Fragment } from 'react';
-import { Card, CardContent, CardHeader, Grid, List, Typography, IconButton } from '@mui/material';
+import { Card, CardContent, CardHeader, Grid, List, Typography } from '@mui/material';
 import ProfileInfoListItem from '../ProfileInfoListItem';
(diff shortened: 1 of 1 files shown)


amos-cha (approved):
Thanks for the fix!

src/components/Profile/components/OfficeInfoList/index.jsx
```diff
@@ -110,24 +104,9 @@ const OfficeInfoList = ({
...
-          href="https://go.gordon.edu/general/myaccount.cfm"
-          target="_blank"
-          rel="noopener noreferrer"
-          style={{ color: gordonColors.primary.blue }}
-        >
-          here
-        </a>
-      </Typography>
-    ) : null;
-
```
amos-cha, line 130: Should we be removing the update?

amos-cha: (nevermind, this was resolved by @antoniavonto )

russtuck: Antonia confirmed that this is old and unused code.

russtuck (approved):
Thanks for finding and fixing this!

PR #2072, opened 2023-07-19
Add recommended extensions and editor defaults

EjPlatzer:
I have added the minimum recommended extensions for VS Code so that users will be prompted to install our linting extensions.

This should probably be complemented by updates to CI to error on Linter warnings.


russtuck (approved):
I'm not an expert in VScode, but this looks good to me.

.vscode/settings.json
```diff
@@ -1,23 +1,11 @@
...
-    // "editor.formatOnSave": true
-  },
-  "[scss]": {
-    // "editor.formatOnSave": true
-  },
-  "extensions.ignoreRecommendations": true,
-  "files.autoSave": "off",
-  "debug.allowBreakpointsEverywhere": true,
-  "gitdoc.enabled": false
+  "editor.rulers": [100],
```
amos-cha, line 3:
```suggestion
  "editor.rulers": [80],
```

is there a specific reason we have this at 100? I just noticed this, I thought most common best practice is 80?

russtuck: I assume it's because so much of the code was done with a 100 character limit that we'll just live with that.  (80 and 100 are both fairly commonly adopted and justifiable limits.  I don't know which is more common.)  I'd personally prefer setting it to 80 and gradually adopting that as we make changes.  But if Evan thinks that's too painful, I won't disagree.  I am interested in his reasoning, especially if I've guessed wrong.

amos-cha (approved):
lgtm

PR #2075, opened 2023-07-19
[Rec-IM] match page update

amos-cha:
Match page now shows recent matches or `This team has not played any matches`. API query limits recent matches to five on 1v1 matches. None are rendered for ladder matches

image

### DIFF
--- a/src/views/RecIM/components/List/index.jsx
+++ b/src/views/RecIM/components/List/index.jsx
@@ -7,6 +7,7 @@ import {
   SurfaceListing,
   TeamListing,
   SportListing,
+  MatchHistoryListing,
 } from './Listing';
 import { useNavigate } from 'react-router-dom';
 import styles from './List.module.css';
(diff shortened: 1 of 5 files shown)


src/views/RecIM/components/List/index.jsx
```diff
@@ -196,6 +197,19 @@ const MatchList = ({ matches, activityID }) => {
   );
 };
 
+const MatchHistoryList = ({ matches, activityID }) => {
+  console.log(matches);
```
amos-cha, line 201:
```suggestion
```

src/views/RecIM/views/Match/index.jsx
```diff
@@ -281,7 +280,7 @@ const Match = () => {
         </Grid>
       </Grid>
     );
-
+    // console.log(match);
```
amos-cha, line 283:
```suggestion
```

PR #2076, opened 2023-07-19
Fix mail location in public profile

russtuck:
Without this, I was getting an error about mail location when looking at facstaff profiles.  Others don't report the same problem, so this is starting a conversation...


EjPlatzer:
This was fixed by #2071

russtuck:
Hmm.  While the problem seemed reproducible before, it isn't now.  Puzzled, but closing.

PR #2082, opened 2023-07-19
Release 07192023 b

russtuck:
Final fixes before presentation.  These PRs will go into production:
* #2066 
* #2070 (which re-adds #1980 with fixes)
* #2058 
* #2074 
* #2069


EjPlatzer (approved):
LTGM!

PR #2086, opened 2023-07-20
Header highlight fix for RecIM

ahwnsals1234:
On the header of Gordon 360, the tab for  RecIM would not highlight when user is using RecIM so it was fixed


ahwnsals1234:
All the works were done by Jay

ArabellaJi (approved):
LGTM!

PR #2093, opened 2023-07-21
Simplify Header and App styling code

EjPlatzer:
The html and CSS to style the Header and the overall app are more complicated than is necessary to achieve the desired style. I have simplified this a little bit.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,139 +1,105 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta name="apple-touch-fullscreen" content="yes">
-    <meta name="apple-mobile-web-app-title" content="Gordon 360">
-    <meta name="mobile-web-app-capable" content="yes">
-    <meta name="apple-mobile-web-app-capable" content="yes" />
-    <meta name="apple-mobile-web-app-status-bar-style" content="black" />
-    <meta name="msapplication-TileColor" content="#ffffff" />
-    <meta name="msapplication-TileImage" content="/images/ms-icon-144x144.png" />
-    <meta name="theme-color" content="#ffffff" />
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
-    <!-- Gordon blue -->
-    <meta name="theme-color" content="#014983" />
-    
-    <base href="/" />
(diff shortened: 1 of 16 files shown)


index.html
```diff
@@ -1,139 +1,105 @@
...
+  <meta name="description" content="Gordon 360 is the student app for Gordon College." />
+</head>
 
-  <body>
-    <noscript> You need to enable JavaScript to use Gordon 360. </noscript>
-    <div id="root"><body style="background-image: linear-gradient(to right, #014983 0%, 
+<body>
+  <noscript> You need to enable JavaScript to use Gordon 360. </noscript>
+  <div id="root">
+    <div style="width: 100%; height: 100%; background-image: linear-gradient(to right, #014983 0%,
```
EjPlatzer, line 102: Most "changes" to this file were just autoformatting. The only contentful changes I made were to the element inside the root div. Previously, this was a duplicate `body` tag (that was unclosed) which is invalid html. I made it a `div` instead and added `width: 100%; height: 100%;` to it's inline styles.

src/components/Header/Header.module.scss
Centerville1, line 1: This looks super good, and so much simpler than it was previously!  One thing I notice, on smaller screens there's a size where the tab text gets cut off slightly, plus there's also no room to add the recim tab to the header (idk what the plan is with that, but we had taken that into consideration for the tab spacing in the previous iteration)

Centerville1: Also, the way on small screens that the people searchbar has no spacing to the header logo looks a little strange, but definitely fixes the previous sizing issues we were having.

Centerville1 (commented):
Looks great, super helpful cleanup and will make this much easier to understand and work with for the future.  I noted a couple of minor spacing issues on certain screens, but otherwise lgtm!

Centerville1:
@EjPlatzer we accidently pushed to this branch and tried to revert the changes in a seperate branch when we accidentally brought your commits into that other branch.  We'll probably have to revert those reverted commits to get these changes in, sorry for that mess!

Centerville1:
> @EjPlatzer we accidently pushed to this branch and tried to revert the changes in a seperate branch when we accidentally brought your commits into that other branch. We'll probably have to revert those reverted commits to get these changes in, sorry for that mess!

lmk if you want me to help get those reverts made

PR #2112, opened 2023-08-22
Update Timesheets Move Notice

EjPlatzer:
Payroll has indicated that no further Timesheets submitted via 360 will be processed. All future timesheets must be submitted to Criterion, and retroactive timesheets must be negotiated with Payroll directly.

### DIFF
--- a/src/views/Timesheets/index.jsx
+++ b/src/views/Timesheets/index.jsx
@@ -613,7 +613,7 @@ const Notice = () => (
   </Container>
 );
 
-const switchOverDate = new Date('2023-08-27 00:00');
+const switchOverDate = new Date('2023-08-22 12:00');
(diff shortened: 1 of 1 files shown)


src/views/Timesheets/index.jsx
```diff
@@ -613,7 +613,7 @@ const Notice = () => (
   </Container>
 );
 
-const switchOverDate = new Date('2023-08-27 00:00');
+const switchOverDate = new Date('2023-08-22 12:00');
```
bennettforkner, line 616: This should probably come from the database in the long term, right?

EjPlatzer: Well no, in the long term we can just delete all the Timesheets code. I was just leaving it in for now in case we need to quickly put timesheets back up.

PR #2121, opened 2023-09-06
Add Alert Message for Unsigned Rec/IM Waiver

bennettforkner:
This PR will add a red alert message to the RecIM homepage when a participant must sign the waiver:
image
image

### DIFF
--- a/src/views/RecIM/views/Home/index.jsx
+++ b/src/views/RecIM/views/Home/index.jsx
@@ -335,7 +335,7 @@ const Home = () => {
   );
 
   let headerAlert = (
-    /*participant?.Status === 'Pending' &&*/ <Alert
+    participant?.Status === 'Pending' && <Alert
       severity="error"
       sx={{ mb: '1rem' }}
       action={
--- a/src/views/RecIM/views/Home/index.jsx
+++ b/src/views/RecIM/views/Home/index.jsx
@@ -1,48 +1,49 @@
-import { useNavigate } from 'react-router-dom';
-import GordonUnauthenticated from 'components/GordonUnauthenticated';
+import AddCircleRoundedIcon from '@mui/icons-material/AddCircleRounded';
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
+import SettingsIcon from '@mui/icons-material/Settings';
 import {
-  Grid,
-  Typography,
-  Card,
-  CardHeader,
-  CardContent,
-  Button,
-  Tabs,
-  Tab,
-  Menu,
-  MenuItem,
+  Accordion,
+  AccordionDetails,
+  AccordionSummary,
+  Alert,
   Badge,
   Box,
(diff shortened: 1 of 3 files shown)


src/views/RecIM/views/Home/index.jsx
```diff
@@ -333,15 +334,38 @@ const Home = () => {
...
-    return loading ? <GordonLoader /> : <GordonUnauthenticated feature={'the Rec-IM page'} />;
-  } else {
-    return (
-      <>
-        <Header>{headerContents}</Header>
-        {loading ? (
-          <GordonLoader />
-        ) : (
+  let headerAlert = (
+    /*participant?.Status === 'Pending' &&*/ <Alert
```
EjPlatzer, line 338: Looks like you commented this out for testing? This should be restored before we merge.

EjPlatzer (changes requested):

amos-cha:
I like it, but I think if we choose a warning color that would be better. The whole point is that you can hypothetically just view and watch Rec-IM as a spectator indefinitely. The error color makes it feel like you can't do anything unless you proceed.

src/views/RecIM/views/Home/index.jsx
bennettforkner: Yea noticed that :\

bennettforkner, line 338:
```suggestion
    participant?.Status === 'Pending' && <Alert
```

bennettforkner:
> I like it, but I think if we choose a warning color that would be better. The whole point is that you can hypothetically just view and watch Rec-IM as a spectator indefinitely. The error color makes it feel like you can't do anything unless you proceed.

Greg is hoping to have all students submit the waiver in this way in the long term, no matter if they are participating in Rec/IM or not. Given that, they should never see the red unless they need take some action. I considered the warning, but ended up deciding to use the alert colors to stand out as an issue.

EjPlatzer:
> I like it, but I think if we choose a warning color that would be better. The whole point is that you can hypothetically just view and watch Rec-IM as a spectator indefinitely. The error color makes it feel like you can't do anything unless you proceed.

That's a good point. @bennettforkner what do you think about using a different color (warning or possibly even info?) instead so that people don't feel pressured to sign the waiver if they aren't planning on participating in RecIM?

bennettforkner:
> > I like it, but I think if we choose a warning color that would be better. The whole point is that you can hypothetically just view and watch Rec-IM as a spectator indefinitely. The error color makes it feel like you can't do anything unless you proceed.
> 
> Greg is hoping to have all students submit the waiver in this way in the long term, no matter if they are participating in Rec/IM or not. Given that, they should never see the red unless they need take some action. I considered the warning, but ended up deciding to use the alert colors to stand out as an issue.

You said this but liked my reply. Where do you land now?

PR #2124, opened 2023-09-22
Quick Search: Use MUI Autocomplete, refactor

EjPlatzer:
The QuickSearch component was previously using the `downshift` package, which is a good 3rd-party combobox provider (although we were using the old pre-hooks implementation with poor accessibility still). However, MUI provides their own `Autocomplete` component that accomplishes the same goal and is visually consistent with our other MUI interface elements, with less code.

So I refactored the QuickSearch component to use the MUI Autocomplete component instead of the old Downshift component. In the process, I discovered that RecIM had duplicated the entire QuickSearch component with only one change (what service method is used to fetch search results), so I updated QuickSearch to accept a `searchFunction` prop (which still defaults to the `quickSearchService.search` funtion). That way I was able to delete the otherwise entirely duplicated RecIMQuickSearch component.

I also streamlined the logic in QuickSearch via `useReducer` to synchronize state updates and reduce the overhead of creating the same Regex repeatedly.


bennettforkner (dismissed):

bennettforkner:
@EjPlatzer Why did the package uninstall warrant a force push?

EjPlatzer:
> @EjPlatzer Why did the package uninstall warrant a force push?

The force push was just for merging in develop. There were conflicts in `package-lock.json`, so I deleted it and recreated it by doing a fresh install via `npm i`, thereby resolving the conflicts, so I committed and pushed the successful merge commit. Then I noticed that the changes to `package-lock.json` weren't what I expected, because it was recreated based on the commits being merged in from develop, meaning the packages that I uninstalled in this branch were still being saved in `package-lock.json`. Since that is part of the merge conflicts, I amendend the original merge commit with the fix and force pushed to replace the unamended merge commit in GitHub.

PR #2125, opened 2023-09-23
Refactor schedule: simplify data flow, streamline calendar and dialog, use typescript, remove `moment`.

EjPlatzer:
I have refactored the `Schedule` component of Profiles. I set out first to remove the unused code leftover from the previous Schedule implementation (that allowed adding custom events). In the process, I managed to simplify the dataflow and reduce the markup without affecting behavior or appearance.

I did add one feature: when choosing a default session, the schedule will now default to the most recent session if the current session has no schedule. Previously, it always defaulted to the current session, even if the profile being viewed had no schedule for that session.

I also updated the components to use TypeScript (which took very little effort after cleaning up the dataflow). Sadly, the `add-to-calendar-button` web component package does not have proper typescript support as we're currently using. They mention [two ways to do this in their documentation](https://add-to-calendar-button.com/use-with-react):
1. Use the official React wrapper package, `add-to-calendar-button-react`. This is what we should do in the future, but it requires react v18, which we can't upgrade to because some of our less quality dependencies don't support it.
2. Add a global type declaration `global.d.ts` that adds a basic type stub for the web component. This should be straightforward in theory, but in practice I spent a couple of hours trying to make it work with no results

For the time being, I have added a `@ts-ignore` comment to turn off type-checking for the lines where the web component is used, so that at least the rest of the file gets proper type-checking.

I also changed our usage of `react-big-calendar` to use `Luxon` for localization instead of `moment`. The difference is completely opaque to us since all the localization happens inside the package, but I did not observe any issues. This change is desirable because `moment` is officially deprecated and this was the very last place we use it in our code - plus, `Luxon` is a smaller package.


src/app.global.scss
```diff
@@ -63,6 +63,7 @@ a {
 .gc360_header {
   background-color: var(--mui-palette-primary-main);
   color: var(--mui-palette-primary-contrastText);
+  border-radius: 4px;
```
bennettforkner, line 66: Is this change relevant to the topic of this PR?

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+const localizer = luxonLocalizer(DateTime, { firstDayOfWeek: 1 });
+
+type Props = {
+  schedule: Schedule;
+  onSelectEvent: (event: CourseEvent) => void;
+};
+
+const GordonScheduleCalendar = ({ schedule, onSelectEvent }: Props) => {
+  const dayStart = new Date();
+  dayStart.setHours(8, 0, 0, 0);
```
bennettforkner, line 15: Are there no classes that start before 8am, labs or otherwise? I assume this is limiting the hours displayed on the calendar.

EjPlatzer: I did not implement this, so I don't know without investigating. I know that this was considered when they updated the schedule this summer.

src/app.global.scss
EjPlatzer: Yes, because I updated the Schedule Panel to use this site-wide header style instead of a component-specific copy. The site-wide style typically has (almost) this border-radius regardless, because it's typically used with the CardHeader component. That border radius wasn't present here because the actual header is an AccordionSummary, not a CardHeader.

I have now gone in and removed the defunct header style from the component, as well fixing the border radius so that it's only applied to the top corners.

bennettforkner (approved):

PR #2128, opened 2023-10-11
Remove unused pages and dependencies

EjPlatzer:
We have a few large pages that have been hidden from the User Interface but left in the code in case they needed to be returned quickly - most notably Wellness and Timesheets. Since both have become obsolete and are unlikely to be needed again, and since we can easily restore this code via `git revert` if needed, I have decided to trim this dead code from our code base. In the process, I removed the last vestiges of some of our dependencies, and noticed several others that were included in our dependency list but never used. I have removed these unused dependencies.

I also began work on removing dependencies that are currently used, but are not truly necessary. The most straight-forward (and sparsely used) of this category is `react-media`, which allows you to render a component only if a CSS media query (passed a string prop) is true. This behavior can be trivially replicated in a number of ways (including via the `useMediaQuery` hook provided by MUI), but turned out to be not needed at all in the places where it was used.

Other candidates for removal are:
- lodash
- luxon
- Prop-types
- react-device-detect
- react-imask
- react-visibility-sensor
However, these require a more in-depth replacement strategy and so likely belong in their own pull request.


bennettforkner (approved):
LGTM!

PR #2130, opened 2023-10-23
Fix jotform waiver url

bennettforkner:
CTS Ticket 166874


EjPlatzer (approved):
Ideally we would make these "buttons" actual `a` tags so the user can preview the url, but it would take more effort than it's worth at the moment, especially since this is currently broken.

PR #2142, opened 2023-11-20
Hl agreements

andrew-wzj:
Make a PR for start code review


andrew-wzj (commented):
It looks good to me, but it needs to figure out why the code change cannot be build

ArabellaJi:
This looks good, but could you try to have an agreement component and put all the code into that so that there will be less code in the HousingLottery/index.jsx?

andrew-wzj (commented):
Remove the empty comments. Small issue but causing build fail

PR #2145, opened 2023-11-26
Hl prefer box

andrew-wzj:
Update code based from comments from Dr. Tuck. id name, remove commented code.


src/views/HousingLottery/components/StudentApplication/components/ApplicantList/components/ApplicantListItem/index.jsx
ArabellaJi, line 1: You don't need to do any work on applicant box in the branch as well, right? Are you able to remove this file?

src/views/HousingLottery/components/StudentApplication/components/ApplicantList/index.jsx
ArabellaJi, line 1: This file also not needed. Can you remove it?

src/views/HousingLottery/index.jsx
```diff
@@ -1,7 +1,99 @@
...
+  CardContent,
+  Grid,
+} from '@mui/material';
+import housingService from 'services/housing';
+import styles from './HousingLottery.module.css';
 import PreferredHall from './studentView/PreferredHall';
 
 const HousingLottery = () => {
-  return <PreferredHall />;
-};
```
ArabellaJi, line 5: I guess you get these two lines from merging from senior-project. You shouldn't remove these lines. If you remove them, the preferred hall box will disappear. Instead, you should keep both these lines and your code for prefer box. It seems like you also already deleted the files related to PreferredHall component. I would recommend you merge from senior-project again to get those files. Does that make sense?

src/views/HousingLottery/index.jsx
```diff
@@ -1,7 +1,99 @@
...
+              <Button variant="contained" onClick={handleClick}>
+                Submit
+              </Button>
+            </Grid>
+          </Grid>
+        </Card>
+      </Grid>
+    </Grid>
+  );
+  };
```
ArabellaJi, line 97: This looks nice! However, can you put these codes into another file, such as Preference/index.jsx, and only have <Preference /> in this file? We will have more and more code in this main index.jsx file, so we shouldn't put all codes in here.

ArabellaJi (dismissed):

src/views/HousingLottery/index.jsx
andrew-wzj: sure thing. I should place these in the other file instead piling in the index.jsx file

src/views/HousingLottery/HousingLottery.jsx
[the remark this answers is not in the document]
andrew-wzj: I removed the HouisngLottery.jsx and combine its content with current index.jsx

PR #2148, opened 2023-12-02
store preference as list

andrew-wzj:
Make the preference as list


andrew-wzj:
Could you also know the problem of submit button of preference box @ArabellaJi

PR #2151, opened 2023-12-03
Added Instructions Card

JayBae9903:
I added the instructions card.

### DIFF
--- a/src/views/HousingLottery/HousingLottery.module.scss
+++ b/src/views/HousingLottery/HousingLottery.module.scss
@@ -1,6 +1,5 @@
 .submit_preference_button {
-  position: 'absolute';
-  margin: 'auto';
+  margin: auto;
   top: 15px;
   bottom: 0;
   left: 45em;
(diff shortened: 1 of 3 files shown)


src/views/HousingLottery/HousingLottery.module.scss
```diff
@@ -1,6 +1,6 @@
 .submit_preference_button {
-  position: "absolute";
-  margin: "auto";
```
ahwnsals1234, line 3: I'm pretty sure that you do not need the single quotes around absolute and auto.

src/views/HousingLottery/HousingLottery.module.scss
```diff
@@ -10,8 +10,8 @@
 }
 
 .add_applicant_button {
```
ahwnsals1234, line 12: same here

ahwnsals1234 (commented):
There are couple of parts that I left comment on so you can make changes, and can you also change the name of the branch to 'hl-instructions' instead of 'instructions' so we can distinctly tell that this branch is for our senior project?

ahwnsals1234:
Also just a reminder that you have to resolve conflicts as well before merging to senior-project

PR #2153, opened 2023-12-03
Hl connect front back

ArabellaJi:
send the application data from frontend to backend and from backend to frontend


ArabellaJi:
hl-connect-front-back should push the senior-project instead of develop, so I will close this PR.

PR #2158, opened 2023-12-04
Hl sort user

ahwnsals1234:
Added a feature to distinguish if the user is logged in and if the logged in user is a student or an admin.


src/views/HousingLottery/index.jsx
andrew-wzj, line 1: It seems related files from ApartApp. And starting having the studentView and adminView which is good.
If you're completed this PR ready to merge, we could have a final review later

PR #2165, opened 2024-01-18
HotFix: Fix RecIM Username Search

bennettforkner:
I am not exactly sure why the username is coming up as null here, but handling the null case fixes the issue for the hotfix's sake.


EjPlatzer (approved):
This hotfix looks fine.

PR #2169, opened 2024-01-22
Upgrade dependencies

EjPlatzer:
Dependabot inspired me to do another round of upgrading our dependencies. Lots of inconsequential minor/patch upgrades. I also completed most of our pending major version upgrades, with the exception of react itself (we still have a few dependencies that don't support react v18), and chart-js (this is major breaking change that merits it's own branch).

I reviewed the release notes for all minor and major versions and tested each upgrade. I also did a tiny bit of updating and cleanup where it was appropriate and not significant.


bennettforkner (commented):
It looks like you have some browser compatibility concerns in the lint and build step here. Is that something you are concerned with?

bennettforkner (approved):
LGTM

PR #2172, opened 2024-02-04
set and get due date

ArabellaJi:
Update and get the due date of the housing applications

### DIFF
--- a/src/views/HousingLottery/studentView/index.jsx
+++ b/src/views/HousingLottery/studentView/index.jsx
@@ -47,7 +47,7 @@ const StudentView = () => {
   const handleClick = async () => {
     try {
       let application_id = nanoid(8),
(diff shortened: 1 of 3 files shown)


src/views/HousingLottery/studentView/index.jsx
```diff
@@ -41,7 +46,13 @@ const StudentView = () => {
 
   const handleClick = async () => {
     try {
-      console.log(application_id);
+      let application_id = nanoid(8),
+        timeTarget = new Date(dueDate + ' 11:24:00 PM').getTime(),
```
ahwnsals1234, line 50: Is there a specific reason the time '11:24:00 PM' has been linked to timeTarget?

ahwnsals1234 (commented):
I did have one question, but aside from that section it looks great to me

PR #2180, opened 2024-02-19
Fixed bug to be able to clear date input

ahwnsals1234:
It was a very simple bug that did not allow us to completely remove the input field text and always left a '0' in there. Now fully fixed.


EjPlatzer:
Closing since housing lottery was never deployed.

PR #2181, opened 2024-02-19
Applicants'emails now has combined in one dropdown button for each ap…

andrew-wzj:
…plication


EjPlatzer:
Closing since housing lottery was never deployed.

PR #2183, opened 2024-02-22
Remove unused admin service

EjPlatzer:
The admin service is unused because site admins are managed as a security group, not in the database. This removes the admin service that had no references in the project.


bennettforkner (approved):
LGTM

PR #2186, opened 2024-02-23
Send code. not description, updating office location

EjPlatzer:
Depends on gordon-cs/gordon-360-api#1023.

Fix the UpdateOfficeLocation component to send building code instead of building description.


bennettforkner (approved):
LGTM

PR #2188, opened 2024-02-26
Better UI for the React Table

ahwnsals1234:
Added UI changes in the Admin react table. Fixed header and added vertical scroll so it is more convenient for users to read data. Every other columns also shows in different colors to make the data more easily distinguishable for the user.


andrew-wzj (approved):
LGTM

PR #2198, opened 2024-03-15
Removing Selected Halls from Preferred Hall Options

JayBae9903:
A feature to dynamically remove halls from the selection list once chosen to present only the remaining options for subsequent selections.


andrew-wzj (approved):
It works correctly on my localhost, and the code looks fine. Thanks for your work

PR #2200, opened 2024-03-18
remove the JSON array button and console log, making the column name …

andrew-wzj:
…as the bold style


andrew-wzj:
Also the unnecessary Instruction part from studentView

andrew-wzj (commented):
self-reviewed

PR #2219, opened 2024-05-20
S24 fixing typos

dli505:
**Changes fixed in PR:**
- Fixed typos
- Fixed padding issue
Before:
image
After:
image

### DIFF
--- a/src/views/IDUploader/index.jsx
+++ b/src/views/IDUploader/index.jsx
@@ -92,7 +92,7 @@ const IDUploader = () => {
             titleTypographyProps={{ align: 'center' }}
             className={styles.ID_card_header}
           />
-          <CardContent>
+          <CardContent sx={{ padding: 5 }}>
             <List
               component="ol"
               sx={{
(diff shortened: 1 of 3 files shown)


src/views/IDUploader/index.jsx
```diff
@@ -92,7 +92,7 @@ const IDUploader = () => {
             titleTypographyProps={{ align: 'center' }}
             className={styles.ID_card_header}
           />
-          <CardContent>
+          <CardContent sx={{ padding: 5 }}>
```
amos-cha, line 95: Probably best practice to put `css` adjustments within a .css file. However, this is a very isolated change so I think this is more of a preference for @EjPlatzer.

amos-cha (approved):
Besides the comment (probably just wait for Evan to respond), LGTM

src/views/IDUploader/index.jsx
EjPlatzer: I think it's a matter of opinion which way is better in this case. I'm fine with merging this as is.

jsenning (commented):
LGTM -- but, were you going to add graduation years for the Summer 2023 team?

amos-cha:
> LGTM -- but, were you going to add graduation years for the Summer 2023 team?

I think that should just be in a different PR. No need to combine too many issues into a PR.

PR #2220, opened 2024-05-21
S24 closing mailbox instructions

dli505:
**Changes made in PR:**
- Re-added close button functionality to GordonDialogBox component
- Added ability to close mailbox opening instructions by clicking off

### DIFF
--- a/src/components/GordonDialogBox/index.tsx
+++ b/src/components/GordonDialogBox/index.tsx
@@ -15,8 +15,6 @@ type Props = {
   buttonClicked: (event: {}) => void;
   buttonName?: string;
   isButtonDisabled?: boolean;
-  closeButtonClicked?: (event: {}) => void;
-  closeButtonName?: string;
   cancelButtonClicked?: (event: {}) => void;
   cancelButtonName?: string;
   severity?: AlertColor;
@@ -27,8 +25,6 @@ const GordonDialogBox = ({
   buttonClicked,
   buttonName,
   isButtonDisabled = false,
-  closeButtonClicked,
-  closeButtonName,
   cancelButtonClicked,
   cancelButtonName,
   severity,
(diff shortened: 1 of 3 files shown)


src/components/GordonDialogBox/index.tsx
```diff
@@ -69,6 +69,11 @@ const GordonDialogBox = ({
             {cancelButtonName ?? 'Cancel'}
           </Button>
         )}
+        {closeButtonClicked && (
+          <Button variant="contained" color="neutral" onClick={closeButtonClicked}>
+            {closeButtonName ?? 'Close'}
+          </Button>
+        )}
```
EjPlatzer, line 76: I'm having trouble imagining a scenario where a dialog should have both a Cancel button and a Close button. Is there some benefit I'm not thinking of in allowing both of these to exist?

amos-cha (approved):
lgtm! good fix

src/components/Profile/components/PersonalInfoList/index.jsx
```diff
@@ -442,6 +442,7 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }) => {
                 <GordonDialogBox
                   open={isJoinDialogOpen}
                   title={`Mailbox Instructions`}
+                  onClose={() => setIsJoinDialogOpen(false)}
                   closeButtonClicked={() => setIsJoinDialogOpen(false)}
```
EjPlatzer, line 446: Could the mailbox instructions dialog use the `cancelButtonClicked` instead, and just pass a `cancelButtonName` of `"Close"`?

```suggestion
                  onClose={() => setIsJoinDialogOpen(false)}
                  cancelButtonClicked={() => setIsJoinDialogOpen(false)}
                  cancelButtonName="Close"
```

EjPlatzer (changes requested):

src/components/GordonDialogBox/index.tsx
dli505: Looking through the rest of 360 it seems that almost every instance of GordonDialogBox was a form that requested user input, so the Cancel button was there along with a Submit button just in case the user changed their mind. The mailbox instructions is not a form, however, so this would probably be the only place right now where the user would use a Close button instead of the Cancel button.

amos-cha: ^agreed with above, I didn't realize there was a cancel button.

src/components/Profile/components/PersonalInfoList/index.jsx
dli505: This works well as well, I'll commit the changes and remove the close button functionality from GordonDialogBox.

PR #2224, opened 2024-05-21
Fixed my profile involvements text on mobile dark mode

Luke-W-Hart:
Fixed the color of the my profile involvements text for dark mode on mobile, changing the color from "--mui-palette-neutral-light" to "--mui-palette-primary-contrastText".

### DIFF
--- a/src/components/Profile/components/MembershipsList/MembershipsList.module.scss
+++ b/src/components/Profile/components/MembershipsList/MembershipsList.module.scss
@@ -33,7 +33,7 @@
       padding-left: 0rem;
       justify-content: center;
       background-color: var(--mui-palette-primary-main);
-      color: var(--mui-palette-neutral-light);
+      color: var(--mui-palette-primary-contrastText);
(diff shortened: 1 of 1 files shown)


src/components/Profile/components/MembershipsList/MembershipsList.module.scss
```diff
@@ -33,7 +33,7 @@
       padding-left: 0rem;
       justify-content: center;
       background-color: var(--mui-palette-primary-main);
-      color: var(--mui-palette-neutral-light);
+      color: var(--mui-palette-primary-contrastText);
       border-top-left-radius: 4px;
       border-top-right-radius: 4px;
```
EjPlatzer, line 38: Since the header already has color, background-color, and border-radius set outside of this media query, the better fix would be to not overwrite those properties inside the media query. The media query should only change the properties that are specific to the smaller screen size.

Luke-W-Hart:
Just getting rid of the color changes in the media query fixed it.

PR #2228, opened 2024-05-21
Changed the quick search text color to white.

JayBae9903:


src/components/Header/components/QuickSearch/QuickSearch.module.scss
```diff
@@ -19,6 +21,7 @@
 
   & :global(.MuiInputBase-root) {
     padding: 2px 10px;
+    color: var(--mui-palette-primary-contrastText);
```
EjPlatzer, line 24: Unfortunately, this breaks the text color of QuickSearch in the RecIM Participant Search:
image

We might have to come up with a different way to style the QuickSearch so that it's visible on either background color.

EjPlatzer (changes requested):
Thanks for working on this! This is a good fix for QuickSearch as it appears in the app bar. Unfortunately, QuickSearch is also used elsewhere on the site, namely in RecIM and in ApartApps, so the styling needs to somehow work for wherever it's placed. You might need to enforce that the background color of the quick search is the Primary color.

src/components/Header/components/QuickSearch/QuickSearch.module.scss
JayBae9903: Thank you for your review! I'm wondering how to get the text color of the parent. The parent's text color usually contrasts with its background color, I think it would be better to use the parent's text color instead of using primary-contrastText color.

JayBae9903: I just found that color: inherit; get the parent's color.

PR #2229, opened 2024-05-21
Update to reload page on edit

tsion-tez:
At first I was trying to update the page and set it but that had to do with the whole profile page so I chose to force reload the page when clicking the submit button giving us the change that we want.


src/components/Profile/components/Identification/components/SocialMediaLinks/components/LinksDialog/index.jsx
```diff
@@ -60,8 +60,8 @@ const LinksDialog = ({ links, createSnackbar, onClose, setLinks }) => {
       console.log('Failed Updates: ', failedUpdates);
       createSnackbar('Failed Updating Link(s)', 'error');
     } else {
-      onClose();
       createSnackbar('Social Media Links Updated', 'success');
+      window.location.reload();
```
EjPlatzer, line 64: The issue here is that we want to reload the user's profile now that their links have been updated, right? This is one way to solve that problem that definitely works, but there is another way that might be better.

There is a custom hook, [`useUserActions`](https://github.com/gordon-cs/gordon-360-ui/blob/develop/src/hooks/useUserActions.js), which references the [`UserActionsContext`](https://github.com/gordon-cs/gordon-360-ui/blob/a727d77abcafedcfe2bd109a576a7854388f2fba/src/contexts/UserContext.tsx#L19). The hook returns an object that has an `updateProfile` function which, when called, forces the front-end to refresh the current user's profile information. So, if you call the `useUserActions` hook at the top of this component to get the `updateProfile` function, you can call `updateProfile` on this line instead. That will refresh the user's profile info (including their links), and React will re-render only the components that depend on the user's profile.

This would be a much more performant solution, since we're only reloading the profile and only re-rendering the components that depend on the profile. It would also be easier to understand, so people looking at this code in the future will know what we're trying to achieve.

If you want to look at an example that uses `useUserActions` for a similar purpose, you can look at the [`IDUPloader`](https://github.com/gordon-cs/gordon-360-ui/blob/a727d77abcafedcfe2bd109a576a7854388f2fba/src/views/IDUploader/index.jsx#L54). The `IDUploader` component needs to refresh the user's profile images when they submit an ID image, so it uses the `updateImage` function that is returned from `useUserActions`, but it's the same basic idea as using `updateProfile`.

EjPlatzer (changes requested):
Thanks for working on this! This turns out to be a difficult bug to fix, because of the way profile information is all bundled together currently. I suggested an alternative solution that might be better. Realistically, the best solution would be to separate a user's social media links from their overall profile, since the social media links are only ever referenced from the profile page, whereas the user's profile is needed in lots of places across the site. However, that would be a much bigger change (although possibly one worth considering for this summer).

src/components/Profile/components/Identification/components/SocialMediaLinks/components/LinksDialog/index.jsx
```diff
@@ -60,9 +64,11 @@ const LinksDialog = ({ links, createSnackbar, onClose, setLinks }) => {
       console.log('Failed Updates: ', failedUpdates);
       createSnackbar('Failed Updating Link(s)', 'error');
     } else {
-      onClose();
       createSnackbar('Social Media Links Updated', 'success');
+      updateProfile();
+      //  setIsSubmitted(true);
```
EjPlatzer, line 69: Can you remove this commented out code, so that it doesn't confuse future programmers?

src/components/Profile/components/Identification/components/SocialMediaLinks/components/LinksDialog/index.jsx
```diff
@@ -9,14 +9,18 @@ import {
...
 import { platforms, socialMediaInfo } from 'services/socialMedia';
 import user from 'services/user';
+import { useUserActions } from 'hooks';
 import styles from './LinksDialog.module.css';
 
 const LinksDialog = ({ links, createSnackbar, onClose, setLinks }) => {
   const [formErrors, setFormErrors] = useState([]);
   const [updatedLinks, setUpdatedLinks] = useState(links);
   const [failedUpdates, setFailedUpdates] = useState([]);
+  //const [isSubmitted, setIsSubmitted] = useState(false);
```
EjPlatzer, line 19: Can you remove this commented out code, so that it doesn't confuse future programmers?

EjPlatzer (changes requested):
Great work, thanks for fixing this! I just had one note, it seems that you left some commented out code - possibly from testing.

EjPlatzer (approved):
LGTM! Thanks for fixing this!

tsion-tez:
Thanks for the assistance.

PR #2231, opened 2024-05-21
Fix Broken Victory Promise to have Color

Luke-W-Hart:
Setting colors[index] = Colors[key] on line 26 set colors[index] = var(--mui-palette-error-main, #B53228), which is why it kept setting the colors to black.
How I fixed:
- Used RegEx to pull out the hex from var(--mui-palette-error-main, #B53228)
- Added a check to return black (#000000) if the match function returns null, else return the appropriate colors.


Luke-W-Hart:
Trying to deal with the "Object is possibly 'null'" error. Used single line if to deal with the null case but now just trying to get the error to go away.

PR #2232, opened 2024-05-22
deleted color mode text from appearance

### LINKED ISSUE #2221: Appearance Settings bug/visual fixes
Image

-Switching between System Default (if Dark) and Dark flashes text as white for whatever reason.
- [ ] Color mode text does not look good. 
- [ ] Disable option that is not System Default? i.e. If my system default is Dark, do not let me choose dark.

Dopey12:
There was a small text when choosing appearance settings, it looks better without it. 
Closes #2221

### DIFF
--- a/src/components/PaletteSwitcherDialog/index.tsx
+++ b/src/components/PaletteSwitcherDialog/index.tsx
@@ -69,7 +69,7 @@ const PaletteSwitcherDialog = ({ dialogOpen, handleClose }: Props) => {
       buttonName="Close"
     >
       <FormControl>
-        <FormLabel id="palette-mode-group-label">Color Mode</FormLabel>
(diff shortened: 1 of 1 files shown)


src/components/PaletteSwitcherDialog/index.tsx
```diff
@@ -69,7 +69,7 @@ const PaletteSwitcherDialog = ({ dialogOpen, handleClose }: Props) => {
       buttonName="Close"
     >
       <FormControl>
-        <FormLabel id="palette-mode-group-label">Color Mode</FormLabel>
+        <FormLabel id="palette-mode-group-label"></FormLabel>
```
amos-cha, line 72:
```suggestion
```
If the goal is to remove the redundancy of the text, simply remove the whole element.

PR #2234, opened 2024-05-22
Corrected variable name VITE_API_URL

jsenning:
This variable name was probably not updated correctly in the documentation when we started using VITE.

All other changes to the file seem to have been made automatically when it was committed.


jsenning (commented):
Self review - nearly all changes are the result of automatic adjustments during committing, presumably to update the markdown used.

jsenning:
I am approving and merging this in blatant disregard for protocol :)

PR #2236, opened 2024-05-23
Deleted-color-mode-text

Dopey12:
Before:
Screenshot from 2024-05-23 10-17-23
After:
Screenshot from 2024-05-23 10-16-55
It seems for now the Color Mode text doesn't fit well with the header and it seems like there isn't a use for it at the moment. I have commented out the code for now because I believe if there are more color modes implemented, then possibly there could be good use for this text in the future,
Issue#2221


amos-cha:
One thing of note is that without the `Color Mode` text, the spacing feels cramped. Possibly add top padding to the text

EjPlatzer (changes requested):
I think having the options all on one line makes it harder to parse quickly, and it doesn't seem necessary since there are only 3 options. Can you remove the `row` prop that is currently being passed to the `RadioGroup` component on line ~76? Maybe add some new screenshots to compare how that looks.

Dopey12:
On one Line: 
Screenshot from 2024-05-23 10-16-55
Without the row prop:
Screenshot from 2024-05-28 09-31-30

It seems that it possibly looks less cramped without the row prop.

amos-cha:
I think the new change is solid. However, it once again just looks cramped and small. I think some style touches with padding will help

Dopey12:
I also enjoy the way it looks without the row. If this seems like a better alternative I can make the change and push it.

EjPlatzer (approved):
LGTM!

PR #2237, opened 2024-05-23
schedule-status

Hls5235:
Stores the status of the 'Schedule' accordion so users can toggle open or closed based on preference. Also changed the default status of the schedule on the profile page to be expanded so it is easier to notice its existence and creates a higher chance of being used.

### DIFF
--- a/src/components/Profile/components/SchedulePanel/index.tsx
+++ b/src/components/Profile/components/SchedulePanel/index.tsx
@@ -28,6 +28,7 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
   const [allSchedules, setAllSchedules] = useState<Schedule[]>([]);
   const [selectedSchedule, setSelectedSchedule] = useState<Schedule | null>(null);
   const [selectedCourse, setSelectedCourse] = useState<CourseEvent | null>(null);
+  const [accordionState, setAccordionState] = useState<boolean>(true);
 
   useEffect(() => {
     setLoading(true);
@@ -48,11 +49,27 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
     });
   }, [profile.AD_Username]);
 
+  useEffect(() => {
+    const getAccordionState = localStorage.getItem('accordionState');
+    if (getAccordionState !== null) {
+      setAccordionState(getAccordionState === 'true');
+    }
+  }, []);
+
+  const toggleAccordionState = () => {
+    setAccordionState(!accordionState);
+    localStorage.setItem('accordionState', String(!accordionState));
+  };
+
   return loading ? (
     <GordonLoader />
   ) : (
     <>
-      <Accordion TransitionProps={{ unmountOnExit: true }}>
+      <Accordion
+        expanded={accordionState}
+        onChange={toggleAccordionState}
+        TransitionProps={{ unmountOnExit: true }}
+      >
         <AccordionSummary
           expandIcon={<ExpandMoreIcon className={styles.expandIcon} />}
           aria-controls="schedule"


src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -28,6 +28,7 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
   const [allSchedules, setAllSchedules] = useState<Schedule[]>([]);
   const [selectedSchedule, setSelectedSchedule] = useState<Schedule | null>(null);
   const [selectedCourse, setSelectedCourse] = useState<CourseEvent | null>(null);
+  const [accordionState, setAccordionState] = useState<boolean>(true);
```
EjPlatzer, line 31: Rather than needing a separate useEffect to get the stored state, you can just check localStorage here:
```suggestion
  const [accordionState, setAccordionState] = useState<boolean>(localStorage.getItem('accordionState') === 'true' ?? true);
```

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -48,11 +49,27 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
...
 
+  useEffect(() => {
+    const getAccordionState = localStorage.getItem('accordionState');
+    if (getAccordionState !== null) {
+      setAccordionState(getAccordionState === 'true');
+    }
+  }, []);
+
+  const toggleAccordionState = () => {
+    setAccordionState(!accordionState);
```
bennettforkner, line 60: I believe react states can have some strange synchronization issues when trying to invert variables like this. The recommended way to avoid those is to use a callback inside of the set function like so.
```suggestion
    setAccordionState(a => !a);
```
Correct, @EjPlatzer ?

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -48,11 +49,27 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
...
+    localStorage.setItem('accordionState', String(!accordionState));
```
EjPlatzer: This is actually introducing a subtle bug, due to the way React queues state updates. This is one of the major footguns of React, unfortunately. It's unlikely to matter in this case, but it's worth doing correctly to prevent user confusion. Also, it's good to learn about it now so you can avoid introducing more impactful bugs in the future.

Rather than spend several minutes writing a confusing explanation of the bug, I will point you to the React docs where they explain this common issue and how to work around it: [https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state)

The extremely short version is that passing a value to `setState` that is computed from the current state (e.g. the negation of the current `accordionState` in this case) will not always work the way you think it should. The correct way is to instead pass an _updater function_ that receives the current state as a parameter and uses that parameter to compute the new state, like so:
```suggestion
    setAccordionState((wasOpen) => !wasOpen);
    localStorage.setItem('accordionState', String(!accordionState));
```

Moreover, to ensure that localStorage stays in-sync with React's state, you should call `setItem` within the updater function:
```suggestion
    setAccordionState((wasOpen) => {
        localStorage.setItem('accordionState', String(!wasOpen));
        return !wasOpen;
    });
```

src/components/Profile/components/SchedulePanel/index.tsx
EjPlatzer: Could we come up with a clearer name for this state variable? As it is currently, I'm unsure whether `accordionState === true` means the accordion is open or closed. Maybe `isScheduleOpen` would be better?
```suggestion
  const [isScheduleOpen, setIsScheduleOpen] = useState<boolean>(localStorage.getItem('accordionState') === 'true' ?? true);
```

src/components/Profile/components/SchedulePanel/index.tsx
Hls5235: Thank you for the input, I updated the code with your suggestion.

src/components/Profile/components/SchedulePanel/index.tsx
EjPlatzer: For keys like this that need to be the same everywhere they're referenced, the best practice is to make it a constant value outside of the component, e.g. on line 25:
```typescript
    const scheduleOpenKey = 'profile.schedule.isOpen';
```
Since the key needs to be unique across all of 360 for all time, I made it a little bit more specific as well.

EjPlatzer (changes requested):
Great work! Just a couple of ways to improve the code, as well as addressing the most common React bug that everyone makes.

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -48,11 +49,27 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
...
+  useEffect(() => {
+    const getAccordionState = localStorage.getItem('accordionState');
+    if (getAccordionState !== null) {
+      setAccordionState(getAccordionState === 'true');
+    }
+  }, []);
+
+  const toggleAccordionState = () => {
+    setAccordionState((a) => !a);
+    localStorage.setItem('accordionState', String(!accordionState));
```
bennettforkner, line 61: I'm not sure how this should be set assuming that the above change works. Perhaps it should just not be inverted?
```suggestion
    localStorage.setItem('accordionState', String(accordionState));
```

EjPlatzer: This will exacerbate the bug. See my review for more background on why the bug exists and the best way to fix it.

src/components/Profile/components/SchedulePanel/index.tsx
Hls5235: I am a little confused on what this is suggesting and i'm hoping you can clarify for me, otherwise I have made changes based on the other comments.

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -45,14 +49,30 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
       setSelectedSchedule(defaultSchedule);
 
       setLoading(false);
+
+      const getIsScheduleOpen = localStorage.getItem('isScheduleOpen');
+      if (getIsScheduleOpen !== null) {
+        setIsScheduleOpen(getIsScheduleOpen === 'true');
+      }
```
EjPlatzer, line 56: You shouldn't need this code, since the default value of `isScheduleOpen` is whatever was stored in local storage.

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -23,11 +23,15 @@ type Props = {
...
 };
 
+const scheduleOpenKey = 'profile.schedule.isOpen';
 const GordonSchedulePanel = ({ profile, myProf }: Props) => {
   const [loading, setLoading] = useState(true);
   const [allSchedules, setAllSchedules] = useState<Schedule[]>([]);
   const [selectedSchedule, setSelectedSchedule] = useState<Schedule | null>(null);
   const [selectedCourse, setSelectedCourse] = useState<CourseEvent | null>(null);
+  const [isScheduleOpen, setIsScheduleOpen] = useState<boolean>(
+    localStorage.getItem('isScheduleOpen') === 'true' ?? true,
```
EjPlatzer, line 33: Rather than typing the string `'isScheduleOpen'` every time you want to get/set local storage, you should instead use the string stored in the variable `scheduleOpenKey`. That way, if you need to change the key in the future, you only have to change the string once, on line 26 where `scheduleOpenKey` is defined, rather than having to find everywhere that `'isScheduleOpen'` was typed and updating that.

```suggestion
    localStorage.getItem(scheduleOpenKey) === 'true' ?? true,
```

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -45,14 +49,30 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
...
+      const getIsScheduleOpen = localStorage.getItem('isScheduleOpen');
+      if (getIsScheduleOpen !== null) {
+        setIsScheduleOpen(getIsScheduleOpen === 'true');
+      }
     });
   }, [profile.AD_Username]);
 
+  const toggleIsScheduleOpen = () => {
+    setIsScheduleOpen((wasOpen) => {
+      localStorage.setItem('isScheduleOpen', String(!isScheduleOpen));
```
EjPlatzer, line 62: Use the constant here:
```suggestion
      localStorage.setItem(scheduleOpenKey, String(!isScheduleOpen));
```

EjPlatzer (changes requested):
Thanks for updating, this is looking good. I just left a couple comments on how you can further simplify this.

src/components/Profile/components/SchedulePanel/index.tsx
Hls5235: Thank you for the explanation, everything should be updated and simplified now, I also changed line 67 to be so the schedule starts open like suggested during the practicum.

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -23,11 +23,15 @@ type Props = {
...
 };
 
+const scheduleOpenKey = 'profile.schedule.isOpen';
 const GordonSchedulePanel = ({ profile, myProf }: Props) => {
   const [loading, setLoading] = useState(true);
   const [allSchedules, setAllSchedules] = useState<Schedule[]>([]);
   const [selectedSchedule, setSelectedSchedule] = useState<Schedule | null>(null);
   const [selectedCourse, setSelectedCourse] = useState<CourseEvent | null>(null);
+  const [isScheduleOpen, setIsScheduleOpen] = useState<boolean>(
+    localStorage.getItem('scheduleOpenKey') === 'true' ?? true,
```
EjPlatzer, line 33: This is using the **string** `'scheduleOpenKey'` as the key. What we want is to use the string `'profile.schedule.isOpen'` as the key, and the string is stored in the variable `scheduleOpenKey` (notice the lack of quotes).
```suggestion
    localStorage.getItem(scheduleOpenKey) === 'true' ?? true,
```

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -48,11 +52,22 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
     });
   }, [profile.AD_Username]);
 
+  const toggleIsScheduleOpen = () => {
+    setIsScheduleOpen((wasOpen) => {
+      localStorage.setItem('scheduleOpenKey', String(!isScheduleOpen));
```
EjPlatzer, line 57:
```suggestion
      localStorage.setItem(scheduleOpenKey, String(!isScheduleOpen));
```

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -48,11 +52,22 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
...
+    });
+  };
+
   return loading ? (
     <GordonLoader />
   ) : (
     <>
-      <Accordion TransitionProps={{ unmountOnExit: true }}>
+      <Accordion
+        expanded={!isScheduleOpen}
```
EjPlatzer, line 67: I'm confused by this. `isScheduleOpen` default to `true` when the user hasn't set anything yet, so by default this will set the `expanded` prop to false. Shouldn't it be the reverse?

EjPlatzer (changes requested):

src/components/Profile/components/SchedulePanel/index.tsx
Hls5235: For some reason when i had scheduleOpenKey in quotes, setting isScheduleOpen to false made it expanded by default.  After fixing that and removing the quotes I had to change it back. Could have also been faulty testing on my part.

EjPlatzer (approved):
Looks good to me! Great work, thanks for your patience in iterating this to the best solution!

Hls5235:
Thank you so much for helping with the cleaning up and simplifying of the code.

PR #2240, opened 2024-05-23
S24 alumni year range

dli505:
**Changes made in PR:**
- Replaced slider to choose graduation year range with two SearchFields
- Added an option to the SearchField component to choose whether or not the default menu option _All_ is shown in the menu
- Disabled toggle to choose graduation year or graduation year range when alumni checkbox is not selected

### DIFF
--- a/src/views/PeopleSearch/components/SearchFieldList/index.tsx
+++ b/src/views/PeopleSearch/components/SearchFieldList/index.tsx
@@ -2,6 +2,7 @@ import {
   Accordion,
   AccordionDetails,
   AccordionSummary,
+  Autocomplete,
   Button,
   Card,
   CardActions,
@@ -11,6 +12,7 @@ import {
   FormControlLabel,
   FormLabel,
   Grid,
+  TextField,
   Typography,
 } from '@mui/material';
 import { ExpandMore } from '@mui/icons-material';
@@ -147,6 +149,8 @@ const SearchFieldList = ({ onSearch }: Props) => {
   const [halls, setHalls] = useState<string[]>([]);
   const currentYear = new Date().getFullYear();
   const [graduationYearRange, setGraduationYearRange] = useState<number[]>([1889, currentYear]);
+  const [gradYearBegin, setGradYearBegin] = useState<number[]>([1889]);
+  const [gradYearEnd, setGradYearEnd] = useState<number[]>([currentYear]);
   // 1889 is the establish date of Gordon
   const [switchYearRange, setSwitchYearRange] = useState(true);
   const [involvements, setInvolvements] = useState<string[]>([]);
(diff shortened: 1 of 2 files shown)


src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -323,8 +326,8 @@ const SearchFieldList = ({ onSearch }: Props) => {
   const handleSwitchChange = () => {
     setSearchParams((sp) => ({
       ...sp,
-      initial_year: '',
-      final_year: '',
+      initial_year: '1889',
```
amos-cha, line 329:
```suggestion
      initial_year: '1889', //gordon est
```

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -521,28 +524,66 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+                        <SearchField
+                          name="initial_year"
+                          value={searchParams.initial_year}
+                          updateValue={handleUpdate}
+                          options={Array.from(
+                            {
+                              length:
+                                Number(searchParams.final_year) > 0
+                                  ? Number(searchParams.final_year) - 1889 + 1
+                                  : currentYear - 1889 + 1,
```
amos-cha, line 538:
```suggestion
                                searchParams.final_year
                                  ? Number(searchParams.final_year) - 1889 + 1
                                  : currentYear - 1889 + 1,
```

Javascript has falsey zero (0 == false, all other numbers == true), my personal preference is that we utilize this. (This also applies to all the uses of  `Number(searchParams.final_year) > 0` found below

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -521,28 +524,66 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+                            },
+                            (_, i) => ({
+                              value:
+                                Number(searchParams.final_year) > 0
+                                  ? (Number(searchParams.final_year) - i).toString()
+                                  : (currentYear - i).toString(),
```
amos-cha, line 544:
```suggestion
                                  ? `${searchParams.final_year - i}`
                                  : `${currentYear - i}`
```

Personal preference for how code looks. However, I think @EjPlatzer should have a final say on this. I like to just accept that javascript is cursed and thus use the fact that the typeless string numbers can be read by the compiler as integers during any form of subtraction. However, I can see how this could be confusing for code reviewers. 

Also I've always preferred using formatted strings over using casting functions. Once again, personal preference for code prettiness

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -521,28 +524,66 @@ const SearchFieldList = ({ onSearch }: Props) => {
...
+                          name="final_year"
+                          value={searchParams.final_year}
+                          updateValue={handleUpdate}
+                          options={Array.from(
+                            {
+                              length:
+                                Number(searchParams.initial_year) > 0
+                                  ? currentYear - Number(searchParams.initial_year) + 1
+                                  : currentYear - 1889 + 1,
+                            },
```
amos-cha, line 568:
```suggestion
                          options={Array.from(
                            {
                              length:
                                Number(searchParams.initial_year) > 0
                                  ? currentYear - Number(searchParams.initial_year) + 1
                                  : currentYear - 1889 + 1,
                            },
```
The fact that the exact same code for length exists twice makes me think there can be an auxiliary function to do this?

amos-cha (changes requested):
Great work, got a couple nitpicks regarding style

amos-cha (approved):
lgtm

PR #2241, opened 2024-05-23
Testing Format for Mobile Device

Hls5235:
We decided it would be beneficial to see your building and room number for each class when viewing the schedule because many people don't know you can click to view it. We decided a separate format would look nicer on mobile rather than computer and vice versa. I can't seem to figure out how to test if the mobile format works so we decided the best options was to view it on 360 train through a mobile device and remove it shortly after.


bennettforkner:
Hi @Hls5235 ,

You can check what a feature looks like on mobile by using the dev-tools on most major browsers. Here is an article showing how to do it in Chrome.

https://developer.chrome.com/docs/devtools/device-mode

I would request that you do not deploy code to train just for the sake of viewing it if you are able to, as that branch is intended to be a clean pipeline to prod.

Hls5235:
Perfect. Thanks for the help.

PR #2243, opened 2024-05-23
Fixed Starting Status

Hls5235:
The starting status of the schedule accordion was closed and line 33 was giving a warning. After editing line 33 the problem seems to be fixed, it displays no warnings and after clearing my cache, the starting status seems to be open. If this seems inefficient or could possibly cause bugs I would appreciate another approach.

### DIFF
--- a/src/components/Profile/components/SchedulePanel/index.tsx
+++ b/src/components/Profile/components/SchedulePanel/index.tsx
@@ -30,7 +30,7 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
   const [selectedSchedule, setSelectedSchedule] = useState<Schedule | null>(null);
   const [selectedCourse, setSelectedCourse] = useState<CourseEvent | null>(null);
   const [isScheduleOpen, setIsScheduleOpen] = useState<boolean>(
-    localStorage.getItem('scheduleOpenKey') === 'true' ?? true,
+    localStorage.getItem(scheduleOpenKey) === 'true' ?? true,
   );
 
   useEffect(() => {
@@ -54,7 +54,7 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
 
   const toggleIsScheduleOpen = () => {
     setIsScheduleOpen((wasOpen) => {
-      localStorage.setItem('scheduleOpenKey', String(!isScheduleOpen));
+      localStorage.setItem(scheduleOpenKey, String(!isScheduleOpen));
       return !wasOpen;
     });
   };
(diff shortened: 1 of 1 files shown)


src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -54,7 +54,7 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
 
   const toggleIsScheduleOpen = () => {
     setIsScheduleOpen((wasOpen) => {
-      localStorage.setItem(scheduleOpenKey, String(!isScheduleOpen));
+      localStorage.setItem(scheduleOpenKey, String(isScheduleOpen));
```
EjPlatzer, line 57: Looks like you're storing the string `'profile.schedule.isOpenOrNot'`, instead of the `true`/`false` that is intended. Not sure how I missed that on the previous PR, sorry.
```suggestion
      localStorage.setItem(scheduleOpenKey, String(!wasOpen));
```

Fixing this here should make the change to the default value of the `useState` above unnecessary.

src/components/Profile/components/SchedulePanel/index.tsx
```diff
@@ -30,7 +30,7 @@ const GordonSchedulePanel = ({ profile, myProf }: Props) => {
   const [selectedSchedule, setSelectedSchedule] = useState<Schedule | null>(null);
   const [selectedCourse, setSelectedCourse] = useState<CourseEvent | null>(null);
   const [isScheduleOpen, setIsScheduleOpen] = useState<boolean>(
-    localStorage.getItem(scheduleOpenKey) === 'true' ?? true,
+    localStorage.getItem(scheduleOpenKey) !== 'true',
```
EjPlatzer, line 33: I believe this is backwards. Using `!== 'true'` will cause `isScheduleOpen` to default to `false` when the localStorage item `profile.schedule.isOpen` equals `'true'` (meaning the schedule will be closed), and `isScheduleOpen` will default to `false` when `profile.schedule.isOpen` equals `'false'`. Here are screenshots showing this behavior:
The schedule defaulted open when the local storage value is 'true'
image
The schedule defaulted closed when the local storage value is 'false'
image

We want instead to check whether the local storage value is equal to the string `'false'`:
```suggestion
    localStorage.getItem(scheduleOpenKey) !== 'false',
```
That will evaluate to the boolean value `false` only when the local storage value has been set to the string `'false'`. If the local storage value is set to the string `'true'`, or if there is no value in local storage, this will evaluate to the boolean value `true`, meaning the schedule will be open.

Hls5235: Yes, this seems to work, thanks again.

EjPlatzer (approved):
LGTM! Great work

PR #2244, opened 2024-05-24
Updated Header Profile Button Alt text for screen reader

### LINKED ISSUE #2134: Profile Icon Link Inaccurate Alt Text
The profile icon link on the header currently says "button: more."  on a screen reader.  This is likely left over from when it was a nav menu, but it should be updated to something more descriptive.

CollyWills16670:
An issue was mentioned in the Triage board where when hovering over the profile icon in the header of Gordon 360 it's alternate text only read "More". I changed the aria label in "src/components/Header/components/NavAvatarRightCorner/index.jsx" to say "My Profile" then tested it with a screen reader.

Fixes #2134

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -7,4 +7,4 @@
  VITE_API_URL=https://360ApiTrain.gordon.edu/
 
 # @LOCALHOST
-#  VITE_API_URL=http://localhost:51626/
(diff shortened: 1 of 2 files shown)


amos-cha (approved):
lgtm

PR #2249, opened 2024-05-24
took-off-coming-soon-text-for-recim

Dopey12:
It seems that on the about page, in the section where it states what students have worked on for 360, there was "(coming soon)" before the brief description about RECIM, since RECIM is now completed and implemented into 360 I removed the "(coming soon)" text from the About section.


amos-cha (approved):
lgtm! nice find

PR #2252, opened 2024-05-29
S24 quicklinks improvements

dli505:
**Changes made in PR:**
- created a Links page that holds useful links for users
- replaced Timesheets in the header and on the navigation bar with the Links page
- Removed the Quick Links dialogue box from the navigation bar


bennettforkner:
Hey @dli505, do you have any screenshots of the design that you could share here? It would be helpful to see without having to run the code!

dli505:
@bennettforkner Here they are! Sorry I forgot!
image
image

src/components/Nav/components/NavLinks/index.jsx
```diff
@@ -124,18 +135,18 @@ const GordonNavLinks = ({ onLinkClick }) => {
...
-      openUnavailableDialog={setDialog}
-      divider={false}
-      onLinkClick={() => {
-        onLinkClick();
-        setAreLinksOpen(true);
-      }}
-      linkName={'Links'}
-    />
-  );
+  // const linksButton = (
```
bennettforkner, line 138: I would remove this comment from the code. It's in the git history, so it can always be referenced relatively easily.

src/components/Nav/components/NavLinks/index.jsx
```diff
@@ -100,6 +101,16 @@ const GordonNavLinks = ({ onLinkClick }) => {
     />
   );
 
+  const LinksButton = (
+    <GordonNavButton
+      onLinkClick={onLinkClick}
+      linkName={'Links'}
+      linkPath={'/links'}
```
EjPlatzer, line 108: When the value of a prop is just a string, you can pass it as a plain string:
```suggestion
      linkName="Links"
      linkPath="/links"
```

src/components/Nav/components/NavLinks/index.jsx
EjPlatzer: In fact, you can remove all of the QuickLinks code, including the component, the `areLinksOpen` state, etc.

You should also move the images used on the links page to within the `views/Links` folder.

EjPlatzer (changes requested):

amos-cha (approved):
lgtm!

EjPlatzer (approved):
LGTM! Nice improvement!

PR #2255, opened 2024-05-30
add error screen for events

dli505:
**Changes made in PR:**
- when there is an error in retrieving events there should now be an error screen instead of an infinite spinner
- Something like so:
image


amos-cha:
@EjPlatzer I don't think there will ever be an error now with the API update to pass in an empty array right?

russtuck:
It appears that recent backend changes hide the distinction between "unable to get event data" and "there are no matching  events". It would be nice to make this clear to the user.  Priority???

jsenning:
The fix in this PR was made unnecessary due to changes in the API.

PR #2257, opened 2024-05-30
S24 delete chapel credits

yushinj:
delete codes and file of chapel credit card on the Home page, also delete related code from other folders.
I accidentally contained previous changes, so please check "delete chapel credit card from Home page".


src/theme.ts
```diff
@@ -278,6 +278,7 @@ export enum ColorSetting {
   System = 'system',
   Dark = 'dark',
   Light = 'light',
+  Custom = 'custom',
```
EjPlatzer, line 281: I'm not sure why these were added. Is this related to removing the Chapel Credit display?

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -24,6 +24,8 @@ const CoCurricularTranscript = () => {
   const isAuthenticated = useIsAuthenticated();
   const { profile: userProfile, loading: loadingProfile } = useUser();
   const profile = isAuthenticated ? userProfile : exampleStudentProfile;
+  console.log('transcript');
+  console.log(transcriptItems);
```
EjPlatzer, line 28: Can you remove theses console logs? They should not be merged into production.

EjPlatzer (changes requested):
This is a nice change, now that we have iAttended to track chapel credits even better than 360 did.

I left comments about a couple of things that don't seem related to removing chapel credits.

EjPlatzer (approved):
LGTM! Great work!

PR #2259, opened 2024-05-30
Addressed Bugs

Hls5235:
I noticed a bug with the sizing on the phone and mac so fixed that along with improving the code so that you don't have to refresh the page to see the different format.


EjPlatzer (approved):
LGTM!

PR #2261, opened 2024-05-30
Lazy load images when sensible

EjPlatzer:
Add the `loading="lazy"` attribute to `img` components, so that the browser can delay loading images that are not yet in the viewport (i.e. haven't been scrolled into the visible section of the page). See [the `loading` attribute of the `img` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading) for more info.

PS: This attribute is not yet well supported on Firefox. Also, for some reason, the MUI Avatar component is not respecting the loading attribute in the for the img slot props. Nevertheless, this serves as a great progressive enhancement for the browsers/components that employ it, and won't cause any issues in cases where it isn't supported.


dli505:
lgtm!

dli505 (approved):
lgtm!

Issue #2265, opened 2024-05-31
Add iAttended to Links page


EjPlatzer:
Since we have removed Chapel Credits from the homepage dashboard, it would be good to have a link to iAttended. The webpage is [iAttendedApp.com](https://iattendedapp.com/). That page has a link to the app store and google play store for people who don't have the app to get it. Users can also sign in to the website and access most or all of iAttended's features that way.

SeanAverS:
Hello, 

I am interested in working on this issue. Could I please be assigned to it? Thanks.

SeanAverS:
Hello,

I am interested in working on this issue. Could I please be assigned? Thanks.

PR #2269, opened 2024-06-03
Add iAttended to Links Page

Luke-W-Hart:
Added iAttended to links page with link provided by Evan in #2265 
Looks like this now...
Screenshot from 2024-06-03 11-24-10


amos-cha (approved):
lgtm

Issue #2135, opened 2023-11-01
Events not focusable


erandersen21:
In the events page, events cannot be accessed or expanded when using the tab key.

Luke-W-Hart:
Need

- [Events accessed when using tab key]
- Event to expand using enter key (I would assume)

erandersen21:
?

On Jun 3, 2024, at 4:04 PM, Luke-W-Hart ***@***.***> wrote:



Need

  *   [Events accessed when using tab key]
  *   Event to expand using enter key (I would assume)

Issue #2133, opened 2023-10-27
Accessibility issues on Events tab


russtuck:
Making the list of events a table, instead of a ton of <div>s, would describe the semantics and let screen readers do a much better job.

This probably applies in other places, like Involvements, too.

Luke-W-Hart:
Related to #2135

PR #2271, opened 2024-06-03
Make Events Focusable with Tab

Luke-W-Hart:
#2135 
Made event item focusable with tab key and description of event focusable with tab key, **but** not able to make event expand with enter key.
Just found this was related to #2133 , not sure if the way I'm doing it is the best


amos-cha (approved):
Good work! LGTM

PR #2273, opened 2024-06-04
fixed-takethetest-colorissue-darkmode

Dopey12:
It seems that on color mode there was an issue with the "Take the test" visual appearance in dark mode, it was an odd blue shade and it wasn't clearly visible. I have made a change to where now the text "Take the test" matches the link colors. 
Before the change: 
Screenshot from 2024-06-04 12-11-55
After the change:
Screenshot from 2024-06-04 12-12-08
The change matches the outline of also other links in the profile page, overall it looks much cleaner.


EjPlatzer (approved):
LGTM!

PR #2274, opened 2024-06-05
Make Enter Key Expand Event

Luke-W-Hart:
Related to #2133 and #2135 
Can now expand an event using the enter key **BUT** anytime you press enter while focused on an event it opens or closes it no matter what. So if you tab to "Add to Calendar" but press enter it just closes the event.

You can however press space while focused on "Add to Calendar" and that doesn't close the event. So it's not perfect, but you can at least open the event now.

### DIFF
--- a/src/components/EventList/components/EventItem/index.jsx
+++ b/src/components/EventList/components/EventItem/index.jsx
@@ -16,7 +16,11 @@ const EventItem = ({ event }) => {
       onClick={() => setExpanded((e) => !e)}
       className={styles.event_item}
       tabIndex={0}
-      onKeyDown={(e) => (e.key === 'Enter' ? setExpanded((e) => !e) : {})}
+      onKeyDown={(e) => {
+        if (e.key === 'Enter' && e.currentTarget === e.target) {
+          setExpanded((e) => !e);
+        }
+      }}
     >
       <Grid item xs={12} sm={4}>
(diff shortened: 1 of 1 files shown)


src/components/EventList/components/EventItem/index.jsx
```diff
@@ -16,6 +16,7 @@ const EventItem = ({ event }) => {
       onClick={() => setExpanded((e) => !e)}
       className={styles.event_item}
       tabIndex={0}
+      onKeyDown={(e) => (e.key === 'Enter' ? setExpanded((e) => !e) : {})}
```
EjPlatzer, line 19: You can check if the `target` of the event is the same as the `currentTarget`, to see whether this event was triggered on this element or if it bubbled up from a child element. More info: https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget
```suggestion
      onKeyDown={(e) => (e.key === 'Enter' && e.currentTarget === e.target ? setExpanded((e) => !e) : {})}
```

Also, I think this anonymous function would be easier to read if we use a normal function body:
```suggestion
      onKeyDown={(e) => {
          if (e.key === 'Enter') {
              setExpanded((e) => !e);
          }
      }}
```

Luke-W-Hart:
Should be all good now. Thanks so much for the help and resource Evan!

EjPlatzer (approved):
LGTM!

Issue #2137, opened 2023-11-01
Accessability issue on schedule section of profile.


hd7970:
When navigating through schedule section via keyboard, you can only select the first course of each day.

Luke-W-Hart:
It seems you can navigate through the courses but can't press enter when your focused on them to add them to your calendar. 
Also pressing tab doesn't navigate to the "ONLN ASYNC" but these aren't really necessary to add to your calendar (since there's no set meeting time) so I would assume that's fine.

PR #2276, opened 2024-06-05
Make Enter Key Open Course Details & Add Course to Calendar Pop-up

Luke-W-Hart:
Related to #2137 
On MyProfile Schedule, adds a simple onKeyPressEvent so when focused on a specific course, any time a key besides tab, esc, or arrow keys are pressed the add course to schedule pop-up comes up.
Could not figure out how to pop-up only on enter key press (for now), but adding the course to your schedule is now possible with just the keyboard.
Also ONLN ASYN courses are not accessible via keyboard navigation, but I would assume this isn't necessary.


jsenning (commented):
During testing I found that now any key (including tab, esc, and enter) will cause the selection event, so I cannot navigate though calendar entries using tab.

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx
```diff
@@ -44,6 +44,7 @@ const GordonScheduleCalendar = ({ schedule, onSelectEvent }: Props) => {
         dayHeaderFormat: () => schedule.session.SessionDescription,
       }}
       onSelectEvent={onSelectEvent}
+      onKeyPressEvent={onSelectEvent}
```
EjPlatzer, line 47: Looking at the docs for `react-big-calendar`, it appears that `onKeyPress` passes two parameters: firstly, the "calendar event" that was selected (i.e. the course in our case), and secondly the "keypress event", i.e. details about the keypress that occurred. (Event is such an overloaded word in programming...)

We can use the info on the keypress event to decide whether the key press is equivalent to selecting an event or not. For example:
```suggestion
      onKeyPressEvent={(selectedEvent, keyPressEvent) => {
          if (event.Key === 'Enter') {
              onSelectEvent(selectedEvent);
          }
      }}
```

Luke-W-Hart: Tried looking at the ```react-big-calender``` documentation as well, didn't realize the "SyntheticEvent" mentioned as the second parameter was actually a keypress event. Thanks Evan!

Luke-W-Hart:
The code I just pushed worked as intended (only opened the pop-up when I pressed enter), just need to see how to deal with the error.

jsenning:
LGTM - This now seems to work correctly

jsenning (approved):
LGTM

jsenning (changes requested):
Need to resolve lint error -  no key field in onKeyPressEvent

Luke-W-Hart:
@EjPlatzer we keep receiving the error that "key is not a property on type 'SyntheticEvent<HTMLElement, Event>", but if we console.log the key press event we receive this
Screenshot from 2024-06-06 09-58-43
Testing this out all of these properties give the same error, any idea as to why it would be doing this?

EjPlatzer:
> @EjPlatzer we keep receiving the error that "key is not a property on type 'SyntheticEvent<HTMLElement, Event>", but if we console.log the key press event we receive this Screenshot from 2024-06-06 09-58-43 So there is a key property, and testing this out all of these properties give the same error, any idea as to why it would be doing this?

@Luke-W-Hart this appears to be a bug (or at least poor design) in the `react-big-calendar` package (or more accurately, [the community-contributed types for the package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b59dc3ac1e2770fbd6cdbb90ba52abe04c168196/types/react-big-calendar/index.d.ts#L443), since `react-big-calendar` doesn't use TypeScript).

They have declared the type of `onKeyPressEvent`'s second parameter to be `React.SyntheticEvent<HTMLElement>`, which is the most general concept of any event in React. In other words, they have said that it will be some kind of event, but not what kind of event it will be. So as far as TypeScript is concerned, this could be any kind of event and we only know that we have access to the properties that all events have in common (e.g. `target`), but not any of the properties specific to a KeyboardEvent.

We could try to contribute a patch to the types for this package, but in the meantime the simplest fix currently is check whether the property exists and is of the correct type. This will tell TypeScript that it's definitely there. Technically, this is the correct thing to do if we don't know that that event will always be of type `KeyboardEvent` and therefore always have `key: string` property.

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx
```diff
@@ -44,6 +44,11 @@ const GordonScheduleCalendar = ({ schedule, onSelectEvent }: Props) => {
         dayHeaderFormat: () => schedule.session.SessionDescription,
       }}
       onSelectEvent={onSelectEvent}
+      onKeyPressEvent={(selectedEvent, keyPressEvent) => {
+        if (keyPressEvent.key === 'Enter') {
```
EjPlatzer, line 48:
```suggestion
        if (
            'key' in keyPressEvent &&
            typeof keyPressEvent.key === 'string' &
            keyPressEvent.key === 'Enter'
        ) {
```

Luke-W-Hart:
Thanks for clearing that up Evan!

jsenning (approved):
LGTM

PR #2280, opened 2024-06-05
reduce redundancy in transcript

yushinj:
Some departments have the same string on Job_Title and Job_Department_Name. Also, transcript repeats the Job_title for each session. So, remove redundant information from transcript.

Example:

train_trancsript

transcript1

### DIFF
--- a/src/views/CoCurricularTranscript/Components/Experience/index.tsx
+++ b/src/views/CoCurricularTranscript/Components/Experience/index.tsx
@@ -8,9 +8,7 @@ type Props = {
 
 const Experience = ({ Experience }: Props) => (
   <div className={styles.experience_transcript_activities}>
-    <div className={styles.organization_role}>
-      {Experience.Job_Department_Name}, {Experience.Job_Title}
-    </div>
+    <div className={styles.organization_role}>{newJobTitle(Experience)}</div>
     <div className={styles.date}> {formatDuration(Experience)} </div>
   </div>
 );
@@ -32,6 +30,21 @@ const formatDuration = ({ Job_Start_Date, Job_End_Date }: StudentEmployment) => 
     return `${format(startDate, 'MMM')} - ${format(endDate, 'MMM yyyy')}`;
   } else {
     return `${format(startDate, 'MMM yyyy')} - ${format(endDate, 'MMM yyyy')}`;
+  }
+};
+
+let prev_job_title = '';
+
+const newJobTitle = ({ Job_Department_Name, Job_Title }: StudentEmployment) => {
+  if (Job_Title != prev_job_title) {
+    prev_job_title = Job_Title;
+    if (Job_Department_Name == Job_Title.split(':')[0]) {
+      return Job_Title;
+    } else {
+      return Job_Department_Name + ', ' + Job_Title;
+    }
+  } else {
+    return '';
   }
 };
 


src/views/CoCurricularTranscript/Components/Experience/index.tsx
```diff
@@ -35,4 +33,19 @@ const formatDuration = ({ Job_Start_Date, Job_End_Date }: StudentEmployment) =>
...
+    prev_job_title = Job_Title;
+    if (Job_Department_Name == Job_Title.split(':')[0]) {
+      return Job_Title;
+    } else {
+      return Job_Department_Name + ', ' + Job_Title;
+    }
+  } else {
+    return '';
+  }
+};
```
amos-cha, line 49:
```suggestion
let prev_job_title = '';

const newJobTitle = ({ Job_Department_Name, Job_Title }: StudentEmployment) => {
  if (Job_Title != prev_job_title) {
    prev_job_title = Job_Title;
    return Job_Department_Name === Job_Title.split(':')[0] 
      ? Job_Title
      : `${Job_Department_Name}, ${Job_Title}`;
  }
  return '';
};
```

Couple things:
1) I'm pretty sure your variable `prev_job_title` as scoped variable can be finicky. @EjPlatzer probably has an opinion on this
2) there is some logic cleanup you can employ

src/services/transcript.ts
```diff
@@ -22,7 +22,9 @@ const getItems = (username: string) =>
   Promise.all([
     userService.getMembershipHistory(username),
     http.get<StudentEmployment[]>('studentemployment/'),
-  ]).then(([memberships, jobs]) => categorizeItems(memberships, jobs));
+  ])
+    .then(([memberships, jobs]) => categorizeItems(memberships, jobs))
+    .then();
```
amos-cha, line 27:
```suggestion
```
I don't think this is necessary

src/services/transcript.ts
```diff
@@ -70,8 +72,14 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
 
   groupedMembershipHistory.activities = memberships;
 
-  groupedMembershipHistory.experiences.sort(
-    (a, b) => getExperienceEndDate(b) - getExperienceEndDate(a),
+  // sorting job title by alphabetical order
+  groupedMembershipHistory.experiences.sort((a, b) => getJobTitle(a).localeCompare(getJobTitle(b)));
```
amos-cha, line 76:
```suggestion
```
this is made redundant by the sort below

src/services/transcript.ts
```diff
@@ -141,3 +149,9 @@ const transcriptService = {
 };
 
 export default transcriptService;
+
+const getJobTitle = (experience: MembershipHistory | StudentEmployment) => {
+  const title = 'Sessions' in experience ? experience.ActivityCode : experience.Job_Title;
+
+  return title;
+};
```
amos-cha, line 157:
```suggestion
const getJobTitle = (experience: MembershipHistory | StudentEmployment) => 
   'Sessions' in experience ? experience.ActivityCode : experience.Job_Title;
```

amos-cha (changes requested):
almost there, just a bit of cleanup

src/services/transcript.ts
yushinj: the order of job titles is not alphabetical when the first sorting code is deleted. 

This is a transcript with sorting alphabetically
train_after

This is a transcript after deleting the first sorting codes. The gray lines are for css test, ignore them.
Screenshot from 2024-06-12 08-45-09

src/services/transcript.ts
```diff
@@ -70,8 +71,13 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
 
   groupedMembershipHistory.activities = memberships;
 
-  groupedMembershipHistory.experiences.sort(
-    (a, b) => getExperienceEndDate(b) - getExperienceEndDate(a),
+  // sorting job title by alphabetical order
+
+  // sorting the same job title by end date
+  groupedMembershipHistory.experiences.sort((a, b) =>
+    getJobTitle(a).localeCompare(getJobTitle(b)) != 0
```
amos-cha, line 78:
```suggestion
    getJobTitle(a).localeCompare(getJobTitle(b)) !== 0
```

src/views/CoCurricularTranscript/index.tsx
```diff
@@ -32,6 +33,7 @@ const CoCurricularTranscript = () => {
           return;
         }
         setLoading(true);
+        setPreviousTitles([]);
```
amos-cha, line 36:
```suggestion
        setPreviousTitles([]);
```
make a comment about this bugging out BECAUSE of the default transcript item. or else this looks confusing

amos-cha (commented):
2 more things i just saw

amos-cha (approved):
lgtm

src/services/transcript.ts
```diff
@@ -141,3 +147,6 @@ const transcriptService = {
 };
 
 export default transcriptService;
+
+const getJobTitle = (experience: MembershipHistory | StudentEmployment) =>
```
EjPlatzer, line 151: The name of this function is misleading. Maybe `getName` would be clearer?
```suggestion
const getName = (experience: MembershipHistory | StudentEmployment) =>
```

src/services/transcript.ts
```diff
@@ -70,8 +70,14 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
 
   groupedMembershipHistory.activities = memberships;
 
-  groupedMembershipHistory.experiences.sort(
-    (a, b) => getExperienceEndDate(b) - getExperienceEndDate(a),
+  /**
+   * Sorting job titles by alphabetical order.
+   * Sorting job titles by end date if there are the same job titles.
+   */
```
EjPlatzer, line 76:
```suggestion
   // Sort experiences by experience name then by end date
```

src/services/transcript.ts
```diff
@@ -70,8 +70,14 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
...
+  groupedMembershipHistory.experiences.sort((a, b) =>
+    getJobTitle(a).localeCompare(getJobTitle(b)) !== 0
+      ? getJobTitle(a).localeCompare(getJobTitle(b))
+      : getExperienceEndDate(b) - getExperienceEndDate(a),
   );
```
EjPlatzer, line 81:
```suggestion
  groupedMembershipHistory.experiences.sort((a, b) => {
     const nameComparison = getName(a).localeCompare(getName(b));
     return nameComparison !== 0 
        ? nameComparison
        : getExperienceEndDate(b) - getExperienceEndDate(a);
  });
```

EjPlatzer (changes requested):
Thanks for working on this! I haven't gotten the chance to review the rest of the code yet, but I have a couple of comments on the parts in `services/transcript.ts`.

src/views/CoCurricularTranscript/Components/Experience/Experience.module.scss
```diff
@@ -31,8 +31,29 @@
...
+    text-align: left;
+  }
+
+  .date {
+    grid-column: 3;
+    grid-row: 1;
+    text-align: left;
+  }
+}
+
```
EjPlatzer, line 54: It's not ideal to duplicate the whole styles, just to add the `position: relative` and `top: -20px`. It would be better to keep those in a separate class and simply apply both classes when needed:
```suggestion
.empty_title {
  position: relative;
  top: -20px;
}

```

src/views/CoCurricularTranscript/Components/Experience/index.tsx
```diff
@@ -1,19 +1,27 @@
...
-    <div className={styles.organization_role}>
-      {Experience.Job_Department_Name}, {Experience.Job_Title}
+const Experience = ({ Experience, previousTitles, setPreviousTitles }: Props) => {
+  const jobTitles = newJobTitle(Experience, previousTitles, setPreviousTitles);
+  const experienceTranscript =
+    jobTitles === ''
+      ? styles.experience_transcript_activities_empty_titles
+      : styles.experience_transcript_activities;
+  return (
+    <div className={experienceTranscript}>
```
EjPlatzer, line 19:
```suggestion
  return (
    <div className={jobTitles === ''
      ? `${styles.experience_transcript_activities} ${styles.empty_title}`
      : styles.experience_transcript_activities}>
```

src/views/CoCurricularTranscript/Components/Experience/index.tsx
```diff
@@ -35,4 +43,19 @@ const formatDuration = ({ Job_Start_Date, Job_End_Date }: StudentEmployment) =>
...
+  if (!previousTitles.includes(Job_Title)) {
+    previousTitles.push(Job_Title);
+    setPreviousTitles(previousTitles);
+    return Job_Department_Name === Job_Title.split(':')[0]
+      ? Job_Title
+      : `${Job_Department_Name}, ${Job_Title}`;
+  }
+  return '';
+};
+
```
EjPlatzer, line 60: There is another way to do this, which in my opinion is both easier and more maintainable.

Basically, you can use [`Object.groupBy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) to group the experiences by Job_Title, and display the title once per group of jobs with the same title.

The grouping is as simple as:
```typescript
const grouped_experiences = Object.groupBy(experiences, (experience) => experience.Job_Title);
```

EjPlatzer (changes requested):
I think this is great work, thank you for improving the transcript.

The current way you're handling repeated jobs does seem to work, but it's more complicated than it needs to be. I commented below with an alternate approach where you group the experiences by their job_title so that you can display the job title only once per group of jobs with the same title.

src/views/CoCurricularTranscript/Components/Experience/index.tsx
EjPlatzer: To give a more complete example, here's a simplified experiences array:
```javascript
const jobs = [
    {title: 'Dishwasher', start: new Date('2016-08-20'), end: new Date('2017-05-15')},
    {title: 'RA', start: new Date('2017-08-01'), end: new Date('2018-05-25')},
    {title: 'RA', start: new Date('2018-08-02'), end: new Date('2019-05-10')}
];
```

and we can group it like so:
```javascript
const jobsByTitle = Object.groupBy(jobs, job => job.title);
```

we can convert that object to an array of objects, where each object in the array has a `Job_Title` property which is the title of all the jobs in that group, and a `jobs` property which is the array containing all the jobs in that group:
```javascript
const jobGroups = Object.entries(jobsByTitle).map(([title, jobs]) => ({Job_Title: title, jobs}))
```

then we can render those job groups:
```jsx
const Experience = ({Job_Title, jobs}) => {
return (
    <div className={styles.experience_transcript_activities}>
        <div className={styles.organization_role}>{Job_Title}</div>
        {jobs.map((job) => (
           <div>{formatDuration(job)}</div>
        )}
    </div>
)
```

yushinj: Thank you for your suggestion. Can you explain why using group is easier and more maintainable? This code takes an array of information one by one from the ordered list. If I use group, it need an extra function to extract data from the group and a function that is similar to the upper code to display one title per group and to ignore job titles in the group, which is more complicate.

EjPlatzer: Because your current code is using React's `useState` hook to keep track of whether a certain title has been seen yet or not, you have to update state (and thus re-render the entire transcript) once for each experience. If you instead group experiences by job title before passing the data to React (i.e. within `services/transcript.ts`), you will already have all the data calculated and won't have to re-render any extra times.

yushinj: Thank you for your answer! I tried to group experiences within `services/transcript.ts` (line 51-78), but the type of data will be changed from `groupedMembershipHistory` to array if I group by job title. Is there better way to group experience or improve the code without `groupBy`?

src/services/transcript.ts
```diff
@@ -70,10 +76,71 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
...
+  }
+
+  for (let i = 0; i < test.length; i++) {
+    jobs[i] = test[i];
+  }
+
+  //console.log('jobs');
+  //console.log(jobs);
+  //console.log('groupedMembershipHistory');
+  //console.log(groupedMembershipHistory);
```
amos-cha, line 143:
```suggestion
```

Remove unused consolelogs

src/views/CoCurricularTranscript/Components/Experience/index.tsx
```diff
@@ -35,4 +42,11 @@ const formatDuration = ({ Job_Start_Date, Job_End_Date }: StudentEmployment) =>
   }
 };
 
+const newJobTitle = ({ Job_Department_Name, Job_Title }: StudentEmployment) => {
+  if (Job_Title === '') {
+    return '';
+  }
+  return Job_Title === '' ? Job_Title : `${Job_Department_Name}, ${Job_Title}`;
```
amos-cha, line 49:
```suggestion
  if (Job_Title === '') {
    return '';
  }
  return Job_Title === '' ? Job_Title : `${Job_Department_Name}, ${Job_Title}`;
```
that top part and the first half of the ternary are logically identical. Is there a reason both are included?

src/services/transcript.ts
```diff
@@ -70,10 +76,71 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
...
+    }
+    GroupByTitle[i].latestDate = maxDate;
+  }
+
+  // Sorting the grouped job by the latest end date
+  GroupByTitle.sort((a, b) => {
+    return getLatestEndDate(b)! - getLatestEndDate(a)!;
+  });
+
+  console.log(GroupByTitle);
```
amos-cha, line 118:
```suggestion

```

remove console logs

src/services/transcript.ts
```diff
@@ -70,10 +76,71 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
...
-    (a, b) => getExperienceEndDate(b) - getExperienceEndDate(a),
+  // Sort experiences by experience name then by end date
+  groupedMembershipHistory.experiences.sort((a, b) => {
+    const nameComparison = getName(a).localeCompare(getName(b));
+    return nameComparison !== 0
+      ? nameComparison
+      : getExperienceEndDate(b) - getExperienceEndDate(a);
+  });
+
+  //groupedMembershipHistory.experiences = jobs;
```
amos-cha, line 87:
```suggestion

```
remove unused code

amos-cha (changes requested):

package.json
```diff
@@ -59,7 +59,7 @@
     "sass": "^1.71.1",
     "stylelint": "^16.2.1",
     "stylelint-config-recommended-scss": "^14.0.0",
-    "typescript": "^5.3.3",
+    "typescript": "5.5",
```
EjPlatzer, line 62: This locks us to exactly version 5.5 of TypeScript, preventing us from receiving patch updates that may fix bugs before the next major release. Instead, we should tell npm to install the best version of TypeScript that has no breaking changes since 5.5:
```suggestion
    "typescript": "^5.5",
```

src/services/transcript.ts
```diff
@@ -18,6 +18,26 @@ export type TranscriptItems = {
   activities: MembershipHistory[];
 };
 
+export type Groupexperience = {
+  Job_Title: string;
+  job: StudentEmployment[] | undefined;
+  latestDate?: string;
+};
+
+export type NewStudentEmployment = {
```
russtuck, line 27: This needs to become StudentEmployment, and the old one above should be deleted.

src/services/transcript.ts
```diff
@@ -18,6 +18,26 @@ export type TranscriptItems = {
...
+  Job_Department: string;
+  Job_Department_Name: string;
+  Job_Date: session[];
+  Job_Latest_Date: string;
+};
+
+export type session = {
+  Job_Start_Date?: string;
+  Job_End_Date?: string;
+  Job_Expected_Date?: string;
```
russtuck, line 38:
```suggestion
  Job_Expected_End_Date?: string;
```
Keep meaning of this field.

src/services/transcript.ts
```diff
@@ -70,9 +90,65 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
...
+      }
+      if (value) {
+        test.push(value);
+      }
+    }
+  }
+
+  for (let i = 0; i < test.length; i++) {
+    jobs[i] = test[i];
+  }
```
russtuck, line 151: I think this should be deleted.

src/services/transcript.ts
```diff
@@ -70,9 +90,65 @@ const categorizeItems = async (memberships: MembershipHistory[], jobs: StudentEm
...
+        maxDate = value!;
+      }
+    }
+    GroupByTitle[i].latestDate = maxDate;
+  }
+
+  // Sorting the grouped job by the latest end date
+  GroupByTitle.sort((a, b) => {
+    return getLatestEndDate(b)! - getLatestEndDate(a)!;
+  });
```
russtuck, line 131: Return this new GroupByTitle in groupedMembershipHistory, NOT the de-grouped version below. (Why do all the work to combine each job into one entry with multiple times, and then not use it?)

src/views/Home/components/DiningBalance/index.jsx
```diff
@@ -11,7 +11,7 @@ import {
 import EditIcon from '@mui/icons-material/Edit';
 import GordonLoader from 'components/Loader';
 import { useEffect, useState } from 'react';
-import { Doughnut } from 'react-chartjs-2';
+import { Pie } from 'react-chartjs-2';
```
russtuck, line 14: Why is this unrelated change in this pull request.  Please put it in a different branch and make a pull request just for it.  Then it won't need to wait for the other parts you are still working on.

src/views/CoCurricularTranscript/Components/Experience/index.tsx
```diff
@@ -6,14 +6,21 @@ type Props = {
   Experience: StudentEmployment;
 };
 
-const Experience = ({ Experience }: Props) => (
-  <div className={styles.experience_transcript_activities}>
-    <div className={styles.organization_role}>
-      {Experience.Job_Department_Name}, {Experience.Job_Title}
+const Experience = ({ Experience }: Props) => {
```
russtuck, line 9: I think this could be simpler if you passed in the grouped jobs.

russtuck (changes requested):
Please split this into 2 pull requests, so the dining points pie chart won't be in here with jobs.

Issue #2268, opened 2024-06-03
Add "days left" for facstaff


russtuck:
Since the chapel display was removed, non-students don't see days left in term.  It would be nice to show it in some simple way. (Perhaps a bar, with a count.)

It would be nice to have an "info" pop-up explaining what it's counting down to (last day of classes?  last day of finals?).

PR #2281, opened 2024-06-05
Added Days Left Progress Bar

Hls5235:
After deleting the chapel credits from the Home Page the staff and faculty could no longer see how many days are left in the term. I added a progress bar onto the Home Page so the staff and faculty can see it. After discussion it seems this feature would be useful to students for different reason than the wheel that they already see on the dining balance card, so I made it visible for everyone. It was also made clear that users would benefit from seeing the time before the spring term begins during the winter break and the same as the small gap between the start of the spring term and end of the summer term along with the start of the summer term and end of the fall term. I added logic in to take care of this, I also did this so it is easy to adjust if for some reason we would stray from the current terms we use. 
#2268 
Screenshot from 2024-06-05 17-13-43


jsenning (changes requested):
Just started to look at this, but please remove .env.development from the PR

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -1,11 +1,11 @@
...
-import { Doughnut } from 'react-chartjs-2';
-import GordonLoader from 'components/Loader';
 import session from 'services/session';
 import styles from '../Doughnut.module.css';
-import { Card, CardContent, CardHeader, Grid, Typography } from '@mui/material';
+import { Card, CardContent, CardHeader, Grid, Typography, checkboxClasses } from '@mui/material';
 import { theme360 } from 'theme';
 import { format, parseISO } from 'date-fns';
+import LinearProgress from '@material-ui/core/LinearProgress';
+import { withStyles } from '@material-ui/core/styles';
```
amos-cha, line 8:
```suggestion
import LinearProgress from '@material-ui/core/LinearProgress';
import { withStyles } from '@material-ui/core/styles';
```
pretty sure we can do these from scss files

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -1,11 +1,11 @@
 import { useEffect, useState } from 'react';
-import { Doughnut } from 'react-chartjs-2';
-import GordonLoader from 'components/Loader';
 import session from 'services/session';
 import styles from '../Doughnut.module.css';
-import { Card, CardContent, CardHeader, Grid, Typography } from '@mui/material';
+import { Card, CardContent, CardHeader, Grid, Typography, checkboxClasses } from '@mui/material';
```
amos-cha, line 4:
```suggestion
import { Card, CardContent, CardHeader, Grid, Typography } from '@mui/material';
```
I don't see where this import is being used

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -24,65 +33,145 @@ const DaysLeft = () => {
...
-        /(Academic Year)|(Grad)/gm,
-        '',
-      );
+      const fallBegin = new Date((await session.get(`${currentYear - 1}09`)).SessionBeginDate);
+      const fallEnd = new Date((await session.get(`${currentYear - 1}09`)).SessionEndDate);
+      const springBegin = new Date((await session.get(`${currentYear}01`)).SessionBeginDate);
+      const springEnd = new Date((await session.get(`${currentYear}01`)).SessionEndDate);
+      const summerBegin = new Date((await session.get(`${currentYear}05`)).SessionBeginDate);
+      const summerEnd = new Date((await session.get(`${currentYear}05`)).SessionEndDate);
+      const nextFallBegin = new Date((await session.get(`${currentYear}09`)).SessionBeginDate);
```
amos-cha, line 43:
```suggestion
      const fallBegin = new Date((await session.get(`${currentYear - 1}09`)).SessionBeginDate);
      const fallEnd = new Date((await session.get(`${currentYear - 1}09`)).SessionEndDate);
      const springBegin = new Date((await session.get(`${currentYear}01`)).SessionBeginDate);
      const springEnd = new Date((await session.get(`${currentYear}01`)).SessionEndDate);
      const summerBegin = new Date((await session.get(`${currentYear}05`)).SessionBeginDate);
      const summerEnd = new Date((await session.get(`${currentYear}05`)).SessionEndDate);
      const nextFallBegin = new Date((await session.get(`${currentYear}09`)).SessionBeginDate);

```
This seems wildly inefficient. Also the magic numbers are a bit of a concern. You could potentially store all beginning/end dates in a single JSON object with a useMemo hook to prevent outstanding memory usage. (there are plenty of tutorials on how to use [useMemo](https://www.freecodecamp.org/news/how-to-work-with-usememo-in-react/))

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -24,65 +33,145 @@ const DaysLeft = () => {
...
+  const msToDays = 1.15741e-8;
+  const winterBreak = Math.round((springBegin - fallEnd) * msToDays);
+  const daysRemainingWinter = Math.round((springBegin - today) * msToDays);
+  const winterProgress = (Math.round((today - fallEnd) * msToDays) / winterBreak) * 100;
+  const preSummer = Math.round((summerBegin - springEnd) * msToDays);
+  const preSummerProgress = (Math.round((today - springEnd) * msToDays) / preSummer) * 100;
+  const daysRemainingPreSummer = Math.round((summerBegin - today) * msToDays);
+  const postSummer = Math.round((nextFallBegin - summerEnd) * msToDays);
+  const daysRemainingPostSummer = Math.round((nextFallBegin - today) * msToDays);
+  const postSummerProgress = (Math.round((today - summerEnd) * msToDays) / postSummer) * 100;
```
amos-cha, line 75:
```suggestion
  const msToDays = 1.15741e-8;
  const winterBreak = Math.round((springBegin - fallEnd) * msToDays);
  const daysRemainingWinter = Math.round((springBegin - today) * msToDays);
  const winterProgress = (Math.round((today - fallEnd) * msToDays) / winterBreak) * 100;
  const preSummer = Math.round((summerBegin - springEnd) * msToDays);
  const preSummerProgress = (Math.round((today - springEnd) * msToDays) / preSummer) * 100;
  const daysRemainingPreSummer = Math.round((summerBegin - today) * msToDays);
  const postSummer = Math.round((nextFallBegin - summerEnd) * msToDays);
  const daysRemainingPostSummer = Math.round((nextFallBegin - today) * msToDays);
  const postSummerProgress = (Math.round((today - summerEnd) * msToDays) / postSummer) * 100;
```
Once again this is just a bit odd as well as the magic numbers.

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -24,65 +33,145 @@ const DaysLeft = () => {
...
+      position: 'relative',
+    },
+    colorPrimary: {
+      backgroundColor: '#00aeef',
+    },
+    bar: {
+      borderRadius: 5,
+      backgroundColor: '#014983',
+    },
+  }))(LinearProgress);
```
amos-cha, line 154: mentioned earlier, but we can use our css files to handle styling

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -24,65 +33,145 @@ const DaysLeft = () => {
...
+  for (let term of termLoop) {
+    if (
+      today >= term.start &&
+      today <= term.end &&
+      (term.condition === undefined || term.condition)
+    ) {
+      termFinder = term.type;
+      break;
+    }
+  }
```
amos-cha, line 97:
```suggestion
  for (let term of termLoop) {
    if (
      today >= term.start &&
      today <= term.end &&
      (term.condition === undefined || term.condition)
    ) {
      termFinder = term.type;
      break;
    }
  }
  this forloop is executed every time this file is loaded, once again not a very efficient method.
```

src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -24,65 +33,145 @@ const DaysLeft = () => {
...
+    { term: termBreaks[2], label: academicTerms[0] },
+  ];
+
+  let termLabel = academicTerms[0];
+
+  for (let currentLabel of termLabelLoop) {
+    if (termFinder === currentLabel.term) {
+      termLabel = currentLabel.label;
+    }
+  }
```
amos-cha, line 123: same as above. You can put these in functions that are called in a useEffect to prevent recalculations each time the page is loaded.

amos-cha (dismissed):

ChrisImagineer:
Can @EjPlatzer take a look - how does the term start/end connect with the Activity listed in the Activity Table that manages the timing for the Wheel? Reason: there have been habits of Session builders to set session dates out of sync with actual experience of the first day of classes and end day of classses which makes the wheel confusing - so a workaround was made a few years ago to control the days remaining related to the wheel.

Hls5235:
Hey @ChrisImagineer, I believe this is already taken care of because I am pulling the start and end dates from the same place that the wheel gets it from. @EjPlatzer should still check but I believe the dates are correct.

Hls5235:
Here is an updated look at the progress bar.
Screenshot from 2024-06-18 14-35-01

jsenning (approved):
LGTM

PR #2284, opened 2024-06-06
Help Page Extra Navigation

Dopey12:
Below I added a jump to side grid, like an anchor, to make it more accessible for users to navigate to a certain section/header on the help page. The more we keep adding to 360 the longer the help page gets so I assumed this could be a good add for user accessibility. I also added a recIM and links paragraph for the site navigation, since those weren't added yet. I don't believe this will be my final result so I am just looking for any feedback, recommendation's to help this look better. #Issue2278

Screenshot from 2024-06-06 10-12-38

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "react-imask": "^7.4.0",
     "react-router": "^6.22.1",
     "react-router-dom": "^6.22.1",
+    "react-router-hash-link": "^2.4.3",
     "react-to-print": "^2.15.1",
     "react-visibility-sensor": "^5.1.1",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.19.3/xlsx-0.19.3.tgz"
--- a/src/views/Help/index.jsx
+++ b/src/views/Help/index.jsx
@@ -1,10 +1,76 @@
 import styles from './Help.module.css';
-
+import { HashLink as Link } from 'react-router-hash-link';
 import { Typography, Grid, Button, Card, CardHeader, CardContent } from '@mui/material';
 
 const Help = () => {
   return (
     <Grid container justifyContent="center">
+      <Grid item xs={12} md={2}>
+        <Card className={styles.help}>
+          <CardHeader
+            className={styles.help_title}
+            title="Jump To:"
+            titleTypographyProps={{ variant: 'h5' }}
+          />
+          <CardContent container="row">
+            <li>
+              <Link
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "react-imask": "^7.4.0",
     "react-router": "^6.22.1",
     "react-router-dom": "^6.22.1",
+    "react-router-hash-link": "^2.4.3",
```
amos-cha, line 39: @EjPlatzer I'm assuming that this is an approved package

src/views/Help/index.jsx
```diff
@@ -1,20 +1,97 @@
 import styles from './Help.module.css';
-
+import { HashLink as Link } from 'react-router-hash-link';
 import { Typography, Grid, Button, Card, CardHeader, CardContent } from '@mui/material';
 
 const Help = () => {
   return (
     <Grid container justifyContent="center">
+      <Grid item xs={12} lg={3} md={12}>
```
amos-cha, line 8:
```suggestion
      <Grid item lg={3} md={12}>
```
xs tag is unneeded

src/views/Help/index.jsx
```diff
@@ -1,20 +1,97 @@
...
+
+          {/* <CardContent>
+            Jump To:
+          </CardContent>
+          <Link className={`gc360_text_link`} style={{ padding: '0 1vw'}}  to="#site-navigation" smooth>Site Navigation</Link>
+          <Link className={`gc360_text_link`} style={{ padding: '0 0.5vw'}} to="#involvement-user-levels" smooth>Involvement User Levels</Link>
+          <Link className={`gc360_text_link`} style={{ padding: '0 0.5vw'}} to="#management-&-editing-involvements" smooth>Management & Editing Involvements</Link>
+          <Link className={`gc360_text_link`} style={{ padding: '0.5vw'}} to ="#issues-&-troubleshooting" smooth>Issues & Troubleshooting</Link>
+          <Link className={`gc360_text_link`} style={{ padding: '0.5vw'}} to="#supported-platforms" smooth>Supported Platforms</Link>
+          <Link className={`gc360_text_link`} style={{ padding: '0.5vw'}} to="#faq" smooth>FAQ</Link> */}
```
amos-cha, line 91:
```suggestion
```

amos-cha (changes requested):

PR #2288, opened 2024-06-06
S24 help page fixes

Dopey12:
Below I added a jump to side grid, like an anchor, to make it more accessible for users to navigate to a certain section/header on the help page. The more we keep adding to 360 the longer the help page gets so I assumed this could be a good add for user accessibility. I also added a recIM and links paragraph for the site navigation, since those weren't added yet. I don't believe this will be my final result so I am just looking for any feedback, recommendation's to help this look better. #Issue2278

Screenshot from 2024-06-06 10-12-38


Dopey12:
Screenshot from 2024-06-06 10-12-38

src/views/Help/index.jsx
```diff
@@ -1,20 +1,92 @@
 import styles from './Help.module.css';
-
+import { HashLink as Link } from 'react-router-hash-link';
 import { Typography, Grid, Button, Card, CardHeader, CardContent } from '@mui/material';
+import React, { useState, useEffect } from 'react';
+
+const FloatingHashlink = () => {
+  const [floating, setFloating] = useState(false);
+};
```
amos-cha, line 8:
```suggestion
```
this is unused?

src/views/Help/index.jsx
```diff
@@ -1,20 +1,92 @@
...
 
 const Help = () => {
   return (
     <Grid container justifyContent="center">
+      <Grid item xs={12} lg={3} md={12}>
```
amos-cha, line 13:
```suggestion
      <Grid item lg={3} md={12}>
```
xs is unnecessary

amos-cha (changes requested):

src/views/Help/index.jsx
Dopey12: Its needed so that the sidebar can take up the whole width of the screen when on mobile.

amos-cha: right but lg md xs are all tags, since md and xs are the same they don't really do anything. You could probably remove the md tag then if it breaks for mobile without the xs tag

EjPlatzer: @amos-cha yes, removing the `md` will keep the same behavior, but removing the `xs` will change the layout on smaller screens. According to [the MUI Grid docs](https://mui.com/material-ui/react-grid/#basic-grid):
> Column widths are integer values between 1 and 12; they apply at any breakpoint and indicate how many columns are occupied by the component.
>
>A value passed to any given breakpoint also applies to all wider breakpoints (if they have no values explicitly defined). For example, xs={12} sizes a component to occupy the full width of its parent container, regardless of the viewport size.

Dopey12: When I was experimenting with different sizes, there was a certain size where the Jump to did not take the whole size of the page like the way it would with xs, but when i added md it seemed to fix the problem for some reason.

EjPlatzer: Hmm @Dopey12 that would be strange because the `xs={12}` means that it will take up all 12 columns of available space until the you reach a higher screen size that has a smaller breakpoint set, in this case `lg={3}`. The `md={12}` doesn't change anything either way.

Dopey12: I made sure to double check @EjPlatzer, I removed the md and there was no issue with all sizes so in the recent commit i removed it.

amos-cha (approved):
lgtm

PR #2290, opened 2024-06-07
S24 migrate components typescript pt1

dli505:
**Changes made in PR:**
- Began migration of components to Typescript
- So far done (I think):
  - AppRedirect
  - BirthdayMessage
  - DynamicButton
  - Error
  - ErrorBoundary
  - EventList
  - GordonConfetti
  - GordonDialogBox (made some changes in future commits)
  - GordonLimitedAvailability
  - GordonOffline
  - GordonTooltip
  - GordonUnauthenticated


src/components/Header/index.tsx
```diff
@@ -111,8 +124,9 @@ const GordonHeader = ({ onDrawerToggle }) => {
           className={styles.tab}
           icon={icon}
           label={name}
-          component={ForwardNavLink}
-          to={route}
+          component={forwardRef<any, Omit<RouterLinkProps, 'to'>>((props, ref) => (
+            <NavLink ref={ref} to={route} {...props} />
+          ))}
```
EjPlatzer, line 129: I think you can just use plain old `NavLink` here, although I'm not sure:
```suggestion
          component={NavLink}
          to={route}
```

src/components/Loader/index.tsx
```diff
@@ -2,11 +2,11 @@ import { CircularProgress, Grid } from '@mui/material';
 import styles from './Loader.module.css';
 
 const GordonLoader = ({ size = 100, color = 'primary', inline = false }) => {
-  if (inline) return <CircularProgress size={size} color={color} />;
+  if (inline) return <CircularProgress size={size} color={'primary'} />;
```
EjPlatzer, line 5: This is a regression for the GordonLoader, since the color prop is now ignored and the loader's color is now just always 'primary'.

src/components/Nav/components/NavLinks/index.tsx
```diff
@@ -46,10 +50,10 @@ const GordonNavLinks = ({ onLinkClick }) => {
     }
     return (
       <GordonDialogBox
-        open={dialog}
-        onClose={() => setDialog(null)}
+        open={dialog ? true : false}
```
EjPlatzer, line 53:
```suggestion
        open={Boolean(dialog)}
```

EjPlatzer (changes requested):
This is coming along great! I have a couple of questions and a couple minor suggestions.

src/components/Loader/index.tsx
EjPlatzer: Since you aren't declaring the types of the props for `GordonLoader`, typescript is doing its best to infer them. In this case, it infers the type of `color` to be string because as far as it's aware you can pass any kind of string in as the color. However, the `CircularProgress` component is much more strict about the type of it's `color` prop. So typescript is warning you that `GordonLoader`'s color could be any string whatsoever but `CircularProgress` only accepts a narrow subset of all possible strings, which could cause an error.

The fix, in this case, is to be explicit about the types of `GordonLoader`'s props. You can add a `Props` type like you have in most other components. Rather than trying to recreate the logic of `CircularProgress`'s props types, you can just use those props and add the `inline` prop that is specific to `GordonLoader`:
```typescript
  type Props = CircularProgressProps & {
      inline?: boolean
  }
```

dli505: It worked, thank you!

src/components/Header/index.tsx
```diff
@@ -12,7 +12,13 @@ import RecIMIcon from '@mui/icons-material/SportsFootball';
 import GordonDialogBox from 'components/GordonDialogBox';
 import { useNetworkStatus } from 'hooks';
 import { forwardRef, useEffect, useState } from 'react';
-import { useNavigate, NavLink, useLocation } from 'react-router-dom';
+import {
+  useNavigate,
+  NavLink,
+  useLocation,
+  Link as RouterLink,
```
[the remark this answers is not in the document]
dli505: Removing them seems to give a lot of errors elsewhere

EjPlatzer:
@dli505 I see that there are some Conflicting files, which will prevent this PR from being merged. Can you resolve these conflicts?

src/components/Header/index.tsx
```diff
@@ -12,7 +12,13 @@ import RecIMIcon from '@mui/icons-material/SportsFootball';
 import GordonDialogBox from 'components/GordonDialogBox';
 import { useNetworkStatus } from 'hooks';
 import { useEffect, useState } from 'react';
-import { useNavigate, NavLink, useLocation } from 'react-router-dom';
+import {
+  useNavigate,
+  NavLink,
+  useLocation,
+  // Link as RouterLink,
```
EjPlatzer, line 19: Please remove this commented out import.

src/components/Header/components/QuickSearch/index.tsx
```diff
@@ -80,9 +80,9 @@ const performSearch = debounce(
 
 type Props =
   | {
-      disableLink: true;
-      customPlaceholderText: string;
-      onSearchSubmit: (person: SearchResult) => void;
+      disableLink?: true;
+      customPlaceholderText?: string;
+      onSearchSubmit?: (person: SearchResult) => void;
```
EjPlatzer, line 85: Why did these get marked as optional? `customPlaceholderText` and `onSearchSubmit` are required when `disableLink` is `true`.

EjPlatzer (changes requested):
This is looking good! One nit and one question.

src/components/Header/components/QuickSearch/index.tsx
dli505: image
When I mark them as required this line of code complains, is there a way to resolve this nicely?

EjPlatzer: The solution is to mark them as optional in the `disableLink: false` case of the union (and also mark `disableLink` itself as optional). This tells TypeScript that the props are not required to be set in the case that `disableLink` is either `false` or `undefined`. I guess typescript doesn't allow a prop that can only be `undefined` to not be passed, you have to explicitly mark that prop as optional as well.

The other way to handle this would be to remove the two undefined props from the case where `disableLink` is false or undefined. However, if we do that, we can't destructure those props, since they only exist when `disableLink` is true. We could handle that by spreading the props other than `searchFunction` into a `props` object:
```typescript
const QuickSearch = ({searchFunction, ...props}: Props) => {...}
```
And then we'd have to check `props.disableLink` before referencing the other two props. e.g. 
```javascript
const linkProps = props.disableLink
      ? { onClick: () => props.onSearchSubmit!(person) }
      : {
          component: Link,
          to: `/profile/${person.UserName}`,
        };
```

Personally, I think it's easier to just mark the other two props as optional when `disableLink` isn't `true`.

EjPlatzer (approved):
LGTM!

Issue #2022, opened 2023-07-14
Collapse "Membership Requests" when there's nothing to do


russtuck:
In the Involvements tab, it would be better to collapse "Membership Requests" when there is nothing to do, so it doesn't waste space, and also so it's more noticeable when there IS something that requires action.  

If it's simpler to only collapse it when it's actually empty, that's almost as good, since people can delete things that are completed.

PR #2292, opened 2024-06-07
collapse membership requests when nothing to do

Dopey12:
Resolves Issue #2022 , when there is no membership request sent, it is automatically closed but when there is a request sent it is automatically open.


EjPlatzer (approved):
LGTM!

PR #2293, opened 2024-06-07
Streamline typing and composability of DynamicButton

EjPlatzer:
The DynamicButton component is a thin wrapper around MUI's `Button`, so I made it behave like one, in that it accepts all `Buton` props and forwards them to the underlying `Button`.


dli505 (approved):
lgtm!

EjPlatzer:
There was an error introduced in #2290 with `GordonOffline` that was causing this PR to fail the linting checks, so I fixed that too.

PR #2295, opened 2024-06-10
S24 Reduce Startup Console Errors

Luke-W-Hart:
Each of these error screenshots should line up with the order of the files that were changed.

Screenshot from 2024-06-10 09-07-35

Screenshot from 2024-06-10 09-07-56

Screenshot from 2024-06-10 09-31-54
The "sx" component on line 151 was unnecessary and still behaves the same when you remove it.


EjPlatzer (approved):
LGTM!

Issue #2132, opened 2023-10-25
Accessibility issues


russtuck:

Luke-W-Hart:
@EjPlatzer for the 2nd point (Disabled tabs don't have enough contrast), would it be best to not have the disabled tabs (People, Rec-IM) on the header when the user is not signed in?
I'm not sure whether there is a reason they are still displayed when they are disabled, just wanted to check.

Luke-W-Hart:
@EjPlatzer for the 2nd point (Disabled tabs don't have enough contrast), would it be best to not have the disabled tabs (People, Rec-IM) on the header when the user is not signed in?
I'm not sure whether there is a reason they are still displayed when they are disabled, just wanted to check.

I also was wondering because of the guest transcript we have that doesn't have a button anywhere, so maybe we could put it on the header, replacing people or Rec-IM?


![Image](https://github.com/gordon-cs/gordon-360-ui/assets/123050501/1c2f4d6e-4399-4021-aa2c-0af4ed766809)

EjPlatzer:
@Luke-W-Hart I believe the intent of showing the disabled tabs is two-fold:
1. Reduce confusion for users who do not realize they are not logged in. For example, if a current student who has an active Gordon account opens 360 but does not realize they aren't logged in, they would see the disabled tabs and if they tried to click one, be asked to login. Whereas if the tabs simply didn't appear, the student might instead assume that something was broken (or perhaps had changed) on 360, and not immediately realize that they aren't logged in.
2. Preview the uses of 360. If a guest user (e.g. a prospective student) accesses the site, they can see the disabled tabs to know what 360 can be used for, but they can't access the content that should only be available to members of campus.

I am not sure if those reasons are valid reasons to keep the disabled tabs behavior, but as far as I'm aware that's why the decision was initially made.

As for the guest transcript - one place we could put that is under the profile button that currently links to `myprofile`, assuming we did want to keep the disabled tabs.

Luke-W-Hart:
@EjPlatzer I guess we'll keep the disabled tabs unless you think otherwise.

This is the color they would have to be in order to have enough contrast. The problem is now the disabled and enabled tab colors don't look very different.

![Image](https://github.com/gordon-cs/gordon-360-ui/assets/123050501/eb428ad6-d25f-4389-8824-2e1e4a41bf9a)
This would also make the dark mode tabs look disabled instead of all being the same color


[...]

PR #2296, opened 2024-06-10
Make Top Tabs Accessible

Luke-W-Hart:
Reference #2132 
Top navigation tabs were inaccessible via keyboard, now they work like before...
- Hitting tab once focuses the page tabs, and then you can use the arrow keys to move focus within the tabs. 
- Hitting tab again jumps focus to the people search box. 
- Shift+tab will take you back to the tabs.

This does not break any of the other page tab functionality (still highlights and underlines correct icon when on the corresponding page).


amos-cha (approved):
lgtm

PR #2297, opened 2024-06-10
S24 Update to REACT 18.x.x

amos-cha:
The React update to 18.x.x has been something queued up by @EjPlatzer primarily bottle necked by outdated packages used by Rec-IM as well as changes usage of react-chart-js2 which now uses objects formatted loosely to look like those of chart-js.Temporarily:
- Hall Rankings (Rec-IM) has been disabled
- Brackets (Rec-IM) has been disabled

This PR contains updates to all outdated packages as well as the update to React 18 which will enable me to find a suitable new package for Rec-IM's brackets.

I also had to update the usage of react-chartjs-2 on victory promise's Polar Area chart which no longer supports callbacks the way we used to handle labels. My solution was to handle all calculations within the label which seems to be exist in the outermost layers of the JSON object passed to chartjs 2. This is definitely more of a 'bandage' solution. However, I do not know if it is worth putting more effort into such an minor visual component of 360.

### OLD:
image

### NEW:
image


russtuck (commented):
I know this change isn't creating the situation of specifying so many patch levels in package.json, but it would be nice to try to fix them, if possible, while we're doing the update.  Are there any that you know really are needed?

The other changes look good to me, but I'm not familiar with these charting packages.

package.json
```diff
@@ -4,64 +4,63 @@
...
-    "@mui/icons-material": "^5.15.10",
-    "@mui/material": "^5.15.10",
-    "@mui/x-date-pickers": "^6.19.4",
-    "add-to-calendar-button": "^2.5.10",
-    "caniuse-lite": "^1.0.30001589",
-    "chart.js": "^2.9.4",
-    "cropperjs": "^1.6.1",
-    "date-fns": "^3.3.1",
+    "@azure/msal-browser": "^3.16.0",
+    "@azure/msal-react": "^2.0.18",
```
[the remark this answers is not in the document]
EjPlatzer: Thanks for thinking about this @russtuck. To answer your first question, our dependencies have specified patch versions because that is how `npm` installs them.

For example, if I run `npm install react` in a brand new project, `npm` puts the following in my `package.json`:
```json
{
  "dependencies": {
    "react": "^18.3.1"
  }
}
```

So you can see that `npm`s default behavior is to install the package specified to the latest patch, but with a Caret Range specifying that the package can be automatically updated to any version that isn't expected to be a breaking change. This is (almost) equivalent to `react: 18`. The one exception is that Caret Ranges handle pre-1.0 packages more carefully.

package.json
```diff
@@ -4,64 +4,63 @@
...
-    "@emotion/styled": "^11.10.8",
-    "@mui/icons-material": "^5.15.10",
-    "@mui/material": "^5.15.10",
-    "@mui/x-date-pickers": "^6.19.4",
-    "add-to-calendar-button": "^2.5.10",
-    "caniuse-lite": "^1.0.30001589",
-    "chart.js": "^2.9.4",
-    "cropperjs": "^1.6.1",
-    "date-fns": "^3.3.1",
+    "@azure/msal-browser": "^3.16.0",
```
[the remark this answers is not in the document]
EjPlatzer: @russtuck I think you may be missing the nuance of how caret ranges handle `0.x` versions. `node-semver` is not at odds with semver.org, and actually caret ranges are slightly looser with presumed non-breaking changes than what semver.org specifies.

Specifically, semver.org says that:
> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Whereas node-semver says of it's handling of caret ranges (emphasis mine):
> Allows changes that do not modify the left-most non-zero element in the [major, minor, patch] tuple. In other words, this allows **_patch and minor_** updates for versions 1.0.0 and above, **_patch updates for versions 0.X >=0.1.0_**, and no updates for versions 0.0.X.
>
>Many authors treat a 0.x version as if the x were the major "breaking-change" indicator.

Notice that, for example, the caret range `^18.3.1` would allow upgrading `18.3.5` or `18.4.0`, which agrees with semver.org (and which you may have been misunderstanding). Furthermore, the caret range `^0.4.3` would allow upgrading to `0.4.4`, whereas according to semver.org "Major version zero (0.y.z) is for initial development. Anything MAY change at any time.".

I think the crucial point is that, as `node-semver` points out, "Many authors treat a 0.x version as if the x were the major "breaking-change" indicator", so caret ranges protect you from upgrading `0.4.3` to `0.5.0`, because `0.4 -> 0.5` is expected to have breaking changes.

Overall, I would say that npm's default way of specifying package versions is helpful in stating which package was intentionally installed (i.e. at the time this was last carefully reviewed and manually upgraded, we installed "18.3.1"), while allowing upgrades to non-breaking versions automatically (and we currently have "18.5.1" installed).

Does that address your concerns @russtuck?

src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
```diff
@@ -18,60 +19,57 @@ const GraphDisplay = ({ scores }: Props) => {
    */
   const emptySliceValue = minimumScore * 0.6;
 
+  Chart.register(RadialLinearScale, ArcElement, Tooltip, Legend);
+  Chart.defaults.plugins.tooltip.callbacks.label = () => '';
+  Chart.defaults.plugins.tooltip.titleMarginBottom = 0;
```
EjPlatzer, line 24: In my testing, both of these are unnecessary.

src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
```diff
@@ -18,60 +19,57 @@ const GraphDisplay = ({ scores }: Props) => {
    */
   const emptySliceValue = minimumScore * 0.6;
 
+  Chart.register(RadialLinearScale, ArcElement, Tooltip, Legend);
```
EjPlatzer, line 22: This can be moved outside the component so that the plugins are only registered once.

EjPlatzer: You can disable the Legend plugin and not have to specify in the options that it shouldn't display anything.
```suggestion
  Chart.register(RadialLinearScale, ArcElement, Tooltip);
```

src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
```diff
@@ -18,60 +19,57 @@ const GraphDisplay = ({ scores }: Props) => {
...
+        display: false,
+      },
+    },
+    scales: {
+      r: {
+        display: false,
+      },
+    },
+    layout: {
+      padding: 30,
```
EjPlatzer, line 68: I think this is more padding than necessary:
```suggestion
      padding: 10,
```

src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
```diff
@@ -18,60 +19,57 @@ const GraphDisplay = ({ scores }: Props) => {
...
-      backgroundColor: colors,
-      borderWidth: 3,
+  const dataOBJ = {
+    labels: labels,
+    datasets: [
+      {
+        data: data,
+        backgroundColor: colors,
+        borderWidth: 3,
+        borderColor: '#fff',
```
EjPlatzer, line 50: In my testing, this was not needed.
```suggestion
```

EjPlatzer (changes requested):
I left several comments about the victory promise graph, but they are all addressed by #2304.

I did have one important question about `package.json`.

Also, I noticed that you didn't commit any changes to `package-lock.json`. When you upgrade packages, you need to include `package-lock.json` so that all future devs (including yourself in the future) can install the versions of our dependencies that you installed and tested against.

package.json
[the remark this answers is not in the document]
amos-cha: I just updated by using a package updater. I can probably just make it `3.x.x` or whatever. Solid find

src/components/Profile/components/VictoryPromise/components/GraphDisplay.tsx
amos-cha: yeah I realized that yesterday after I left, default is white

package.json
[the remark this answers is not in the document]
amos-cha: I'm currently not in office but I have all the changelogs pulled up. I could update just the 4 major ones, but I felt that we might as well just update all of them. If you disagree, I'll just make the simple update.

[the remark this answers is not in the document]
EjPlatzer: When it comes to upgrading dependencies, even minor and even sometimes patch versions, the best practice is "trust but verify".

In other words, if you want to upgrade all of these, you should look through all of the release notes/changelogs for each version change to see if any breaking changes were made, if anything was deprecated, if there are any big changes we want to take advantage of, etc.

I think that belongs in a different pull request, both to keep this one focused on one specific major upgrade, and also so you don't have to do all that just to get this change completed.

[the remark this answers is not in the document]
amos-cha: As far as I can tell, there are no deprecations. But there was a single tag change on @emotions:
image

When I get back to office I'll revert the other package updates and test. Then we can get a merge in tomorrow most likely

package.json
```diff
@@ -13,25 +13,24 @@
...
     "react-cropper": "^2.3.3",
     "react-csv": "^2.0.3",
     "react-device-detect": "^2.2.3",
-    "react-dom": "^17.0.2",
+    "react-dom": "^18.3.1",
     "react-dom-confetti": "^0.2.0",
     "react-dropzone": "^14.2.3",
     "react-ga4": "^2.1.0",
-    "react-icons": "^5.0.1",
+    "react-icons": "^5.2.1",
```
EjPlatzer, line 33: Was this upgrade required for React 18?

package.json
```diff
@@ -44,12 +43,12 @@
...
     "@types/luxon": "^3.4.2",
     "@types/react-big-calendar": "^1.8.8",
-    "@types/react-dom": "^18.2.19",
+    "@types/react-dom": "^18.3.0",
     "@types/react-router-dom": "^5.3.3",
     "@vitejs/plugin-react": "^4.2.0",
     "eslint-config-react-app": "^7.0.1",
     "eslint-plugin-import": "^2.29.1",
-    "eslint-plugin-jsdoc": "^48.2.0",
+    "eslint-plugin-jsdoc": "^48.2.9",
```
EjPlatzer, line 51: Was this upgrade required for React 18?

EjPlatzer (approved):
A couple of (I believe) unnecessary upgrades were left in, but I don't think ever of them will matter.

package.json
amos-cha: missed one

package.json
amos-cha: I looked at the changelog and it just looks like all positive things that were updated with no name changes.

Issue #2287, opened 2024-06-06
back-to-back quad courses wrap in course schedule


jsenning:
There are some formatting issues when two back-to-back quad courses appear in the same timeslot on the schedule; here is an example:

![Image](https://github.com/gordon-cs/gordon-360-ui/assets/6454529/f1b902d8-384e-4664-8a6e-a578148cee10)

Maybe we need to change the font size?  Also, while we're looking at this, is there some way to better distinguish the course code from the building/room number?

Hls5235:
I actually just started working on this, however I believe the course code and room number issue that your talking about actually won’t be a problem once this is fixed and correct me if I’m wrong, MAT 227 is the title and KOSC is the building but the room code actually isn’t on the screen and that’s why it’s hard to distinct the two.

jsenning:
MAT 227 should be in the same room as MAT 226.  I think the program displays "MAT 227" but it wraps so  that when it tries to display "KOS 126" we get just the "KOS" but "126" would be below that if the schedule card was larger.  You could maybe test this by changing the hours we show courses for (i.e. start at 12:00pm rather than 8:00am).

Just a thought, but we often write course codes without a space, so maybe using "MAT226" and "MAT227" would help.  Maybe also reducing the font size would help.

That said, The problem could be worse - if the course has a section code or lab code.   The example that comes to mind is BUS105FA and BUS105FB (look at Kent Siebert's fall teaching schedule).

PR #2298, opened 2024-06-10
Schedule visual bugs s24

Hls5235:
There were issues with the text display for quad classes and I was able to fix it by having the size of the text shrink when the screen gets smaller. I also have the location display on a new line and instead of it showing 'ONLN ASY' we decided it would be easier to understand if it just said 'ASYNC' and it fit on the screen better. I also eliminated the spaces between the course and course code, MAT 210 -> MAT210. This is now more consistent with how it is printed everywhere else throughout Gordon. I also added an else if to account for courses like Discovery that have a room number and location of null. Amos previously mentioned that null refers to an object, which is true but the best way I found to prevent it from printing 'null null' was the else if statement that I added.
#2287

### DIFF
--- a/src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/ScheduleCalendar.scss
+++ b/src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/ScheduleCalendar.scss
@@ -68,12 +68,6 @@ button.rbc-input::-moz-focus-inner {
   max-width: 100%;
 }
 
-@media (max-width: 1152px) {
-  .rbc-row-segment .rbc-event-content {
-    font-size: 70%;
-  }
-}
-
 .rbc-day-slot .rbc-event-content {
   background-color: var(--mui-palette-primary-main);
   color: var(--mui-palette-primary-contrastText);
@@ -86,6 +80,7 @@ button.rbc-input::-moz-focus-inner {
   min-height: 1em;
 }
 @media (max-width: 1152px) {
+  .rbc-row-segment .rbc-event-content,
   .rbc-day-slot .rbc-event-content {
     font-size: 70%;
   }
--- a/src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/ScheduleCalendar.scss
+++ b/src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/ScheduleCalendar.scss
@@ -68,18 +68,28 @@ button.rbc-input::-moz-focus-inner {
   max-width: 100%;
 }
 
+@media (max-width: 1152px) {
+  .rbc-row-segment .rbc-event-content {
+    font-size: 70%;
+  }
+}
+
 .rbc-day-slot .rbc-event-content {
   background-color: var(--mui-palette-primary-main);
   color: var(--mui-palette-primary-contrastText);
   font-size: 90%;
-  width: 60%;
+  width: 75%;
   flex: 1 1 0;
-  word-wrap: break-word;
+  word-wrap: normal;
   line-height: 1;
   height: 100%;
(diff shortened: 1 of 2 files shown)


src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx
```diff
@@ -3,6 +3,7 @@ import { CourseEvent, Schedule, scheduleCalendarResources } from 'services/sched
 import './ScheduleCalendar.css';
 import { DateTime } from 'luxon';
 import { useWindowSize } from 'hooks';
+import { isNull } from 'lodash';
```
amos-cha, line 6:
```suggestion
```

unused import

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/ScheduleCalendar.scss
```diff
@@ -68,18 +68,28 @@ button.rbc-input::-moz-focus-inner {
   max-width: 100%;
 }
 
+@media (max-width: 1152px) {
+  .rbc-row-segment .rbc-event-content {
+    font-size: 70%;
+  }
+}
+
```
amos-cha, line 76:
```suggestion
```

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/ScheduleCalendar.scss
```diff
@@ -68,18 +68,28 @@ button.rbc-input::-moz-focus-inner {
...
   line-height: 1;
   height: 100%;
   min-height: 1em;
 }
-
+@media (max-width: 1152px) {
+  .rbc-day-slot .rbc-event-content {
+    font-size: 70%;
+  }
+}
```
amos-cha, line 92:
```suggestion
@media (max-width: 1152px) {
  .rbc-row-segment .rbc-event-content,
  .rbc-day-slot .rbc-event-content {
    font-size: 70%;
  }
}
```

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx
```diff
@@ -21,17 +22,21 @@ const GordonScheduleCalendar = ({ schedule, onSelectEvent }: Props) => {
...
     let title;
+
     if (course.location.includes('ASY')) {
-      title = `${course.title}${width >= 1500 ? ' | ' : ' '} ${course.location}`;
+      title = `${course.title.replaceAll(' ', '')} | ASYNC`;
+    } else if (course.location.includes('null')) {
+      title = `${course.title.replaceAll(' ', '')}`;
     } else {
-      title = `${course.title}${width >= 2212 ? ' | ' : ' '} ${course.location}`;
+      title = `${course.title.replaceAll(' ', '')} \n${course.location}`;
```
amos-cha, line 31: I think this can most elegantly be handled with a switch case. 
https://www.w3schools.com/js/js_switch.asp
This will also allow for trickle down processes like ``${course.title.replaceAll(' ', '')}`` that is being used in both your else if and else statements

amos-cha (changes requested):

src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx
Hls5235: We talked yesterday in class and decided it would be best done using ternary operators. The error we found has also been fixed by adding tempTitle instead of having title equal something.

amos-cha (approved):
lgtm

PR #2299, opened 2024-06-11
Improve Login/Home Page Accessibility

Luke-W-Hart:
Reference #2132 
- Login main image now has alt text.
- Screen reader now more clearly states whether you are on the home page when 360 logo is focused on.
- Other tabs on the header and in hamburger already indicate whether you are on a page and which page you are on.


jsenning (commented):
Nice work.  I wonder if "Home Page" might be better than "You are on the Home Page".

jsenning (approved):
LGTM

PR #2304, opened 2024-06-11
Victory Promise Improvements

EjPlatzer:
I cleaned up the victory promise graph slightly, and fixed the `minimumScore` so that empty scores are again proportional to the lowest non-empty score.


amos-cha (approved):
lgtm, thanks!

PR #2309, opened 2024-06-13
Fix Disabled Top Tab Contrast

Luke-W-Hart:
Reference #2132 
When someone is not logged in/authenticated the disabled top tabs (People and Rec-IM) now pass contrast tests. This is what they look like now...
Screenshot from 2024-06-10 16-24-48
When on dark mode tabs were the same color, now they look like this...
Screenshot from 2024-06-12 11-24-49

The difference isn't fairly obvious from the enabled tabs but it at least passes the contrast checker.

### DIFF
--- a/src/components/Header/Header.module.scss
+++ b/src/components/Header/Header.module.scss
@@ -18,7 +18,7 @@
   flex-grow: 1;
 
   &.disabled_tab {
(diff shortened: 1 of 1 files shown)


jsenning (approved):
LGTM

PR #2310, opened 2024-06-17
Release: June 17th, 2024

amos-cha:
These PRs will go into production:


EjPlatzer (approved):
I have reviewed these changes and everything looks good to me. Technically, this release is a regression in terms of RecIM features because we have disabled the affiliations leaderboard as well as the bracket view in #2297. I think this is okay since neither feature is vital and RecIM is not seeing a lot of use over the summer.

Issue #2289, opened 2024-06-07
Fix warnings and errors in console at startup


russtuck:

Luke-W-Hart:
#2295

russtuck:
All gone except one error about top tab values (repeated 5 times).  Fixing this would be complicated, so declaring victory for now and moving on.

PR #2313, opened 2024-06-17
Reduce Console Errors

Luke-W-Hart:
Reference #2289 
Addresses this error...
Screenshot from 2024-06-17 15-41-44

Navigation by the header tabs and side nav links still works, but I'm not sure if removing the innerRefs breaks anything else.


src/components/Header/index.jsx
```diff
@@ -27,7 +27,7 @@ const headerLogo64dpi = 'images/gc_' + angleMode + '_yellow_logo_64.png';
 const headerLogo56dpi = 'images/gc_' + angleMode + '_yellow_logo_56.png';
 const headerLogo56dpiNoText = 'images/gc_' + angleMode + '_yellow_logo_56_vert.png';
 
-const ForwardNavLink = forwardRef((props, ref) => <NavLink innerRef={ref} {...props} />);
+const ForwardNavLink = forwardRef((props, ref) => <NavLink {...props} />);
```
EjPlatzer, line 30: I think this became unnecessary when we upgraded react-router past version 5.1, but no one has removed the unnecessary ref forwarding.

At this point, `ForwardNavLink` is at best useless and at worst silently swallowing the `ref`, so we should get rid of it and just pass a plain old `NavLink` where it was used.

src/components/Nav/components/NavAvatar/index.jsx
```diff
@@ -42,7 +42,6 @@ const GordonNavAvatar = ({ onLinkClick }) => {
   const buttonLink = forwardRef((props, ref) => (
     <Link
       {...props}
-      innerRef={ref}
       to={profile ? `/myprofile` : '/'}
```
EjPlatzer, line 45: As above, we should just get rid of `forwardRef` here and use a plain old `Link`.

EjPlatzer (changes requested):
Nice work! Thanks for doing this useful maintenance. I think this has knock on opportunities to simplify the code even further.

Luke-W-Hart:
Should be good now, and also added a key to the event header columns to resolve this error...
Screenshot from 2024-06-17 16-27-05

Thanks for the help Evan!

EjPlatzer (approved):
LGTM!

PR #2315, opened 2024-06-18
Rename Activities to Involvements

Luke-W-Hart:
Finishes #261 
Just renames the "activities.ts" file to "involvements.ts", and changes the import path in all corresponding files.


EjPlatzer (approved):
LGTM!

PR #2316, opened 2024-06-18
Guest Transcript Button for Logged Out Only

Luke-W-Hart:
Switch guest transcript button to show only for users who aren't logged in. Was showing for logged in users... (Whoops).


EjPlatzer (approved):
Oops that's my bad :)

Issue #1703, opened 2023-02-26
Update date libraries for birthday logic to use native JS dates


bennettforkner:
I would love to see the birthday logic cleaned up to not use the `Luxon` library. Everything you're doing with these dates can be done really easily with native JS dates:
```typescript
const getBirthdate = async (): Promise<Date> =>
  http.get('profiles/birthdate').then(Date.parse) // Date.parse handles ISO 8601 format natively

const isBirthdayToday = async () => {
  const now = Date.now();
  const birthdate = await getBirthdate();
  return birthdate.getMonth() === now.getMonth() && birthdate.getDate() === now.getDate();
}
```

_Originally posted by @EjPlatzer in https://github.com/gordon-cs/gordon-360-ui/pull/1694#discussion_r1115788438_

PR #2317, opened 2024-06-18
Switched Birthday Logic from Luxon to Native JS

Luke-W-Hart:
Addresses #1703 
getMonth() returns 0-11
getDate() returns 1-31

### DIFF
--- a/src/services/user.ts
+++ b/src/services/user.ts
@@ -256,7 +256,7 @@ const setHomePhonePrivacy = (makePrivate: boolean) =>
 const setImagePrivacy = (makePrivate: boolean) =>
   http.put('profiles/image_privacy/' + (makePrivate ? 'N' : 'Y')); // 'Y' = show image, 'N' = don't show image
 
-const getBirthdate = async (): Promise<Date> =>
+const getBirthdate = (): Promise<Date> =>
   http.get<string>('profiles/birthdate').then((birthdate) => new Date(birthdate));
 
 const isBirthdayToday = (): Promise<boolean> => getBirthdate().then(isToday);
--- a/src/services/user.ts
+++ b/src/services/user.ts
@@ -1,4 +1,3 @@
-import { DateTime } from 'luxon';
 import { Platform, platforms, socialMediaInfo } from 'services/socialMedia';
 import CliftonStrengthsService, { CliftonStrengths } from './cliftonStrengths';
 import http from './http';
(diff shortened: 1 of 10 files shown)


src/services/user.ts
```diff
@@ -256,12 +255,14 @@ const setHomePhonePrivacy = (makePrivate: boolean) =>
 const setImagePrivacy = (makePrivate: boolean) =>
   http.put('profiles/image_privacy/' + (makePrivate ? 'N' : 'Y')); // 'Y' = show image, 'N' = don't show image
 
-const getBirthdate = async (): Promise<DateTime> =>
-  DateTime.fromISO(await http.get('profiles/birthdate'));
+const getBirthdate = async (): Promise<Date> => await http.get('profiles/birthdate');
```
EjPlatzer, line 258: `http.get` returns a string, but you have typed it as returning a Date, which could cause bugs down the line. I think it would be better to convert the string to a proper Date here.

Also, when you await a promise and then immediately return the value of the promise, you can instead remove both `await` and the `async` keyword and just return the original promise directly. As this is currently written, it awaits the promise, unwraps it, and then wraps in another promise. If we get rid of both async and await, it simply returns the original promise.
```suggestion
const getBirthdate = (): Promise<Date> => http.get<string>('profiles/birthdate').then((birthdate) => new Date(birthdate));
```

src/services/user.ts
```diff
@@ -256,12 +255,13 @@ const setHomePhonePrivacy = (makePrivate: boolean) =>
...
 
 const isBirthdayToday = async () => {
-  const birthday = await getBirthdate();
-  return birthday?.month === DateTime.now().month && birthday?.day === DateTime.now().day;
+  const birthday = new Date(await getBirthdate());
+  return (
+    birthday.getMonth() === new Date().getMonth() && birthday.getDate() === new Date().getDate()
+  );
 };
```
EjPlatzer, line 265: Since we have committed to using `date-fns` for out date utilities in 360, it would make sense to use their `isToday function to simplify here:
```suggestion
const isBirthdayToday = async () => {
  const birthdate = new Date(await getBirthdate());
  return isToday(birthdate);
};
```

In fact, you could then fit this onto one line, and more importantly make (in my opinion) more expressive and clear:
```suggestion
const isBirthdayToday = (): Promise<boolean> => getBirthdate().then(isToday)
```

EjPlatzer (changes requested):
This is great! I left a couple comments on opportunities to improve it further.

src/services/user.ts
```diff
@@ -256,13 +256,10 @@ const setHomePhonePrivacy = (makePrivate: boolean) =>
 const setImagePrivacy = (makePrivate: boolean) =>
   http.put('profiles/image_privacy/' + (makePrivate ? 'N' : 'Y')); // 'Y' = show image, 'N' = don't show image
 
-const getBirthdate = async (): Promise<DateTime> =>
-  DateTime.fromISO(await http.get('profiles/birthdate'));
+const getBirthdate = async (): Promise<Date> =>
```
EjPlatzer, line 259: Since this function does not use the `await` keyword, it does not need to be `async`.
```suggestion
const getBirthdate = (): Promise<Date> =>
```

Because `async` functions always return a Promise, even if the function never actually `await`s anything, this is adding an extra Promise object for no benefit.

EjPlatzer (approved):
LGTM!

PR #2321, opened 2024-06-20
Hides Progress Bar on Train

Hls5235:
Train does not retrieve the current or correct dates for the progress bar to work. This means that to test it correctly you need to be on PROD, to minimize clutter and keep train looking nice I added logic to not display it when using train.


jsenning (approved):
LGTM

PR #2323, opened 2024-06-20
Removes schedule for non student/teachers

Hls5235:
For a someone without a schedule like an athletic coach, they will see the schedule tab and a small drop-down menu with unreadable text inside. I removed the schedule tab entirely if the user has had no courses past or present, the schedule will only display on the users profile if they have been a part of at least one course. I also made the title of the schedule tab say 'Course Schedule' instead of just 'Schedule' this will hopefully add a little more clarity for people such as athletes who might be looking for practice times or any other reason.
[#2324](https://github.com/gordon-cs/gordon-360-ui/issues/2324)


jsenning (approved):
LGTM

Issue #2325, opened 2024-06-21
Align the text for each tab on the Profile page for Mobile


Hls5235:
When the screen gets smaller the text for tabs such as 'Schedule' are all aligned differently. Changing theses to be the same alignment would add a little more order and pattern to the profile page.

russtuck:
Done by #2326

PR #2326, opened 2024-06-21
Centers all text when screen gets smaller

Hls5235:
This centers the text in every tab on the profile page when the screen gets smaller. This adds consistency and looks overall cleaner and more uniformed.
[#2325](https://github.com/gordon-cs/gordon-360-ui/issues/2325)


jsenning (commented):
This works as described for me.

jsenning (approved):
LGTM

Issue #2329, opened 2024-06-24
People and Rec-IM Hamburger Tabs Disabled for Logged in Users


Luke-W-Hart:
When a user is logged in, the People and Rec-IM tabs are disabled in the hamburger menu...

![Image](https://github.com/gordon-cs/gordon-360-ui/assets/123050501/3339ad4a-edba-4e60-a360-ea5761dff87c)

You can still navigate to these pages via the header navigation tabs. 
The tabs are still disabled and grayed out when the user is **not** logged in.

PR #2330, opened 2024-06-24
Fix People Rec-IM Tab Error

Dopey12:
When a user is logged in, the People and Rec-IM tabs are disabled in the hamburger menu...
image
Now the user is able to navigate through the hamburger menu for both people and RECIM the rest worked prior to this issue so those were left alone. #2329


jsenning (approved):
LGTM

PR #2332, opened 2024-06-24
Fix Header Tab Keyboard Navigation

### LINKED ISSUE #1390: Fix Keyboard Accessibility for Top Tabs
The non-home tabs on 360 are not accessible via keyboard. Using <tab> to access them fails because focus jumps straight to the search box (and never comes back).  

The arrow keys should also work to move to these tabs.

Luke-W-Hart:
Fixes #1390 
Allows all header tabs to be accessed via keyboard by pressing the tab key. The conversation about this was a while ago, so I'm not sure if we still want every header tab to be accessible by pressing the enter key or if the original functionality (just using the arrow keys to navigate the header tabs) is good. If the original functionality is fine then I will just close the issue referenced.

### DIFF
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -132,7 +132,7 @@ const GordonHeader = ({ onDrawerToggle }: Props) => {
           label={name}
           component={NavLink}
           to={route}
-          tabIndex={name == 'People' ? 5 : 7}
+          tabIndex={0}
         />
       );
     }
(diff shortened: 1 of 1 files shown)


EjPlatzer (changes requested):
[MDN says](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) this about values > 0 for the `tabindex` attribute:
> Warning: You are recommended to only use 0 and -1 as tabindex values. Avoid using tabindex values greater than 0 and CSS properties that can change the order of focusable HTML elements ([Ordering flex items](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items)). Doing so makes it difficult for people who rely on using keyboard for navigation or assistive technology to navigate and operate page content. Instead, write the document with the elements in a logical sequence.

Furthermore, [WebAIM says](https://webaim.org/techniques/keyboard/tabindex#overview) that "`tabindex="1"` (or any number greater than 1) defines an explicit tab or keyboard navigation order. **This must always be avoided**."

If we do want the header tabs to be keyboard navigable, we should use `0` values, rather than positive values,

Luke-W-Hart:
That's good to know, looks like I need to read up on this a little more. :slightly_smiling_face: Thanks Evan!

jsenning (approved):
LGTM

jsenning (approved):
The tabindex > 0 issue has been addressed

PR #2334, opened 2024-06-25
S24 mailbox information

jsenning:
Adjusted API route name from `/api/profiles/mailbox-combination` to `/api/profiles/mailbox-information` since other information than just the combination may be present in the JSON received.  This change is probably not important enough to warrant a change on its own, but matches changes in the API that were made to clean things up while removing an unnecessary thrown exception.

See PR 1056 in gordon-360-api


russtuck (approved):
LGTM

PR #2338, opened 2024-06-27
Paginate Events Page

Luke-W-Hart:
Paginating the Events Page by Date (ie. "This Week", "Two Months", "One Month").
Screenshot from 2024-06-28 10-48-01
Screenshot from 2024-06-28 10-48-13
Screenshot from 2024-06-28 10-48-48
Screenshot from 2024-06-28 10-49-06

### DIFF
--- a/src/services/event.ts
+++ b/src/services/event.ts
@@ -187,20 +187,6 @@ const makeMatchesFilters =
             return true;
           }
           break;
-        case 'This Week':
-          //console.log(new Date((new Date()).setDate(new Date().getDate() + 8)));
-          if (new Date(event.StartDate) <= new Date(new Date().setDate(new Date().getDate() + 7))) {
-            return true;
-          }
-          break;
-        case '2 Weeks':
-          //console.log(new Date((new Date()).setDate(new Date().getDate() + 8)));
-          if (
-            new Date(event.StartDate) <= new Date(new Date().setDate(new Date().getDate() + 14))
-          ) {
-            return true;
-          }
-          break;
         default:
           break;
       }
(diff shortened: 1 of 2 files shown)


Luke-W-Hart:
Current status is that only "This Week" and "2 Weeks" filter, and they only filter when search text box or type filters ("CEC", "Admissions" are updated.

ChrisImagineer:
Brilliant solution to focusing the search for the user!

russtuck:
Would it be better for users overall for the default to be 2 weeks, instead of all?  I think it would, but we might need to make it clear in the results that a time filter was applied.  Otherwise, it could be puzzling and people might struggle to find how to see more events.  Maybe a message at the bottom of the results: "Limited to 2 weeks.  To see more events, open the Filters and change the time filter".  Of course, the "2 weeks" part would need to reflect the current filter value.

When I read "This Week", especially next to "2 Weeks", think from now until Saturday.  But it's actually the next 7 days.  I think it would be clearer and more consistent to call it "1 Week".

In my opinion, we don't think we need so many options, and a smaller dropdown would be better.  How about reducing the choices to "1 Week", "2 Weeks", "1 Month", either 4 or 6 months, and "All"?

russtuck (changes requested):
Please change "This Week" to "1 Week" and shrink the choices a bit.  This is clearly an improvement over now (offering more options without hurting anything).  The code looks reasonable to me.

Changing the default and making the new behavior clear with a message at the bottom isn't required.  It would be a good follow-on change in a separate PR.

russtuck (approved):
LGTM

PR #2340, opened 2024-07-01
remove material-ui, replaced with @mui

amos-cha:
Material-ui/* was deprecated a while ago when updating to MUI5. However, it appears that material-ui/core was not deprecated and with the update to React 18 has now caused package conflicts on new `npm install`s. This PR fully removes material-ui/core and updates remaining @mui/material @mui/styles.


jsenning (approved):
LGTM!

Issue #2342, opened 2024-07-02
Victory Promise Graph Display Bug


Dopey12:
If a user has no points towards their victory promise, the graph wont be displayed when clicking on Show Graph. This seems to have been fixed but due to recent react changes and implementations the chart seems to be off, I was able to fix the display issue but the size is still a little off.

PR #2343, opened 2024-07-02
graph-does-display

Dopey12:
When  a user had zero victory promise points the graph wont display, that is fixed but apparently the graph cannot be changed to its original size, there are two sizes I was able to get and I am curious of your thoughts. #2342 
Screenshot from 2024-07-02 09-41-03
Screenshot from 2024-07-02 09-42-37


jsenning (approved):
As you pointed out there are two options, neither of which is perfect - but better than doing nothing at all.

LGTM.

PR #2344, opened 2024-07-03
Change Past Events Checkbox Color

Luke-W-Hart:
Changed to ScottieCyan so that contrast tests passed.
From this...
Screenshot from 2024-07-03 09-09-01
To this...
Screenshot 2024-07-03 at 9 10 19 AM


amos-cha (commented):
I somehow feel like this should be a global change and not localized

russtuck:
Checkboxes in advanced people search look fine.  Do they have this same code on each one, or do they get it implicitly?  If implicitly, can the same thing be done here?   Otherwise, setting a color to secondary seems reasonable.

russtuck (approved):
LGTM

PR #2352, opened 2024-07-10
Swap the colors for days remaining

yushinj:
swap the day left color with day finished color, so every color layers are counterclockwise.


jsenning (approved):
LGTM

PR #2353, opened 2024-07-10
Return to top of page hashlink-help-page

Dopey12:
The Return to Top of Page works as intended and makes it easier for users to go back to the top of the page after using an anchor to go to certain titles on the Help Page, Only problem is the ident i can't figure it out but I got the spacing out. 
Screenshot from 2024-07-10 15-26-10
Screenshot from 2024-07-10 15-26-27


russtuck (changes requested):
The code looks good, but when I test it and click on "Issues & Troubleshooting", it goes a little too far down, so that header isn't visible.  It doesn't do this on train, so let's try to fix it before pushing.

russtuck:
This only happens occasionally, but not every time and I don't see a reason in the code.  Overall, it's still an improvement.

russtuck (approved):

PR #2031, opened 2023-07-14
User Privacy Settings

ArabellaJi:
Note: This is the new PR for https://github.com/gordon-cs/gordon-360-ui/pull/1931. The reason for creating this new PR is that when I wanted to merge develop into my old branch, I mistakenly did a `git rebase develop` in that branch, which pushed some irrelevant changes and commits made by other people into that old branch. Even though those changes were not harmful, they could still lead to some confusion, so I decided to open this new branch and PR.

Allow users to choose which phone numbers they would like to show in their profiles and some other privacy options.

Fixes https://github.com/gordon-cs/gordon-360-ui/issues/1308
Fixes https://github.com/gordon-cs/gordon-360-ui/issues/1897

API: https://github.com/gordon-cs/gordon-360-api/pull/1042

PR #2355, opened 2024-07-10
S24 privacy combined profile

jsenning:
This PR replaces #2031 and is linked to PR https://github.com/gordon-cs/gordon-360-api/pull/1060 in the API repo.

@ArabellaJi began working on this in 2023 and did most of the heavy lifting. She began with the intention of giving FacStaff a mechanism to optionally display their mobile phone number.  This then expanded to provide privacy controls for other items on the FacStaff personal profile.  Before she handed it off at the start of the summer, she had also expanded her work so that student data (currently mobile phone and home address) used the same mechanisms.

In the past weeks we have continued this work to more completely integrate the new privacy settings and adjusting profile view models to allow individual profile elements have individual boolean values to indicate the user has requested they be private.  Doing this meant that many of the profile fields have changed from strings to objects that contain a string and a privacy flag.  For example, in both the UI and API the mobile phone field is now has type `ProfileItem` that is a class with two fields `value` and `isPrivate`.  The value field is the string representing the phone number, and the `isPrivate` field is a boolean that is true if the should be treated as confidential (and shown in red).  Of course, data the viewing user is not allowed to see is not sent from the API, so the `isPrivate` field being true means that the viewing user is authorized to see the information but the color reminds them that the information is restricted and should be treated as such.

This change deprecates the use of the "KeepPrivate" profile field for faculty staff accounts and the "isMobilePhonePrivate" field for student accounts (although one outstanding issue is the need to ensure that student mobile phone privacy settings here are appropriately reflected in the Jenzabar database.)

Some profile information still use other types of privacy controls.  One example is Clifton Strengths (although this might be an easy fix).  Another is the privacy settings for alumni.  We hope that in the near future these can be modified to use the new privacy settings.


src/components/Profile/components/PersonalInfoList/index.tsx
```diff
@@ -78,52 +76,62 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }: Props)
   const [advisorsList, setAdvisorsList] = useState<AdvisorType[]>([]);
   const [showMailCombo, setShowMailCombo] = useState(false);
   const [graduationInfo, setGraduationInfo] = useState<Graduation | null>(null);
-  const isStudent = profile.PersonType?.includes('stu');
-  const isFacStaff = profile.PersonType?.includes('fac');
-  const isAlumni = profile.PersonType?.includes('alu');
+  const isStudent = checkIsStudent(profile); //profile.PersonType.includes('stu');
+  const isFacStaff = checkIsFacStaff(profile); //profile.PersonType.includes('fac');
+  const isAlumni = checkIsAlumni(profile); //profile.PersonType.includes('alu');
```
russtuck, line 81:
```suggestion
  const isStudent = checkIsStudent(profile);
  const isFacStaff = checkIsFacStaff(profile);
  const isAlumni = checkIsAlumni(profile);
```
Better to delete the old code

src/components/Header/components/QuickSearch/QuickSearch.module.scss
```diff
@@ -2,6 +2,7 @@
 @import '../../../../vars';
 
 .root {
+  color: var(--mui-palette-primary-contrastText);
```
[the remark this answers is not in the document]
jsenning: I agree with you.  I have no idea why this is in the PR!

src/components/Header/components/NavAvatarRightCorner/index.tsx
```diff
@@ -35,7 +35,7 @@ export const GordonNavAvatarRightCorner = ({ onClick }: Props) => {
       <Avatar className={styles.root} src={`data:image/jpg;base64,${image}`} sizes="70px" />
     ) : (
       <Avatar className={styles.root}>
-        {user.profile?.FirstName?.[0]} {user.profile?.LastName?.[0]}
+        {user.profile?.FirstName?.Value[0]} {user.profile?.LastName?.Value[0]}
```
EjPlatzer: If `Value` is `null` or `undefined`, this will throw a TypeError. We need to extend the optional chaining to the indexing to preserve the current behavior:
```suggestion
        {user.profile?.FirstName?.Value?.[0]} {user.profile?.LastName?.Value?.[0]}
```

src/components/Nav/components/NavAvatar/index.tsx
```diff
@@ -36,7 +36,7 @@ const GordonNavAvatar = ({ onLinkClick }: Props) => {
       <Avatar className={`${styles.avatar}`} src={`data:image/jpg;base64,${image}`} />
     ) : (
       <Avatar className={`${styles.avatar} ${styles.placeholder}`}>
-        {profile.FirstName?.[0]} {profile.LastName?.[0]}
+        {profile.FirstName?.Value[0]} {profile.LastName?.Value[0]}
```
EjPlatzer: As above:
```suggestion
        {profile.FirstName?.Value?.[0]} {profile.LastName?.Value?.[0]}
```

vite.config.ts.timestamp-1716474284345-f7f0b86136517.mjs
EjPlatzer, line 1: Not sure why this file was generated or checked in, but it shouldn't be included in the repo.

src/components/Profile/components/PersonalInfoList/UpdateUserPrivacyDropDownMenu/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+const UpdateUserPrivacy = ({ username, fieldList, excludedVisibilityList }) => {
+  const [privacySettingList, setPrivacySettingList] = useState();
+  const [groupList, setGroupList] = useState([]);
+  const [visibleTo, setVisibleTo] = useState('Private'); // default private while loading
+
+  const handlePrivacy = async (event) => {
+    try {
+      await userService.setUserPrivacy({ Field: fieldList, VisibilityGroup: event.target.value });
+    } catch {}
```
EjPlatzer: This empty catch block will silently swallow errors, which is not a great way to handle errors - especially errors updating a user's data in relation to privacy, leading them to think that they changed their privacy settings but they didn't. We should probably tell the user if there was an error (we've been using a toast message for this historically).

src/components/Profile/components/PersonalInfoList/UpdateUserPrivacyDropDownMenu/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+      userService.getPrivacySetting(username).then(setPrivacySettingList);
+    };
+    loadData();
+  }, [username]);
+
+  useEffect(() => {
+    if (privacySettingList) {
+      setVisibleTo(privacySettingList.find((f) => fieldList.includes(f.Field))?.VisibilityGroup);
+    }
+  }, [privacySettingList]);
```
EjPlatzer: This useEffect is only used to update `setVisibleTo` when `userService.getPrivacySetting` returns data. There is no reason that I can see for it to be a separate useEffect, rather than doing this work directly in the `.then()` call of the getPrivacySetting promise (or awaiting the promise to handle it within loadData).

src/components/Profile/components/PersonalInfoList/index.tsx
```diff
@@ -184,38 +197,7 @@ const PersonalInfoList = ({ myProf, profile, isOnline, createSnackbar }: Props)
...
-        isHomePhonePrivate
-          ? 'Personal Info Visible (This change may take several minutes)'
-          : 'Personal Info Hidden (This change may take several minutes)',
-        'success',
-      );
-    } catch {
-      createSnackbar('Privacy Change Failed', 'error');
-    }
-  };
+  }, [isStudent, myProf, graduationInfo, setPlannedGradDate, profPlannedGradYear, createSnackbar]);
```
EjPlatzer: Not sure why this change was included in this PR, but GitHub is rightfully pointing a violation of React's rules of hooks here:
> The 'setPlannedGradDate' function makes the dependencies of useEffect Hook (at line 200) change on every render. To fix this, wrap the definition of 'setPlannedGradDate' in its own useCallback() Hook

EjPlatzer (commented):
I haven't yet reviewed the two most significant files (PersonalInfoList and services/user.ts), but here are some comments for the other changes.

src/components/Header/components/NavAvatarRightCorner/index.tsx
jsenning: Done

src/components/Nav/components/NavAvatar/index.tsx
jsenning: Done.

vite.config.ts.timestamp-1716474284345-f7f0b86136517.mjs
jsenning: Sorry about that -- it's been removed.

src/components/Profile/components/PersonalInfoList/index.tsx
jsenning: Commit message for this change indicated the change was suggested by github.  The change has been reverted.

src/components/Profile/components/PersonalInfoList/UpdateUserPrivacyDropDownMenu/index.jsx
jsenning: Fixed by @EjPlatzer.

src/components/Profile/components/PersonalInfoList/UpdateUserPrivacyDropDownMenu/index.jsx
jsenning: Fixed by @EjPlatzer

src/components/Header/components/QuickSearch/QuickSearch.module.scss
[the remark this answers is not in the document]
jsenning: This change has been removed from the PR.

jsenning:
All known issues have been dealt with (thanks @EjPlatzer).

russtuck (approved):
Approving based on earlier reviews and work.

PR #2318, opened 2024-06-18
S24 migrate components typescript pt2

dli505:
A continuation of https://github.com/gordon-cs/gordon-360-ui/pull/2290
**Changes made in PR:**
- finished migration of Components to Typescript
- files finished:
    - Header
    - Loader
    - Nav
    - NavButton
    - PaletteSwitcherDialog
    - PhotoCropper
    - Profile
    - Providers
    - PWAInstructions
    - Snackbar
    - SpreadsheetUploader

PR #2360, opened 2024-07-11
Revert "S24 migrate components typescript pt2"

amos-cha:
Reverts gordon-cs/gordon-360-ui#2318 (temporary for the presentation)


jsenning (approved):
LGTM

PR #2362, opened 2024-07-12
Accurate Alt Texts for Victory Promise Icons

Luke-W-Hart:
Screen Reader now says accurate categories for victory promise icons rather than just "image".
Screenshot 2024-07-12 at 9 50 42 AM


jsenning (approved):
LGTM

PR #2366, opened 2024-08-09
Fixes bar when between summer and fall semester

Hls5235:
The progress bar didn't know what to do when the current date is between the summer and fall semesters. The issue was that postSummer's start and end dates were accidentally switched, but everything seems to work properly after changing them back.

### DIFF
--- a/src/views/Home/components/DaysLeft/index.jsx
+++ b/src/views/Home/components/DaysLeft/index.jsx
@@ -102,8 +102,8 @@ const DaysLeft = () => {
           name: 'Post-Summer',
           label: 'Fall',
           type: 'Break',
-          start: addDays(termDates.summer?.end, 1),
-          end: addDays(termDates.fall?.start, -1),
+          start: addDays(termDates.fall?.start, -1),
+          end: addDays(termDates.summer?.end, 1),
         },
       };
       const termLoop = [
@@ -115,8 +115,8 @@ const DaysLeft = () => {
         termValues.postSummer,
       ];
 
-      let currentTerm = 'Fall';
-      let currentLabel = 'Fall';
+      let currentTerm = '';
+      let currentLabel = '';
       let currentDaysLeft = '';
       let currentProgress = '';
       let currentType = '';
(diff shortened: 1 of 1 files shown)


src/views/Home/components/DaysLeft/index.jsx
```diff
@@ -102,8 +102,8 @@ const DaysLeft = () => {
           name: 'Post-Summer',
           label: 'Fall',
           type: 'Break',
-          start: addDays(termDates.summer?.end, 1),
-          end: addDays(termDates.fall?.start, -1),
+          start: addDays(termDates.fall?.start, -1),
```
jsenning, line 105: The new logic here doesn't make sense with the way the other breaks are handled.  Why is the start of post summer defined in terms of the start of the fall semester?  The code you had before seems more correct.

Or - if this logic is in fact correct, does this mean that the logic for winter and preSummer should also be changed?

jsenning (changes requested):

src/views/Home/components/DaysLeft/index.jsx
Hls5235: I just noticed the real issue, since the end of the summer and the start of the fall are one day apart, when I add and subtract one from them it essentially switches the two dates. Now I am thinking of either not adding or subtracting for postSummer, but that seems like bugs can appear easily if the dates aren't one day apart. I think the best option would be to add logic that if the days are greater than one day apart then add and subtract accordingly, otherwise don't.

Hls5235: I added logic that if the difference between the two dates is greater than 1 it would properly add and subtract the dates, otherwise, it will have them remain as they are and display one day left until fall term when on that day.

jsenning (changes requested):
There are two issues here that I think are important.

First, as I recall, the sessions that 360 uses are actually based on Housing rather than Academic sessions.  It looks like 360 is finding that the summer term ended on August 9, 2024 and the fall term began on August 10, 2024.  While we can't do anything about that now, it seems to me that this component would benefit from using the actual academic session start/end dates.  Of course, this means we'd need a way to find them...

Second, I think the original code for computing the start and end of postSummer session are okay and will work correctly as long as only the date portion of the `today` date object are used.  `today = new Date()` returns the current date & time.  This means that at any time more that one second into August 9, `today` is greater than the summer term end date (2024-08-09 00:00:00) but less than the fall term start date (2024-08-10 00:00:00).

I think a better fix is to make `today` variable rather than a const and use `today.setHours(0)`, `today.setMinutes(0)`, and `today.setSeconds(0)` to force the time field to match the timefield in all the term start and end dates.  This means that you'll always be comparing just dates.

PR #2379, opened 2024-10-07
Release: Potentially fix auth redirects, add account selection

### LINKED ISSUE #2370: 360 Site Extreme Loading Time and Loading Loop Issue
I looked around the project board, and didn't see an issue for this already, so I figured I'd make one.  This is a known major usability and performance issue, and I've personally heard many students complain about it, as well as experienced it myself on occasion.

**Anecdotes:**
A few choice quotes from students on why this is a serious UX issue:
"360 is so broken"
"why does it take so long to load"
"that spinny thing is so annoying, it just keeps popping up over and over again"

**The Issue:**
When first loading the home page, often the circle loader will appear and disappear rapidly many times in a row before the page finally loads.  I've seen (on other student's devices, usually phones) it happen for up to 10 seconds before finally loading the page.  It only happens sometimes, which makes it difficult to reproduce reliably, but it's often enough to create the impression (anecdotally) that the site doesn't work very well.  I've heard this sentiment from many students now.

**Solution Thoughts:**
Obviously I don't know what the cause is, and don't want to speculate.  It's potentially a difficult issue to fix since it's so hard to reproduce.  I might try to do some testing to see if I can find a setup to reliably (or at least somewhat reliably) reproduce the issue.

I do know that on the user experience side, this is made worse by the addition of the animated circle loader.  Before I implemented that during Summer Practicum two summers ago, a simple blank screen would show during that loading period (the loader is pure html/css which is displayed before the React components mount).  The issue may have been present prior to that summer, but was considerably less noticeable and obnoxious in lieu of the loader's repeating appearance.

Any thoughts or input anyone has would be appreciated, as fixing this would greatly improve the UX in my opinion.  Also, if anyone is able to reliably recreate this (or has it happen to them more often than it does for me), methodology and screen recording of the issue would be useful.

EjPlatzer:
This release includes PR #2377, which may fix #2370 and adds some temporary logging of authentication events as well as letting users select which Microsoft account they want to use for 360 when they have multiple active accounts.


bennettforkner (approved):
LGTM

PR #2386, opened 2024-10-08
Added Several Links on Safety Page

Luke-W-Hart:
Updated
Screenshot 2024-10-10 at 12 36 16 PM

### DIFF
--- a/src/views/CampusSafety/index.tsx
+++ b/src/views/CampusSafety/index.tsx
@@ -26,12 +26,28 @@ const CampusSafety = () => {
               titleTypographyProps={{ variant: 'h4' }}
             />
             <CardContent>
+              <Typography variant="h5" align="center">
+                Emergency?
+              </Typography>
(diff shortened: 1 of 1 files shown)


src/views/CampusSafety/index.tsx
```diff
@@ -26,7 +26,15 @@ const CampusSafety = () => {
               titleTypographyProps={{ variant: 'h4' }}
             />
             <CardContent>
-              <Typography variant="body1">Gordon Police Resources</Typography>
+              <Link
+                href="https://www.gordon.edu/police"
```
Centerville1, line 30: Before we commit this, do you also want to add the police phone numbers to this card, maybe add some simple styling to it, but don't worry about getting it to look 100% yet.

Centerville1 (changes requested):
LGTM, could you just add the GCPD phone numbers to the police card before we commit, no need to complete the styling, just content would be helpful.

PR #2387, opened 2024-10-08
WMR RD View Setup

Mrand03:
Changes made in PR:

- Changed the housing index.jsx to go to the RD View only if you are faculty ( We can change this when we make a group for RD's). 
- Created an RDView folder within the housing/components folder. This new folder is where all of the RDView components will go as well as where the index.jsx for the RD View is.
- Created a card to edit documents in he RD View which has links to SharePoint and our room range program, as well as a dialog box that pops up when you click the edit button next to "RA/AC Task List". The code for the edit button is in a folder within RDView called components which will hold any components needed for the RD View.
- The code to edit tasks is just the dialog box that pops up, you cannot type anything yet.

### DIFF
--- a/src/views/Housing/components/RDView/updateTasks/index.jsx
+++ b/src/views/Housing/components/RDView/updateTasks/index.jsx
@@ -1,0 +1,27 @@
+import React, { useState } from 'react'; // Import useState for managing open state
+import { IconButton } from '@mui/material';
+import EditIcon from '@mui/icons-material/Edit';
+import GordonDialogBox from 'components/GordonDialogBox';
+
+const UpdateTasks = () => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <div>
+      <IconButton onClick={() => setOpen(true)} size="large" aria-label="Update RA/AC Tasks">
+        <EditIcon fontSize="small" />
+      </IconButton>
+
+      <GordonDialogBox
+        open={open}
+        onClose={() => setOpen(false)}
+        title="Update RA/AC Tasks"
+        buttonName="UPDATE"
+        cancelButtonName="CANCEL"
+        cancelButtonClicked={() => setOpen(false)}
+      />
+    </div>
+  );
+};
+
+export default UpdateTasks;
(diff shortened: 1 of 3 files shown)


src/views/Housing/index.jsx
```diff
@@ -1,9 +1,22 @@
...
+import { AuthGroup } from 'services/auth';
+import RDView from './components/RDView';
 
-const Housing = () => (
-  <Grid Container>
-    <Card></Card>
-  </Grid>
-);
+const Housing = () => {
+  const isFaculty = useAuthGroups(AuthGroup.Faculty);
```
dli505, line 7: I have an `isFacStaff` function that we can import...
`import { Profile as profileType, isFacStaff as checkIsFacStaff } from 'services/user';`
Hopefully this works?

src/views/Housing/index.jsx
```diff
@@ -1,9 +1,22 @@
...
+
+  if (isFaculty) {
```
dli505, line 9: If it does work then this would be `isFacStaff` instead I think

Again not sure if it does work but worth a try? If not then we can use the auth groups

src/views/Housing/components/RDView/updateTasks/index.jsx
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react'; // Import useState for managing open state
+import { IconButton } from '@mui/material';
+import EditIcon from '@mui/icons-material/Edit';
+import GordonDialogBox from 'components/GordonDialogBox';
+
+const UpdateTasks = () => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <div>
```
dli505, line 10: I think in React instead of using divs we just use empty jsx fragments, like
`<></>`

dli505 (changes requested):
Yay Mya this looks really great! I added some quick comments, and if you could paste a screenshot of what the screen looks like that would look really awesome!

Mrand03:
Hi Danya,

Thank you for your suggestions! I changed how we check if the user is faculty to what you suggested and it worked so we can use it. I also changed the div to <> </> and it caused no issues. Let me know if you have any other suggestions!

dli505 (approved):
Looks good to me!

PR #2389, opened 2024-10-13
Create more components for ResidentView

jtasonye:
**Changes made in PR:**

* Created `MyHall` component
* Created `MyRA` component
* Created `OnDuty` component
* Created `Resources` component
* Added all components to the `Housing` index.jsx


dli505 (approved):
LGTM! The use of `<strong></strong>` tags is nagging me at the back of my head (I think...? there's another way to do it in MUI/React) but that's a problem for later :)

PR #2391, opened 2024-10-14
Added Lost and Found Section to Safety Page

Luke-W-Hart:
Button does not navigate anywhere since I need to figure out how to do navigation like the Rec-IM page.
Screenshot 2024-10-13 at 9 54 24 PM

### DIFF
--- a/src/views/CampusSafety/index.tsx
+++ b/src/views/CampusSafety/index.tsx
@@ -1,4 +1,14 @@
-import { Typography, Card, CardContent, CardHeader, Grid, Link } from '@mui/material';
+import {
+  Typography,
+  Box,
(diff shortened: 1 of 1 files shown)


Centerville1 (changes requested):
Actually, I should have mentioned this to you, but I don't want to push any of the lost and found code to this Safety Page Branch, the thought is we can push just the safety page (without the lost and found features) into develop just to go through the process with Evan's review, and catch issues with our coding early.

Let's make a separate shared repo for the lost and found features, feel free to make that branch based on the SafetyPage branch, and then turn this PR to point to that.

Centerville1 (approved):
LGTM

PR #2395, opened 2024-10-19
Lost and found end2 end jones

Centerville1:
Added initial lost and found page for end to end demo.

Screen Shot 2024-10-18 at 16 50 27


Luke-W-Hart:
This is just for adding the page right? We're worrying about the navigation showing up later (Like the RecIM navigation)?

Centerville1:
> This is just for adding the page right? We're worrying about the navigation showing up later (Like the RecIM navigation)?

Correct, I started work on the breadcrumb style navigation (like Rec-IM has), but it only works to get back to the home page currently.  So there's the base work for that, but it'll need more work @Luke-W-Hart .

PR #2396, opened 2024-10-20
wmr - Ra view actual

dli505:
**Changes made in PR:**
- Basic skeleton of some of the RA View!
- A work in progress but there is stuff
- Added component folders for different things


Mrand03:
This looks great Danya, thank you! I will merge it now.

PR #2399, opened 2024-10-23
UI with sample data for Lost and Found Page

richmondnao:
This is a sample code for the Lost and Found UI. This shows sample data that is hard coded into it. Yet to add changes to allow it to pull data from the backend.

### DIFF
--- a/src/views/CampusSafety/views/LostAndFound/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFound/index.tsx
@@ -4,8 +4,8 @@ import EditIcon from '@mui/icons-material/Edit';
 import DeleteIcon from '@mui/icons-material/Delete';
 import { useState, useEffect } from 'react';
 import Header from '../../components/Header';
-import styles from './LostAndFound.module.css';
-import { useTheme } from '@mui/material/styles'; // Import theme
+import styles from './LostAndFound.module.css'; // Import the external CSS
+import { useTheme } from '@mui/material/styles'; // Access theme
 
 // Define the type for Missing Item
 type MissingItem = {
@@ -21,7 +21,7 @@ const LostAndFound = () => {
   const [activeReports, setActiveReports] = useState<MissingItem[]>([]);
   const [pastReports, setPastReports] = useState<MissingItem[]>([]);
   const [loading, setLoading] = useState(true);
-  const theme = useTheme(); // Access theme
+  const theme = useTheme(); // Access theme if needed
 
   // Mock Data for demonstration
   const mockActiveReports: MissingItem[] = [
@@ -103,11 +103,6 @@ const LostAndFound = () => {
                   <span style={{ color: theme.palette.warning.main }}>Gordon</span> Lost and Found
                 </Typography>
               }
-              sx={{
-                backgroundColor: theme.palette.primary.main,
-                color: theme.palette.primary.contrastText,
-                padding: '20px',
-              }}
             />
             <CardContent>
               <Grid container>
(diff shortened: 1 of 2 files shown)


src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -1,25 +1,125 @@
...
+      <Header />
+
+      <Grid container justifyContent="center" marginTop={3}>
         <Grid item sm={10}>
           <Card className={styles.card}>
-            <CardHeader className={styles.title} title="Gordon Lost and Found"></CardHeader>
+            <CardHeader
+              className={styles.title}
+              title={
+                <Typography variant="h4" align="center">
```
Centerville1, line 102: This styling should be done in a separate CSS file using a class name in general, it's better to have the styling in one place rather than inline, makes it easier to find and change later

src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -1,25 +1,125 @@
...
+                  <span style={{ color: theme.palette.warning.main }}>Gordon</span> Lost and Found
+                </Typography>
+              }
+              sx={{
+                backgroundColor: theme.palette.primary.main,
```
Centerville1, line 107: Same thing here, this should be done in a separate CSS file

src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -28,6 +128,194 @@ const LostAndFound = () => {
...
         </Grid>
       </Grid>
+
+      {/* Button to Report a Missing Item */}
+      <Grid container justifyContent="center" marginTop={3}>
+        <Grid item>
+          <Button
+            variant="contained"
+            sx={{
+              backgroundColor: theme.palette.info.main, // Correct theme color
```
Centerville1, line 138: Same here, separate CSS is better style

src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -28,6 +128,194 @@ const LostAndFound = () => {
...
+                Reports
+              </Typography>
+
+              {/* Table Header */}
+              <Grid
+                container
+                spacing={2}
+                sx={{
+                  marginTop: '10px',
+                  backgroundColor: theme.palette.primary.main,
```
Centerville1, line 172: Separate CSS here too

Luke-W-Hart:
Just a comment, looking at the files for the Events page (views and services files) will probably be super helpful as the style and some functionality is the same.

Centerville1 (changes requested):
Overall looks good, but one thing to change is to replace inline styling (like style= and sx= props) with a separate CSS file putting all the CSS in one place.  Using a separate CSS file makes it a lot easier to edit and find the custom styling later, and this is the normal style of 360. (the sx prop in 360 is usually reserved for specific cases where external CSS doesn't work, but not for basic things like colors and padding)

richmondnao:
@Luke-W-Hart yup I looked at the events file for the initial one I have, but yes definitely using that for figuring out how it connects with backend would be really helpful. Thanks Luke

src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -28,6 +128,194 @@ const LostAndFound = () => {
...
+                            <Typography align="center">{report.dateLost}</Typography>
+                          </Grid>
+                          <Grid item xs={2.5}>
+                            <Typography align="center">{report.location}</Typography>
+                          </Grid>
+                          <Grid item xs={2.5}>
+                            <Typography align="center">{report.category}</Typography>
+                          </Grid>
+                          <Grid item xs={2.5}>
+                            <Typography align="center">{report.description}</Typography>
```
ben-1010, line 216: Might be nice to truncate the description text so that the report boxes can all be the same size. Could be something like:
{report.description.substring(0, 20)} {report.description.length >= 20 && '...'}

src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -28,6 +128,194 @@ const LostAndFound = () => {
...
+                  <Typography align="center">Date Lost</Typography>
+                </Grid>
+                <Grid item xs={2.5}>
+                  <Typography align="center">Location</Typography>
+                </Grid>
+                <Grid item xs={2.5}>
+                  <Typography align="center">Category</Typography>
+                </Grid>
+                <Grid item xs={2.5}>
+                  <Typography align="center">Description</Typography>
```
ben-1010, line 275: Same truncating code could be used here

ben-1010 (commented):
Looks good to me, added some suggested code to truncate description text

Centerville1 (approved):
LGTM!

PR #2402, opened 2024-10-28
Wmr - RD View Room Range With API Calls 

Mrand03:
**The following is in this PR:**
- `Housing/RoomRanges` was created and then edited to include API calls, fetching data as well as pushing data to the API whenever a room range is created, deleted or assignes. 
- I was not able to test this code due to issues connecting frontend to backend so I do not know if any of it it works correctly or not but was hoping that Ross would be able to test it and see if it works since he is able to connect our UI to our API through the ssh tunnel. 
- This is what the room range page look like right now on localhost:
RoomRange1
RoomRange2


src/views/Housing/components/RDView/components/RoomRanges/index.jsx
jtasonye, line 1: Great work on this Mya! It was a pleasure pair programming with you to figure out how to only show the RAs of the selected building.

Mrand03:
**More added to this PR:**

- Every function of room range works correctly except for deleting RA assignments to ranges because the endpoint still needs to be made.
- User can now add room ranges which are then displayed for the user to see, delete room ranges, assign an RA to a room range and when you select the building you want to create the ranges for, the building stays selected while you create room ranges and only RAs in that building show up as available to assign to those ranges instead of the whole list of RAs.
- I am now able to test the API calls as I make them on the front end because I run everything on the VM instead of ssh tunneling

src/views/Housing/index.jsx
jtasonye, line 1: Your branch does not yet have the updated ResidentView structure. Once we merge your changes into the main branch, we can work out the merge conflicts to show your work and the new ResidentView structure.

jtasonye (approved):
LGTM!

jtasonye (approved):
LGTM.

PR #2403, opened 2024-10-29
wmr - add housing button to side navigation

dli505:
**Changes made in PR:**
- added Housing Button to the side navigation


Mrand03 (approved):
Thanks for doing this Danya, it looks great!

PR #2407, opened 2024-10-31
General User L&F Home Page Mobile View + Header Breadcrumbs Functioning

richmondnao:
Breadcrumbs are working and mobile view works great as well

### DIFF
--- a/src/views/CampusSafety/components/Header/Header.module.scss
+++ b/src/views/CampusSafety/components/Header/Header.module.scss
@@ -27,7 +27,7 @@
   position: sticky;
   top: -1rem;
   margin-bottom: 1rem;
-  z-index: 1; // this component defaults to a higher z-index that overlays quick search
+  z-index: 10; // this component defaults to a higher z-index that overlays quick search
   width: calc(100% + 1rem);
   padding: 0.5rem 0.5rem;
   // clip-path removes box shadow above this nav bar
@@ -52,8 +52,31 @@
   }
 }
 
-@media (min-width: $break-sm) {
-  .breadcrumbText {
+/* Medium screens */
+@media (max-width: $break-md) {
+  .headerMain {
+    padding: 1rem 2rem;
+  }
+  .title {
+    font-size: 1.3em;
+  }
+  .subtitle {
     font-size: 0.9em;
+  }
+}
+
+/* Small screens */
+@media (max-width: $break-sm) {
+  .headerMain {
+    padding: 0.5rem 1rem;
+  }
+  .title {
+    font-size: 1.1em;
+  }
+  .subtitle {
+    font-size: 0.8em;
+  }
+  .breadcrumbText {
+    font-size: 0.7em;
   }
 }
--- a/src/views/CampusSafety/components/Header/Header.module.scss
+++ b/src/views/CampusSafety/components/Header/Header.module.scss
@@ -37,18 +37,39 @@
 .breadcrumbContainer {
   padding: 0.5rem;
   color: var(--mui-palette-primary-contrastText);
+
+  .MuiBreadcrumbs-separator {
+    color: var(--mui-palette-primary-main);
+  }
 }
 
 .breadcrumbText {
   font-size: 0.8em;
+  color: var(--mui-palette-primary-main);
+
+  @media (max-width: $break-sm) {
+    font-size: 0.7em;
+  }
+
+  @media (min-width: $break-sm) {
+    font-size: 0.9em;
+  }
 }
 
 .breadcrumbLink {
   text-decoration: none;
-  color: var(--mui-palette-info-main);
+  color: var(--mui-palette-link-main);
   &:hover {
-    color: var(--mui-palette-info-dark);
+    color: var(--mui-palette-warning-main);
     text-decoration: underline;
+  }
+
+  // Dark mode-specific adjustments
+  @media (prefers-color-scheme: dark) {
+    color: var(--mui-palette-link-main);
+    &:hover {
+      color: var(--mui-palette-link-light);
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


ben-1010 (commented):
Mobile view on Missing item report has text that appears in header when you scroll down. Also the header takes up almost half the page on the confirmation screen.
Screenshot 2024-10-31 at 11 41 01 AM
Screenshot 2024-10-31 at 11 40 37 AM

richmondnao:
> Mobile view on Missing item report has text that appears in header when you scroll down. Also the header takes up almost half the page on the confirmation screen.
> 
> Screenshot 2024-10-31 at 11 41 01 AM
> 
> Screenshot 2024-10-31 at 11 40 37 AM
> 
> 

Hey Ben, the header isn't responsive on mobile yet, that why it looks like that, that hasn't been added to the headers scss yet, I can work on that but I thought that'll be a different thing. But I'm pretty sure once the header is made properly responsive that should solve that issue.

ben-1010:
Makes sense, sounds good.

ben-1010 (approved):
LGTM

richmondnao:
> LGTM

Hey Ben made the header responsive, it was a quick fix, didnt realize it would be, can you check it again from your end

ben-1010:
Awesome, looks good on my end

Centerville1:
The color of the breadcrumb links is very hard to read, in dark mode it's a bit too dark with not enough contrast, and in light mode it just looks gray (tested on Chrome):
Screen Shot 2024-10-24 at 16 29 39

Centerville1:
Looks like the middle screen sizes aren't quite working currently, between 600px and 900px across the missing reports list looks like this:
Screen Shot 2024-10-31 at 14 38 24

Centerville1:
On larger screens, the user should probably be able to click anywhere in the list item to expand the whole thing, like how the events view works, it feels a bit confusing the way the expansion works overall.  Even on mobile, it would probably be better to expand when you click anywhere in the box, rather than just when you click on the little arrow.
Screen Shot 2024-10-31 at 14 42 07

src/views/CampusSafety/components/Header/Header.module.scss
```diff
@@ -27,19 +27,56 @@
   position: sticky;
   top: -1rem;
   margin-bottom: 1rem;
-  z-index: 1; // this component defaults to a higher z-index that overlays quick search
+  z-index: 10; // this component defaults to a higher z-index that overlays quick search
```
Centerville1, line 30: I'm just curious, why did the z-index value need to be changed?

src/views/CampusSafety/components/Header/index.tsx
```diff
@@ -1,66 +1,90 @@
...
-    children && <Grid className={styles.mainHeader}>{children}</Grid>
-  );
+        <AppBar className={styles.stickyNav}>
+          <Breadcrumbs
+            aria-label="breadcrumb"
+            separator={<NavigateNextIcon fontSize="small" />}
+            className={styles.breadcrumbContainer}
+          >
+            {/* Home breadcrumb */}
+            <CampusSafetyBreadcrumb link={pathnames.length > 0 ? '/campussafety' : null}>
```
Centerville1, line 60: This line should probably check if the pathname length is > 1, not > 0.  With this, the home page breadcrumb is a link even when you are on the home page, which there is purposeless since you're already on the page it links to.

src/views/CampusSafety/components/Header/index.tsx
```diff
@@ -1,66 +1,90 @@
...
+            {/* Only render additional breadcrumbs beyond "Campus Safety Home" */}
+            {pathnames
+              .slice(1) // Start from the second item to avoid redundancy
+              .map((value, index) => {
+                const isLast = index === pathnames.length - 2;
+                const to = `/${pathnames.slice(0, index + 2).join('/')}`;
+
+                return (
+                  <CampusSafetyBreadcrumb key={to} link={!isLast ? to : null}>
+                    {value.replace(/-/g, ' ')}
```
Centerville1, line 76: What is the purpose of the .replace here?  None of our routes have dashes in them, so why are we replacing dashes with spaces?

Centerville1 (changes requested):
Overall looks great, but there are a couple of lines that appear unused, and a minor issue with the breadcrumb while on the campus safety home page.  I also had questions about a couple of lines.

src/views/CampusSafety/components/Header/Header.module.scss
richmondnao: Changing that prevents the FirstName and Lastname from showing when you scroll, like it was in Bens initial comment/screenshot

src/views/CampusSafety/components/Header/index.tsx
richmondnao: I had that there because i was thinking of adding hyphens to the links that were showing on the frontend, just so it looks cleaner, so if we had lets say Lost-And-Found, it would show as Lost And Found on the breadcrumb rather than what it is currently lostandfound all together. But we can change that if we think its not neccessary but that was my initial thought for having that in there

src/views/CampusSafety/components/Header/index.tsx
richmondnao: Yup i can change that, that was an oversight

src/views/CampusSafety/components/Header/index.tsx
Centerville1: Gotcha, that's what I was wondering, and that makes sense to leave it in for now.  One thing we will have to deal with, and this was an issue I didn't catch in a previous PR, but urls should be all lowercase (by convention, and by what everything else on the web does), so the routes we wrote will need to be changed to be that way too, so you may have to write some fancier formatting code to capitalize them down the line (we don't need to fix that rn).

src/views/CampusSafety/components/Header/Header.module.scss
Centerville1: Gotcha, makes sense 👍

Centerville1 (approved):
LGTM

PR #2410, opened 2024-11-01
Data Entry Missing Item Database Page

richmondnao:
A page containing the list of all missing items, with filters to search the list.  Each item entry is clickable to view more details.

Screen Shot 2024-11-03 at 13 49 51

Screen Shot 2024-11-03 at 13 50 18

### DIFF
--- a/src/services/auth.ts
+++ b/src/services/auth.ts
@@ -109,6 +109,7 @@ export enum AuthGroup {
   AcademicInfoView = '360-AcademicInfoView-SG',
   LostAndFoundAdmin = '360-LostAndFoundAdmin-SG',
   LostAndFoundKiosk = '360-LostAndFoundKiosk-SG',
+  LostAndFoundDevelopers = '360-LostAndFound-Developers-SG',
 }
 
 export {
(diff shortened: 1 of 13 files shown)


Centerville1:
Made a change to use CSS, instead of JS to cutoff the text with ellipses.  This way, the text cutoff will respond to the width of the screen, instead of always being set to a constant width.

src/services/auth.ts
```diff
@@ -107,6 +107,9 @@ export enum AuthGroup {
   Staff = '360-Staff-SG',
   Student = '360-Student-SG',
   AcademicInfoView = '360-AcademicInfoView-SG',
+  LostAndFoundAdmin = '360-LostAndFoundAdmin-SG',
+  LostAndFoundKiosk = '360-LostAndFoundKiosk-SG',
+  LostAndFoundDevelopers = '360-LostAndFound-Developers-SG',
```
Centerville1, line 112: Evan added a lost and found developers security group, so we can use this one for now, pending our conversation with him.  I tested it and it works for me 👍

Luke-W-Hart (approved):
LGTM!

PR #2413, opened 2024-11-02
Delete Report Button 

erandersen21:
Sets an item's status to deleted and moves it to past reports, but does not include the confirmation pop-up. PR also contains status update front-end service.
API: https://github.com/gordon-cs/gordon-360-api/pull/1083

### DIFF
--- a/src/views/CampusSafety/views/LostAndFound/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFound/index.tsx
@@ -29,6 +29,7 @@ const LostAndFound = () => {
   const [expandedFields, setExpandedFields] = useState<{
     [id: string]: { [field: string]: boolean };
   }>({});
+  const [pageUpdates, setPageUpdates] = useState(0);
   const navigate = useNavigate();
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


Luke-W-Hart (approved):
LGTM! Watched as it worked.

Centerville1 (changes requested):
LGTM too, but let's wait to merge until we add the pop-up confirmation modal.

src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -109,16 +111,25 @@ const LostAndFound = () => {
     console.log(`Editing report: ${reportId}`);
   };
 
+  const handleDeleteClick = (reportId: string) => {
+    setReportToDelete(reportId);
+    setDeleteModalOpen(true);
+    console.log(reportId);
```
Luke-W-Hart, line 117: Do we still need this console.log?
If so then LGTM, if not then it can be removed and that also LGTM!

Luke-W-Hart (changes requested):

PR #2415, opened 2024-11-08
Wmr button hall select

dli505:
**Changes made in PR:**
image
image

Thank you to @Centerville1 and @Mrand03 for all y'all's help ^^


Mrand03 (approved):
This looks great Danya, I am glad you were able to fix the bug and make the check-in system work. If you need any help when connecting the  code to the API let me know!

PR #2416, opened 2024-11-09
wmr- Completed room range delete RA Assignment

Mrand03:
**Changes in PR:**

- The `RoomRanges` component now has a way to delete an RA/AC assignment to a room range by simply pressing the remove button next to the assignment.

Assignment_Remove


jtasonye (approved):
Good job Mya, everything looks great!

PR #2417, opened 2024-11-10
wmr - Display the full name of resident's RA

jtasonye:
Screenshot 2024-11-10 at 2 41 36 AM
**Changes made in PR:**
- Created the function `fetchRacName` in the `MyRAC` file, which sets the resident's RA based on the resident's room range


Mrand03 (approved):
This looks great Jason, thanks for doing this and it looks like you are getting an even greater understanding of how calling the API works from the front end! This will help when we need to display the RA status for the residents based on who their RA is.

PR #2418, opened 2024-11-11
Added cards and matched to existing formatting (unfinished)

erandersen21:
Added cards to links page, added logos, and changed card formatting. Still need to change formatting for Gordon Police card (how do we think that should look?) and get rid of all the white space in the grid (switch to something other than MUI Grid?)


erandersen21:
Screenshot 2024-11-12 at 4 00 30 PM

Centerville1:
@erandersen21 one thing that should get included in this PR is removing the Safety Page file, header button, side menu button, and route from the system, since the links page will now take it's place.  This will also probably have to involve a change to the breadcrumbs too, to work without the safety page.

ben-1010 (approved):
LGTM, only note is that is might be worth considering removing all visible occurences of "Campus Safety" since the page is only for the lost and found now

richmondnao (approved):
LGTM

PR #2420, opened 2024-11-11
Updated layout for RoomRanges in RD view ()

RossClark01:
This was done with **help from ChatGPT please take extra caution** when reviewing changes. All functionality has been tested

image

### DIFF
--- a/src/views/Housing/components/RDView/components/RoomRanges/index.jsx
+++ b/src/views/Housing/components/RDView/components/RoomRanges/index.jsx
@@ -8,6 +8,9 @@ import {
   Box,
   FormControl,
   InputLabel,
+  Card,
+  CardContent,
+  CardActions,
 } from '@mui/material';
 import { Select, MenuItem } from '@mui/material';
 import http from '../../../../../../services/http';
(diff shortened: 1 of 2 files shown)


src/views/Housing/components/RDView/components/RoomRanges/index.jsx
```diff
@@ -149,132 +137,192 @@ const RoomRanges = () => {
...
+                  >
+                    <Box>
+                      {range.Hall_ID}: {range.Room_Start} - {range.Room_End}
+                    </Box>
+                    <Button
+                      variant="outlined"
+                      color="secondary"
+                      size="small"
+                      onClick={(e) => {
+                        e.stopPropagation(); // Prevent triggering list item click
```
Mrand03, line 234: What does line 234 do when you use `e.stopPropagation` ? I am a little confused on what it is preventing.

Mrand03 (commented):
Everything looks great, I only had one comment about something that was confusing to me but nothing major, once that gets answered it should be good to merge!

src/views/Housing/components/RDView/components/RoomRanges/index.jsx
RossClark01: helps to make sure the event does not trigger anything inadvertently

PR #2421, opened 2024-11-12
wmr - Fixed function to return a RA_StudentsViewModel object

jtasonye:
**Changes made in PR:**
- I renamed the function `fetchRacName` to `fetchRaInfo` did some re-working on the API endpoint to return a `RA_StudentsViewModel` object, rather than just returning the Ra's name or ID like we discussed earlier. This is the new front-end logic to display the information now.


Mrand03:
LGTM!

PR #2423, opened 2024-11-13
L&F Home Page Widget

Luke-W-Hart:
Finished

Just 3 reports
Desktop:
Screenshot 2024-11-24 at 3 31 27 PM
Mobile:
Screenshot 2024-11-24 at 3 32 11 PM

More than 3 reports Text
Desktop:
Screenshot 2024-11-24 at 3 28 25 PM
Mobile:
Screenshot 2024-11-24 at 3 29 35 PM

No Reports Text
Desktop:
Screenshot 2024-11-23 at 9 30 12 PM
Mobile:
Screenshot 2024-11-23 at 9 30 59 PM


Luke-W-Hart:
What this currently looks like. Still need to have 
- Text for no reports
- Limit reports to 3 (then view all button if more than 3)
Screenshot 2024-11-23 at 9 00 55 PM
Screenshot 2024-11-23 at 9 01 33 PM

Luke-W-Hart:
No Reports Text
Screenshot 2024-11-23 at 9 30 12 PM

Luke-W-Hart:
Max of three reports with text if user has more then three.
Screenshot 2024-11-24 at 3 19 16 PM

src/views/Home/components/LostAndFoundCard/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+        <Typography align="center">{formatDate(report.dateLost)}</Typography>
+      </Grid>
+      <Grid item xs={width >= 900 ? 4 : 12}>
+        <Typography
+          align="center"
+          style={{ cursor: report.locationLost.length > 15 ? 'pointer' : 'default' }}
+        >
+          {report.locationLost.length <= 15
+            ? report.locationLost
+            : `${report.locationLost.slice(0, 15)}...`}
```
Centerville1, line 76: There's a way to do this with pure CSS that's a lot cleaner and more adaptive than the JS method used here.  It's a bit annoying to get working, but it works a lot better because it dynamically adjusts the cutoff point based on the width of the screen.
@Centerville1 can work on that this week.

src/views/Home/components/LostAndFoundCard/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+        </Typography>
+      </Grid>
+      <Grid item xs={width >= 900 ? 4 : 12}>
```
Centerville1, line 79: No need to do the width math here, the Grid API handles this for you.  You can just use md={} or sm={} to set the grid width on larger screens.  xs={} stands for extra small, which is the grid size on extra small screens.

src/views/Home/components/LostAndFoundCard/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+      } finally {
+        setLoading(false);
+      }
+    };
+    fetchMissingItems();
+  });
+
+  const renderReportContent = (report: MissingItemReport) => (
+    <>
+      <Grid item xs={width >= 900 ? 4 : 12}>
```
Centerville1, line 66: Same here, use md={} or sm={}

src/views/Home/components/LostAndFoundCard/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+        <Typography align="center">{formatDate(report.dateLost)}</Typography>
+      </Grid>
+      <Grid item xs={width >= 900 ? 4 : 12}>
```
Centerville1, line 69: Same here, use md={} or sm={}

src/views/Home/components/LostAndFoundCard/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+                  <Typography align="center">Location Lost</Typography>
+                </Grid>
+                <Grid item xs={4}>
+                  <Typography align="center">Description</Typography>
+                </Grid>
+              </Grid>
+            )}
+        {activeReports?.length > 3 ? (
+          <>
+            {activeReports.slice(0, 3).map((report) => (
```
Centerville1, line 137: I don't actually think we have to do the slicing and checking here.  You can instead do the slice in the useEffect where you first load the data, and just set a flag to whether you had to cut down the length, to use when determining whether to display the "more items not shown".

src/views/Home/components/LostAndFoundCard/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+                        {width < 900 ? (
+                          <>
+                            {/* Display two items per row on mobile */}
+                            <Grid item xs={3}>
+                              <Typography align="center">{formatDate(report.dateLost)}</Typography>
+                            </Grid>
+                            <Grid item xs={3}>
+                              <Typography align="center">
+                                {report.locationLost.length > 15
+                                  ? report.locationLost.slice(0, 15) + '...'
```
Centerville1, line 155: There's a way to do the ... using pure CSS that is more reactive.  You can find examples of that on the data entry list page (and maybe the general user list page?).

Centerville1 (changes requested):

Centerville1:
@Luke-W-Hart FYI, I'm implementing a bunch of the changes I recommended in [UI #2442](url)

Centerville1 (approved):
Approved because most of the changes I requested are coming in [UI #2442](https://github.com/gordon-cs/gordon-360-ui/pull/2442)

PR #2428, opened 2024-11-19
Fixed indexing issue

RossClark01:
When filtering data the index used for selected list item was off, causing unexpected behavior. switch from using the index to passing the ID of the list item to the API function call

button for remove room range added back


Mrand03 (approved):
This looks great Ross, thank you for doing this, the frontend for room ranges is very intuitive and user friendly now.

PR #2431, opened 2024-11-21
Populated On duty table with data

RossClark01:
Adjusted the RA checkin confirmation to show the hall name instead of the building code using a map like @dli505 was doing before


jtasonye (approved):
Looks good, I think making the map was a really good idea as well!

PR #2433, opened 2024-11-21
Minor Edits to Links Page

erandersen21:
Changed look of Gordon Police card and resized cards to fix on one page


src/views/Links/index.jsx
```diff
@@ -19,17 +19,30 @@ import FeedbackIcon from '@mui/icons-material/Feedback';
...
+      padding-bottom: 8px;
+    }
+  `);
+
   // Academics Resources UI
   const academicsCard = (
     <Card>
       <CardHeader title={`Academics`} className="gc360_header" />
-      <CardContent>
+      <CardContentWithPadding>
```
Centerville1, line 45: Probably use CSS classes here, instead of a custom styled component if possible, just to be more in line with how the rest of 360 approaches styling.  And for the other cards too

Centerville1 (changes requested):
Overall looks good, the only thing is changing out the custom styled component for CSS classes if possible, simply because that's how most of 360's styling is done, and we try to stay away from inline styling as much as possible.

PR #2434, opened 2024-11-22
Rd on call table

RossClark01:
Added on call table to mobile view component

Added standard staff view and pulled in table and mobile view components

Added auth group to use for physical plant staff to get housing info

@EjPlatzer for the auth groups will we need to make 360-RA-sg and 360-PLT-SG groups to use? we currently are just pulling in the main groups


Mrand03 (approved):
Hey Ross, this looks great! Thank you for doing the different Auth groups for RD, and standard access, the table looks great!

PR #2436, opened 2024-11-24
Display RD Name on My Hall Component Card

jtasonye:
**Changes made in PR**
Screenshot 2024-11-24 at 4 37 37 PM
- Renamed the typescript file 
- Added `fetchRdInfo` function in the typescript file
- Took out the hardcoded times on the table (the table is not part of MVP, so the times are not necessarily needed
- Screenshot is from Mya's login since my login (Wilson Hall) does not have an assigned RD in the View that the API endpoint  is pulling from

### DIFF
--- a/src/services/residentLife/ResidentStaff.ts
+++ b/src/services/residentLife/ResidentStaff.ts
@@ -11,8 +11,19 @@ type RA = {
   ID: string;
 };
 
+type RD = {
+  HallName: string;
+  BuildingCode: string;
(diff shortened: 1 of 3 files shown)


Mrand03 (approved):
This looks great Jason, the UI look very clean and user friendly. Thank you for implementing the RD name component, we can talk with the team and Melanie about what she wants for the components we are not going to implement before the MVP (whether she wants TBD or not) but otherwise it looks great!

PR #2438, opened 2024-11-25
wmr - RA can set preferred contact

Mrand03:
Created a simple way for RAs and ACs to be able to set their preferred contact which is the method in which residents will be able to contact them. It gets the RA/AC's ID and what contact they choose, then it posts this information to the database to be pulled by the resident and RD view.

Preferred Contact

### DIFF
--- a/src/services/residentLife/RA_Info.ts
+++ b/src/services/residentLife/RA_Info.ts
@@ -11,8 +11,16 @@ type RA = {
   ID: string;
 };
 
+type Contact = {
+  raId: string;
+  preferredContactMethod: string;
+};
+
 // Fetches the information of an RA from the API endpoint "Housing/ra/{hallId}/{roomNumber}"
 const fetchRaInfo = (hallId: string, roomNumber: number): Promise<RA[]> =>
   http.get(`Housing/ra/${hallId}/${roomNumber}`);
 
-export { fetchRaInfo };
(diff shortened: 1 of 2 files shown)


jtasonye (commented):
It looks like you haven't pulled the most recent updates from our `wheres-my-ra` branch. The path `/residentLife/RA_Info.ts` should be `/residentLife/ResidentStaff.ts`. Try pulling and then pushing up again, that might be why we have the 'conflicting files' issue.

I think it would also be good to put the submit button under the radio buttons, that is the most common way that I've seen from experience.

Other than those things, everything else looks good, good job Mya!

Mrand03:
> It looks like you haven't pulled the most recent updates from our `wheres-my-ra` branch. The path `/residentLife/RA_Info.ts` should be `/residentLife/ResidentStaff.ts`. Try pulling and then pushing up again, that might be why we have the 'conflicting files' issue.
> 
> I think it would also be good to put the submit button under the radio buttons, that is the most common way that I've seen from experience.
> 
> Other than those things, everything else looks good, good job Mya!

Thank you for the feedback, I can definitely impliment that quick change to the submit button.

Mrand03:
The submit button is now under the radio buttons as shown below. Let me know what you think and if there is anything else that should be changed. Otherwise it should be good to merge. 

Preferred Contact2

PR #2439, opened 2024-11-25
Added profile links to RD and RA with 360 link styling

jtasonye:
**Changes made in PR:**
Screenshot 2024-11-25 at 1 31 19 AM

- Added profile links to both the RD and RA upon Ross's suggestion

I have hardcoded data in the table because it looks more visually appealing, but we can decide with Melanie how we want that to look for our MVP


Mrand03 (approved):
This looks great, I agree that the table does look better than having empty space but like you said we can see what Melanie thinks and go from there. I like the link hover color, it is easy to see and goes with the 360 colors which is very nice. Thanks for doing this.

PR #2440, opened 2024-11-25
wmr - Fixed RA UI error and fixed user profile links

jtasonye:
**Changes made in PR:**
- @RossClark01 brought up a good point that the username of an RA might not always be their "first.last" name. Sometimes, they might have just an initial as a first name, or have a number in the last name, so pulling the username from the email would be the best way to get the username for all cases.
- Mistakenly put RD instead of RA, so I fixed that real quick

### DIFF
--- a/src/views/Housing/components/ResidentView/components/MyHall/index.jsx
+++ b/src/views/Housing/components/ResidentView/components/MyHall/index.jsx
@@ -22,11 +22,13 @@ const MyHall = () => {
 
   useEffect(() => {
     if (rdInfo) {
-      const fullName = rdInfo.RD_Name || '';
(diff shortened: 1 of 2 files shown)


Mrand03 (approved):
This looks great Jason, good job noticing that UI error of saying RD rather than RA and also good job with the change of including the RD's first and last name as it is displayed on their email, the code looks clean and concise. Thanks for noticing that as well @RossClark01.

RossClark01 (approved):
Looks good to me! It is certainly worth it to make this change now. Using just the firstname and lastname may have worked for a while but using this method will help with edge cases later on (i.e when there is john.smith and john.smith2). Good work!

PR #2441, opened 2024-11-26
Report Found Item Page

Centerville1:
Report found item page, with buttons linking to it from the home page widget and the main lost and found page.
Screen Shot 2024-11-26 at 17 11 19

### DIFF
--- a/src/views/CampusSafety/views/LostAndFound/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFound/index.tsx
@@ -1,7 +1,5 @@
 import { Card, CardContent, CardHeader, Grid, Typography, IconButton, Button } from '@mui/material';
-import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined';
-import EditIcon from '@mui/icons-material/Edit';
-import DeleteIcon from '@mui/icons-material/Delete';
+import { InfoOutlined, Edit, Delete, ExpandMore, ExpandLess } from '@mui/icons-material';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
 import { Collapse } from '@mui/material';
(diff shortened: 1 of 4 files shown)


src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -169,34 +167,47 @@ const LostAndFound = () => {
...
                 <Grid item xs={12} sm={11}>
                   <Typography variant="body1" align="center">
-                    Police staff will view reports, and you will be notified if your item is found.
+                    <InfoOutlined /> Police staff will view reports, and you will be notified if
+                    your item is found.
                   </Typography>
+                  {/* Button to Report a Missing Item */}
+                  <Grid
+                    container
+                    justifyContent="center"
```
erandersen21, line 178: The grids don't seem to be properly spaced on the page, so this line doesn't work as intended (at least when I run it). 
Screenshot 2024-11-29 at 10 08 46 PM

src/views/CampusSafety/views/LostAndFound/views/ReportFound/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const navigate = useNavigate();
+
+  return (
+    <>
+      <Header />
+      <Card>
+        <CardHeader title={'Report a Found Item'} className="gc360_header" />
+        <CardContent>
+          <Typography>If you've found an item somewhere on campus:</Typography>
+          <ul>
```
erandersen21, line 16: This is just my opinion, but I think this looks a little weird as a bulleted list.

src/views/CampusSafety/views/LostAndFound/views/ReportFound/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useNavigate } from 'react-router';
+import Header from 'views/CampusSafety/components/Header';
+
+const ReportFound = () => {
+  const navigate = useNavigate();
+
+  return (
+    <>
+      <Header />
+      <Card>
```
erandersen21, line 12: This card should probably be centered and sized relative to the content, but I assume you're working on that.

erandersen21 (approved):
LGTM except for the minor issues I mentioned.

src/views/CampusSafety/views/LostAndFound/index.tsx
Centerville1: I just pushed a fix to UI #2442 which fixes this :+1:

src/views/CampusSafety/views/LostAndFound/views/ReportFound/index.tsx
Centerville1: Yeah, if you have ideas for how to design this page better, let me know.  I kinda agree, but I also don't really know what would work better.  I wrote this page in about 15 minutes, so it's kinda just thrown together.

src/views/CampusSafety/views/LostAndFound/views/ReportFound/index.tsx
Centerville1: Fixed, good point!

src/views/CampusSafety/views/LostAndFound/views/ReportFound/index.tsx
erandersen21: I'm not exactly sure either. Maybe we can all figure out how it should look on Tuesday? I would say you could merge it for now.

PR #2442, opened 2024-11-26
General User UI Visual Improvements

Centerville1:
This PR overhauls and refactors a great deal of the code for the general user UI components.  This includes visual changes, functional tweaks, and a bunch of code quality and readability improvements.  This also includes updated mobile views, and simplified mobile functionality.

New lost and found page list styling:
Screen Shot 2024-11-27 at 03 46 52

And on mobile:
Screen Shot 2024-11-27 at 03 47 18

Home page widget updated design:
Screen Shot 2024-11-27 at 03 48 22


Luke-W-Hart:
Code looks good! I'm not too sure about how the report list mobile view UI looks. I'm assuming location and description are two separate rows so more text can be shown, but I feel like it would be better if we made the reports two rows instead of three. Let me know what you think.

src/views/Home/components/LostAndFoundCard/LostAndFoundCard.module.scss
```diff
@@ -0,0 +1,4 @@
+.linkHeader {
+  cursor: pointer;
+  text-decoration: underline;
```
Luke-W-Hart, line 3: This is leading to the Lost and Found card title being underlined, not sure if we want this. Otherwise, LGTM!

Centerville1: Changed to only underline on hover.

richmondnao (approved):
LGTM

PR #2443, opened 2024-11-27
Reset page scroll position before client side navigation

Centerville1:
It's been bothering me for a while that the scroll position of the "main" page element doesn't reset when navigating between pages.  Most of 360 is fine on larger screens (as the whole page can fit vertically), but on mobile this can cause issues when navigating between pages using buttons, and finding yourself halfway down the page without context.

This code seems to work, and calls the ScrollToTop method each time the react useNavigate method is called (although there may be a better way to implement this, so I'm open to suggestions, but I'd like to make this change to navigation).

### DIFF
--- a/src/app.jsx
+++ b/src/app.jsx
@@ -2,7 +2,7 @@ import { useMsal } from '@azure/msal-react';
 import AppRedirect from 'components/AppRedirect';
 import BirthdayMessage from 'components/BirthdayMessage';
 import { useWatchSystemColorScheme } from 'hooks';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useLocation } from 'react-router';
 import { Route, Routes, useNavigate } from 'react-router-dom';
 import { CustomNavigationClient } from 'services/NavigationClient';
@@ -17,6 +17,7 @@ import routes from './routes';
 const App = () => {
   useWatchSystemColorScheme();
   const location = useLocation();
+  const mainRef = useRef(null);
 
   useEffect(() => {
     analytics.onPageView();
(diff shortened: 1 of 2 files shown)


EjPlatzer:
Hi @Centerville1 , thanks for working on this. I think there is a cleaner way to implement this that doesn't require us to change the custom navigation client. I'm actually a little bit surprised that the current version of the code works, because the custom navigation client is only used when MSAL redirects the page, not when the user clicks on a link or causes a programmatic navigation.

This article presents a really simple alternative that I think works in our environment, although I haven't tested it yet: [https://medium.com/@caden0002/fixing-scroll-position-issues-in-react-router-scroll-to-top-on-navigation-86bcfbdfc9db](https://medium.com/@caden0002/fixing-scroll-position-issues-in-react-router-scroll-to-top-on-navigation-86bcfbdfc9db)

Centerville1:
> Hi @Centerville1 , thanks for working on this. I think there is a cleaner way to implement this that doesn't require us to change the custom navigation client. I'm actually a little bit surprised that the current version of the code works, because the custom navigation client is only used when MSAL redirects the page, not when the user clicks on a link or causes a programmatic navigation.
> 
> This article presents a really simple alternative that I think works in our environment, although I haven't tested it yet: https://medium.com/@caden0002/fixing-scroll-position-issues-in-react-router-scroll-to-top-on-navigation-86bcfbdfc9db

Solid, that works, and is a lot simpler!  Just pushed that implementation, the one difference from the article is the need to use a reference to the main element, since it's that element which holds the scroll position on 360, rather than window.

EjPlatzer (changes requested):
Looking good! I left a comments about small fixes.

EjPlatzer (approved):
LGTM! Thanks for improving this!

PR #2444, opened 2024-11-27
wmr - Display On Duty Component Info

jtasonye:
**Changes made in PR:**
- Added API call to `RA_OnCall.ts` file to pull the On Duty RA's information
- Filled in the needed information for `On Duty` component (name, room #, contact info, photo, profile link)
- @Mrand03 or @dli505, one of you should test it to see if the information displays in your resident view. On the backend, the `Housing.Current_On_Call`, is the view that we pull from to display this information. Since my dorm (Wilson) does not have an RD (look at `RD_Info` view), the RA on duty for Wilson does not show up in the `Housing.Current_On_Call` view, so I cannot see whether it displays, but the logic to display it should be right
- This is a joint pull request with ( https://github.com/gordon-cs/gordon-360-api/pull/1103 ), so that needs to be approved before the results from this pull request can show up.

### DIFF
--- a/src/services/residentLife/RA_OnCall.ts
+++ b/src/services/residentLife/RA_OnCall.ts
@@ -1,5 +1,17 @@
 import http from 'services/http';
 
+type RA = {
+  Hall_ID: string;
+  Hall_Name: string;
+  RoomNumber: string;
+  RA_Name: string;
+  PreferredContact: string;
+  Check_in_time: Date;
+  RD_Email: string;
+  RA_Profile_Link: string;
+  RD_Profile_Link: string;
+};
+
 const fetchOnDutyData = async () => {
   try {
     return await http.get('Housing/ra/on-call/all');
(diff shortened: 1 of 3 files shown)


src/services/residentLife/RA_OnCall.ts
```diff
@@ -9,4 +21,7 @@ const fetchOnDutyData = async () => {
   }
 };
 
-export { fetchOnDutyData };
+// Fetches the information of an On Call RA from the API endpoint "Housing/ra/on-call/{hallId}"
+const fetchOnDutyRA = (hallId: string): Promise<RA[]> => http.get(`ra/on-call/${hallId}`);
```
dli505, line 25: This should be `Housing/ra/on-call/${hallId}`!

dli505 (changes requested):
Hi Jason, great work here! Just a small bug fix where I commented and we should be good to go!

jtasonye:
Thanks for catching that @dli505 , I made the fix on the frontend and added some things and added some things in the backend so now it displays!

Screenshot 2024-11-30 at 12 11 02 AM

dli505 (approved):
LGTM!

PR #2449, opened 2024-12-01
Dark Mode Form Label Text Contrast Fix

Centerville1:
This has been an issue for several years now.  Figured I'd finally fix it, since it's kinda annoying, and relatively easy to fix :).  Added a style override to the theme, fixing the colors of the label and border elements when focusing form fields.

### Old: 
Screen Shot 2024-12-01 at 00 30 59

### New:
Screen Shot 2024-12-01 at 01 00 31


EjPlatzer (approved):
LGTM! Thanks for fixing this!

Issue #97, opened 2018-06-22
Margins should be consistent between pages


softwary:
With the same breakpoints, etc

PR #2450, opened 2024-12-02
Fix for Bug on Project Board #97

richmondnao:
Fix for Bug #97


Centerville1 (approved):
Looks Splendid to Me :)

PR #2451, opened 2024-12-02
Default Actions for Missing Item Reports

ben-1010:
This code adds a default action on every missing item report that is created or altered by a general user.  Each time a user creates, edits, or deletes their reports, an action is added to the report, recording the interaction.

[Actions Design Doc](https://docs.google.com/document/d/129QPHFhe1eaR-TtU0kXGKn7iMFlirIHKKwXDV9dJFig/edit?usp=sharing)


src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -127,7 +156,37 @@ const MissingItemFormEdit = () => {
       dateLost: new Date(formData.dateLost).toISOString() || DateTime.now().toISO(),
       forGuest: false,
     };
+    const originalFields = Object.keys(formData);
+    const newFields = Object.keys(formDataCopy);
```
Centerville1, line 160: What's the purpose of getting newFields as well as original fields?  Could you just use original fields twice, since the two objects are identical in format?

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -127,7 +156,37 @@ const MissingItemFormEdit = () => {
...
+    let newActionNote = '';
+    for (let i = 0; i < originalFields.length; i++) {
+      if (
+        // @ts-ignore
+        JSON.stringify(formDataCopy[originalFields[i]]) != JSON.stringify(formData[newFields[i]])
```
Centerville1, line 165: You can remove this ts-ignore using the following line.
```
JSON.stringify(formDataCopy[originalFields[i] as keyof typeof formDataCopy]) !==
        JSON.stringify(formData[newFields[i] as keyof typeof formData])
```

Since you know the keys in originalFields[i] newFields[i] will exist on the corresponding types, ```as keyof typeof ____``` tells typescript you are accessing a valid index.

[https://stackoverflow.com/questions/57350092/string-cant-be-used-to-index-type](https://stackoverflow.com/questions/57350092/string-cant-be-used-to-index-type)

Also, notice the use of !==, not !=.  != checks for inequality with type conversion, and not true inequality, so can lead to all sorts of weird, nasty, unexpected JS bugs (JS is a beautiful language :)

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -127,7 +156,37 @@ const MissingItemFormEdit = () => {
...
+      ) {
+        // @ts-ignore
```
Centerville1, line 167: This ts-ignore appears to be doing nothing, I got no errors when running without it.

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -127,7 +156,37 @@ const MissingItemFormEdit = () => {
...
+        newActionNote +=
+          originalFields[i] +
+          ': OLD: ' +
+          // @ts-ignore
```
Centerville1, line 171: You can solve this ts-ignore using the same trick as above.

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -127,7 +156,37 @@ const MissingItemFormEdit = () => {
...
+          formDataCopy[originalFields[i]] +
+          ', NEW: ' +
+          // @ts-ignore
```
Centerville1, line 174: same here

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -127,7 +156,37 @@ const MissingItemFormEdit = () => {
...
+          formData[newFields[i]] +
+          ' ';
+      }
+    }
     await lostAndFoundService.updateMissingItemReport(requestData, parseInt(itemid || ''));
+    const actionRequestData = {
+      missingID: itemid,
```
Centerville1, line 181:
```suggestion
      missingID: parseInt(itemid || ''),
```
This tells Typescript that this will always be a number, which lets you remove the ts-ignore on line 188.

Centerville1 (changes requested):
Just need to fix the Typescript errors, and a couple of variable naming and use things to fix.  I included example code to fix the various typescript errors, which should work, lmk if you have any questions!

src/views/CampusSafety/views/LostAndFound/index.tsx
```diff
@@ -95,6 +108,19 @@ const LostAndFound = () => {
   const handleModalSubmit = async () => {
     try {
       await lostAndFoundService.updateReportStatus(parseInt(reportToDelete || ''), 'deleted');
+      let actionRequestData = {
+        ...newActionFormData,
+        missingID: reportToDelete,
```
Centerville1, line 113:
```suggestion
        missingID: parseInt(reportToDelete || ''),
```
This will fix the Typescript error on line 122.

Centerville1 (changes requested):

Centerville1 (approved):
LGTM 👍

PR #2464, opened 2024-12-05
Fix for Issue with Creating Missing Item Reports in Production

Centerville1:
This fixes the issue with calling the POST route in production returning 404.  Not sure how we missed this, apparently this route call works fine in the development environment, but as soon as you build and run for production, it doesn't work with the leading /.  Not sure how we missed this inconsistency, but it's fixed now :).

Thanks @EjPlatzer for spotting that!


erandersen21 (approved):
LGTM

PR #2470, opened 2024-12-05
wmr- Minor changes to UI on Room Ranges

Mrand03:
**Changes made in this PR**

- Room ranges UI was made to be cleaner and more appealing 
- Added instructions and clarified the instructions that were there for room ranges so that everything is esay to do for the user
image
image


dli505 (approved):
LGTM, I'm not sure if the css stuff should be in a separate `.scss` file or not, but it looks good for now. Thanks Mya!

PR #2471, opened 2024-12-05
WMR - cleaning up ra UI

dli505:
**Changes made in PR:**
- added mobile view for RA view
- removed TaskList and Scheduling
- added links for work requests and GoPo
- made button prettier
- formatted everything

### DIFF
--- a/src/views/Housing/components/RAView/components/Schedule/index.jsx
+++ b/src/views/Housing/components/RAView/components/Schedule/index.jsx
@@ -1,7 +1,53 @@
 import ScheduleCalendar from '/src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx';
+import { Calendar, luxonLocalizer } from 'react-big-calendar';
+import { DateTime } from 'luxon';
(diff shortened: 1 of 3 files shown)


src/views/Housing/components/RAView/components/Schedule/index.jsx
```diff
@@ -1,7 +1,53 @@
 import ScheduleCalendar from '/src/components/Profile/components/SchedulePanel/components/ScheduleCalendar/index.tsx';
+import { Calendar, luxonLocalizer } from 'react-big-calendar';
+import { DateTime } from 'luxon';
```
RossClark01, line 3: We should be using built in functions, 360 is trying to phase out Luxon (According to comment from Evan)

RossClark01 (changes requested):
looks good to me but noticed you are using Luxon for scheduling component. removing luxon was a requested change on lost and found so we will need to align ourselves with that change as well

src/views/Housing/components/RAView/components/Schedule/index.jsx
dli505: Removed `Schedule` for now, since it's not going to be in the MVP, I'll figure out a replacement for Luxon next semester.

PR #2477, opened 2024-12-08
Change Admin List to sort by Created Date

Centerville1:
Change the admin list view to sort by dateCreated, that way new reports always show up on top.


erandersen21 (approved):
LGTM

PR #2480, opened 2024-12-09
Added logic for displaying housing tab

RossClark01:
Checks if a user is a residential student or a member of a related group before displaying the housing tab in the header or side navigation bar


dli505 (approved):
LGTM!

PR #2481, opened 2024-12-09
Status for items and uneditable fields

richmondnao:
Finished the Status and Made fields for other status uneditable

### DIFF
--- a/src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
@@ -75,6 +75,7 @@ const MissingItemFormEdit = () => {
   });
 
   const [showConfirm, setShowConfirm] = useState(false);
+  const isEditable = formData.status === 'active';
 
   useEffect(() => {
     const fetchUserData = async () => {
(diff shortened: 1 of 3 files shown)


src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -339,6 +342,38 @@ const MissingItemFormEdit = () => {
...
+              {/* Display a chip for items with statuses other than "active" */}
+              {formData.status.toLowerCase() !== 'active' &&
+                formData.status.toLowerCase() !== 'found' && (
+                  <Grid container xs={9.7} className={styles.foundContainer} rowGap={2}>
+                    <Grid container item xs={12} md={6} rowGap={2}>
+                      <Grid item xs={12}>
+                        <Chip
+                          className={styles.largeChip}
+                          sx={{
+                            height: 'auto',
```
Centerville1, line 354: Can we move height: 'auto' into the CSS largeChip class?  I know the mui override section here needs to be in the sx, but this line doesn't as far as I can tell.

Centerville1 (changes requested):
Overall looks good, just one minor CSS fix, then should be good to go 👍

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
richmondnao: Sure

Centerville1 (approved):
LGTM

PR #2484, opened 2024-12-09
Removed non MVP code

RossClark01:
Added switch case for housing developers to select their view


src/views/Housing/index.jsx
jtasonye, line 1: The switch case is wild! But for presentation purposes, this is an ingenious add.

jtasonye (approved):
LGTM! I will go ahead and merge it.

PR #2485, opened 2024-12-09
wmr - Refining ResidentUI for MVP

jtasonye:
**Changes made in PR:**

Screenshot 2024-12-09 at 3 51 50 PM

- Added a heading (I can turn it into a component later if we want to use it on other views)
- Added util folder for some of our most repeated functions (`formatPhoneNumber`, `isMobile`, `staffType`)
- Changed Resources format based on our customer's feedback


RossClark01 (approved):
Looks good to me!

src/views/Housing/utils/staffType/staffType.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Utility function to determine the staff type based on the building code.
+ * @returns {string} - Returns 'RA' for Residence Assistant or 'AC' for Apartment Coordinator based on the building code.
+ */
+export const staffType = {
+  BRO: 'AC',
+  CHA: 'RA',
+  EVN: 'RA',
+  FER: 'RA',
```
russtuck, line 9: I have a vague memory that there are apartments in the basement of Ferrin.  Is that true?  Do they have an AC?

PR #2487, opened 2024-12-09
Wmr display hall images on sp

jtasonye:
**Changes made in PR:**
- Changed hall mascot photo paths so that they are relative to the root, rather than the directory. This is most likely why it was not showing up in our Sp-branch.


RossClark01 (approved):
LGTM!

PR #2489, opened 2024-12-10
More work for cleaning up MVP

RossClark01:
made the welcome banner  a separate component

Added welcome banner to ra page. Made alternate text for ra/resident for the welcome

RD links removed Sharepoint and groups.gordon

used previously unused editdocs component for RD view

Removed phone links from RD table on desktop

Added a no on call ra view for the table so it is not just empty


dli505 (approved):
LGTM! Thanks for your work Ross.

PR #2490, opened 2024-12-10
Fix Report Column Color Contrast

Luke-W-Hart:
Figured it was best to play it safe with plain old white.
Screenshot 2024-12-10 at 2 00 20 AM
Screenshot 2024-12-10 at 2 00 41 AM


erandersen21 (approved):
LGTM

PR #2495, opened 2024-12-10
Clarified Wording when Creating a Report

Luke-W-Hart:
Button in bottom right used to be "SUBMIT"
Screenshot 2024-12-10 at 6 00 15 PM
Used to say "File a Report" in the header and then "Confirm the Details of your Report" below outside of the header.
Screenshot 2024-12-10 at 6 01 26 PM


Centerville1 (approved):
LGTM!

PR #2497, opened 2024-12-11
wmr - Refactoring API routes

jtasonye:
**Changes made in PR:**
- Mirrored the api route name changes on the UI to follow RESTful conventions


src/services/residentLife/ResidentStaff.ts
```diff
@@ -28,16 +28,16 @@ type method = {
   PreferredContact: string;
 };
 
-// Fetches the information of an RA from the API endpoint "Housing/ra/{hallId}/{roomNumber}"
+// Fetches the information of an RA from the API endpoint "Housing/ras"
 const fetchRaInfo = (hallId: string, roomNumber: number): Promise<RA[]> =>
-  http.get(`Housing/ra/${hallId}/${roomNumber}`);
+  http.get(`Housing/ras?hallId=${hallId}&roomNumber=${roomNumber}`);
```
EjPlatzer, line 33: String interpolation is not the best tool for adding search params to a URL, because there's a lot of characters that have to be specially encoded. Browsers provide the `URLSearchParams` class to make working with search params easier, but in 360 we've made a simple utility function to handle most cases:
```suggestion
  http.get(`Housing/ras${http.toQueryString({hallId, roomNumber})`);
```

EjPlatzer (changes requested):
One small note on handling search params in the query string of a URL.

src/services/residentLife/ResidentStaff.ts
RossClark01: Updated accordingly

RossClark01 (approved):
Thanks for starting this sir! Adjusted routes a bit more to account for a few more changes. I have a few more points to address on the API side so we can merge this into WMRA after changes are squared away

jtasonye:
Thanks for catching that Ross!

RossClark01:
@jtasonye added a few commits here, can you take a look and make sure everything looks good. Once done we should be able to merge this into our WMRA branch and get it into SP. The API side has another PR open that once reviewed can be merged into WMRA then SP for further user testing

jtasonye:
@RossClark01 LGTM!

russtuck (commented):
Why do we have "post" instead of "put" here?  

It looks like all the post commands are idempotent and modify existing data.  That means that a put would be more appropriate.  Post should generally only be used when asking the API server to create something new and assign it a unique URL.

PR #2498, opened 2024-12-11
Lost and Found UI 1.0 Release

Centerville1:
We have spent the last week working through changes from review by @EjPlatzer.  This PR represents our release candidate for our MVP of the Lost and Found senior project.

Changes include all the UI changes needed for the MVP of the lost and found project, covering all of the user stories listed as completed on our project board.


src/services/lostAndFound.ts
```diff
@@ -0,0 +1,138 @@
+import http from './http';
+import userService from './user';
+
+/**
+ * Missing item report object, representing the model of a report for communication with the
+ * backend service.
+ */
+export type MissingItemReport = {
+  recordID?: number;
```
EjPlatzer, line 9: Why is `recordID` nullable?

src/services/lostAndFound.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Fetch an array containing the full list of all missing item reports for
+ * the currently logged in user.
+ * @returns MissingItemReport[] array of all missing item reports.
+ */
+const getMissingItemReportUser = (): Promise<MissingItemReport[]> => {
+  const profile = userService.getProfileInfo();
+  return profile.then((profileValue) =>
+    http.get<MissingItemReport[]>(`lostandfound/missingitems?user=${profileValue?.AD_Username}`),
+  );
+};
```
EjPlatzer, line 61: I think it would be better to make the username a parameter of this function, so that it can be re-used for getting reports for a given user other than the current one. Also, `userService.getProfileInfo` sends ~5 API requests and does a lot of work to build the current user's whole profile, so this route becomes much less efficient than it could be.

It should be almost trivial to call `useUser` in the components that call `getMissingItemReportUser` so that we can pass in `profile.AD_Username` as a parameter.

Finally, putting URL Search Params in an interpolated string is bad practice, because URL params have to be encoded carefully. 

```suggestion
const getMissingItemReportUser = (username: string): Promise<MissingItemReport[]> => 
    http.get<MissingItemReport[]>(`lostandfound/missingitems?${http.toQueryString({user: username})}`)
```

EjPlatzer (changes requested):
A couple of small things in the lostAndFound service code.

src/services/lostAndFound.ts
EjPlatzer: If the type of the objects is different, we should use different types for them. I see that you're actually using the type `Omit<MissingItemReport, 'recordID'>` when posting, which is the correct approach. You can save that as a new type, e.g. 
```typescript
export type NewMissingItemReport = Omit<MissingItemReport, 'recordID'>
```

But if the value returned from the API never has a null/undefined `recordID`, we should model that property as never being null/undefined. Otherwise, you have to fight with typescript downstream because you know that it's defined but typescript doesn't and will complain when you try to access it.

The same thing applies to the ID fields on the other objects you model in this file.

src/services/lostAndFound.ts
Centerville1: Just resolved this in the most recent merge from senior-project-2.

Centerville1:
@EjPlatzer just pushed changes to resolve the two things you noticed, let me know if there's anything else we should do before merging, thanks again!

PR #2500, opened 2024-12-12
Fixed get user service (currently broken)

erandersen21:


src/views/Home/components/LostAndFoundCard/components/ActiveReports/index.tsx
```diff
@@ -33,17 +34,30 @@ const ActiveReports = () => {
...
   const numReportsToDisplay = 4;
   const [snackbar, setSnackbar] = useState({ message: '', severity: undefined, open: false });
+  const [user, setUser] = useState({ AD_Username: '' });
   const navigate = useNavigate();
 
   const createSnackbar = useCallback((message, severity) => {
     setSnackbar({ message, severity, open: true });
   }, []);
 
+  useEffect(() => {
```
Centerville1, line 44: Evan mentioned: "It should be almost trivial to call useUser in the components that call getMissingItemReportUser."  I believe there is a hook called useUser, that will get this information.  That will avoid having to call userService.getProfileInfo() every time.  Like he mentioned, calling this has to perform ~5 API requests, which is much slower than it needs to be to just get the username.  See if you can figure out using useUser instead of this, should be much more performant.

src/views/Home/components/LostAndFoundCard/components/FoundReports/index.tsx
```diff
@@ -15,17 +16,30 @@ const formatDate = (date: string) => {
...
+  const [user, setUser] = useState({ AD_Username: '' });
   const navigate = useNavigate();
   const isMobile = useMediaQuery('(max-width:375px)');
 
   const createSnackbar = useCallback((message, severity) => {
     setSnackbar({ message, severity, open: true });
   }, []);
 
+  useEffect(() => {
+    const fetchUserData = async () => {
```
Centerville1, line 28: Same thing here, use useUser instead if possible (to avoid excessive API requests)

Centerville1 (dismissed):

Centerville1 (approved):
LGTM

PR #2503, opened 2024-12-16
Wmr UI fixes

jtasonye:
**Changes in PR**
- Removed all instances of isMobile utility function since Evan said it was not needed
- Changed staff type to be deeply immutable so that it cannot be changed when imported


Mrand03:
Thanks for making these changes! looks good to merge.

PR #2510, opened 2024-12-17
Updates to my hall to remove characters when calling the api

RossClark01:
if a resident was in B20 'B20' was sent to the API causing some errors. Now only ints are sent to the API


src/views/ResLife/components/ResidentView/components/MyRA/index.jsx
jtasonye, line 1: LGTM!

jtasonye (commented):
LGTM! Have you tested the room ranges function to see if this change works as intended?

jtasonye (approved):
I tested out the change in making a room range and it works. I can merge.

PR #2512, opened 2024-12-17
WMRA UI  updates for deployment

RossClark01:
Logic for hiding the page when browsed directly by non resident students
notice added when no RA assigned on resident view


Mrand03 (approved):
Thanks Ross for making this logic for commuters and if there is no RA assigned, this will be important when we merge into production since there will be no RA assignments right away.

PR #2517, opened 2024-12-17
Wmr UI fixes

RossClark01:
Updated Links: Adjusted all links to be relative to the environment, removing hardcoded URLs for better flexibility

Improved RD/RA Assignment Logic: Added logic to display a message when no RD/RA is assigned, instead of leaving the field blank.

ResLife View Update: Updated the ResLife view to check the student’s residential status before returning the page. Previously, commuter students could not access the ResLife tab but could still directly browse the page. 

MyHall Update: Resolved an issue where room numbers containing extra characters (e.g., B20 for basement rooms or 101A for apartments with multiple rooms) were causing API errors. The update ensures that only the root room number is extracted and passed to the API.


Centerville1 (approved):
@RossClark01  LGTM, feel free to wait for Evan's review, but I'd like to get this merged ASAP assuming you've tested so we can release first thing tomorrow morning.

.env.production
```diff
@@ -6,3 +6,6 @@ VITE_ANALYTICS_ID=G-2FE78G0CBN
 
 # @TRAIN
 VITE_API_URL=https://360ApiTrain.gordon.edu/
+
+# @LOCALHOST
+# VITE_API_URL=http://localhost:51658/
```
EjPlatzer, line 11: We should revert this change before merging to develop.

EjPlatzer (changes requested):
Just need to clean up `.env.production` and then this is ready to merge.

.env.development
```diff
@@ -6,6 +6,3 @@
 # @TRAIN
  VITE_API_URL=https://360ApiTrain.gordon.edu/
 
-# @LOCALHOST
-#  VITE_API_URL=http://localhost:51660/
-
```
EjPlatzer, line 11: This needs to be reverted - `.env.development` should have a Localhost section that is commented out by default

PR #2525, opened 2025-01-16
Fix Last Checked Date

Centerville1:
Last checked date would not always update to the most recent, and was simply based on the order the admin actions were fetched from the database.

Screen Shot 2025-01-16 at 11 32 24

Screen Shot 2025-01-16 at 11 32 09

### DIFF
--- a/src/views/CampusSafety/views/LostAndFoundAdmin/views/MissingItemList/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFoundAdmin/views/MissingItemList/index.tsx
@@ -124,9 +124,15 @@ const MissingItemList = () => {
 
   // Find and format the last checked date based on the list of admin actions for a given report.
   const displayLastCheckedDate = (report: MissingItemReport) => {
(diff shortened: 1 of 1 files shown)


Luke-W-Hart:
@Centerville1 when I tested if the report had already been checked on a previous date and then I did a check admin action again the date would be the previous checked date.
Screenshot 2025-01-17 at 3 54 21 PM
Screenshot 2025-01-17 at 3 54 43 PM

Centerville1:
> @Centerville1 when I tested if the report had already been checked on a previous date and then I did a check admin action again the date would be the previous checked date. Screenshot 2025-01-17 at 3 54 21 PM Screenshot 2025-01-17 at 3 54 43 PM

@Luke-W-Hart the report you mentioned shows up correctly for me.  Unless this is a safari issue?  Are you sure you're running the right code/any other information about what's going wrong.

Centerville1:
Going to make the changes on the backend instead, to send actions in date order.

PR #2527, opened 2025-01-22
Disabled date input via keyboard

erandersen21:
Users are no longer able to type in dates, which was causing problems for the MUI date picker.


Centerville1 (approved):
LGTM

PR #2529, opened 2025-01-22
wmr - UI update for Resident View

jtasonye:
**Old Resident View**
old_res_view

**Updated Resident View**
new_res_view

**Changes made in PR:**
- Updated the Chase, Nyland, and Tavilla hall mascot images since the previous images did not display well in dark mode.
- Fixed the cards so that even when they a resized, they remain the same size (Used LLM to figure out what type of styling approach to take, which lead to me playing around with the card elevation styling. Those sections should be closely reviewed).
- Renamed `Contacts` to `Email Contacts` since a user reported that they did not want to click the links since they did not know what form of contact would be made (either by call or by email).
- Removed the hyperlink from the avatar component when an RA has not been assigned to a resident.


.env.development
dli505, line 1: Did we mean to push this?

.env.production
dli505, line 1: Same question for here

src/views/ResLife/components/ResidentView/components/MyHall/index.jsx
dli505, line 1: Not sure if the styles (like the sizes and stuff) have to go into an scss file? Otherwise this is fine

src/views/ResLife/components/ResidentView/components/OnDuty/index.jsx
```diff
@@ -67,23 +61,30 @@ const OnDuty = () => {
           <Grid container spacing={2} alignItems="center">
             {/* Text Section */}
             <Grid item xs={8}>
-              <Typography variant="h5" color="warning.main">
-                No one is on duty right now!
+              <Typography variant="subtitle1" color="warning.main">
+                No {staffTypeLabel ? staffTypeLabel : 'one'} is on duty right now! 🐾
```
dli505, line 65: I like this a lot!

src/views/ResLife/components/ResidentView/components/OnDuty/index.jsx
```diff
@@ -67,23 +61,30 @@ const OnDuty = () => {
...
+              justifyContent="center"
+              sx={{ marginTop: { xs: 1, sm: 2, md: 2 } }}
+            >
               <Avatar
                 src={ScottieMascot}
                 alt="Scottie"
                 sx={{
-                  width: { xs: 80, sm: 110, md: 80, lg: 120 },
-                  height: { xs: 80, sm: 110, md: 80, lg: 120 },
+                  width: { xs: 80, sm: 90, md: 100, lg: 130 },
```
dli505, line 84: Same question about whether these need to be moved to an `.scss` file

dli505 (changes requested):
Looks good to me! Just left some comments and questions, but if they aren't pertinent, then lmk and I can approve

dli505 (approved):
LGTM!

PR #2530, opened 2025-01-24
feature for auto expiring items after 6 months

richmondnao:


src/views/CampusSafety/index.tsx
```diff
@@ -41,6 +43,44 @@ export const CampusSafetyRoutes: CampusSafetyRoutesObject = {
 const CampusSafetyApp = () => {
   const { profile, loading: loadingProfile } = useUser();
 
+  useEffect(() => {
+    const updateAndFixReports = async () => {
+      try {
+        const sixMonthsAgo = new Date();
+        sixMonthsAgo.setMonth(sixMonthsAgo.getMonth() - 6);
+        sixMonthsAgo.setHours(0, 0, 0, 0); // Normalize time
+        console.log(`Six months ago date: ${sixMonthsAgo.toISOString()}`);
```
Centerville1, line 52: Should remove this debugging console.log

src/views/CampusSafety/index.tsx
```diff
@@ -41,6 +43,44 @@ export const CampusSafetyRoutes: CampusSafetyRoutesObject = {
...
+          }
+          if (reportDate >= sixMonthsAgo && report.status === 'expired') {
+            await lostAndFoundService.updateReportStatus(
+              parseInt((report.recordID ?? '').toString()),
+              'active'
+            );
+          }
+        }
+      } catch (err) {
+        console.error('Error in updateAndFixReports:', err);
```
Centerville1, line 73: Same here, should probably remove this console.log if it's for debugging, maybe show an error snackbar so users can see it and report if there's an issue (although maybe that would be weird in this case)?

src/views/CampusSafety/index.tsx
```diff
@@ -41,6 +43,44 @@ export const CampusSafetyRoutes: CampusSafetyRoutesObject = {
 const CampusSafetyApp = () => {
   const { profile, loading: loadingProfile } = useUser();
 
+  useEffect(() => {
```
Centerville1, line 46: Putting this in this here means it will break for any user who doesn't have admin permissions every time they try to run it, since the API will return a 401 when they try to access all reports.  I think we should include a check for if the user is an admin user before attempting to run this method.

src/views/CampusSafety/index.tsx
```diff
@@ -41,6 +43,44 @@ export const CampusSafetyRoutes: CampusSafetyRoutesObject = {
...
+        for (const report of reports) {
+          const reportDate = new Date(report.dateCreated);
+          reportDate.setHours(0, 0, 0, 0);
+          if (reportDate < sixMonthsAgo && report.status === 'active') {
+            await lostAndFoundService.updateReportStatus(
+              parseInt((report.recordID ?? '').toString()),
+              'expired'
+            );
+          }
+          if (reportDate >= sixMonthsAgo && report.status === 'expired') {
```
Centerville1, line 65: Maybe add a comment here explaining the reasoning for this section so we remember, something like double checking that all expired reports are meant to be expired, in case one user's clock is set incorrectly.

Centerville1 (changes requested):
A couple things to finish up.  Also, just checking, did you test this on the database (I just didn't see any expired reports in train, so just making sure you've seen this work).

Centerville1:
@richmondnao I also discussed this with Karl this morning, but I think we should change this so reports expire sooner (we'll also need to change to message in the form confirmation page.  This would be try to limit the number of active reports in the system at a time, to keep it manageable.  We were thinking 2 months, was reasonable, but we should maybe add the ability for admin staff to extend that for more valuable/stolen items.

src/views/CampusSafety/index.tsx
richmondnao: Yes i tested this with console logs, it sets the status to expired, if its more than 6 months, uses the new Date() which mean i could test by changing the date on my computer, and it changed those to expired, then changed it back and it checks the date and makes them active if they need to be active

src/views/CampusSafety/index.tsx
richmondnao: I left that there for when you are reviewing this and want to test the code on your end, but yes i do plan to remove them before merging

richmondnao:
> @richmondnao I also discussed this with Karl this morning, but I think we should change this so reports expire sooner (we'll also need to change to message in the form confirmation page. This would be try to limit the number of active reports in the system at a time, to keep it manageable. We were thinking 2 months, was reasonable, but we should maybe add the ability for admin staff to extend that for more valuable/stolen items.
> 
> I know this is potentially a bit of an addition, but if you want to work on some ideas for that, I'd be happy to talk it through as well, but up to you how much we do for this PR.

Not sure what you mean by this , (we'll also need to change to message in the form confirmation page)

We could probably put the admin extension in a different PR since we aren't sure how long that would take, and its not a priority, but that's my guess. This one tool a little bit more thinking about to figure out than i expected, that why

Centerville1:
@richmondnao Overall looks good, but we do need to change the text on the confirmation pages, where it says "Your Report will Expire Automatically in 6 months" to instead say "in 2 months."  This would be for all of the confirmation pages, on the general user and admin user side, and both creating and editing, I believe.

Centerville1 (approved):
LGTM!

PR #1132, opened 2021-05-19
Re-centered Elements

cpabbot:
Re-centered a few elements affected by change in global styling.
Centered people search, 404 mascot, activity profile title, and lock icon for private profile memberships.

Issue #1130, opened 2021-05-19
Membership Requests Cancel vs Hide


cpabbot:
There is a separate cancel icon depending if a request has been approved, which makes sense functionally but might be poor visual design. The 'x' is functionally like a hide button while the 'cancel' button removes a sent request. If we ever created a notification (or push notification) system we could simply put the approved request notifications there instead. Something to think about.

![image](https://user-images.githubusercontent.com/17221652/118860664-ac9ebc80-b8a9-11eb-92cd-e257bf647d8d.png)

PR #2532, opened 2025-01-27
Frontend Use API Filters

Luke-W-Hart:
Goes with [API #1130](https://github.com/gordon-cs/gordon-360-api/pull/1130)
Also now [API #1132](https://github.com/gordon-cs/gordon-360-api/pull/1132)
Grabs reports from back end with API side filtering.
- Original API request grabbed all reports and then set filtered reports on the client side.
- Now, when filters are changed, API request grabs only the reports matching the filter query.


src/views/CampusSafety/views/LostAndFoundAdmin/views/MissingItemList/index.tsx
```diff
@@ -21,6 +21,7 @@ import styles from './MissingItemList.module.scss';
 import { DateTime } from 'luxon';
 import { useNavigate } from 'react-router';
 import useMediaQuery from '@mui/material/useMediaQuery';
+import { sort } from 'services/utils';
```
ben-1010, line 24: Unused import

ben-1010 (changes requested):
Tested and works, one small change needed with imports

Centerville1:
@Luke-W-Hart Fixed code with imports, but I think this code is in need of refactoring.  Since we're doing pagination, I think we also need to move all the filtering to the API side.  This will not only greatly simplify the logic (which currently has to update both from the API and the frontend filters, which is a bit of a mess of state control dependencies currently), but will also pave the way for pagination to work, so I'm going to work on that API update tomorrow, and then push the UI updates here, so I think we'll wait to merge this until that happens.

Centerville1:
Ready for review, and tested, should be working and good to go!

ben-1010 (approved):

PR #2535, opened 2025-01-27
Report Filters Saved in URL Query Params

Centerville1:
Uses URL query parameters to store the filter values, and maintains the filter values across navigation.


Centerville1:
> LGTM, one thing I noticed in testing is that if you select a report and then navigate back to the lost item database, the active filter goes away. Is this intentional? Or is it always supposed to default to active reports?

Right, good point, I think it's just the need to update the link in the breadcrumb, gonna try to push that change now.

PR #2539, opened 2025-01-29
Require Description for Stolen Items

Luke-W-Hart:
Still need to test


Centerville1 (approved):
LGTM when I tested it, the only thing is the disabled button looks a little weird (like it stays blue, instead of being fully grayed out), but otherwise it works well.

PR #2541, opened 2025-02-02
Fixed case-sensitive issues for item categories

erandersen21:
Changed admin missing item form to match the lowercase formatting of the item category from the general user form and changed the edit form to be case insensitive so that previous upper case entries don't cause issues.


Centerville1 (approved):
LGTM

PR #2542, opened 2025-02-02
Made location lost wrap to another line

erandersen21:
Fixed the admin side item view so that the location lost text wraps to another line instead of cutting off text


Centerville1 (approved):
LGTM, love a good one word change 👌.

PR #2545, opened 2025-02-03
wmr - RD View Task List UI

Mrand03:
**Changes made in this PR:**

- Built the UI for the RD view task manager.
- This allows the RD to create, edit and delete tasks so that the RA/AC can see what they need to do while on duty.
- This is not yet using API calls due to my backend not running with my frontend but I will try to figure that out and work on getting the API calls associated with this code tomorrow. 

image
image
image


dli505 (approved):
LGTM!

PR #2546, opened 2025-02-03
Lazy Loading for Missing Item Reports

richmondnao:
I still have to remove the code for the sort thing, i only commented out the code did not remove it yet.

### DIFF
--- a/src/views/CampusSafety/views/LostAndFoundAdmin/views/MissingItemList/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFoundAdmin/views/MissingItemList/index.tsx
@@ -71,7 +71,7 @@ const MissingItemList = () => {
   const location = useLocation();
   const isMobile = useMediaQuery('(max-width:900px)');
 
-  // Lazy loading state and ref 
+  // Lazy loading state and ref
   const [lazyLoading, setLazyLoading] = useState(false);
   const [hasMore, setHasMore] = useState(true);
   const loadMoreRef = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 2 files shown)


Luke-W-Hart (dismissed):
The pagination does work, but the reports are ordered with the report with the oldest Date Created returned first and the reports with the newest Date Created at the end. It should be the opposite (I tested with the up to date API code).

Centerville1:
> The pagination does work, but the reports are ordered with the report with the oldest Date Created returned first and the reports with the newest Date Created at the end. It should be the opposite (I tested with the up to date API code).

That'll be an API change, I'll make that now.

src/views/CampusSafety/views/LostAndFoundAdmin/views/MissingItemList/index.tsx
```diff
@@ -448,9 +502,9 @@ const MissingItemList = () => {
                           {formatDate(report.dateLost)}
                         </Grid>
                         <Grid item xs={2}>
-                          <div
-                            className={styles.dataCell}
-                          >{`${report.firstName} ${report.lastName}`}</div>
+                          <div className={styles.dataCell}>
```
Centerville1, line 505: Did anything actually change here?

Centerville1 (changes requested):
Overall looks pretty good, just a couple of questions and other comments.

src/views/CampusSafety/views/LostAndFoundAdmin/views/MissingItemList/index.tsx
richmondnao: Nope just a space lol

Centerville1 (approved):

Centerville1 (approved):

PR #2547, opened 2025-02-05
Added RAs on call to RA view

RossClark01:
RAs now have the option to view who is on call for each hall if they are on call.

Added logic to housing developer view selector to avoid a looping issue


dli505 (approved):
LGTM!

PR #2549, opened 2025-02-07
Fixes from UI Review

Centerville1:
Fixes from review of #2548

### DIFF
--- a/src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
@@ -167,7 +167,7 @@ const MissingItemFormEdit = () => {
       }
     });
     if (dateError !== null) {
-      errors['dateLost'] = "You can't select a date in the future!";
+      errors['dateLost'] = dateError;
     }
     setValidationErrors(errors);
     return Object.keys(errors).length === 0;
--- a/src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
+++ b/src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import { useState, useEffect, useMemo } from 'react';
 import {
   Card,
   CardHeader,
@@ -228,9 +228,9 @@ const MissingItemFormEdit = () => {
     navigate('/lostandfound');
   };
 
-  const [dateError, setDateError] = React.useState<DateValidationError | null>(null);
+  const [dateError, setDateError] = useState<DateValidationError | null>(null);
 
-  const errorMessage = React.useMemo(() => {
+  const errorMessage = useMemo(() => {
     switch (dateError) {
       case 'invalidDate': {
         return 'Invalid Date';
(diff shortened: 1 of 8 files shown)


src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -180,7 +197,7 @@ const MissingItemFormEdit = () => {
   const handleReportSubmit = async () => {
     const requestData: MissingItemReport = {
       ...formData,
-      dateLost: new Date(formData.dateLost).toISOString() || DateTime.now().toISO(),
+      dateLost: new Date(formData.dateLost).toDateString(),
```
EjPlatzer, line 200: Is there a reason you switched from `toISOString` to `toDateString`? I'm not sure that `toDateString` is really what you want, since that returns a browser-specific human-readable date representation (e.g. "Fri Feb 07 2025"), not a standard machine-readable date string. Maybe you wanted something like `formatISO(new Date(formData.dateLost), {representation: 'date'})` from [`date-fns`](https://date-fns.org/v4.1.0/docs/formatISO#examples)?

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -201,7 +218,7 @@ const MissingItemFormEdit = () => {
     await lostAndFoundService.updateMissingItemReport(requestData, parseInt(itemId || ''));
     const actionRequestData = {
       missingID: parseInt(itemId || ''),
-      actionDate: DateTime.now().toISO(),
+      actionDate: new Date().toDateString(),
```
EjPlatzer, line 221: As above, are you sure `toDateString` does what you want?

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
```diff
@@ -488,6 +524,8 @@ const MissingItemFormEdit = () => {
                       name="description"
                       value={formData.description}
                       onChange={(e) => setFormData({ ...formData, description: e.target.value })}
+                      error={!!validationErrors.description}
```
EjPlatzer, line 527: A more readable way to get a boolean from a non-boolean value is with the Boolean constructor:
```suggestion
                      error={Boolean(validationErrors.description)}
```

(Note: the Boolean constructor without the `new` keyword returns the value coerced to a primitive `boolean` type in JS. If you use the `new` keyword, you get a non-primitive `Boolean` object.)

EjPlatzer (commented):
These are good fixes! I have one important question about your use of `toDateString` that may be blocking.

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
EjPlatzer: If it's just in the lost-and-found code, you can add it the backlog of changes to make for the next release. If it's more widespread throughout 360, I would say don't worry about it.

src/views/CampusSafety/views/LostAndFound/views/MissingItemEdit/index.tsx
Centerville1: Surprisingly, from my testing, it seems to work either way.  My guess is the API does some form of date string parsing when accepting the object send by the frontend, because the date gets updated either way.  For consistency though, I modified it to use toISOString instead 👍

ben-1010 (approved):
LGTM

PR #2555, opened 2025-02-14
Added RA option to check in to additional halls

RossClark01:
As usual RA will have a button to check in to their shift

image

When the RA is checked in the button text will change asking if they want to check in to additional halls. Previously the button would be disabled

image

When clicked any halls the RA is currently checked in to will be checked and disabled as they are unable to check out of a hall and unable to check into a hall they are already in.

image

The API call will only deal with the newly selected halls

image

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -4,8 +4,8 @@
 # VITE_API_URL=https://360Api.gordon.edu/
 
 # @TRAIN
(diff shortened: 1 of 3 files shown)


Mrand03 (approved):
Thanks for making this change Ross, based on our user testing feedback, the UI looks great and so does the code. Everything seems good to merge.

PR #2564, opened 2025-02-22
Found Item Clean Out Page

erandersen21:
This is a messy, non-functional version of the clean-out page. It will need many styling changes and added logic, but this was my general idea. I would appreciate any feedback!

Known issues:
- Does not work on mobile


ben-1010 (approved):
LGTM

PR #2566, opened 2025-02-24
Remove Under Construction Text

Luke-W-Hart:
Removed the "Under Construction" text as it's no longer under construction :)
Looks like this now...
Screenshot 2025-02-24 at 4 43 51 PM


Centerville1 (approved):
LGTM

PR #2569, opened 2025-02-26
Comparison View on Admin Page

Luke-W-Hart:


erandersen21 (changes requested):
The list of reports should refresh after a change. There probably should also be some sort of confirmation after a match (but this isn't probably needed for approval).

erandersen21 (approved):
The thing where you can't click on the same report again after marking it "no match found" should be fixed before release, but otherwise LGTM

PR #2570, opened 2025-02-27
Moved Contact change to be a separate component

RossClark01:
Added back call to TaskList as it was lost in a previous commit from an outdated branch


Mrand03 (approved):
Peer reviewed during team meeting, everything looks great.

PR #2573, opened 2025-02-28
Found Confirmation Page

richmondnao:
Changes for confirmation page


src/views/LostAndFound/index.tsx
```diff
@@ -53,10 +58,18 @@ export const CampusSafetyRoutes: CampusSafetyRoutesObject = {
     element: <FoundItemList />,
     formattedName: 'Found Items Database',
   },
+  '/lostandfoundadmin/founditemform/founditemdatabase': {
```
Centerville1, line 61: Why do we have two routes going to the found item list view?  Why is this one under founditemform?

src/views/LostAndFound/index.tsx
```diff
@@ -53,10 +58,18 @@ export const CampusSafetyRoutes: CampusSafetyRoutesObject = {
...
+    element: <FoundItemList />,
+    formattedName: 'Found Items Databases',
+  },
   '/lostandfoundadmin/founditemdatabase/:itemId': {
     element: <FoundItemFormEdit />,
     formattedName: 'View Found Item #~',
   },
+  '/lostandfoundadmin/founditemform/founditemdatabase/:itemId': {
```
Centerville1, line 69: Why do we have two routes going to the found item edit view?

src/views/LostAndFound/components/Header/index.tsx
```diff
@@ -53,23 +53,25 @@ const pathSubstringToFormattedName = (substring: string, pathnames: string[]) =>
...
+    const routeConfig = CampusSafetyRoutes[breadcrumbRoute];
+    if (routeConfig) {
+      formattedName = routeConfig.formattedName;
+      let queryString = '';
+      if (routeConfig.queryString) {
+        queryString = routeConfig.queryString;
+      }
+      if (formattedName) {
+        formattedName = formattedName.replace('~', substring);
+        return [formattedName, queryString];
```
Centerville1, line 66: Just out of curiosity, is this change just to clean up the logic here, or was there some other issue?

src/views/LostAndFound/views/FoundItemConfirm/index.tsx
```diff
@@ -0,0 +1,520 @@
...
+   * When the "Mark as Match" button is clicked,
+   * open the match confirmation dialog with the selected report.
+   */
+  const handleMatchButtonClick = (report: MissingItemRow) => {
+    setSelectedMatchReport(report);
+    setMatchDialogOpen(true);
+  };
+
+  /**
+   * If the user confirms the match, update the found item with matchingMissingID.
```
Centerville1, line 98: Probably also needs to update the missing item to be status found, and add the found admin action to the missing item.

src/views/LostAndFound/views/FoundItemConfirm/index.tsx
```diff
@@ -0,0 +1,520 @@
...
+          <Grid container justifyContent="flex-end" padding={2}>
+            <Grid item xs={6} sm={3} md={2}>
+              <Button variant="contained" color="primary" fullWidth onClick={handleFinish}>
+                Finish
+              </Button>
+            </Grid>
+          </Grid>
+        </Card>
+      )}
+      {/* MATCH CONFIRMATION DIALOG */}
```
Centerville1, line 461: Doesn't have to be in this PR, but we'll need to add fields to record how you contacted the owner, which will then be used to create Admin Actions on both items recording the match, and how the owner was contacted.

src/views/LostAndFound/index.tsx
```diff
@@ -40,6 +41,10 @@ export const CampusSafetyRoutes: CampusSafetyRoutesObject = {
     element: <MissingItemReportData />,
     formattedName: 'View Report #~',
   },
+  '/lostandfoundadmin/founditemform/:recordID': {
```
Centerville1, line 44: These route names seem a little bit confusing.  Why is the confirmation page under foundItemForm?  Maybe it should be it's own url, like foundItemCross-Check (we might end up linking to this page from other places too.

Centerville1 (changes requested):
A couple comments, and things I had questions about, but overall looking pretty good 👍

src/views/LostAndFound/index.tsx
richmondnao: Resolved in speech

src/views/LostAndFound/index.tsx
richmondnao: Resolved in speech

src/views/LostAndFound/components/Header/index.tsx
richmondnao: resolved in speech

Centerville1 (approved):
Merging functional version now, more changes/improvements to this page will come in future PRs

PR #2574, opened 2025-02-28
Wmr ra UI fixes

jtasonye:
**Changes made in PR:**
* Tasklist will always be displayed, but if an RA/AC is not on duty, they will get a message that says that tasks are unavailable to them unless they are on duty
* Added `CheckIn` button component into the `MyHall` component. With this, line 165 in `MyHall`, the `CheckIn` will only show if the user is an RA. I have it commented out to check that the `CheckIn` component works, but for the final push, it should be uncommented
* It would be great if someone could test how the tasks would look when things are added to it, I was not able to get to that. My thinking is that we will do the same thing we did with tasklist and create a scrollbar, so it does not go above a specified height

Screenshot 2025-02-28 at 1 53 01 PM
Screenshot 2025-02-28 at 1 53 44 PM

### DIFF
--- a/src/views/ResLife/components/RAView/components/CheckIn/index.jsx
+++ b/src/views/ResLife/components/RAView/components/CheckIn/index.jsx
@@ -187,7 +187,7 @@ const CheckIn = () => {
 
   return (
     <Grid container item justifyContent="center" alignItems="center">
-      <Grid item xs={12} md={4}>
+      <Grid item xs={12} md={12}>
         <Button variant="contained" fullWidth={true} onClick={() => setOpen(true)}>
           {isCheckedIn ? 'check in to additional Halls?' : 'Check In To Your Shift'}
         </Button>
--- a/src/views/ResLife/components/RAView/components/CheckIn/index.jsx
+++ b/src/views/ResLife/components/RAView/components/CheckIn/index.jsx
@@ -189,7 +189,7 @@ const CheckIn = () => {
     <Grid container item justifyContent="center" alignItems="center">
       <Grid item xs={12} md={12}>
         <Button variant="contained" fullWidth={true} onClick={() => setOpen(true)}>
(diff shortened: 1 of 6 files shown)


src/views/ResLife/components/RAView/components/CheckIn/index.jsx
```diff
@@ -187,9 +187,9 @@ const CheckIn = () => {
 
   return (
     <Grid container item justifyContent="center" alignItems="center">
-      <Grid item xs={12} md={4}>
+      <Grid item xs={12} md={12}>
         <Button variant="contained" fullWidth={true} onClick={() => setOpen(true)}>
-          {isCheckedIn ? 'check in to additional Halls?' : 'Check In To Your Shift'}
+          {isCheckedIn ? 'check in to extra Halls' : 'Check In To Your Shift'}
```
RossClark01, line 192: May just be me but additional sounds a little more professional than 'extra'

src/views/ResLife/components/ResidentView/components/MyHall/index.jsx
```diff
@@ -149,6 +162,8 @@ const MyHall = () => {
               />
             </Tooltip>
           </Grid>
+          {/* {isRA ? <CheckIn /> : <></>} */}
```
RossClark01, line 165: Will need to figure this out before push to SP. I can take a look as well

RossClark01 (approved):
looks good to me! couple minor details we will need to address before working towards roll out

src/views/ResLife/components/ResidentView/components/MyHall/index.jsx
RossClark01: remembering now it wasn't giving us trouble, we just needed an RA to test it... We might be able to get Richmond to sign in for us

PR #2578, opened 2025-03-06
Rd checkin

RossClark01:
Added the component for scheduling RDs to be on duty and for RAs to have access to the on call info

RD On-Duty Scheduler: 

image

Accessible through Edit Docs:

image

Available on the RA page:

image

Matched existing style where able and followed logic from the task component for RDs and the Ra on duty for residents. The RD on call phone is passed from RD to RD so it will be one number. We can edit this once we get it.


dli505 (approved):
LGTM!

PR #2580, opened 2025-03-13
Kiosk User Features Done

richmondnao:
Finished Features for kiosk User.


Luke-W-Hart (changes requested):
Looks awesome!
Just a few things:
- Breadcrumbs need to be formatted rather than just lowercase with no spaces
- Should we remove the New Action button for Kiosk users? It doesn't click so maybe it's better for it to not be there.
- I'm able to change the date in kiosk view (but not submit the date change)
- Status filter should probably be default to "active" on kiosk lists

These aren't things that will break the system but it'd be good to at least write them down. If you think we just need to address these later let me know and I'll approve this PR.

richmondnao:
> Looks awesome!
> 
> Just a few things:
> 
> - Breadcrumbs need to be formatted rather than just lowercase with no spaces
> 
> - Should we remove the New Action button for Kiosk users? It doesn't click so maybe it's better for it to not be there.
> 
> - I'm able to change the date in kiosk view (but not submit the date change)
> 
> - Status filter should probably be default to "active" on kiosk lists
> 
> 
> 
> These aren't things that will break the system but it'd be good to at least write them down. If you think we just need to address these later let me know and I'll approve this PR.

Yh I can get to these changes later tonight they should be quick fixes

richmondnao:
> Looks awesome! Just a few things:
> 
> * Breadcrumbs need to be formatted rather than just lowercase with no spaces
> * Should we remove the New Action button for Kiosk users? It doesn't click so maybe it's better for it to not be there.
> * I'm able to change the date in kiosk view (but not submit the date change)
> * Status filter should probably be default to "active" on kiosk lists
> 
> These aren't things that will break the system but it'd be good to at least write them down. If you think we just need to address these later let me know and I'll approve this PR.

Hey Luke, i think the Breadcrumbs should be fine for now, it might be a bigger change to fix it than we want it happen between now and tomorrow, and the new Actions Button has been disabled. Also the Kiosk user should be able to submit any changes to the found item, there's no way to disable the date without not showing it completely, and that would mean having a different view if the page for the kiosk user which i don't believe we would want to do. Also the status filter for found items currently doesn't have any defaults, both Admin and Kiosk User, we could add it to the project board and have it be a different project, since its not exactly a feature we need immediately.

Luke-W-Hart:
Sounds good, I'll put some of those things on the project board tomorrow.

Luke-W-Hart (approved):
LGTM! Tested and approved based on previous comments.

PR #2581, opened 2025-03-13
Category edit richie

richmondnao:
Allow admin to change category of missing reports


Luke-W-Hart (commented):
Erik and I both tried to test this and got an 401 Unauthorized when we tried to edit the category. Was it working for you?

richmondnao:
> Erik and I both tried to test this and got an 401 Unauthorized when we tried to edit the category. Was it working for you?

Yh it woudn't work for now, if youre changing it for something thats not your item, some API stuff need to be changed to allow give permissions for the change to be possible, so basically this is awaiting an API change, Matt and I mentioned this earlier though

Luke-W-Hart (approved):
LGTM!

PR #2583, opened 2025-03-16
Changed links on mobile to open in the same tab

RossClark01:
This will resolve an issue when using the 360 'app' clicking on a link takes you over to your default browser


dli505 (approved):
LGTM!

PR #2584, opened 2025-03-17
Staff View updates

RossClark01:
Add RD on call to Staff view

edit RD on call to better match other card styling/spacing


jtasonye:
I completely agree, it does have a friendly feel, like we actually care about them or something. Good job, I approve!

PR #2588, opened 2025-03-17
Quick fix match richie

richmondnao:
[When you go into the edit page for the found item entered i want to navigate back to the found item entry confirmation page. And commented out the last change for going to the admin page after a confirmation match.


src/views/LostAndFound/views/FoundItemConfirm/index.tsx
```diff
@@ -110,6 +110,7 @@ const FoundItemConfirmation = () => {
       });
       setMatchDialogOpen(false);
       setSelectedMatchReport(null);
+      //navigate('/lostandfound/lostandfoundadmin')
```
Luke-W-Hart, line 113: Remove this?

Luke-W-Hart (approved):
LGTM! Just see the one comment

PR #2592, opened 2025-03-21
Mobile View on Clean Out Page

erandersen21:
The pages now work on mobile (as in they are functional and none of the info is cut off), but they are difficult to understand and poorly laid out (aka ugly). We could merge this if we wanted since it's better than before, but I'm going to work on it more.


Centerville1 (approved):
LGTM

PR #2594, opened 2025-03-23
Fixed Details Button Contrast on CrossCheck Page

Luke-W-Hart:
Looks like this (on the far left) (matches the Figma now):
Screenshot 2025-03-23 at 4 35 52 PM

Also seems like the columns are broken for the data cells :)


ben-1010 (approved):
LGTM

PR #2595, opened 2025-03-23
Found Item List Status Defaults to Active

Luke-W-Hart:
Also includes formatting changes from prettier (since whoever did FoundItemEdit didn't have prettier installed I assume).


ben-1010 (approved):
LGTM

PR #2602, opened 2025-03-24
added day of week to 1 time event

RossClark01:


jtasonye (approved):
Thanks for catching that Ross! The logic looks good, forget the sp-branch, let's push to prod right now!

PR #2607, opened 2025-03-25
updates status UI

RossClark01:
resolved a data issue when editing an event

added logic to determine which days have active events so the user knows which days to look at:

image

added a confirm delete option that lets the user know all events in the series are deleted:

image

added back option to remaining pages:

image


RossClark01 (commented):
Peer reviewed with team in person

PR #2618, opened 2025-04-04
Added Phone number management

RossClark01:
Added functionality for storing phone numbers in the backend. 

When loading or there is an error text is displayed

image

When pulled succesfully

image

There is an edit option for the RDs. I am fighting with styling it functions just fine

image

Very simple interface

image


dli505 (approved):
LGTM!

PR #2622, opened 2025-04-07
Using Link Report Function on Missing Item Details Page

Luke-W-Hart:


src/views/LostAndFound/views/MissingReportDetails/index.tsx
```diff
@@ -332,12 +336,25 @@ const MissingItemReportData = () => {
...
-    if (!checkedItemNotFound && newActionFormData.action === 'Checked') {
+    if (typeof checkedActionFormData.foundID !== 'string') return;
+
+    if (!checkedItemNotFound && newActionFormData.action === 'Checked' && isValidForm()) {
       await lostAndFoundService.updateReportStatus(parseInt(itemId ? itemId : ''), 'found');
       setReportUpdated(reportUpdated + 1);
-    }
-
-    if (isValidForm()) {
+      lostAndFoundService.linkReports(
```
Centerville1, line 344: One issue, what happens if I put in a found ID that doesn't exist.  Currently, I can enter any found ID I want, and the linking to the found item will break, but the rest of the process will work fine.  Probably need to add error handling to the linkReports method so if the API fails on any of the steps, it will rollback/undo the whole linking step.

Centerville1 (changes requested):

src/views/LostAndFound/views/FoundItemEdit/index.tsx
```diff
@@ -683,6 +683,25 @@ const FoundItemFormEdit = () => {
                 </Select>
               </div>
 
+              <div style={{ marginBottom: '1rem' }}>
+                {/* Delete line below before merging */}
+                {console.log(foundItem.matchingMissingID)}
```
richmondnao, line 688: Says delete line before merge, lol

Luke-W-Hart: Glad you caught that haha

richmondnao (approved):
LGTM!

Centerville1 (approved):
LGTM (to clear the issue)

PR #2624, opened 2025-04-13
Changes for Found Confirmation Page

richmondnao:


Centerville1 (changes requested):
Screen Shot 2025-04-13 at 13 57 24

A couple thoughts just visually:
Serial number looks good, although I think we should put that in a box of a different color or something to make it stand out even more...  Maybe just a border is fine.

The text from the entered item details is too small, we should make that bigger.  I also think it would be helpful to put the date found, location, category, and description fields for the found item in a row, so it's easier to compare the entered found item with the missing items in the scrollable list (not absolutely necessary, but would make the top-bottom comparison step much easier, especially if your entering a lot of items).

The category filter dropdown is empty when the page first loads, even though the list is filtered by electronics.  It also appears that the filter, when you set it manually, is case sensitive, and only looks for items with the upper case category (so when you select to filter by Electronics, it only shows 2 reports, when in reality there are a lot).  Also, the missing report list doesn't filter by active reports only, so it includes some expired (and maybe other statuses) missing item reports when it should not.  Also, is there a reason we removed the text in Figma that says "Check if item has been reported missing"?  I think that would be helpful just as a reminder of what this page does, for people who don't use it often.
Screen Shot 2025-04-13 at 14 11 40


Also, we'll have to update the link reports functionality to use the new service method Ben and Luke wrote (see #2622).  This performs all the needed linking functionality (which requires a lot of steps), and is designed to be reused wherever we do linking.

Screen Shot 2025-04-13 at 14 06 40

The order of the fields in the matching pop-up should be the same on both sides, to make it easier to compare and double check.

We'll also need to show the owner information, and record how we contacted the owner when we link the reports.  The pop-up to do that is being designed by Ben/Luke, so maybe wait for them to create that and then just reuse it in both places 9would be better).

Let me know if you have any questions!

richmondnao:
> Screen Shot 2025-04-13 at 13 57 24
> A couple thoughts just visually: Serial number looks good, although I think we should put that in a box of a different color or something to make it stand out even more... Maybe just a border is fine.
> 
> The text from the entered item details is too small, we should make that bigger. I also think it would be helpful to put the date found, location, category, and description fields for the found item in a row, so it's easier to compare the entered found item with the missing items in the scrollable list (not absolutely necessary, but would make the top-bottom comparison step much easier, especially if your entering a lot of items).
> 
> The category filter dropdown is empty when the page first loads, even though the list is filtered by electronics. It also appears that the filter, when you set it manually, is case sensitive, and only looks for items with the upper case category (so when you select to filter by Electronics, it only shows 2 reports, when in reality there are a lot). Also, the missing report list doesn't filter by active reports only, so it includes some expired (and maybe other statuses) missing item reports when it should not. Also, is there a reason we removed the text in Figma that says "Check if item has been reported missing"? I think that would be helpful just as a reminder of what this page does, for people who don't use it often. Screen Shot 2025-04-13 at 14 11 40
> 
> Also, we'll have to update the link reports functionality to use the new service method Ben and Luke wrote (see #2622). This performs all the needed linking functionality (which requires a lot of steps), and is designed to be reused wherever we do linking.
> 
> Screen Shot 2025-04-13 at 14 06 40
> The order of the fields in the matching pop-up should be the same on both sides, to make it easier to compare and double check.
> 
> We'll also need to show the owner information, and record how we contacted the owner when we link the reports. The pop-up to do that is being designed by Ben/Luke, so maybe wait for them to create that and then just reuse it in both places 9would be better).
> 
> Let me know if you have any questions!

Added all requested changes only thing left if the link for matching items pending merge into the main branch, everything else is good.

Centerville1 (approved):
LGTM

PR #2627, opened 2025-04-14
Found Item Finder/Owner Info on Edit Page

erandersen21:
Makes finder and owner visible/editable on the found item edit page.

Issues:


Luke-W-Hart (changes requested):
A few things:
- If the finder or owner is cleared then I can't save the changes
- I am unable to put anything into the Finder phone number text field (if they want the item)
- If there is a phone number and email associated with an owner (or finder I would assume), and then I change the person, there is no longer a phone number or email

PR #2633, opened 2025-04-28
Changed link to use the new Link Reports Service link

richmondnao:
Added New Service Link

### DIFF
--- a/src/views/LostAndFound/views/FoundItemConfirm/index.tsx
+++ b/src/views/LostAndFound/views/FoundItemConfirm/index.tsx
@@ -46,7 +46,6 @@ const FoundItemConfirmation = () => {
   const [foundItem, setFoundItem] = useState<FoundItem | null>(null);
   const [missingReports, setMissingReports] = useState<MissingItemReport[]>([]);
   const [loading, setLoading] = useState(true);
-  const [missingItem, setItem] = useState<MissingItemReport | null>(null);
 
   // Default filter: show missing reports matching the found item's category.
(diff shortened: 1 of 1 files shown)


src/views/LostAndFound/views/FoundItemConfirm/index.tsx
```diff
@@ -140,70 +140,60 @@ const FoundItemConfirmation = () => {
...
-      username: foundItem?.finderUsername || 'unknown',
-      isPublic: false,
-    };
+
     try {
-      await lostAndFoundService.createAdminAction(selectedMissingReport.recordID, requestData);
+      await lostAndFoundService.linkReports(
+        missingID,
+        foundID,
+        // owner info comes from your foundItem
```
Centerville1, line 170: Surely the owner info should come from the missing item?  Since that's the report submitted by the owner?  Most of the time we don't know the owner of the foundItem, but we always know the owner of the missing item...

src/views/LostAndFound/views/FoundItemConfirm/index.tsx
```diff
@@ -140,70 +140,60 @@ const FoundItemConfirmation = () => {
...
+        foundItem.finderUsername || 'unknown',
```
Centerville1, line 171: Should probably be empty string if unknown, to avoid issues in the API.

Centerville1 (changes requested):
A couple things to change...  These might have come from the other use of this linkReports function, and if so these changes should be made there too, and were probably missed earlier

Centerville1 (approved):

PR #2635, opened 2025-04-30
Admin Page Dashboard and Color Key

ben-1010:
Added a dashboard with legends to visualize last checked dates on missing reports and statuses of found items

Screenshot 2025-05-01 at 3 33 55 PM


Luke-W-Hart (changes requested):
I think we want the missing reports we’re counting to be active reports, other than that it looks good!

Luke-W-Hart (approved):
LGTM!

PR #2645, opened 2025-05-20
Lf release2.1

richmondnao:
Added Bug Fixes that Gordon Police discovered, when submitting a report for someone they were unable to change the change to a date in the past and it only worked when the date was the current date. 
Also fixed bug for double submission for a missing report for others.

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -1,8 +1,8 @@
 # Refers to a production environment, which may be either train or prod
 
 # @PROD
-# VITE_API_URL=https://360Api1.gordon.edu/
+VITE_API_URL=https://360Api.gordon.edu/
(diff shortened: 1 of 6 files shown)


.env.production
EjPlatzer, line 1: Looks like this branch has had `master` merged into it for a full release. The changes from master can't be merged into `develop` without causing a bunch of environment issues for both train and eventually production.

Please create a clean branch and cherry pick the relevant commits into it.

EjPlatzer (changes requested):
The fixes look good, but this branch can't be merged because it contains `master` commits. Please create a new branch with the two relevant commits and create a new PR with a clearer description.

PR #2646, opened 2025-05-21
Date lost fix

richmondnao:
This Pull Request looks to solve two specific bugs in the Lost and Found System. Firstly, it solves the error where an admin user is not able to submit a report on behalf of someone else if the date selected is in the past and not the current date.
Secondly, it solves the error where submitting a report on behalf of someone else creates two reports for that person.


richmondnao:
> LGTM!

Awesome, Also i still have have authorization to merge , :/

Issue #1085, opened 2021-04-22
ApartApp (2) Help dropdown additions


joshuaRogers24:
- [ ] Change X info to say “use the ‘x’ icon next an item to remove it from the list”
- [ ] Add section for info about the Rank functionality

PR #2651, opened 2025-05-27
Does not include punctuation in search

Jshmill:
Kind of a rudimentary fix of #1085. This PR takes the last name search and strips it of all punctuation, which does allow us to search last names with periods and apostraphes, but there could be a more in-depth attempt at this to make it more polished.


jsenning (changes requested):
This seems to remove hyphens.  It might be better to just remove periods.

Jshmill:
Closing pull request because we are going to try to tackle this problem in the backend

Issue #2653, opened 2025-05-28
Involvement card sizes cause big words to be split


Jshmill:
When on the involvement pages, depending on the size of the window, some of the cards have text that splits a word into two different lines in order to fit. It could benefit from having bigger cards such that the text doesn't need to split down a word in order to fit on the card.

PR #2654, opened 2025-05-28
involvement cards fit for any size window

Jshmill:
Quick fix to #2653, I had it working already I just wanted to document the problem to keep things organized


jsenning (commented):
Does this work as expected on different window sizes (mobile, desktop, etc)?
Also, when I look at the "Files Changed" tab I see multiple warnings suggesting small changes.  These would be good to address. while you're working in this file.

russtuck (approved):

Issue #2658, opened 2025-05-29
Menu button off center


Jshmill:
The "TODAYS MENU" text on the button is not centered when viewing from a very specific window. Easy CSS fix.

PR #2659, opened 2025-05-29
aligned "Todays Menu" text to be centered on the button

Jshmill:
Quick fix to #2658. Just wanted to document it!


jsenning (commented):
This is a small change, and I presume it works okay.  To verify, however, I'd need to checkout the branch and see the change - and given that dining balances might show differently for fac/staff than for students, I'm not sure I could see what you see.

A better approach is to have "before" and "after" screenshots as part of the PR.  That makes it much easier to see impact of the change and gives some context for the change. 

Nice job creating a project board entry and linking to it.

Jshmill:
That makes sense,
here is the before:

Screenshot 2025-06-02 at 10 11 49 AM


and here is the after:

Screenshot 2025-06-02 at 10 12 53 AM

PR #2663, opened 2025-05-30
RecIM admin activities filter sorted by status, name, and start date

justin123-lee:
RecIMadminAfter
RecIMadminBefore
In the admin command center, I added a filter that sorts the activities based on their status, name, and start date.


src/views/RecIM/views/Admin/index.jsx
```diff
@@ -368,7 +391,54 @@ const Admin = () => {
...
+                  <ActivityList
+                    activities={[...activities].sort((a, b) => {
+                      if (activitySort === 'Name') {
+                        return (a.Name || '').localeCompare(b.Name || '');
+                      }
+                      if (activitySort === 'StartDate') {
+                        return new Date(a.StartDate) - new Date(b.StartDate);
+                      }
+                      return 0;
+                    })}
```
EjPlatzer, line 435: Since you're doing the sorting within the JSX, it will re-run on every render. This is bad practice, because React chooses to re-render things fairly often, so you're doing potentially lots of redundant work.

The alternative is to do the sorting once in the event handler (i.e. a function you pass as the `onChange` prop) of the Sort By input, and store the sorted activities as state.

justin123-lee: That makes sense. I will try to work on it following your suggestion. Thank you

jsenning (changes requested):
Justin,

The code looks okay to me.

This PR also deletes .env.development -- which is not what you wanted to do.  Typically we don't want to commit changes to this file, but this PR will actually remove the file from the develop branch.  I suggest you use `git checkout develop .env.development` followed by a `git add .env.development` and a `git commit` to restore the file.

PR #2668, opened 2025-06-04
S25 fix DaysLeft

### LINKED ISSUE #2664: Fix DaysLeft
The DaysLeft feature requires access to future sessions in a calendar year; apparently these don't always exist for future sessions.  This either causes the DaysLeft bar to be corrupted or not shown at all.  PR #2668

jsenning:
Fixes #2664.  The DaysLeft bar displayed on the homepage stopped working during the winter break of the 2024-2025 academic year.  This was due to it trying to get information from sessions that did not yet have session codes in the database.

The newly refactored version gets a list of all available sessions, sorted from newest to oldest.  A linear search then determines which of three cases applies:
  1. today's date falls after the end of the most recent session
  2. today's date falls with a session
  3. today's date falls between two existing sessions
Once we know what session or break we are in, we construct an appropriate message for the progress bar.

Testing has become easier since we never ask for a session that does not exist.  To try different dates we just need to change how the `today` variable is initialized.

### DIFF
--- a/src/services/session.ts
+++ b/src/services/session.ts
@@ -83,7 +83,7 @@ const getTermCode = (): string => {
     year -= 1;
   }
 
-  return `${year.toString()}${term}`;
+  return `${year.toString().substring(2)}${term}`;
 };
 
 const parseSessionCode = (sessionCode: string) => parse(sessionCode, 'yyyyMM', new Date());


Jshmill (approved):
LGTM!

PR #2669, opened 2025-06-05
Streamline loading and calculation of days left

EjPlatzer:
I simplified the loading process to use less intermediate state (and therefore less re-renders), and tried to streamline the date logic by using Date primitives and our standard `date-fns` library, with the goal of making the logic easier to understand for future maintainers.


jsenning (approved):
Evan,  Thanks for looking at this and all the refactoring.  The date-fns library is quite handy and certainly does clean things up a bit.

PR #2671, opened 2025-06-05
S25 fix days left again

jsenning:
Adjusted the computations for numbers of days.  For example, between a term ending on the 5th and another starting on the 8th there are only 2 days.  Likewise, a term with the first day on the 5th and the last day on the 8th has 4 days.  This PR makes these adjustments.


jsenning:
Thanks Evan!

PR #2672, opened 2025-06-05
S25 fixed change mobile phone number

JasonWang432:
For the UI now when clicking the pencil button it will let you edit the existing phone number with the number you inputted and cancel it if you missed click on it. Before hand when click on the pencil icon you where able to press cancel if you did not input anything in the text field, however if you did input anything at all regardless of where you click even the cancel button it would crash the 360 website saying something went wrong and email cts etc.


JasonWang432:
https://github.com/gordon-cs/gordon-360-api/pull/1207
Link to the backend pull request for the Fixed change mobile phone number

src/components/Profile/components/PersonalInfoList/components/UpdatePhoneDialog/index.tsx
```diff
@@ -9,98 +9,129 @@ import {
...
 
   const handleSubmit = async () => {
+    setLoading(true);
     try {
       await userService.setMobilePhoneNumber(mobilePhoneNumber);
-      createSnackbar('Your phone number will update within a couple hours.', 'success');
-    } catch {
-      createSnackbar('Phone number failed to update. Please contact CTS.', 'error');
+      if (onUpdateSuccess) onUpdateSuccess(mobilePhoneNumber);
+      const updatedProfile = await userService.getProfileInfo();
```
jsenning, line 63: This line generates a warning in Github that indicates this variable is never used.  Can you look into this and make the necessary changes to remove the warning?

jsenning (changes requested):
The code LGTM except for a github warning.  Can this be tested with 360.studenttest?  if not, perhaps a code reviewer (Josh?) could verify that this works as expected.

src/components/Profile/components/PersonalInfoList/components/UpdatePhoneDialog/index.tsx
JasonWang432: I just removed the line as it's not used and is already being updated by the onUpdateSuccess callback.

PR #2674, opened 2025-06-06
Adds an effect for hovering over the tabs in Gordon360

Jshmill:
Small change that I think adds an immersive level to Gordon360... When the cursor is hovered over the tabs on the top, the tab lifts up a little bit. In the screenshot below, the cursor is hovered over the people tab.

Screenshot 2025-06-06 at 9 15 26 AM

This makes the UI for the header a little bit more interactive and immersive, and I think it spices things up nicely.

### DIFF
--- a/src/components/Header/Header.module.scss
+++ b/src/components/Header/Header.module.scss
@@ -16,6 +16,24 @@
   min-width: 55px;
   max-width: 125px;
   flex-grow: 1;
+  position: relative;
+  transition:
+    transform 0.2s ease,
+    z-index 0.2s ease,
+    box-shadow 0.2s ease;
+
+  &:hover {
+    transform: translateY(-4px);
+    z-index: 10;
+  }
+
+  :global(svg) {
+    transition: transform 0.2s ease;
+  }
+
+  &:hover :global(svg) {
+    transform: translateY(-2px);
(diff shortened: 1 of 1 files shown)


Jshmill:
I added some more animations along the header and along the navigation bar on the side as well, to add an extra immersive level. I also think having the logo move as well is important, and me personally I didn't realize that it was clickable for a long time, so having that little feature should help the user learn that clicking the logo does something - in this case, takes them to home.

Another video to document it all:


Uploading Screen Recording 2025-06-06 at 10.20.22 AM.mov…

Jshmill:
I added some more animations along the header and along the navigation bar on the side as well, to add an extra immersive level. I also think having the logo move as well is important, and me personally I didn't realize that it was clickable for a long time, so having that little feature should help the user learn that clicking the logo does something - in this case, takes them to home.

Another video to document it all:



https://github.com/user-attachments/assets/4dbdb7c0-cd60-4b43-89aa-7e05aeb93eb5

Centerville1 (changes requested):
Love this change!  Overall looks great, just a bit of CSS can be simplified and still seems to achieve the same effect.

russtuck (approved):

PR #2692, opened 2025-06-17
S25 metz link

abelyshev22:
Change the link in DiningBalance.index.jsx to the metz menu link.


EjPlatzer (dismissed):
As long as this page is stable (i.e. the URL of the menu is not changing frequently), this looks good to me.

I wonder if it's worth linking to `www.gordonmetz.com` from the Links page (probably within Services) since it has lots of information beyond the current menu.

PR #2696, opened 2025-06-18
Update index.tsx

JasonWang432:
Changed the options={gender.sort()} to options={[...gender].sort()} to help avoid unintended state mutation.

### DIFF
--- a/src/views/PeopleSearch/components/SearchFieldList/index.tsx
+++ b/src/views/PeopleSearch/components/SearchFieldList/index.tsx
@@ -239,7 +239,7 @@ const SearchFieldList = ({ onSearch }: Props) => {
       setDepartments(departments);
       setBuildings(buildings.map((b) => b.Description));
       setInvolvements(involvements);
-      setGender(gender);
+      setGender([...gender].sort());
       setLoading(false);
     };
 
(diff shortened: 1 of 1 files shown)


EjPlatzer (commented):
Is there a reason that you don't sort the gender before it's stored, so that it's only sorted once, instead of every time the component renders?

JasonWang432:
it was the suggestion that copilot made so it seem like the easiest fix however I was unsure if it was necessary to even do in the first place

src/views/PeopleSearch/components/SearchFieldList/index.tsx
```diff
@@ -239,7 +239,7 @@ const SearchFieldList = ({ onSearch }: Props) => {
       setDepartments(departments);
       setBuildings(buildings.map((b) => b.Description));
       setInvolvements(involvements);
-      setGender(gender);
+      setGender([...gender].sort());
```
EjPlatzer, line 242: You could use [the `toSorted` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) instead, which is easier to understand. It's semi-new to JavaScript, but has been supported in all major browsers since July 2023, and [has 91% support](https://caniuse.com/mdn-javascript_builtins_array_tosorted) across all users globally, so it should be safe to use.

```suggestion
      setGender(gender.toSorted());
```

EjPlatzer (approved):
LGTM!

PR #2704, opened 2025-06-24
Add metz link to services tab in Gordon360 links page

abelyshev22:
Since we went through with the today's menu link change, I thought it would be nice to have the original link redirect located somewhere else, as that page holds additional info.


EjPlatzer (approved):
LGTM!

PR #1222, opened 2021-06-29
(S21) Adjusted layout of people search options and fixed class year

cpabbot:
Also changed 'show large images' to be a toggle instead of a checkbox

image
<pre>
      ^
      |
      |
</pre>
image

PR #2706, opened 2025-06-25
S25 registration info

JasonWang432:
https://github.com/gordon-cs/gordon-360-api/pull/1222 is the matching PR in the API.

Front end for the Registration window where is would show the session date like spring and get the start and end of there registration date with any holds they might have and then lets then click on it where it would lead to the info for how to fix the hold they have


JasonWang432:
the api request for the backend
S25 registration date #1222

jsenning (dismissed):
There is a merge conflict that needs to be resolved -- looks like you may have lost a closing `</div>` tag.

PR #2707, opened 2025-06-27
s25 Quick search highlights and advanced search

JasonWang432:
Fixed an issue where when pressing enter would select the first option without visually highlighting it. I also added an advanced search option to the bottom of the quick search where it would bring you to the people's tab


src/components/Header/components/QuickSearch/index.tsx
```diff
@@ -166,14 +203,54 @@ const GordonQuickSearch = ({
...
+      renderOption={(props, option) => {
+        if (option.UserName === '__ADVANCED__') {
+          return (
+            <Tooltip title="Can't find who you're looking for? Try our Advanced Search.">
+              <MenuItem
+                {...props}
+                onClick={() => {
+                  setCurrentQuery('');
+                  dispatch({ type: 'RESET' });
+                  navigate('/people');
```
ben-1010, line 234: Minor issue, but the autocomplete dropdown will still be active after the user clicks on the advanced search option. It looks like this immediately after clicking:
Screenshot 2025-07-08 at 3 09 51 PM
You'll probably need to blur it when the user clicks this option.

ben-1010 (changes requested):
Just a few small things need improvement here, but great job!

src/components/Header/components/QuickSearch/index.tsx
```diff
@@ -221,37 +288,35 @@ const getHighlightedDetails = (person: SearchResult, inputPartsRegex: RegExp) =>
...
- * - `'`
- * - `-`
- * - `.`
- */
-const specialCharactersRegex = /[^a-zA-Z0-9'\-.\s]/gm;
-/**
- * Match all instances of whitespace and `.` at the beginning and end of a string.
- */
+// Regex for invalid characters in search
+const specialCharactersRegex = /[^a-zA-Z0-9'\-\.\s]/gm;
```
ben-1010, line 304: The "." doesn't need to be escaped here. The expression `[^a-zA-Z0-9'.-\s]` should work fine.

JasonWang432:
just did the changes suggested and it works

russtuck (approved):

PR #2714, opened 2025-06-30
s25 small fix fetching currentTerm

justin123-lee:
`export const getCurrentTerm = (): Promise<AcademicTerm> => http.get('academicterm/currentterm');`

removed ' / ' from ' `/academicterm/currentterm`


jsenning (approved):
LGTM

PR #2716, opened 2025-07-01
s25 change daysleft from sessions to terms

justin123-lee:
The daysleft bar in the home page was using sessions instead of terms to count the days. I changed sessions to terms and used the filtered undergrad terms from the backend for the list of terms in daysleft. Undergrad terms are only terms of Fall, Spring and Summer.

### DIFF
--- a/src/services/academicTerm.ts
+++ b/src/services/academicTerm.ts
@@ -12,8 +12,10 @@ export const getAllTerms = (): Promise<AcademicTerm[]> => http.get('academicterm
 
 export const getCurrentTerm = (): Promise<AcademicTerm> => http.get('academicterm/currentterm');
 
-export const getUndergradTerms = (): Promise<AcademicTerm[]> =>
-  http.get('academicterm/undergradTerms');
+export const getUndergradTerms = async (): Promise<AcademicTerm[]> => {
+  const allTerms = await getAllTerms();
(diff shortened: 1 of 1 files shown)


jsenning (commented):
Reviewing this PR has led me down a trail of discovery and this note documents some of what I've learned, listed in no particular order.

- Getting a list of all terms from the API is not very efficient, as there are over 200 terms, most of which are in the past.  To show the daysLeft bar we only need information from the current term or, if between terms, the previous term and the next term.
- Related to the last point, is there anything we need to do to cache the appropriate data for the PWA to work properly when it's offline?
- Term codes: FA, JN, SP, and SU stand for fall, January, spring, and summer terms.  These apply for undergraduates and graduates -- except graduate education.  SF and WS stand for summer-fall and winter-spring and apply to graduate education.
- The `academicTerm/current` route returns the first academic term object in which the value in the `TermConfigYear` column of the Identify.dbo.TermConfig table corresponding to the row where the `TermConfigName` column contains `PeopleSearchCurrentTerm`.  (This was suggested by @EjPlatzer).  I presume that the table on Production is kept up-to-date, but the table on Train is now over a year out-of-date.  Also, when does the listed term change -- the first day of the new term?
- The `sessions/current` route returns a session object for the session which includes the current date.  This seems more reasonable, especially for the daysLeft task.
-

PR #2719, opened 2025-07-01
s25 Show Current or Next Term Schedule

JasonWang432:
This update uses academic term data to automatically select the most relevant schedule for the student. If the current date falls within an active term and the student has courses during that term, that schedule will be shown by default. If not, it will look ahead and display the next upcoming term that includes courses for the student. If neither condition is met, it will fall back to showing the most recent term where the student had any courses.

### DIFF
--- a/src/components/Profile/components/SchedulePanel/index.tsx
+++ b/src/components/Profile/components/SchedulePanel/index.tsx
@@ -24,6 +24,7 @@ type Props = {
 };
 
 const scheduleOpenKey = 'profile.schedule.isOpen';
+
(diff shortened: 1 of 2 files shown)


jsenning (changes requested):
Jason - this will be a nice feature to have -- especially the ability to see the schedule for the upcoming term.  I'll note as well that this should (and does!) work for faculty.

At least in the case of faculty, we sometimes have schedule data for more than one future term (e.g. both fall and spring terms), even though the schedules for at least one of the terms has not been finalized.  We should not make schedules available until after the point in time that the registrar publishes the schedule, so we need to keep this from happening.

PR #2724, opened 2025-07-08
Make name change feature private-only

### LINKED ISSUE #2722: S25 Name change button appearing on public profile.
It seems that when searching for another student on 360 the name "edit" button is displayed on their profile. While this wouldn't allow students to change another student name it's still important to keep these features private.

abelyshev22:
Fixes #2722


EjPlatzer (approved):
LGTM!

PR #2730, opened 2025-07-10
A couple more fixes needed to make sure everything shows up

Jshmill:
Forgot to take out a couple slashes in the other pull request, so this one takes care of them!

### DIFF
--- a/src/views/MarketPlace/components/AdminPage/index.jsx
+++ b/src/views/MarketPlace/components/AdminPage/index.jsx
@@ -11,6 +11,7 @@ import {
   CircularProgress,
   TextField,
   Box,
+  Button,
 } from '@mui/material';
 import {
   Dialog,
@@ -117,17 +118,22 @@ const AdminMarketplaceThreads = () => {
   return (
     <Paper>
       <Box p={2}>
-        <TextField
-          label="Search ID Number"
-          variant="outlined"
-          value={searchQuery}
-          onChange={(e) => setSearchQuery(e.target.value)}
-          onKeyDown={(e) => {
-            if (e.key === 'Enter') {
-              setSearchTrigger(searchQuery.trim());
-            }
-          }}
-        />
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
+          <TextField
+            label="Search ID Number"
+            variant="outlined"
+            value={searchQuery}
+            onChange={(e) => setSearchQuery(e.target.value)}
+            onKeyDown={(e) => {
+              if (e.key === 'Enter') {
+                setSearchTrigger(searchQuery.trim());
+              }
+            }}
+          />
+          <Button variant="contained" onClick={() => setSearchTrigger(searchQuery.trim())}>
+            Search
+          </Button>
+        </Box>
       </Box>
       <Table>
         <TableHead>
--- a/src/views/MarketPlace/components/MarketPlacePopup/index.jsx
+++ b/src/views/MarketPlace/components/MarketPlacePopup/index.jsx
@@ -174,8 +174,7 @@ Thank you
       // Fetch profile info
       userService
         .getInformalName(item.PosterUsername)
-        .then(setProfileInfo);
-        })
+        .then(setProfileInfo)
         .catch(() => setProfileInfo(null));
     }
   }, [item?.PosterUsername]);
--- a/src/views/MarketPlace/index.jsx
+++ b/src/views/MarketPlace/index.jsx
@@ -38,6 +38,16 @@ import GordonLogo from './images/gordoncollegelogo.jpg';
 const sorts = ['Date', 'Price', 'Title'];
 const order = ['Ascending', 'Descending'];
 
+const handlePriceChange = (value, setter) => {
+  if (value === '' || /^\d*\.?\d*$/.test(value)) {
+    if (value.includes('.')) {
+      const [, decPart] = value.split('.');
+      if (decPart.length > 2) return; // block input beyond 2 decimals
+    }
+    setter(value === '' ? undefined : Number(value));
+  }
+};
+
 /**
  * Marketplace component renders the main UI for browsing and managing marketplace listings.
  *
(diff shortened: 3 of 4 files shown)


src/services/marketplace.ts
```diff
@@ -54,7 +54,7 @@ export const getProfileImage = async (username: string) => {
     throw new Error('No access token found');
   }
 
-  const response = await fetch(`/api/profiles/image/${username}`, {
+  const response = await fetch(`api/profiles/image/${username}`, {
```
EjPlatzer, line 57: Why are you using raw `fetch` instead of the wrapped `http` service? This is not aligned with the rest of the codebase, and causes issues like this that have already been solved in `http`. This also makes long-term maintenance a pain because our HTTP code is now spread around the codebase instead of centralized in the `http` service like it's supposed to be.

EjPlatzer: More to the point, why did you create new `getProfileInfo` and `getProfileImage` methods in the marketplace service, instead of using the existing ones in `user.ts`?

russtuck (dismissed):
See Evan's comments.

src/services/marketplace.ts
```diff
@@ -74,7 +74,7 @@ export const getProfileInfo = async (username: string) => {
     throw new Error('No access token found');
   }
 
-  const response = await fetch(`/api/profiles/${username}`, {
+  const response = await fetch(`api/profiles/${username}`, {
```
russtuck, line 77: I'd suggest replacing this method (getProfileInfo()) with a simple method in user.ts that looks something like this:

const getInformalName = async (username: string = ''): Promise<Profile | undefined> => {
  const profile = await getProfile(username);

  if (!profile) return undefined;
  return {
    NickName: profile.NickName,
    LastName: profile.LastName,
  };

Jshmill: I believe it is fixed in my commit

src/services/user.ts
```diff
@@ -307,6 +307,21 @@ const isBirthdayToday = (): Promise<boolean> => {
...
+  NickName: string;
+  LastName: string;
+};
+
+const getInformalName = async (username: string = ''): Promise<InformalName | undefined> => {
+  const profile = await getProfile(username);
+  if (!profile) return undefined;
+
+  return {
+    NickName: profile.NickName,
```
EjPlatzer, line 320:
```suggestion
    NickName: profile.NickName ?? profile.FirstName,
```

src/views/MarketPlace/components/AdminPage/index.jsx
```diff
@@ -63,13 +68,71 @@ const AdminMarketplaceThreads = () => {
...
+        <TextField
+          label="Search ID Number"
+          variant="outlined"
+          value={searchQuery}
+          onChange={(e) => setSearchQuery(e.target.value)}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter') {
+              setSearchTrigger(searchQuery.trim());
+            }
+          }}
```
EjPlatzer, line 129: There needs to be some way for a mouse user to trigger the search, and preferably a visible, obvious way to trigger the search because not every user will think to hit enter.

Also, since you're only setting `searchTrigger` in response to this `onKeyDown`, you should just make `trySearchById` a regular function and directly call it when the event occurs, rather than setting an intermediate state:
```suggestion
          onKeyDown={(e) => {
            if (e.key === 'Enter') {
              trySearchById(searchQuery.trim());
            }
          }}
```

src/views/MarketPlace/components/AdminPage/index.jsx
```diff
@@ -63,13 +68,71 @@ const AdminMarketplaceThreads = () => {
...
+              setIsModalOpen(true);
+              setSearchQuery('');
+              setSearchTrigger('');
+              setLoading(false);
+              return;
+            }
+          }
+
+          hasMore = threadsPage.length === pageSize;
+          currentPage++;
```
EjPlatzer, line 101: This seems like more work than is necessary to find a specific database object by Id. Can you not just create an API route that gets a specific ThreadEditHistory object by Id? Even if you can't query the database by id directly (although I suspect you definitely can), you should at least move this nested searching logic into the API where it can be performed much more efficiently and not hold up the rendering on the front end.

src/views/MarketPlace/components/MarketPlacePopup/index.jsx
```diff
@@ -172,9 +172,12 @@ Thank you
...
       // Fetch profile info
-      getProfileInfo(item.PosterUsername)
-        .then((info) => {
-          setProfileInfo(info);
+      userService
+        .getInformalName(item.PosterUsername)
+        .then((profile) => {
+          setProfileInfo(
+            profile ? { NickName: profile.NickName, LastName: profile.LastName } : null,
+          );
```
EjPlatzer, line 180: Since `getInformalName` returns a `{NickName, LastName}` object or undefined, isn't this functionally identical to:
```suggestion
        .then(setProfileInfo);
```

src/views/MarketPlace/index.jsx
```diff
@@ -131,6 +130,16 @@ const Marketplace = () => {
...
+  const handlePriceChange = (value, setter) => {
+    if (value === '' || /^\d*\.?\d*$/.test(value)) {
+      if (value.includes('.')) {
+        const [, decPart] = value.split('.');
+        if (decPart.length > 2) return; // block input beyond 2 decimals
+      }
+      setter(value === '' ? undefined : Number(value));
+    }
+  };
+
```
EjPlatzer, line 142: Since this is a pure function, please declare it outside the component.

EjPlatzer (dismissed):
It seems like there was some scope creep on this fixes PR, but that's understandable. Overall, this looks good.

I left a few code quality suggestions, but nothing that is show stopping if this needs to be merged.

src/views/MarketPlace/components/AdminPage/index.jsx
Jshmill: Fixed!!

Jshmill:
So I did everything except the backend implementation, because that would take a little bit more work which I think deserves its own pull request, but I can create a tile for it and work on it tomorrow!

PR #2732, opened 2025-07-11
Added API thread search by ID logic to Admin Page

Jshmill:
UI addition for [#1240](https://github.com/orgs/gordon-cs/projects/15/views/1?pane=issue&itemId=119507221&issue=gordon-cs%7Cgordon-360-api%7C1239)

Backend found here: https://github.com/gordon-cs/gordon-360-api/pull/1240

### DIFF
--- a/src/views/MarketPlace/components/AdminPage/index.jsx
+++ b/src/views/MarketPlace/components/AdminPage/index.jsx
@@ -22,9 +22,15 @@ import {
   IconButton,
 } from '@mui/material';
 import CloseIcon from '@mui/icons-material/Close';
+import GordonSnackbar from 'components/Snackbar';
 
 const AdminMarketplaceThreads = () => {
   const [threads, setThreads] = useState([]);
+  const [snackbar, setSnackbar] = useState({
+    message: '',
+    severity: 'info',
+    open: false,
+  });
 
   const [totalCount, setTotalCount] = useState(0);
   const [page, setPage] = useState(0);
@@ -82,7 +88,11 @@ const AdminMarketplaceThreads = () => {
           setSearchQuery('');
           setSearchTrigger('');
         } else {
-          alert(`No thread found with ID ${searchTrigger}`);
+          setSnackbar({
+            message: `No thread found with ID ${searchTrigger}`,
+            severity: 'warning',
+            open: true,
+          });
         }
       } catch (err) {
         console.error('Error searching thread history:', err);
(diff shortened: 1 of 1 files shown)


src/views/MarketPlace/components/AdminPage/index.jsx
```diff
@@ -73,48 +68,31 @@ const AdminMarketplaceThreads = () => {
...
-          hasMore = threadsPage.length === pageSize;
-          currentPage++;
+        if (results.length > 0) {
+          const history = await marketplaceService.getThreadEditHistory(results[0].ThreadId);
+          setEditHistory(history);
+          setIsModalOpen(true);
+          setSearchQuery('');
+          setSearchTrigger('');
+        } else {
+          alert(`No thread found with ID ${searchTrigger}`);
```
EjPlatzer, line 85: `alert` is an insecure, deprecated browser function that is often disabled in modern browsers and should not be used. For transient user notifications, use a snackbar instead.

src/views/MarketPlace/components/AdminPage/index.jsx
```diff
@@ -73,48 +68,31 @@ const AdminMarketplaceThreads = () => {
...
+        const results = await marketplaceService.getAdminThreads({
+          id: parseInt(searchTrigger),
+          page: 1,
+          pageSize: 1,
+        });
 
-          hasMore = threadsPage.length === pageSize;
-          currentPage++;
+        if (results.length > 0) {
+          const history = await marketplaceService.getThreadEditHistory(results[0].ThreadId);
```
EjPlatzer, line 79: Does this really have the same behavior? Before, the id in `searchTrigger` was the id of the ThreadEditHistory, but now it seems like it's being used as the id of the Thread.

EjPlatzer (changes requested):

src/views/MarketPlace/components/AdminPage/index.jsx
Jshmill: Fixed!

src/views/MarketPlace/components/AdminPage/index.jsx
Jshmill: I'm pretty sure it does work! I am able to search by ID's that are deep within the edit history of a post and it finds it, and it's super efficient as well... 

here is a demo of the search function, and as you can see, 380 is deep in the history of that one post:


https://github.com/user-attachments/assets/fe948942-2960-4620-b421-c3212d97161d

EjPlatzer (approved):
LGTM!

PR #2751, opened 2025-07-18
added class date in adition to term date

abelyshev22:
quad class were only showing the term dates which didn't really match the dates of the class


jsenning (approved):
LGTM

PR #2753, opened 2025-07-18
update office location for office admins added

### LINKED ISSUE #2752: Office Location Update
This issue is implemented by the pull requests #2753 and https://github.com/gordon-cs/gordon-360-api/pull/1257.

It has been noted by Tina Lippiatt that there are a number of Office Locations that are incorrect for Faculty and Staff. 

A new security group that has permission to change other Faculty/Staff's office location just like how existing users can change their own would be a beneficial feaeture.

chanyoung-hur:
Give members of the 360-OfficeAdmin-SG security group the ability to update faculty/staff office locations from their public profiles.

Paired with API https://github.com/gordon-cs/gordon-360-api/pull/1257


fishyfishfish55:
Everything requested is functional, but there is still some code cleanup to be done (see [api comment](https://github.com/gordon-cs/gordon-360-api/pull/1257#issuecomment-4735118097)). 

Right now we get the Mailstop description from the updateMailStop function but this function should probably either return a table with both the updated Mailstop and the description, or just the Mailstop, with another helper function to get the associated value (see [fae4e4a](https://github.com/gordon-cs/gordon-360-ui/pull/2753/commits/fae4e4a37cd7c9f78945d60d0c802de62872255a))

jsenning (changes requested):
This looks good to me.  except that the file `.env.development` should not be included in the PR.  As a faculty member I can change my mailstop and my office hours with immediate effect but changes to my office location go into a black hole.  The current behavior on production is similar but the requested changes are eventually updated. As the back-end mechanism for updating office location isn't changed, there is good reason to believe the changes here will work correctly.

Once the `.env.develoment` file is removed I think @EjPlatzer should review this just to make sure the approach taken is acceptable.

src/services/user.ts
```diff
@@ -281,12 +281,31 @@ const setPlannedGraduationYear = (value: number | string) => {
...
+const updateOfficeHours = (officeHours: string, username?: string) => {
+  let url = 'profiles/office_hours';
+  if (username) {
+    url += `?username=${encodeURIComponent(username)}`;
+  }
+  return http.put(url, officeHours);
+};
 
-const updateOfficeHours = (value: string) => http.put(`profiles/office_hours`, value);
+// const updateOfficeHours = (value: string) => http.put(`profiles/office_hours`, value);
```
EjPlatzer: Minor nitpick: can we remove this outdated, commented out code before deploying?

EjPlatzer (changes requested):
I think this all looks good. Like I mentioned in my API review, I've updated the database views so that changes to office location, office hours, and mail stop are reflected immediately in CCT.dbo.FacStaff. Can we get rid of the toast message saying that the changes to Office location will take a few hours to be displayed, and wire up the data correctly so that changes are shown to the user immediately?

src/components/Profile/components/OfficeInfoList/UpdateOfficeLocationDialog/index.tsx
```diff
@@ -24,12 +28,10 @@ const UpdateOffice = () => {
...
-      await userService.updateOfficeLocation({ BuildingCode: building, RoomNumber: room });
-      setSnackbar({
-        message: 'Your office location will update within a couple hours.',
-        severity: 'success',
-        open: true,
-      });
+      await userService.updateOfficeLocation(
+        { BuildingCode: building, RoomNumber: room },
+        username,
+      );
```
EjPlatzer: Once the API is updated to return the new office location, we should update the frontend to receive the returned new office location and display it to the user.

fishyfishfish55: Rather than requesting the updated value from the backend, the mailstop and office hours components attempt to update the current fields with the submitted ones when no errors are thrown in the API request, and assume that the database is changed to reflect the frontend.
I have made the UpdateOfficeLocationDialog and OfficeInfoList components to follow the existing pattern, but would it be worth doing a larger refactor to request updated values from the backend?

EjPlatzer: That's a good question! The number one thing we want to avoid is someone thinking they've made a change, but that change hasn't actually taken effect in the backend. That's why the baseline best practice is to display the value as it currently exists from the backend. We could employ the optimistic updates pattern to assume the operation has succeeded while the backend is still processing it (e.g. show the value the user submitted between the time their request is sent to the backend and the time the new value is returned from the backend), but at the end of the day we would still want to show the value that the backend has as the source of truth.

jsenning: I think it makes sense to have updates to mailstop, office room and building, and office hours all work the same way.  We do return updated profile values for the mailstop (need to update the tooltip).  If the PUTs to the API return the updated information then the UI can update the values displayed using that information.

fishyfishfish55: Should be fixed.

EjPlatzer (approved):
LGTM!

jsenning (approved):
LGTM

PR #2765, opened 2025-12-10
Hotfix: Fix erroneous privacy "enforcement"

EjPlatzer:
A summer practicum student tried to enforce privacy for mobile phones in an unrelated PR this summer. Firstly, the front end can never enforce privacy because the data has already been delivered to the user by that point. Secondly, the backend was already enforcing privacy for mobile phones as designed, but the student broke that and hid the mobile phone from staff members (e.g. RDs) who need access to it to perform their jobs.

This fixes CTS Case #00079425.


russtuck:
Thank you!

PR #2774, opened 2026-06-12
fix profile reload issue

woobenskyp:
The myprofile page was using the public profile endpoint to populate itself. This caused some private information to be missing, e.g., Gordon ID.


jsenning (approved):
LGTM

PR #2779, opened 2026-06-22
Restore .env.development

jsenning:


jsenning (commented):
LGTM - just restoring default configuration

